PR #1, opened 2023-12-19
Rebranding: Update working title to actual title

JoCa96:
Updated package names to match.


packages/components/package.json
```diff
@@ -17,11 +17,11 @@
   },
   "repository": {
     "type": "git",
-    "url": "https://github.com/schwarzit/nui",
-    "directory": "packages/nui"
+    "url": "https://github.com/schwarzit/onx",
```
larsrickert, line 20:
```suggestion
    "url": "https://github.com/schwarzit/onyx",
```

README.md
```diff
@@ -6,11 +6,11 @@
   </a>
 </p>
 
-# NUI
+# Onyx
 
-A Vue.js UI component library and Design System created by [Schwarz IT](https://it.schwarz).
+A Design System and Vue.js UI component library created by [Schwarz IT](https://it.schwarz).
```
larsrickert, line 11: We should also apply this change to:
- packages/components/README.md
- packages/components/package.json
- apps/docs/src/index.md
- The GItHub repository description

.changeset/pre.json
```diff
@@ -2,8 +2,8 @@
   "mode": "pre",
   "tag": "alpha",
   "initialVersions": {
-    "@temp-nui/nui": "0.0.0",
-    "@temp-nui/storybook-utils": "0.0.0"
+    "@sit-onyx/components": "0.0.0",
```
larsrickert, line 5: I am not quite sure about this package name.
What do you think about just naming it `sit-onyx`?

I think its common practice to name by main package by the scope itself and use the scope only for additional packages. E.g. like `vue`, `prettier`, `storybook` etc.

larsrickert (changes requested):

packages/components/.cloud-foundry/manifest-dev.yml
```diff
@@ -1,5 +1,5 @@
 ---
 applications:
-  - name: nui-storybook-dev
+  - name: onyx-storybook-dev
```
larsrickert, line 3: This will also change the URL of the deployment, so we should also update it in the `apps/docs/src/.vitepress/env.ts` file :)

README.md
JoCa96: Done, description will be updated

packages/components/package.json
```diff
@@ -17,11 +17,11 @@
   },
   "repository": {
     "type": "git",
-    "url": "https://github.com/schwarzit/nui",
-    "directory": "packages/nui"
+    "url": "https://github.com/schwarzit/onyx",
+    "directory": "packages/components"
```
larsrickert, line 21: Should we also rename the folder from `components` to `sit-onyx`?
So that the folder name is equal to the package name for every package

larsrickert (dismissed):

packages/components/package.json
JoCa96: done

larsrickert (approved):

PR #3, opened 2023-12-19
chore: snapshot release for PRs

JoCa96:
- Added conditiontional snapshot release to pipline
   - we only want to publish a real version on the main branch
   - all other branches/prs should only produce a snapshot
- Set `apps/docs` as private package as we don't want to publish it

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,7 +3,6 @@ on:
   push:
     branches:
       - "main"
-  pull_request:
 
 jobs:
   prerelease:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -49,5 +49,7 @@ jobs:
       - name: 📸 Snapshot and Publish
         if: ${{ github.ref != 'refs/heads/main' }}
         run: |
+          # [Snapshot release fails if repository is in pre-release mode](https://github.com/changesets/changesets/issues/1195)
+          rm -f .changeset/pre.json
           npx changeset version --snapshot
           npx changeset publish --tag snapshot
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -31,12 +32,26 @@ jobs:
...
+        if: ${{ github.ref == 'refs/heads/main' }}
+        run: |
           npx changeset version
           npx changeset publish
-          git add .
+          git add --all
           git commit --amend --no-edit
           git push --follow-tags
+
+      - name: 📸 Snapshot and Publish
```
larsrickert, line 49: Whats the purpose of the snapshot releases?
Why not just releasing an alpha version after merged on main?

I think releasing every PR will lead to a lot of pretty much unused versions on npm

JoCa96: It's useful to be able to actually install and smoke test the would-be package of an PR before its actually merged. It also ensures that there areWe missed this feature in CoreUI quite a few times. Is there an issue with having many snapshot versions on npm?
If there is we could also remove the PR trigger

larsrickert: I don‘t think there is an issue with npm, personally I just don‘t like having hundreds of snapshot versions :)

But its nice to have the option. What do you think about changing it to a manual trigger instead and remove the PR trigger?

JoCa96: done

larsrickert: You‘d need the workflow_dispatch trigger in order to be able to run the workflow manually:)

larsrickert (approved):

PR #5, opened 2023-12-20
chore(headless): set up headless package skeleton

JoCa96:
- added skeleton for headless package `@sit-onyx/headless`
- added missing authors and licenses


packages/headless/README.md
```diff
@@ -0,0 +1,57 @@
...
+    <img src="../../.github/schwarz-group.svg" width="400px" />
+    </div>
+  </a>
+</p>
+
+# Onyx Headless
+
+A composable headless library for Vue created by [Schwarz IT](https://it.schwarz).
+
+Inspired by [melt](https://melt-ui.com/).
```
larsrickert, line 13:
```suggestion
Inspired by [Melt UI](https://melt-ui.com/).
```

packages/headless/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "@sit-onyx/headless",
+  "description": "Headless Composables for Vue",
```
larsrickert, line 3:
```suggestion
  "description": "Headless composables for Vue",
```

packages/headless/package.json
```diff
@@ -0,0 +1,21 @@
...
+  "type": "module",
+  "author": "Schwarz IT KG",
+  "license": "Apache-2.0",
+  "exports": {
+    ".": {
+      "default": "./src/index.ts"
+    }
+  },
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
```
larsrickert, line 14: Would prefer to remove this noop script

packages/headless/package.json
```diff
@@ -0,0 +1,21 @@
...
+    "build": "tsc --noEmit"
+  },
+  "dependencies": {
```
larsrickert, line 17: Shouldn't they be peerDependencies?

packages/headless/src/index.ts
```diff
@@ -0,0 +1 @@
+export { createComboBox } from "./composables/createComboBox";
```
larsrickert, line 1:
```suggestion
export * from "./composables/createComboBox";
```

In case we export types etc. from the file in the future

packages/headless/tsconfig.json
```diff
@@ -0,0 +1,109 @@
+{
```
larsrickert, line 1: We should use the same tsconfig as we do for the `sit-onyx` package which extends the official @vue/tsconfig :)

packages/headless/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "@sit-onyx/headless",
```
larsrickert, line 2: Do we also neec to add this package to the changeset package? I've seen that there is a `initialVersion` config for all packages

larsrickert (changes requested):

packages/headless/package.json
```diff
@@ -0,0 +1,21 @@
...
+  "author": "Schwarz IT KG",
+  "license": "Apache-2.0",
+  "exports": {
+    ".": {
+      "default": "./src/index.ts"
+    }
+  },
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "build": "tsc --noEmit"
```
larsrickert, line 15: This is rather a type-check command since it does not actually build anything, right?

We should also add it to the check workflow that runs for PRs

packages/headless/package.json
JoCa96: According to my tests it isn't necessary for new packages, but I added it anyways

packages/headless/package.json
JoCa96: done

packages/headless/package.json
JoCa96: Out-of-scope for this PR: I added a follow-up task: https://github.com/orgs/SchwarzIT/projects/2/views/1?pane=issue&itemId=48102251

packages/headless/tsconfig.json
JoCa96: We should have a global base tsconfig, so we have consistent rules for all packages. Out-of-scope for this PR. Will create a follow-up ticket.

packages/headless/package.json
JoCa96: done

larsrickert (dismissed):

larsrickert (approved):

PR #9, opened 2024-01-02
feat(forms): add demo for validation behavior

BoppLi:


apps/docs/src/.env.example
```diff
@@ -0,0 +1,2 @@
+# move to apps/docs/src/.env.local file
+# VITE_STORYBOOK_HOST='http://localhost:6006'
```
larsrickert, line 2:
```suggestion
# VITE_STORYBOOK_HOST=http://localhost:6006
```

apps/docs/src/.env.example
```diff
@@ -0,0 +1,2 @@
+# move to apps/docs/src/.env.local file
```
larsrickert, line 1:
```suggestion
# move to .env.local file, e.g. by running:
# cp .env.example .env.local
```

packages/sit-onyx/src/components/TestInput/TestInput.vue
```diff
@@ -8,10 +8,22 @@ const props = withDefaults(
...
   }>(),
   {
     modelValue: "",
     label: "",
+    minlength: undefined,
+    maxlength: undefined,
+    min: undefined,
+    max: undefined,
+    type: undefined,
+    pattern: undefined,
```
larsrickert, line 26: This is caused by eslint, right?^^

packages/sit-onyx/src/components/TestInput/TestInput.vue
```diff
@@ -34,11 +48,22 @@ const handleChange = (event: Event) => {
...
-    <input v-model="value" :required="props.required" @change="handleChange" />
+    <input
+      v-model="value"
+      :required="props.required"
+      :minlength="props.minlength"
+      :maxlength="props.maxlength"
+      :min="props.min"
+      :max="props.max"
+      :type="props.type"
+      :pattern="props.pattern"
```
larsrickert, line 63:
```suggestion
      v-bind="props"
```

We could use this shortcut to remove some redundancies

packages/sit-onyx/src/demos/FormDemo/FormDemo.vue
```diff
@@ -0,0 +1,79 @@
...
+
+const onSubmit = (event: Event) => {
+  event.preventDefault();
+  emit("submit", formData);
+};
+</script>
+
+<template>
+  <form @submit="onSubmit($event)">
+    <TestInput v-model="formData.defaultInput" label="Default"></TestInput>
```
larsrickert, line 49: Don't we have a eslint rule for auto fixing self closing tags?
```suggestion
    <TestInput v-model="formData.defaultInput" label="Default" />
```

larsrickert (changes requested):

apps/docs/src/.env.example
BoppLi:
```
# move to apps/docs/src/.env.local file, e.g. by running:
# cp apps/docs/src/.env.example apps/docs/src/.env.local
```

or
```
# move to src/.env.local file, e.g. by running:
# cp src/.env.example src/.env.local
```

-> usually, you would run scripts from the root folder of the repo / app, right?

packages/sit-onyx/src/components/TestInput/TestInput.vue
BoppLi: yep, eslint wants to have everything defined in the default :( I planned to discuss this with you guys once the PR is ready.

packages/sit-onyx/src/demos/FormDemo/FormDemo.vue
BoppLi: I thought so too, but pnpm lint-fix does not change it somehow.

packages/sit-onyx/src/components/TestInput/TestInput.vue
BoppLi: done :)

BoppLi:
should the PR title be rather `feat(demo): create`?

apps/docs/src/.env.example
larsrickert: The root of the app yes :D Your second suggestion sound good!

larsrickert:
> should the PR title be rather `feat(demo): create`?

It think the current title is fine or something like `feat(forms): add demo story`

packages/sit-onyx/src/demos/FormDemo/FormDemo.vue
larsrickert: We need to add this to our root `.eslintrc`:
```js
rules: {
  "vue/html-self-closing": [
    // see https://eslint.vuejs.org/rules/html-self-closing
    "error",
    {
      html: {
        void: "always",
        normal: "never",
        component: "always",
      },
    },
  ],
},

```

packages/sit-onyx/src/components/TestInput/TestInput.vue
```diff
@@ -1,5 +1,5 @@
 <script lang="ts" setup>
-import { computed } from "vue";
+import { computed, ref } from "vue";
 
 const props = withDefaults(
```
BoppLi, line 4: To be discussed:
1. we planned to export the prop types of all components within the component (if I remember correctly)
2. it would be better to somehow refer to the native html component's props when typing the props of this component.

So I tried the following:
```vue
type NativeInputProps = Pick<
  HTMLInputElement,
  "required" | "minLength" | "maxLength" | "min" | "max" | "type" | "pattern" | "title"
>;
export type TestInputProps = {
  /** The current input value */
  modelValue?: string;
  /** Label to show next to the input */
  label?: string;
} & Partial<NativeInputProps>;

const props = withDefaults(defineProps<TestInputProps>(), {
  modelValue: "",
  label: "",
  type: "text",
});
```

But the vue compiler really does not like it...

image


not sure if it's because of the union type or because it gets overwhelmed by the >350 properties that the HTMLInputElement has... 🤔

BoppLi: seems to be the latter, just using the union type seems to work properly.

BoppLi: doing it properly wihtout the pick gets a bit long ... did we want to move the exported types outside of the component files?

```ts
type HTMLInputTypeAttribute =
  | "button"
  | "checkbox"
  | "color"
  | "date"
  | "datetime-local"
  | "email"
  | "file"
  | "hidden"
  | "image"
  | "month"
  | "number"
  | "password"
  | "radio"
  | "range"
  | "reset"
  | "search"
  | "submit"
  | "tel"
  | "text"
  | "time"
  | "url"
  | "week";
  /** equivalent to Pick<HTMLInputElement,
 * "required" | "minLength" | "maxLength" | "min" | "max" | "type" | "pattern" | "title"
 * >;
 */
type NativeInputProps = {
  required?: boolean | undefined;
  pattern?: string | undefined;
  type?: HTMLInputTypeAttribute | undefined;
  max?: number | string | undefined;
  maxLength?: number | undefined;
  min?: number | string | undefined;
  minLength?: number | undefined;
};
export type TestInputProps = {
  /** The current input value */
  modelValue?: string;
  /** Label to show next to the input */
  label?: string;
} & Partial<NativeInputProps>;

const props = withDefaults(defineProps<TestInputProps>(), {
  modelValue: "",
  label: "",
  type: "text",
});
```

BoppLi: tried to condence this now. but in the end, we should definitely declare which "type" values are allowed, but will probably have separate components for some of those types anyways :)

packages/sit-onyx/src/components/TestInput/TestInput.vue
BoppLi: Partial helped :)

packages/sit-onyx/src/components/TestInput/TestInput.vue
larsrickert: Yes I don't think it makes sense to allow types such as date, file, time etc. when we have dedicated datepickers, filepickers etc.

packages/sit-onyx/src/components/TestInput/TestInput.vue
```diff
@@ -34,12 +62,25 @@ const handleChange = (event: Event) => {
...
 <template>
-  <label class="input">
+  <label class="input" :class="{ 'input--touched': isTouched }">
     <span class="input__label" :class="{ 'input__label--required': props.required }">
       {{ props.label }}
+      <!-- 
+        <span aria-label="required">*</span> 
+        should we use the above instead of the "after"?
+        see https://developer.mozilla.org/en-US/docs/Learn/Forms/Form_validation#full_example
+      -->
```
BoppLi, line 72: To be discussed:
should we use the above instead of the "after"?
see [https://developer.mozilla.org/en-US/docs/Learn/Forms/Form_validation#full_example](https://developer.mozilla.org/en-US/docs/Learn/Forms/Form_validation#full_example)
```suggestion
        <span aria-label="required">*</span>
```

packages/sit-onyx/src/components/TestInput/TestInput.vue
BoppLi: done. let's still discuss tomorrow if we can make it work somehow to connect to the HTMLInputTypeAttribute interface without making vue break (couldn't find it out so far)
and where we want to keep the prop type definitions -> in each .vue file or in (a) separate file(s).

BoppLi:
Add doc for minleght: prefilled values will not invalidate

BoppLi:
todo: provide translations for the common languages for the error cases and let the projects pass a lang key

BoppLi:
create playwright tests for each validation message (without screenshot)

packages/sit-onyx/src/components/TestInput/TestInput.vue
```diff
@@ -1,27 +1,59 @@
...
+  label?: string;
+  /** A custom error message that will be shown (only when the input is invalid) */
+  customErrorMessage?: string;
+} & NativeInputProps;
+
+const props = withDefaults(defineProps<TestInputProps>(), {
+  modelValue: "",
+  label: "",
+  type: "text",
+  customErrorMessage: undefined,
```
BoppLi, line 30: analyze why the vue standard linting rule wants to have a default value for all
vue/require-default-prop

packages/sit-onyx/src/components/TestInput/TestInput.vue
```diff
@@ -1,27 +1,59 @@
...
+const coreElement = ref<HTMLInputElement | null>(null);
+
+const isValid = ref(coreElement.value?.validity.valid || false);
+
+const getErrorMessage = (): string => {
+  if (isValid.value) return "";
+
+  return props.customErrorMessage || coreElement.value?.validationMessage || "";
+};
+const errorMessage = ref(getErrorMessage());
```
BoppLi, line 55: change isValid to contain the object, then the errorMessage can be a computed again.
```suggestion
const validityState = ref(coreElement.value?.validity || null);

const getErrorMessage = (): string => {
  if (isValid.value) return "";

  return props.customErrorMessage || coreElement.value?.validationMessage || "";
};
const errorMessage = ref(getErrorMessage());
```

packages/sit-onyx/src/components/TestInput/TestInput.vue
```diff
@@ -31,15 +63,54 @@ const handleChange = (event: Event) => {
...
+  /* element will always be truthy onMounted */
+  const element = coreElement.value;
+  if (element) {
+    /* we need to watch for value changes, 
+    using computed on element.validity.valid 
+    unfortunately misses the validity resetting */
+    watch(
+      value,
+      () => {
+        /* todo we could use a composable or util for the "only update when it changed" pattern? */
```
BoppLi, line 77: remove the check, instead create a second watcher for emitting the event only when the valid state changed.

packages/sit-onyx/src/components/TestInput/TestInput.vue
BoppLi: not needed, required on the input itself is understandable enough.

packages/sit-onyx/src/components/TestInput/TestInput.vue
BoppLi: 1. for now, we keep the types it in the .vue file.
2. created a follow up task for investigating the compiler crash. https://github.com/orgs/SchwarzIT/projects/5/views/7?filterQuery=task-priority%3Adev&pane=issue&itemId=48984776

larsrickert: This issue is not only a compiler crash. I noticed that Storybook currently does not support generating docs for props when you use a union type such as `export type TestInputProps = { //... } & Partial<NativeInputProps>`

So they will not be documented which is a problem

PR #14, opened 2024-01-02
docs: add headless and storybook-utils to VitePress

larsrickert:


apps/docs/src/packages/headless.md
```diff
@@ -0,0 +1,58 @@
...
+# @sit-onyx/headless
+
+::: warning Work in progress / Active development
+This library is currently in early / active development.
+:::
+
+{{ packageJson.description }}.
+
+Inspired by [Melt UI](https://melt-ui.com).
+
```
BoppLi, line 18: we don't have it yet in npmjs.com right? as soon as we have it, we might want to add the link to the package here :)

apps/docs/src/packages/headless.md
```diff
@@ -0,0 +1,58 @@
...
+Make sure that this chapter is kept up to date with installation steps in
+packages/headless/README.md file.
+ -->
+
+Install the npm package with your corresponding package manager:
+
+::: code-group
+
+```sh [pnpm]
+pnpm add sit-onyx/headless
```
BoppLi, line 31: don't we need the @?

```suggestion
pnpm add @sit-onyx/headless
```

apps/docs/src/packages/headless.md
```diff
@@ -0,0 +1,58 @@
...
+
+{{ packageJson.description }}.
+
+Inspired by [Melt UI](https://melt-ui.com).
+
+## Installation
+
+<!--
+Make sure that this chapter is kept up to date with installation steps in
+packages/headless/README.md file.
```
BoppLi, line 23: can't we import markdown content? and put those steps in a separate file that we link from packages/headless/README.md and import it right here.?

BoppLi: on the other hand, it's not much content so redundance is OK in this case I guess ^^

apps/docs/src/packages/headless.md
BoppLi: also in the next 3 examples.

apps/docs/src/packages/headless.md
```diff
@@ -0,0 +1,58 @@
...
+
+```sh [yarn]
+yarn install sit-onyx/headless
+```
+
+:::
+
+## Composables
+
+### useComboBox
```
BoppLi, line 46: do we really want to show the usage of all the components we'll build or shouldn't we name this more like "Example usage"

apps/docs/src/packages/headless.md
```diff
@@ -0,0 +1,58 @@
...
+<script lang="ts" setup>
+import { createCombobox } from "@sit-onyx/headless";
+
+const {
+  elements: {
+    ...
+  }
+ } = createCombobox();
+</script>
+```
```
BoppLi, line 58: isn't a <template ... missing to fully showcase the usage?

apps/docs/src/packages/index.md
```diff
@@ -0,0 +1,16 @@
+<script lang="ts" setup>
+import headlessPackageJson from "../../../../packages/headless/package.json";
+import storybookUtilsPackageJson from "../../../../packages/storybook-utils/package.json";
+
```
BoppLi, line 4:
```suggestion
<!-- Make sure to also create a file matching the packages name in apps/docs/src/packages to make the link below work when adding another package to this list. -->
```

BoppLi (changes requested):

apps/docs/src/packages/headless.md
larsrickert: Not yet, lets do this after they are released :)

apps/docs/src/packages/headless.md
larsrickert: I plan the remove the redundant content in the README.md files and just put a link to the deployed docs once the deployment works

apps/docs/src/packages/headless.md
larsrickert: I think we should show the basic usage yes, esspecially because other projects will also use them (e.g. ODS) so they need to know which composables exist and how they are used :)

apps/docs/src/packages/headless.md
larsrickert: Yes but I don't know how Jonathan intended the composable to work :D I think its just a placeholder, thats why I added the "Work in progress" warning at the top

apps/docs/src/packages/storybook-utils.md
```diff
@@ -0,0 +1,106 @@
...
+
+# @sit-onyx/storybook-utils
+
+{{ packageJson.description }}.
+
+## Installation
+
+<!--
+Make sure that this chapter is kept up to date with installation steps in
+packages/headless/README.md file.
```
BoppLi, line 17:
```suggestion
packages/storybook-utils/README.md file.
```

apps/docs/src/packages/storybook-utils.md
```diff
@@ -0,0 +1,106 @@
...
+---
+
+<script lang="ts" setup>
+import packageJson from "../../../../packages/storybook-utils/package.json";
+</script>
+
+# @sit-onyx/storybook-utils
+
+{{ packageJson.description }}.
+
```
BoppLi, line 12: Should we add some info like
```suggestion
Required Storybook version 7.6 or higher

```

apps/docs/src/packages/storybook-utils.md
```diff
@@ -0,0 +1,106 @@
...
+};
+
+export default config;
+```
+
+:::
+
+### defineStorybookActionsAndVModels
+
+Utility to define Storybook meta for a given Vue component which will take care of defining argTypes for the given events as well as implementing v-model handlers so that the Storybook controls are updated when you interact with the component.
```
BoppLi, line 80: is the bug still present that the v-model does not synchronize when you are in a docs tab of storybook?

BoppLi (approved):

apps/docs/src/packages/storybook-utils.md
larsrickert: Fixed

apps/docs/src/packages/storybook-utils.md
larsrickert: I thinks its sufficient that its defined in the peerDependencies.
The package works with all versions of storybook >= 7

apps/docs/src/packages/storybook-utils.md
larsrickert: Yes although I am not sure if this is a bug or intended by Storybook because the v-model only works for the primary/first story.

I will add a follow-up task for this

BoppLi (approved):

PR #16, opened 2024-01-03
chore: setup workflow for updating Playwright screenshots

larsrickert:
The new workflow has to be triggered manually. It will then update all screenshots and create a PR targeting the selected branch.

Successfull test workflow run can be found [here](https://github.com/SchwarzIT/onyx/actions/runs/7394540946/job/20116073512).


packages/sit-onyx/playwright/snapshots/TestInput/default-chromium-linux.png
BoppLi, line 1: do we really need 3 screenshots for each component? :O and later lightxdark mode -> 6 per component? :O

CODEOWNERS
```diff
@@ -1 +1,6 @@
-* @JoCa96 @BoppLi @larsrickert
\ No newline at end of file
+# Global code owners (developers)
+* @JoCa96 @BoppLi @larsrickert
+
+# Define designers as code owners for screenshots to automatically
+# request them as reviewer if visual changes exist
+/packages/sit-onyx/playwright/snapshots/ @jannick-ux
```
BoppLi, line 6: will only UX be able to approve those PRs or we as well? should there be a policy that then requires 1 approval from each "team" or else requires a total of 2 approvals?

also, martin/till need to create their accounts so jannick is allowed to have vacation ever again :p

packages/sit-onyx/playwright.config.ts
```diff
@@ -8,6 +8,11 @@ import { fileURLToPath } from "node:url";
 export default defineConfig({
   testDir: "./src/components",
   snapshotDir: "./playwright/snapshots",
+  // custom snapshotPathTemplate to remove the testFileName folder that we don't want
+  snapshotPathTemplate: "{snapshotDir}/{testFileDir}/{arg}-{projectName}-{platform}{ext}",
+  // we don't want to update snapshots on the local machine of each developer.
+  // if you want to update snapshots for your branch, use the corresponding GitHub action.
```
BoppLi, line 14: can you pls create a small follow-up PR afterwards with the link to the github action that you create with this PR? ^^ #devsarelazy

packages/sit-onyx/src/components/TestInput/TestInput.vue
```diff
@@ -34,7 +34,7 @@ const handleChange = (event: Event) => {
 </script>
 
 <template>
-  <label>
+  <label class="input__wrapper">
```
BoppLi, line 37: as I did in my PR (which is still draft), BEM requires this class to be named

```suggestion
  <label class="input">
```

packages/sit-onyx/src/components/TestInput/TestInput.vue
```diff
@@ -45,6 +45,11 @@ const handleChange = (event: Event) => {
 
 <style lang="scss" scoped>
 .input {
+  &__wrapper {
```
BoppLi, line 48: see comment above.

BoppLi (changes requested):

packages/sit-onyx/playwright/snapshots/TestInput/default-chromium-linux.png
larsrickert: Currently we have configured 3 browsers (chromium, firefox and safari) thats why we have 3 screenshots.

BoppLi: ok, UX will have fun with the reviews then :D

CODEOWNERS
larsrickert: Good point, I added martin :) Yes all reviewers can approve so as we currently only require 1 reviewer, either a dev or ux could approve in this case

BoppLi (approved):

packages/sit-onyx/playwright.config.ts
larsrickert: Yes sure 😄

PR #69, opened 2024-01-05
docs(changeset): feat: add breakpoints / viewports

larsrickert:
Add Onyx breakpoints to Storybook so the user can easily select one of them to see how the components behave on different screen sizes.

### DIFF
--- a/packages/storybook-utils/src/theme.ts
+++ b/packages/storybook-utils/src/theme.ts
@@ -1,6 +1,6 @@
 /** All available Storybook breakpoints / viewports supported by Onyx. */
 export const ONYX_BREAKPOINTS = {
-  "2xsmall": {
+  xsmall2 : {
     name: "2xsmall",
     styles: {
       width: "320px",


jannick-ux (dismissed):

packages/storybook-utils/src/theme.ts
```diff
@@ -0,0 +1,45 @@
+/** All available Storybook breakpoints / viewports supported by Onyx. */
+export const ONYX_BREAKPOINTS = {
+  "2xsmall": {
```
BoppLi, line 3: [optional]
```suggestion
  xsmall2 : {
```

would UX also be ok with that? then we wouldn't have to wrap them in "" ^^

BoppLi (dismissed):

packages/storybook-utils/src/theme.ts
BoppLi: or xxsmall

larsrickert: The breakpoint names are defined by @jannick-ux in Figma: https://www.figma.com/file/YfEUBOHk4J4nYrk04geswG/onyx---Conception-%26-DEV-Exchange?type=design&node-id=14-698&mode=dev

jannick-ux: The names of the breakpoints have never been used with "" by myself. So having it without those signs is intended.

BoppLi (approved):

PR #73, opened 2024-01-08
refactor(storybook-utils): rename breakpoints

larsrickert:
Fix breakpoint name to match the key name which was updated by Jannick

### DIFF
--- a/.changeset/shiny-lies-cough.md
+++ b/.changeset/shiny-lies-cough.md
@@ -3,3 +3,12 @@
 ---
 
 refactor: rename breakpoints
+
+| Old     | New |
+| ------- | --- |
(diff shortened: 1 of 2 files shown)


jannick-ux (commented):
I think we can do
2xs - xs - sm - md - lg - xl - 2xl

But I dont want to name it xsmall2 - this is confusing

larsrickert:
> I think we can do 2xs - xs - sm - md - lg - xl - 2xl
> 
> But I dont want to name it xsmall2 - this is confusing

Alright, done. Can you also update the names in the Figma file?

BoppLi (approved):

jannick-ux (commented):
fine. thx

PR #74, opened 2024-01-08
feat(headless): create poc combobox composable

JoCa96:
Implemented "createComboBox" headless composable based on [WAI ARIA Combox Pattern][1] and [W3C ARIA Combobx Spec][2].

So far only `listbox` is supported as `aria-controls` for the combobox.

[1]: https://www.w3.org/WAI/ARIA/apg/patterns/combobox/
[2]: https://w3c.github.io/aria/#combobox


packages/headless/src/utils/builder.ts
```diff
@@ -0,0 +1,43 @@
+import { computed, type ComputedRef, type HtmlHTMLAttributes, type Ref } from "vue";
+
+export type IteratedHeadlessElementFunc<T extends Record<string, unknown>> = (
+  opts: T,
+) => HtmlHTMLAttributes;
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type HeadlessElementAttributes = HtmlHTMLAttributes | IteratedHeadlessElementFunc<any>;
```
larsrickert, line 8: Can we use `unknown` instead of `any`?

packages/headless/src/utils/builder.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export type HeadlessElements = Record<
+  string,
+  HeadlessElementAttributes | ComputedRef<HeadlessElementAttributes>
+>;
```
larsrickert, line 13:
```suggestion
export type HeadlessElements = Record<string, MaybeRef<HeadlessElementAttributes>>;
```

There is a build in vue type for this :)

packages/headless/src/composables/comboBox/createComboBox.ts
```diff
@@ -0,0 +1,119 @@
...
+  }) => {
+    const inputValue = ref("");
+    const inputValid = ref(true);
+    const controlsId = createId("comboBox-control");
+    const labelId = createId("comboBox-label");
+
+    const descendantKeyIdMap: Record<string, string | undefined> = {};
+
+    const getOptionId = (key: string) =>
+      descendantKeyIdMap[key] ?? (descendantKeyIdMap[key] = createId("comboBox-option"));
```
larsrickert, line 33: In the second part of the `??` is the element also returned? Because it is just an assignment

packages/headless/src/composables/comboBox/createComboBox.ts
```diff
@@ -0,0 +1,119 @@
...
+
+    const handleInput = (event: Event) => {
+      const inputElement = event.target as HTMLInputElement;
```
larsrickert, line 36: Should be check if `event.target instanceof HTMLInputElement` to catch cases where the composable is used incorrectly?

packages/headless/src/composables/comboBox/createComboBox.ts
```diff
@@ -0,0 +1,119 @@
...
+          onNext();
+          break;
+        case "Home":
+          onFirst();
+          break;
+        case "End":
+          onLast();
+          break;
+        default:
+          break;
```
larsrickert, line 74:
```suggestion
```

packages/headless/src/utils/builder.ts
```diff
@@ -0,0 +1,43 @@
...
+ *   option: computeIterated<{ key: string; label: string; disabled: boolean }>(
+ *     ({ key, label, disabled }) => ({
+ *       // Do something with the typed props
+ *     }),
+ * }
+ * ```
+ */
+export const computeIterated = <P extends Record<string, unknown>>(
+  iteratedFunc: IteratedHeadlessElementFunc<P>,
+) => computed(() => iteratedFunc);
```
larsrickert, line 43: Whats the benefit of this wrapper function? Cant we just use computed directly?:

packages/headless/src/composables/comboBox/createComboBox.ts
```diff
@@ -0,0 +1,119 @@
...
+            id: getOptionId(key),
+            "aria-selected": activeKey.value === key,
+            "aria-label": label,
+            "aria-disabled": disabled,
+            onClick: () => onSelect(key),
+          }),
+        ),
+        /** this is an input */
+        input: computed(() => ({
+          "v-bind:is": "input",
```
larsrickert, line 99:
```suggestion
          "is": "input",
```

Isn't this the same?

packages/headless/src/utils/builder.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export type HeadlessComposable<Elements extends HeadlessElements, State extends HeadlessState> = {
+  elements: Elements;
+  state: State;
+};
+
+/**
+ * We use this identity function to ensure the correct typings of the headless composables
+ */
+export const createBuilder = <P, Elements extends HeadlessElements, State extends HeadlessState>(
```
larsrickert, line 25: If this is only used for typing, can't we just define the type when creating the composable like so?:
```ts
export const useComboBox: HeadlessComposable = () => {
  // ...
}
```

larsrickert (changes requested):

packages/headless/src/utils/builder.ts
JoCa96: I want a readonly Ref, thats why I use ComputedRef

packages/headless/src/composables/comboBox/createComboBox.ts
JoCa96: Yes, it does both

JoCa96: The expression `(descendantKeyIdMap[key] = createId("comboBox-option"))` is executed and the result is returned.

packages/headless/src/utils/builder.ts
JoCa96: Unfortunately type inference does not work with assigned types.
With an assigned type we would need to to declare all types upfront:
```
type Props = {...};
type Elements = {...};
type State = {...};
export const useComboBox: HeadlessComposable<Props, Elements, State> = () => {
  // ...
}
```

packages/headless/src/utils/builder.ts
JoCa96: It removes the need for nested inline typings:

```ts
computed<
        IteratedHeadlessElementAttributes<{ key: string; label: string; disabled: boolean }>
      >(() => ({ key, label, disabled }) => ({
        role: "option",
        id: getOptionId(key),
        "aria-selected": activeKey.value === key,
        "aria-label": label,
        "aria-disabled": disabled,
        onClick: () => onSelect(key),
      }))
```

vs

```ts
computeIterated<{ key: string; label: string; disabled: boolean }>(
        ({ key, label, disabled }) => ({
          role: "option",
          id: getOptionId(key),
          "aria-selected": activeKey.value === key,
          "aria-label": label,
          "aria-disabled": disabled,
          onClick: () => onSelect(key),
        }),
      ),
```

Looks way better imo

packages/headless/src/composables/comboBox/createComboBox.ts
JoCa96: you are right

packages/headless/src/utils/builder.ts
larsrickert: You restricted the generic of the `IteratedHeadlessElementFunc` with extends to be a Record, so I think it should be `Record<string, unknown>` instead of any

packages/headless/src/utils/builder.ts
larsrickert: Okay, I just found it confusing the first time I looked at the code :)

packages/headless/src/utils/builder.ts
larsrickert: Optionally we could use satisfies like
```ts
export const useComboBox = () => {
  // ...
} satisfies HeadlessComposable;
```

Like that all props will be inferred correctly and we ensure the type.
But then we can probably also keep the createBuilder

packages/headless/src/utils/builder.ts
JoCa96: This would break the generic type inference

larsrickert (dismissed):

packages/headless/src/utils/builder.ts
JoCa96: @larsrickert ping

larsrickert: If it breaks the inference its fine like that :)

larsrickert (approved):

PR #75, opened 2024-01-08
feat(i18n): support i18n / translations

### LINKED ISSUE #71: Define concept for global translations
We want to have common translations included in the Vue library (e.g. German, English) etc. for common texts like "Okay", "Cancel" etc.

### Acceptance criteria
- a general concept is defined how translations are implement (e.g. provide/inject, json files etc.)
- all current translations (if any) are migrated to the new concept
- consider also the support of variables and plural handling (e.g. a text like `{value} option(s) selected`)

larsrickert:
closes #71

## Description

Implement + document i18n concept which is basically the same as `vue-i18n` which means:
- We / the project can define `.json` files which includes the translations
- support for pluralization + value placeholders
- auto-completion for translation keys, so Typescript / build will fail if we e.g. delete/rename a translation but forgot to change it in the places where we actually use it (+ we get type error if we try to use a key that does not exist)

## Checklist


apps/docs/src/i18n/index.md
```diff
@@ -0,0 +1,115 @@
...
+import type { Translation } from "sit-onyx";
+
+const myCustomLanguage: Translation = {
+  // add your translations here...
+  someMessage: "Hello World",
+};
+```
+
+```json [When contributing to Onyx]
+// simply create a new json file for your language (e.g. "en-US")
```
BoppLi, line 108:
```suggestion
// simply create a new json file for your language (e.g. "en-GB")
```

maybe this makes it a bit more realistic bc we already have en-US ^^

BoppLi (changes requested):

apps/docs/src/i18n/index.md
```diff
@@ -0,0 +1,115 @@
...
+import { useI18n } from "vue-i18n";
+
+const { locale } = useI18n();
+
+provideI18n({
+  // The Onyx locale will be updated whenever your vue-i18n locale changes
+  locale,
+  // make sure that the key for each language is the same that you are using
+  // for your vue-i18n JSON files
+  messages: { "de-DE": deDE },
```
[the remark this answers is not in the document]
larsrickert: Currently not because you directly import the `.json` files so there is no index file for them. But the project should be able to achieve this with native Vite import like so:
```ts
const languages = import.meta.glob("sit-onyx/locales/*.json", { import: "default", eager: true });
```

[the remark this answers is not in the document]
BoppLi: can't we do just that for them inside a file in the /locales folder and export the full locales, in terms of DX? ^^

packages/sit-onyx/src/components/TestInput/TestInput.vue
```diff
@@ -1,16 +1,20 @@
 <script lang="ts">
 export const INPUT_TYPES = ["email", "number", "password", "search", "tel", "text", "url"] as const;
+export type InputType = (typeof INPUT_TYPES)[number];
+
+/**
+ * Input types that have a translation for their validation error message.
+ * Must be up to date with i18n/locales/en-US.json.
+ */
+const TRANSLATED_INPUT_TYPES = ["email", "number", "tel", "url"] as const satisfies InputType[];
```
BoppLi, line 9: didn't we have once an implementation (in the blueprint?) where we got the typing info (= which keys exist) directly out of a translation json file? or does this not work without async?

mayby this might be less manual synching in the long run.

but if it makes it too complicated it's also fine to keep it as-is :)

packages/sit-onyx/src/components/TestInput/TestInput.vue
```diff
@@ -45,8 +45,10 @@ export type TestInputProps = {
   maxLength?: number;
   /** For validation: The lower limit of a number value */
   min?: number;
+  /** Step if type is "number" */
```
BoppLi, line 48:
```suggestion
  /** Step size of a number value */
```

packages/sit-onyx/src/components/TestInput/TestInput.vue
```diff
@@ -71,29 +73,43 @@ const emit = defineEmits<{
   validityChange: [state: ValidityState];
 }>();
 
-const isTouched = ref(false);
+const { t } = injectI18n();
 
-const inputElement = ref<HTMLInputElement | null>(null);
+const value = computed({
```
BoppLi, line 78: isn't there a best practice to order the code e.g. 
1. all refs
2. all computed
3. ...
or was just something my previous project liked? ^^

packages/sit-onyx/src/components/TestInput/TestInput.vue
```diff
@@ -71,29 +73,43 @@ const emit = defineEmits<{
...
-});
 
-const value = computed({
-  get: () => props.modelValue,
-  set: (value) => emit("update:modelValue", value),
+  const errorType = getFirstInvalidType(validityState.value);
+  if (!errorType) return "";
+
+  // a custom error message always is considered first
+  if (props.errorMessage || errorType === "customError") return props.errorMessage;
```
BoppLi, line 94: to honor the comment's message, I'd actually put that first in the code aswell ;)
if it's invalid, the actual active error type should not matter. also, customError will only ever be possible when the projects would be able to call the [setCustomValidity](https://developer.mozilla.org/en-US/docs/Web/API/HTMLObjectElement/setCustomValidity) function on the inner input itself, is that even possible with our new component structure?
Additionally, setCustomValidity is messy because it immediately sets valid to false (side effects) and the projects have to reset it by hand. that's why I offered preferably just an error message override option. 
Also this should allow the projects to also ajdust something like a patternMismatch situationally.


```suggestion
  // a custom error message always is considered first
  if (props.errorMessage) return props.errorMessage;
  
  const errorType = getFirstInvalidType(validityState.value);
  if (!errorType) return "";
```

BoppLi: if you want we can have a call to clarfiy this topic :)

packages/sit-onyx/src/components/TestInput/TestInput.vue
```diff
@@ -71,29 +73,43 @@ const emit = defineEmits<{
...
+
+  // if the error is "typeMismatch", we will use an error message depending on the type property
+  if (errorType === "typeMismatch") {
+    const type = TRANSLATED_INPUT_TYPES.includes(props.type as TranslatedInputType)
+      ? (props.type as TranslatedInputType)
+      : "generic";
+    return t.value(`validations.typeMismatch.${type}`);
```
BoppLi, line 101: image

```suggestion
    return t.value(`validations.typeMismatch.${type}`, { value: value.value });
```

BoppLi (changes requested):

packages/sit-onyx/src/components/TestInput/TestInput.vue
larsrickert: In general, I'd say
1. types / props / emits
2. composables
3. refs
4. computed / functions

I moved it

packages/sit-onyx/src/components/TestInput/TestInput.vue
larsrickert: That will give a TypeScript issue since the errorType of the ValidityState can be "customError" where we don't have a translation so I am returning the custom error message in this case

packages/sit-onyx/src/components/TestInput/TestInput.vue
larsrickert: Found a nice way to do it :)

apps/docs/src/i18n/index.md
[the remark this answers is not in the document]
larsrickert: Is there a use case where a project will import ALL languages? I'd prefer to not use this because then you don't have tree-shaking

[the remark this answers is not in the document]
BoppLi: DIA e.g. already reached 28 language files, I know it will take some time until we have that much in onyx included but it's not unrealistic if the projects support :)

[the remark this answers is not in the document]
BoppLi: thx!

packages/sit-onyx/src/components/TestInput/TestInput.vue
BoppLi: I added it to [📜 #6 ](https://github.com/SchwarzIT/onyx/wiki/Technical-Vision-&-Guidelines#component-implementation)

BoppLi: maybe smth we can write a linter rule later ^^

packages/sit-onyx/src/components/TestInput/TestInput.vue
BoppLi: then I'd rather return an empty string as the existence of customError does not mean that the projects have passed over a props.errorMessage

apps/docs/src/i18n/index.md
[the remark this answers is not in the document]
BoppLi: alternatively, using a function with a parameter to pass over only the relevant languages might help?

packages/sit-onyx/src/index.ts
```diff
@@ -1,3 +1,8 @@
 import "@/styles/index.scss";
 
 export { default as TestInput } from "@/components/TestInput/TestInput.vue";
+export * from "@/types";
+
+// export everything except "injectI18n" because we don't want that the import is suggested
+// by projects using this library because they probably use "vue-i18n"
```
BoppLi, line 7:
```suggestion
// to projects using this library because they probably use "vue-i18n"
```

packages/sit-onyx/src/utils/forms.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * Extracts the first invalid validity type from the given HTML ValidityState.
+ * "valueMissing" is prioritized over other types to align with the default browser behavior.
+ */
+export const getFirstInvalidType = (validity: ValidityState) => {
+  // since the types are getters of the ValidityState we need to get the keys using "Object.getOwnPropertyDescriptors"
+  const availableValidityTypes = Object.entries(
+    Object.getOwnPropertyDescriptors(ValidityState.prototype),
+  )
+    .filter(([key, value]) => key !== "valid" && value.enumerable)
```
BoppLi, line 10:
```suggestion
    .filter(([key, value]) => key !== "valid")
```

isn't this redundant?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Enumerability_and_ownership_of_properties?retiredLocale=de
> Most iteration means (such as [for...in](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in) loops and [Object.keys](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys)) only visit enumerable keys.

BoppLi: same for Object.entries
> The Object.entries() static method returns an array of a given object's own **enumerable** string-keyed property key-value pairs.

packages/sit-onyx/src/utils/forms.ts
```diff
@@ -0,0 +1,24 @@
...
+    .map(([key]) => key) as Exclude<keyof ValidityState, "valid">[];
+
+  // sort valueMissing first to align with the default browser behavior
+  availableValidityTypes.sort((a, b) => {
+    if (a === "valueMissing") return -1;
+    if (b === "valueMissing") return 1;
+    return 0;
+  });
```
BoppLi, line 18: technically this could simplify it

```suggestion
  // the error for an empty "required" value gets top priority and is returned first
  if (validity.valueMissing) return "valueMissing";

  // since the types are getters of the ValidityState we need to get the keys using "Object.getOwnPropertyDescriptors"
  const availableValidityTypes = Object.entries(
    Object.getOwnPropertyDescriptors(ValidityState.prototype),
  )
    .filter(([key, value]) => key !== "valid" && value.enumerable)
    .map(([key]) => key) as Exclude<keyof ValidityState, "valid">[];
```

BoppLi (changes requested):

packages/sit-onyx/src/i18n/index.ts
```diff
@@ -0,0 +1,135 @@
...
+    return (
+      key: ObjectToDottedStrings<Translation>,
+      placeholders: Record<string, string | number | undefined> = {},
+    ): string => {
+      // use English message as fallback
+      let message = resolveMessage(key, messages.value) ?? resolveMessage(key, enUS) ?? "";
+
+      if (placeholders) {
+        Object.entries(placeholders).forEach(([key, value]) => {
+          if (value === undefined) return;
```
BoppLi, line 71: didn't you say for undefined placeholder values, a fallback was planned -> to show the key instead? like that it would still show e.g. {value}, like we saw in the screenshot from an older comment in this PR.

Or is this intended in hopes the dev might notice something is wrong?

BoppLi: ah, the case in the older comment was that no placeholders were passed, so it's a different issue. 

but I suppose the behavior would be the same here if you pass e.g. undefined for "value"

BoppLi: I was referring to https://github.com/SchwarzIT/onyx/pull/75/files/f9ea0c2113dd710d03fbeae1bb3392fc4ea58c99#r1450141113

packages/sit-onyx/src/i18n/index.ts
```diff
@@ -0,0 +1,135 @@
...
+          // "gi" is used to replace all occurrences because String.replaceAll() is not available
```
BoppLi, line 72: I don't understand
image

but inside `packages/sit-onyx/tsconfig.vitest.json` ES2021 is already specified?
image

BoppLi: I mean your solution is also totally fine, I just don't get why this setting seems to be ignored

packages/sit-onyx/src/i18n/index.ts
BoppLi: maybe the fallback could be just `""` or just an empty string ¯\_(ツ)_/¯

packages/sit-onyx/src/i18n/index.ts
```diff
@@ -0,0 +1,135 @@
...
+      return resolvePluralization(message, pluralizationValue);
+    };
+  });
+
+  return { locale, t };
+};
+
+/**
+ * Provides a global i18n instance that is used by Onyx.
+ * Must only be called once in the `App.vue` file.
```
BoppLi, line 88:
```suggestion
 * Must only be called once in the `App.vue` file of a project that consumes Onyx.
```
maybe? onyx itself does not have one, might be misleading for contributors ¯\_(ツ)_/¯

BoppLi: btw, how are we even able to inject i18n in onyx wihtout ever providing it 😵‍💫
looking forward to trying it out in a sample project ^^

packages/sit-onyx/src/i18n/index.ts
```diff
@@ -0,0 +1,135 @@
...
+export const provideI18n = (options: ProvideI18nOptions) => {
+  provide(I18N_INJECTION_KEY, createI18n(options));
+};
+
+/**
+ * Injects the Onyx i18n instance.
+ */
+export const injectI18n = () => {
+  const fallbackValue = createI18n();
+  return inject(I18N_INJECTION_KEY, fallbackValue);
```
BoppLi, line 99: why storing it, is this important?
```suggestion
  return inject(I18N_INJECTION_KEY, createI18n());
```

packages/sit-onyx/src/i18n/index.ts
```diff
@@ -0,0 +1,135 @@
...
+ * Resolves the given dotted key (e.g. `a.b.c`) to the translation value of the given messages.
+ * @returns Message value or undefined if translation does not exist.
+ */
+const resolveMessage = (
+  key: ObjectToDottedStrings<Translation>,
+  messages: DeepPartial<Translation>,
+): string | undefined => {
+  // see https://stackoverflow.com/a/6394168
+  const message = key.split(".").reduce<TranslationValue | undefined>((obj, i) => {
+    if (!obj || typeof obj === "string") return obj;
```
BoppLi, line 112: I tried to dig deeper to understand this better, but could never enter this condition when trying it with our "required" test input.. how would this be possible?

```
  console.log("## 1 params", key, messages);
  console.log("## 2 split", key.split("."));
  // see https://stackoverflow.com/a/6394168
  const message = key.split(".").reduce<TranslationValue | undefined>((obj, i) => {
    console.log("## 3 red iteration with", i, "obj is", obj);
    if (!obj || typeof obj === "string") {
      console.log("## 4 result is", obj);
      return obj;
    }
    console.log("## 5 iteration result is", obj[i]);
    return obj[i];
  }, messages);

  console.log("## 6 original messages is", messages);
  return message && typeof message === "string" ? message : undefined;
```
Example: ## 4 was never logged

(fyi, imho a unit test for this function might be helpful and also maybe moving it to a util because this is not just for translations but to traverse into a given object by the given key path)

packages/sit-onyx/src/i18n/index.ts
```diff
@@ -0,0 +1,135 @@
...
+
+/**
+ * Returns the pluralized message (if multiple formats are defined using `" | "` separator.
+ */
+const resolvePluralization = (message: string, value?: number) => {
+  const formats = message.split(" | ").map((i) => i.trim());
+  if (formats.length <= 1) return message;
+
+  let pluralization: 0 | 1 | 2 = 1;
+  if (value === 0) pluralization = 0;
```
BoppLi, line 127:
```suggestion
  if (!value || value === 0) pluralization = 0;
```

right now an empty value would stay with "1", is this intended?

BoppLi (changes requested):

packages/sit-onyx/src/i18n/index.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+
+test("should translate with pluralization", () => {
+  provideI18n({
+    locale: "en-US",
+    messages: {
+      "en-US": {
+        pluralization: "Zero items | 1 item | {n} items",
+        pluralizationWithoutZero: "1 item | {n} items",
+        withoutPluralization: "{n} items",
+      },
```
BoppLi, line 86: didn't you want to add a test case for this? :)

```suggestion
        noPluralizationIntended: 'The pipe "|" is part of the text',
      },
```
...
```
  message = t.value("noPluralizationIntended" as TestTranslationKey, { n: 0 });
  expect(message).toBe('The pipe "|" is part of the text');
```

packages/sit-onyx/src/i18n/index.ts
BoppLi: I see the test case, but imho, an empty value is as good as 0 only. we would affect the data integrity by staying with 1 here.

A fancy thing we could do is use pluralization 0 for undefined but writing the "halbgeviertstrich", e.g. in your test:
```
  message = t.value("pluralization" as TestTranslationKey);
  expect(message).toBe("– items");
```
It's UXes trusted friend for these situations 🙃

packages/sit-onyx/src/i18n/index.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+  expect(message).toBe("1 item");
+
+  message = t.value("pluralization" as TestTranslationKey, { n: 2 });
+  expect(message).toBe("2 items");
+
+  message = t.value("pluralization" as TestTranslationKey, { n: -1 });
+  expect(message).toBe("-1 items");
+
+  message = t.value("pluralization" as TestTranslationKey);
+  expect(message).toBe("1 item");
```
BoppLi, line 105: please refer to my other comment in index.ts for this test case, imho this is not the expected result.

packages/sit-onyx/src/i18n/index.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+  });
+  const { t } = injectI18n();
+
+  let message = t.value("helloWorld" as TestTranslationKey);
+  expect(message).toBe("Hello World");
+
+  locale.value = "de-DE";
+
+  message = t.value("helloWorld" as TestTranslationKey);
+  expect(message).toBe("Hallo Welt");
```
BoppLi, line 147: wouldn't something like this reflect the expectation/real-world-scenario better?

```suggestion
  const message = vue.computed(()=> t.value("helloWorld" as TestTranslationKey));

  expect(message.value).toBe("Hello World");
  locale.value = "de-DE";
  expect(message.value).toBe("Hallo Welt");
```

BoppLi (changes requested):
All in all a very interesting PR, thx! finally fully through 🥳

apps/docs/src/i18n/index.md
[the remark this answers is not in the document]
larsrickert: A function would still not have tree shaking because Vite can not statically analyse it.
If so many locales are imported, we should support lazy loading them anyways.
I'd like to delay this feature until we have so many build-in locales that it makes sense to implement it :)

packages/sit-onyx/src/utils/forms.ts
larsrickert: Does not seem so because keys like `constructor` are still included
image

packages/sit-onyx/src/types/index.ts
```diff
@@ -0,0 +1,76 @@
...
+ */
+export type DeepPartial<T> = T extends object ? { [P in keyof T]?: DeepPartial<T[P]> } : T;
+
+/**
+ * Gets a union type from an object that contains all combinations of nested keys as arrays.
+ *
+ * @see https://stackoverflow.com/a/47058976
+ * @example
+ * ```ts
+ * type Test = PathsToStringProps<{ a: "foo"; b: { c: "bar"; d: "baz" } }>;
```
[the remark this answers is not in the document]
larsrickert: Yes :) Thanks!

packages/sit-onyx/src/types/index.ts
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * // nested value
+ * {
+ *   someKey: {
+ *     someOtherKey: "Hello World"
+ *   }
+ * }
+ * ```
+ */
+export type TranslationValue = string | { [key: string]: string | NestedMessage };
```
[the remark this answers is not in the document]
larsrickert: Yes I think I had some weird TypeScript error before but it seems fine now

packages/sit-onyx/src/i18n/index.ts
larsrickert: Because the `tsconfig.vitest.json` is used for tests/type checking. For the actual build/component library the `tsconfig.app.json` is used :) It was probably a leftover, I removed it

packages/sit-onyx/src/i18n/index.ts
larsrickert: Nope, I just created a variable so that I can name it `fallbackValue` to make clearer that it is only a fallback and not the actual value

packages/sit-onyx/src/i18n/index.ts
```diff
@@ -0,0 +1,135 @@
...
+   * import deDE from "sit-onyx/locales/de-DE.json";
+   * {
+   *   messages: {
+   *     // English is always supported so we don't need to add it here
+   *     'de-DE': deDE
+   *   }
+   * }
+   * ```
+   */
+  messages?: Record<string, DeepPartial<Translation> | Record<string, TranslationValue>>;
```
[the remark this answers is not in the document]
larsrickert: I needed this for the unit tests. I removed it

apps/docs/src/i18n/index.md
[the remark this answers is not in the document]
BoppLi: ok, should we create a follow-up ticket or just keep it in mind? ^^

packages/sit-onyx/src/i18n/index.ts
larsrickert: I wanted to align it with `vue-i18n` and there you will get 1 if n is undefined :)

packages/sit-onyx/src/i18n/index.spec.ts
larsrickert: See https://github.com/SchwarzIT/onyx/pull/75#discussion_r1452051406

packages/sit-onyx/src/components/TestInput/TestInput.vue
BoppLi: There are two approaches
1) we watch the props.errorMessage and call setCustomValidity() with the value -> when the errorMessage is reset, we would automatically reset the error state when calling setCustomValidity with an empty string :)
But the down side is, that the previous workflow does not work anymore which I showcased in the FormDemo component. the workflow was
- the input got a pattern (lowercase characters only) assigned
- when the user typed in an uppercase character, the validityChange Event fired and the patternMismatch was true
- the dev was able to detect that and set customErrorMessage that explained which pattern is expected, unlike our generic standard "your pattern is not ok" message
- now with the above behavior, it would kill itself - when setting errorMessage, the TestInput fires setCustomValidity("You need to use lowercase only") and this fires the validityChanged event again. our condition in the FormElement now sees that patternMismatch is not the active error anymore and resets the errorMessage?
I guess we would need to check if patternMismatch is still kept even though customError is set. if that's the case, no problem exists and we could implement it like that.
Should I try it out on your branch with my FormElement component from my older branch?

2) could we offer setCustomValidity as a method that can be called on the TestInput component?

packages/sit-onyx/src/i18n/index.spec.ts
larsrickert: Yes, like it :)

packages/sit-onyx/src/i18n/index.ts
larsrickert: Yes the placeholders are kept as seen here https://github.com/SchwarzIT/onyx/pull/75#discussion_r1450713080

Or did I got your comment wrong?

larsrickert: If the translation is `Hello {firstName}` and you don't pass a value for firstName, the resulting string should be
`Hello {firstName}`

packages/sit-onyx/src/i18n/index.ts
BoppLi: aah, right 🤦🏻‍♀️

packages/sit-onyx/src/i18n/index.ts
BoppLi: ok

BoppLi:
```suggestion
  const fallbackInstance = createI18n();
  return inject(I18N_INJECTION_KEY, fallbackInstance);
```

apps/docs/src/i18n/index.md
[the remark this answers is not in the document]
larsrickert: I created a draft ticket, but we probably need a few months before we get to that point^^
https://github.com/orgs/SchwarzIT/projects/5/views/7?pane=issue&itemId=49970134

packages/sit-onyx/src/components/TestInput/TestInput.vue
larsrickert: Should we delay this feature then? I think its out of scope of this PR. Maybe we should remove the custom error message for now and add it later when we defined a concept for it? What do you think?

packages/sit-onyx/src/i18n/index.ts
larsrickert: Finding: `vue-i18n` removes empty placeholders so we should align and do the same

packages/sit-onyx/src/components/TestInput/TestInput.vue
BoppLi: created follow-up ticket #113

packages/sit-onyx/src/i18n/index.spec.ts
```diff
@@ -75,11 +75,12 @@ test("should translate with/without placeholders", () => {
   message = t.value("does.not.exist" as TestTranslationKey);
   expect(message).toBe("");
 
-  // shows the original placeholders if no values were provided
+  // removes the original placeholders if no values were provided (same behavior as "vue-i18n")
   message = t.value("placeholder" as TestTranslationKey, {
     firstName: undefined,
+    lastName: undefined,
```
BoppLi, line 81: I left out lastName in this example intentionally because this is what could happen and it should behave still the same?

packages/sit-onyx/src/i18n/index.ts
BoppLi: did not expect that, but good to know :O better not forget it then ^^

BoppLi (approved):
awesome, thank you!

PR #98, opened 2024-01-09
docs: fix missing external link icons

larsrickert:
## Description

Update VitePress to add missing icon for external link, see [VitePress changelog](https://github.com/vuejs/vitepress/blob/main/CHANGELOG.md#100-rc36-2024-1-8).

## Checklist

### DIFF
--- a/apps/docs/src/getting-started.md
+++ b/apps/docs/src/getting-started.md
@@ -4,7 +4,7 @@
 This library is currently in early / active development.
 :::
 
-<div class="external-link--hide">
+<div class="hide-external-link">
 
 [![npm version](https://badge.fury.io/js/sit-onyx.svg)](https://www.npmjs.com/package/sit-onyx)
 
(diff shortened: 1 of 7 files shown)


apps/docs/src/getting-started.md
```diff
@@ -4,8 +4,12 @@
 This library is currently in early / active development.
 :::
 
+<div class="external-link--hide">
```
BoppLi, line 7: strict BEM would need
```suggestion
<div class="external-link external-link--hide">
```

BoppLi (approved):
I actually don't see a difference to before, is it because I use chrome?
(yes I ran pnpm i, but I even could see the badges before just fine)
image

larsrickert:
> I actually don't see a difference to before

The difference is in the page content, not the badges

Before:
image

After:
image

For the npm badges, the div with the extra class is needed because they are technically also links but would look like this:
image

apps/docs/src/getting-started.md
larsrickert: I renamed it to `hide-external-link` so no BEM is needed :D

BoppLi: ok :)

BoppLi (approved):

PR #120, opened 2024-01-15
chore: run deployment after release

larsrickert:
## Description

Run the deployments after the release because we import e.g. package versions (or changelogs in the future) so we need to make sure to deploy after the release has updated the version/changelog.

## Checklist


.github/workflows/release.yml
```diff
@@ -4,6 +4,12 @@ on:
     branches:
       - "main"
   workflow_dispatch:
+    inputs:
+      stage:
+        description: Stage to deploy to
+        type: environment
+        required: true
+        default: dev
```
JoCa96, line 12: While we are in alpha we can deploy to `prod` stage by default. For none main branch triggers we can use the snapshot release and deploy to dev.

After the 1.0.0 release we will change that logic.

What do you think?

larsrickert: So should we remove the "stage" parameter and set it to "prod" if on main and "dev" otherwise?
I guess for now we probably only need one stage, right? So either prod or dev

JoCa96: I would say `prod` for main branch and `dev` for others.

larsrickert: Done :)

JoCa96 (approved):

PR #122, opened 2024-01-15
feat: Create dependabot.yml

JoCa96:
Added and configured `dependabot.yml` to enable dependabot.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "npm" # See documentation for possible values
+    directory: "/" # Location of package manifests
+    versioning-strategy: "increase"
+    schedule:
+      interval: "weekly"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,12 +1,10 @@
-# To get started with Dependabot version updates, you'll need to specify which
-# package ecosystems to update and where the package manifests are located.
 # Please see the documentation for all configuration options:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/"
```
larsrickert, line 7: Does this also include sub directories like `packages/sit-onyx`?

.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
...
+    versioning-strategy: "increase"
```
larsrickert, line 8: Seems like this will also update major versions / packages with breaking changes so we should check the dependabot PRs whether our users are affected by this breaking change so we might not update it yet

image
image
image

larsrickert: Maybe we can use the ignore property for this?: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#ignore

I am also fine if we leave it like this and check how it works in the future

.github/dependabot.yml
JoCa96: I didn't find any definitiv answer about this. Dependabot supports pnpm and the `pnpm-lock.yaml` file, so I think so. We will need to test it to find out.

.github/dependabot.yml
JoCa96: Dependabot creates PRs, so it will build and test. If we don't want an update we can configure it: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#ignore

JoCa96: But generally we SHOULD NOT ignore version udpates

larsrickert: Okay lets see :)

larsrickert (approved):

PR #133, opened 2024-01-16
feat: update dependabot to group updates for minor and patch, but not major

JoCa96:


.github/dependabot.yml
```diff
@@ -7,8 +7,12 @@ updates:
     directory: "/"
     versioning-strategy: "increase"
     schedule:
-      interval: "weekly"
+      interval: "daily"
```
larsrickert, line 10: Do we really want to get daily PRs for dependencies?^^

JoCa96: Can you extend on that? :)

larsrickert: Sure, I am worried that we will have a PR every day then that will only contain 1 or 2 updated versions which might lead to "noise" inside our PRs with irrelevant changes.

I am fine with trying it like this and if it gets annoying we can change in the future^^

larsrickert (approved):

PR #147, opened 2024-01-17
feat(figma-utils): sort variables when parsing

larsrickert:
## Description

Sort variables by:
1. alphabetically
2. if two variables only differ in a number (e.g. `color-100` and `color-1000`), sort them numerically so that 100 is sorted before 1000

## Checklist


packages/figma-utils/src/utils/parse.ts
```diff
@@ -61,6 +61,25 @@ export const parseFigmaVariables = (
...
+    // sort them by number instead of alphabetically so e.g. 100 is sorted before 1000
+    data.variables = Object.keys(data.variables)
+      .sort((a, b) => {
+        const aAsNumber = numberRegex.exec(a)?.[0] ?? "";
+        const bAsNumber = numberRegex.exec(b)?.[0] ?? "";
+        const isSameBaseName = a.replace(aAsNumber, "") === b.replace(bAsNumber, "");
+
+        if (aAsNumber && bAsNumber && isSameBaseName) return +aAsNumber - +bAsNumber;
+        return a.localeCompare(b);
+      })
```
BoppLi, line 78: if we have many, we would do the transformation operation many many times. suggestion:
```suggestion
      .map((key) => {
        const counter = numberRegex.exec(key)?.[0] ?? "";
        return {
          key,
          counter,
          base: key.replace(counter, ""),
        };
      })
      .sort((a, b) => {
        const isSameBaseName = a.base === b.base;

        if (a.counter && b.counter && isSameBaseName) return +a.counter - +b.counter;
        return a.key.localeCompare(b.key);
      })
      .map(bundle => bundle.key)
```

packages/figma-utils/src/utils/parse.ts
```diff
@@ -61,6 +61,25 @@ export const parseFigmaVariables = (
...
     if (data.modeName === DEFAULT_MODE_NAME) delete data.modeName;
+
+    const numberRegex = /\d+/;
+
+    // sort variables by name
+    // for variables with the same name that just end with a different number (e.g. my-var-100 and my-var-200)
+    // sort them by number instead of alphabetically so e.g. 100 is sorted before 1000
+    data.variables = Object.keys(data.variables)
+      .sort((a, b) => {
+        const aAsNumber = numberRegex.exec(a)?.[0] ?? "";
```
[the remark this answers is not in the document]
larsrickert: Imo this should also be sorted numericaly, so e.g. "foo-200-bar", "foo-100-bar" should be sorted as
1. foo-100-bar
2. foo-200-bar

packages/figma-utils/src/utils/parse.ts
larsrickert: Done

BoppLi: Nice!!! I like your change! 😊

BoppLi (approved):

PR #151, opened 2024-01-18
feat: link invalid to custom errorMessage

### LINKED ISSUE #113: Follow-up: custom error validation
depends on merge of #75

context, see https://github.com/SchwarzIT/onyx/pull/75#discussion_r1452056083

-> make the custom error example work properly

BoppLi:
closes #113 

## Description

By linking the errorMessage prop to the setCustomValidity function of the input element, the devs that use the component can trigger an invalid state by their own validation logic whenever needed.
The validityChange event includes all information of invalid states, so when the projects e.g. want to provide a specific error message on patternMismatch, both patternMismatch and customError will turn true in the ValidityState.

## Checklist


packages/sit-onyx/src/components/TestInput/TestInput.vue
```diff
@@ -133,6 +133,17 @@ watch(
...
   { deep: true },
 );
+
+watch(
+  [props.errorMessage, inputElement],
+  () => {
+    if (!inputElement.value) return;
+    // by using setCustomValidity, the ValidityState will turn invalid
+    // as long as it's not an empty string
+    inputElement.value.setCustomValidity(props.errorMessage || "");
```
larsrickert, line 143:
```suggestion
    if (!inputElement.value || !props.errorMessage) return;
    // by using setCustomValidity, the ValidityState will turn invalid
    // as long as it's not an empty string
    inputElement.value.setCustomValidity(props.errorMessage);
```

The custom error should only be set if a message exists, right?

packages/sit-onyx/src/components/TestInput/TestInput.vue
```diff
@@ -133,6 +133,17 @@ watch(
...
+  },
+  { deep: true },
```
larsrickert, line 145:
```suggestion
  { immediate: true },
```

Why is deep needed? Should be immediately show the custom error?

larsrickert (changes requested):

packages/sit-onyx/src/components/TestInput/TestInput.vue
BoppLi: it's important that it is reset again when props.errorMessage is removed / reset to empty string 

that change would be a problem because

packages/sit-onyx/src/components/TestInput/TestInput.vue
BoppLi: whoopsie, good catch, thx!

packages/sit-onyx/src/components/TestInput/TestInput.vue
larsrickert: Ah okay so calling `setCustomValidity` with an empty string will reset the error and make the form valid again?

BoppLi: yep :) as long as no other error is still active

larsrickert: Nice 👍

larsrickert (approved):

PR #154, opened 2024-01-18
fix(i18n): do not reset i18n when provided

BoppLi:
## Description

Ensure that the i18n instance is not reset when it has been provided.

## Checklist


packages/sit-onyx/src/i18n/index.ts
larsrickert, line 1: PR title: Its more a fix than a feat, right?
Can we also add a changeset? :)

larsrickert (changes requested):

larsrickert (approved):

larsrickert (approved):

PR #157, opened 2024-01-19
feat: create alpha-test-app

### LINKED ISSUE #39: Research forms and their validation

BoppLi:
closes #39

## Description

During the alpha phase of onyx, we can make use of a bare vitest starter app to try out onyx components in a user-near environment and detect problems that Storybook can not show.
The alpha-test-app will be removed once onyx leaves the alpha phase. As it is temporary, it will not be mentioned in our documentation app.

## Checklist


apps/alpha-test-app/README.md
```diff
@@ -0,0 +1,18 @@
+# Vue 3 + TypeScript + Vite
+
+This template should help get you started developing with Vue 3 and TypeScript in Vite. The template uses Vue 3 `<script setup>` SFCs, check out the [script setup docs](https://v3.vuejs.org/api/sfc-script-setup.html#sfc-script-setup) to learn more.
```
BoppLi, line 3: todo @myself: write short info what this app is for (see PR description)

apps/alpha-test-app/index.html
```diff
@@ -0,0 +1,13 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <link rel="icon" type="image/svg+xml" href="/vite.svg" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Vite + Vue + TS</title>
```
BoppLi, line 7: todo

apps/alpha-test-app/src/App.vue
```diff
@@ -0,0 +1,33 @@
+<script setup lang="ts">
+import { provideI18n } from "sit-onyx";
+import deDE from "sit-onyx/locales/de-DE.json";
+import { useI18n } from "vue-i18n";
+import HelloWorld from "./components/HelloWorld.vue";
+
+const { locale } = useI18n();
+provideI18n({ locale, messages: { "de-DE": deDE } });
```
BoppLi, line 8: @larsrickert we have to take a look, this gives errors in the IDE that did not happen with the blueprint setup 
image

apps/alpha-test-app/src/App.vue
```diff
@@ -0,0 +1,33 @@
...
+  <div>
+    {{ locale }}
+    <button @click="locale = 'de-DE'">Deutsch</button>
+    <button @click="locale = 'en-US'">English</button>
+  </div>
+  <HelloWorld msg="Vite + Vue" />
+</template>
+
+<style scoped>
+.logo {
```
BoppLi, line 21: todo cleanup

apps/alpha-test-app/src/main.ts
```diff
@@ -0,0 +1,17 @@
...
+import enUS from "./i18n/locales/en-US.json";
+import deDE from "./i18n/locales/de-DE.json";
+
+const i18n = createI18n({
+  legacy: false,
+  locale: "en-US",
+  messages: { "en-US": enUS, "de-DE": deDE },
+});
+
+// await import(`./locales/${locale}.json`);
```
BoppLi, line 15: todo cleanup

apps/alpha-test-app/src/style.css
```diff
@@ -0,0 +1,13 @@
+:root {
```
BoppLi, line 1: @larsrickert I removed most of the styles that were created by the starter app, should I kick everything except the #app style? what do you think

packages/sit-onyx/src/components/TestInput/TestInput.vue
```diff
@@ -122,20 +122,23 @@ const handleChange = (event: Event) => {
 watch([value, inputElement], () => {
   // update validity state when value changes
   if (!inputElement.value) return;
-  validityState.value = inputElement.value.validity;
+  const newValidityState = getValidityStateProperties().reduce<
```
BoppLi, line 125: todo write a comment to explain what's going on

apps/alpha-test-app/src/App.vue
BoppLi, line 9: image
@larsrickert is this error known to you?

BoppLi: solved it by changing the package linking style

BoppLi: deDE as OnyxTranslations

-> typecast solves issues.

Should we add this to the docs that this is needed? :/

apps/alpha-test-app/package.json
```diff
@@ -0,0 +1,22 @@
...
+  "private": true,
+  "version": "0.0.0",
+  "type": "module",
+  "scripts": {
+    "dev": "vite",
+    "build": "vue-tsc && vite build",
+    "preview": "vite preview"
+  },
+  "dependencies": {
+    "sit-onyx": "file:../../packages/sit-onyx",
```
BoppLi, line 12: fyi @larsrickert , when I only used `pnpm link ../../packages/sit-onyx`, the package.json was not modified so I worried that the next dev that wants to install will not get sit-onyx in here at all. that's why I added this line.
There as also alternatively that syntax
```suggestion
    "sit-onyx": "link:../../packages/sit-onyx",
```
but then I got that weird error in https://github.com/SchwarzIT/onyx/pull/157#discussion_r1458850251

I just hope this way it works fine. Could you try to get it running on your laptop if it works fine? :)

apps/alpha-test-app/.gitignore
larsrickert, line 1: I think we can delete this file because we already have a .gitignore in our root folder

apps/alpha-test-app/.vscode/extensions.json
larsrickert, line 1: We already have this file in our root folder so we can delete it here

apps/alpha-test-app/README.md
```diff
@@ -0,0 +1,27 @@
+# Alpha Test App
+
+During the alpha phase of onyx, we use this minimalistic app to try out onyx components in a user-near environment and detect problems that Storybook can not show.
+This app will be removed once onyx leaves the alpha phase. As it is temporary, it will not be mentioned in our documentation app.
+
+**Important**: To test changes in onyx, make sure to run `pnpm build` in `packages/sit-onyx/package.json`. Only then will the changes be reflected in the Alpha Test App.
```
larsrickert, line 6: Alternatively we could just change the start command to build sit-onyx before starting the app

apps/alpha-test-app/package.json
larsrickert:
```suggestion
    "sit-onyx": "workspace:^",
```

Since we are in the monorepo we can use the `workspace` command from pnpm to link the package :)

apps/alpha-test-app/package.json
```diff
@@ -0,0 +1,22 @@
...
+    "vue": "^3.4.15",
+    "vue-i18n": "^9.8.0"
+  },
+  "devDependencies": {
+    "@vitejs/plugin-vue": "^5.0.3",
+    "typescript": "^5.3.3",
+    "vite": "^5.0.11",
+    "vue-tsc": "^1.8.27"
```
larsrickert, line 20: All dependencies except vue-i18n can be removed because we already have them installed in our root package.json

apps/alpha-test-app/public/vite.svg
larsrickert, line 1: Should we keep the alpha app minimal and remove this logo?

apps/alpha-test-app/src/App.vue
```diff
@@ -0,0 +1,18 @@
+<script setup lang="ts">
+import { OnyxTranslations, provideI18n } from "sit-onyx";
+import deDE from "sit-onyx/locales/de-DE.json";
+import { useI18n } from "vue-i18n";
+import HelloWorld from "./components/HelloWorld.vue";
+
+const { t, locale } = useI18n();
+provideI18n({ locale, messages: { "de-DE": deDE as OnyxTranslations } });
```
larsrickert, line 8: Mmh this type cast should not be necessary...

apps/alpha-test-app/public/vite.svg
larsrickert: Same for `assets/vue.svg`

apps/alpha-test-app/src/components/molecules/FormDemoMolecule.vue
```diff
@@ -0,0 +1,102 @@
+<script lang="ts" setup>
+import { TestInput } from "sit-onyx";
+import { ref, watch } from "vue";
+
+type Props = {
```
larsrickert, line 5:
```suggestion
type FormData = {
```

The type is only for a single prop, not for all :D

apps/alpha-test-app/src/components/molecules/FormDemoMolecule.vue
```diff
@@ -0,0 +1,102 @@
...
+  minlengthInput: string;
+  maxInput: number;
+  typeInput: string;
+  patternInput: string;
+};
+const props = defineProps<{
+  formData: Props;
+}>();
+
+const formState = ref<Props | undefined>();
```
larsrickert, line 17:
```suggestion
const formState = ref<Props>();
```

apps/alpha-test-app/src/components/molecules/FormDemoMolecule.vue
```diff
@@ -0,0 +1,102 @@
...
+
+const emit = defineEmits<{
+  submit: [data: object];
```
larsrickert, line 20:
```suggestion
  submit: [data: Props];
```

apps/alpha-test-app/src/components/molecules/FormDemoMolecule.vue
```diff
@@ -0,0 +1,102 @@
...
+  props,
+  () => {
+    formState.value = { ...props.formData };
+  },
+  { immediate: true },
+);
+</script>
+
+<template>
+  <form v-if="formState" ref="formElement" class="demo" @submit="onSubmit($event)">
```
larsrickert, line 48:
```suggestion
  <form v-if="formState" ref="formElement" class="demo" @submit.prevent="emit('submit', props)">
```

(optional) Same behaviour + we could get rid of the `onSubmit` function :)

apps/alpha-test-app/tsconfig.json
larsrickert, line 1: I'd like to use the same base tsconfig as for the other packages :)

```json
{
  "extends": "@vue/tsconfig/tsconfig.dom.json",
  "include": ["src/**/*"],
  "compilerOptions": {
    "rootDir": "./src",
    "baseUrl": ".",
  }
}
```

apps/alpha-test-app/tsconfig.node.json
larsrickert, line 1: I think we can remove this file

packages/sit-onyx/src/components/TestInput/TestInput.vue
```diff
@@ -120,22 +120,17 @@ const handleChange = (event: Event) => {
 };
 
 watch([value, inputElement], () => {
-  // update validity state when value changes
   if (!inputElement.value) return;
-  validityState.value = inputElement.value.validity;
+  const newValidityState = transformValidityStateToObject(inputElement.value!.validity);
+  // only update + emit the validity state when value changes
+  if (JSON.stringify(newValidityState) !== JSON.stringify(validityState.value)) {
```
larsrickert, line 126: If I remember JSON.stringify is not be most reliable way to compare objects.
For the TetInput its fine for me to do it like that but we should change it for our real components

packages/sit-onyx/src/utils/forms.ts
```diff
@@ -1,3 +1,23 @@
+const getValidityStateProperties = () =>
+  Object.entries(Object.getOwnPropertyDescriptors(ValidityState.prototype))
+    .filter(([_, value]) => value.enumerable)
+    .map(([key]) => key) as (keyof ValidityState)[];
+
+/**
+ * The standard HTML ValidityState is an object with getter properties.
+ * This makes it hard to compare changes. With below function, we iterate through all
+ * getters and transform them to properties of a new blanc object.
```
larsrickert, line 9:
```suggestion
 * This makes it hard to compare changes. This function iterates through all
 * getters and transforms them to properties of a new plain object.
```

packages/sit-onyx/src/utils/forms.ts
```diff
@@ -1,3 +1,23 @@
...
+ */
+export const transformValidityStateToObject = (validityState: ValidityState): ValidityState => {
+  return getValidityStateProperties().reduce<Record<keyof ValidityState, boolean>>(
```
larsrickert, line 12:
```suggestion
  return getValidityStateProperties().reduce(
```

Does this work? Since the default value of the reducer is already type casted, it should be the same

packages/sit-onyx/src/utils/forms.ts
```diff
@@ -7,11 +27,9 @@ export const getFirstInvalidType = (validity: ValidityState) => {
...
 
   // since the types are getters of the ValidityState we need to get the keys using "Object.getOwnPropertyDescriptors"
-  const availableValidityTypes = Object.entries(
-    Object.getOwnPropertyDescriptors(ValidityState.prototype),
-  )
-    .filter(([key, value]) => key !== "valid" && value.enumerable)
-    .map(([key]) => key) as Exclude<keyof ValidityState, "valid">[];
+  const availableValidityTypes = getValidityStateProperties().filter(
+    (key) => key !== "valid",
+  ) as Exclude<keyof ValidityState, "valid">[];
```
larsrickert, line 32:
```suggestion
  const availableValidityTypes = getValidityStateProperties().filter(
    (key): key is Exclude<keyof ValidityState, "valid"> => key !== "valid",
  );
```

larsrickert (changes requested):

apps/alpha-test-app/package.json
BoppLi: thank youuu!

apps/alpha-test-app/README.md
BoppLi: I'll add that but will keep the info in the README, as people are used to livereload they might not remember that they have to restart ^^

larsrickert:
Seems like the pnpm lock file is broken on the branch, can you run `pnpm install` to fix it?

apps/alpha-test-app/tsconfig.node.json
BoppLi: image
does not look like that works

apps/alpha-test-app/src/App.vue
larsrickert: Actually the definition of our `OnyxTranslations` is invalid. Can you change it to in `packages/sit-onyx/src/src/i18n/index.ts`:
```ts
/**
 * The type of the imported `enUS` above is a concrete type so value type of each message
 * is e.g. "{ myKey: 'HelloWorld'}" but it should be "{ myKey: string }".
 * This utility type converts all values to be of type string (more generic)
 * so we can actually use other locales as well.
 */
type GetTypeOfTranslations<T> = T extends object
  ? { [P in keyof T]?: GetTypeOfTranslations<T[P]> }
  : string;

/** Available translations that are used by onyx components. */
export type OnyxTranslations = GetTypeOfTranslations<typeof enUS>;
```

apps/alpha-test-app/package.json
BoppLi: you mean only the dev dependencies right? when I remove "vue" vite crashes

larsrickert: Yes sorry :)

apps/alpha-test-app/tsconfig.node.json
larsrickert: Did you also apply this suggestion?: https://github.com/SchwarzIT/onyx/pull/157#discussion_r1459139278

Both must be applied to work

packages/sit-onyx/src/components/TestInput/TestInput.vue
BoppLi: is this better?

```suggestion
  if (
      !validityState.value ||
    Object.values(newValidityState).toString() != Object.values(validityState.value).toString()
  ) {
```

larsrickert: The problem is the same^^ Both rely on that all keys are in the same order so the comparison fails e.g. for `{a: 1, b: 2}` and `{b:2, a:1}` although the objects are the same.

Usually you'd use a package like [fast-deep-equals](https://www.npmjs.com/package/fast-deep-equal) but since we agreed to only have minimal direct dependencies we probably have to implement a minimal version of it on our own that iterates through the objects and compares the content of each key instead of their string representation :)

apps/alpha-test-app/src/App.vue
BoppLi: will do

BoppLi: nice, that's way better 😸

apps/alpha-test-app/tsconfig.node.json
BoppLi: 🤦🏻‍♀️  now me too ran into the trap.. sry 😅
image

packages/sit-onyx/src/components/TestInput/TestInput.vue
BoppLi: oki, my thought was that the order will stay the same because we iterate on the same pattern through the ordered keys we get from the util. but I was also on the brink of writing a compare util, I'll just do that so we don't have to worry about it in the future :3

BoppLi:
> Seems like the pnpm lock file is broken on the branch, can you run `pnpm install` to fix it?

can you try again with the latest version? did that already

larsrickert:
> > Seems like the pnpm lock file is broken on the branch, can you run `pnpm install` to fix it?
> 
> can you try again with the latest version? did that already

Yes it works now 👍

apps/alpha-test-app/index.html
```diff
@@ -0,0 +1,13 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <link rel="icon" type="image/x-icon" href="favicon.ico" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Onyx Alpha Test App</title>
```
larsrickert, line 7:
```suggestion
    <title>onyx Alpha Test App</title>
```

apps/alpha-test-app/public/favicon.ico
larsrickert, line 1: Should we remove this file?

packages/sit-onyx/src/utils/comparator.spec.ts
larsrickert, line 1: Should we use `test.each()` so can add more test cases easily?:
https://vitest.dev/api/#test-each

packages/sit-onyx/src/utils/comparator.ts
```diff
@@ -0,0 +1,13 @@
+type FlatObject = Record<string, string | boolean | number | null | undefined>;
+
+/**
+ * Compares two flat objects (whose values are only primitives, no arrays or objects)
+ * @returns whether all values match, ignoring in which order they appear
+ */
+export const areObjectsFlatEqual = (obj1: FlatObject, obj2: FlatObject): boolean => {
+  if (Object.keys(obj1).length !== Object.keys(obj2).length) return false;
```
larsrickert, line 8: I am not sure if this check is correct, example:
`{ a: "foo", b: undefined }` and `{ a: "foo" }` are equal but have a different number of keys

Can you add a test case for this?

packages/sit-onyx/src/utils/comparator.ts
```diff
@@ -0,0 +1,13 @@
...
+  const notEqualItems = Object.entries(obj1).filter(([name, value]) => {
+    return value !== obj2[name];
+  });
```
larsrickert, line 11: For performance reasons, we don't need to compare all keys. As soo as we have the first mismatch, we can return false

packages/sit-onyx/src/utils/forms.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+test("should prioritize valueMissing over other invalid types", () => {
+  // ARRANGE
+  const testState = getTestState({ badInput: true, valueMissing: true });
+  // ACT
+  const result = getFirstInvalidType(testState);
+  // ASSERT
+  expect(result).toBe("valueMissing");
+});
+
+test("should return a single invalid type", () => {
```
larsrickert, line 37: I don't get this test case :D

packages/sit-onyx/src/utils/forms.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+  // ARRANGE
+  const testState = getTestState({ tooShort: true });
+  // ACT
+  const result = getFirstInvalidType(testState);
+  // ASSERT
+  expect(result).toBe("tooShort");
+});
+
+test("should return the first invalid type that is found", () => {
```
larsrickert, line 46: Imho the test name is a little confusing because "badInput" is returned although its the third key in the object :)
I mean it makes sense because badInput is the first key in the ValidityState object but maybe we can adjust the test name :)

larsrickert (changes requested):

apps/alpha-test-app/public/favicon.ico
BoppLi: should we let the browser tab take the fallback? 
image

packages/sit-onyx/src/utils/comparator.ts
BoppLi: I'll change it to .find :)

apps/alpha-test-app/public/favicon.ico
larsrickert: Just wanted to keep the file count of the demo app minimal. I am also fine with keeping the favicon^^

packages/sit-onyx/src/utils/forms.spec.ts
BoppLi: better?

BoppLi: I was worried that the descriptions get too long, but it's more important that it is understandable so I changed it ^^
better? :)

packages/sit-onyx/src/utils/forms.spec.ts
BoppLi: I was worried that the descriptions get too long, but it's more important that it is understandable so I changed it ^^
better? :)

packages/sit-onyx/src/utils/comparator.ts
```diff
@@ -0,0 +1,20 @@
...
+
+  // { } and { a: undefined } are equal, so we ignored all undefined properties to get a more reliable comparison
+  if (noUndefinedEntries1.length !== noUndefinedEntries2.length) return false;
+
+  return (
+    -1 ===
+    noUndefinedEntries1.findIndex(([name, value]) => {
+      return value !== obj2[name];
+    })
+  );
```
larsrickert, line 19:
```suggestion
  return noUndefinedEntries1.every(([name, value]) => value === obj2[name]);
```

packages/sit-onyx/src/utils/comparator.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+  const compareObj = { ...referenceObj, e: undefined };
+
+  // ACT
+  const result = areObjectsFlatEqual(referenceObj, compareObj);
+
+  // ASSERT
+  expect(result).toBeTruthy();
+});
+
+test.each([
```
larsrickert, line 39: (optional) You could also combine all the test cases above into this test.each by adding a second property `shouldBeEqual`

larsrickert (changes requested):

packages/sit-onyx/src/utils/comparator.spec.ts
BoppLi: I'll try, I just hope it does not make the test label too hard to build :D

BoppLi: image
the test message gets truncated :'(

BoppLi: image
even if I make them fail. imho it's more important that the messages are complete then to get all in one test.each.

but I'll bundle up the truthy tests, better than nothing ^^

larsrickert (dismissed):

larsrickert (approved):

pnpm-lock.yaml
```diff
@@ -9317,7 +9357,7 @@ packages:
     dev: true
 
   /stackback@0.0.2:
-    resolution: {integrity: sha512-1XMJE5fQo1jGH6Y/7ebnwPOBEkIEnT4QF32d5R1+VXdXveM0IBMJt8zfaxX1P3QhVwrYe+576+jkANtSS2mBbw==}
+    resolution: {integrity: sha512-1XMJE5fQo1jGH6Y/7ebnwPOBEkIEnT4QF32d5R1+VXdXveM0IBMJt8zfaxX1P3QhVwrYe+576+jkANtSS2mBbw==, tarball: https://schwarzit.jfrog.io/schwarzit/api/npm/npm/stackback/-/stackback-0.0.2.tgz}
```
larsrickert, line 9360: Tarball is internal, we need to remove it

larsrickert:
> image
> not again 😭
> 
> what should I do? restart it?

Yes you can restart for now...I am working on a fix or maybe we should just remove the contributor count

BoppLi:
> Yes you can restart for now...I am working on a fix or maybe we should just remove the contributor count

restart worked :)
yeah, disabling at least until we have it stable would make sense

larsrickert (approved):

PR #178, opened 2024-01-23
feat(vitepress-theme): add `breakpoint` scss mixin

larsrickert:
## Description

Add a scss mixin for applying CSS only to specific VitePress breakpoints

## Checklist

### DIFF
--- a/apps/docs/src/packages/vitepress-theme.md
+++ b/apps/docs/src/packages/vitepress-theme.md
@@ -104,7 +104,7 @@ Applies CSS only to the given VitePress breakpoint.
   }
 
   @include mixins.breakpoint(min, xl, 1) {
-    // your styles for grater than xl breakpoint (exclusive)
+    // your styles for greater than xl breakpoint (exclusive)
   }
 }
 ```
--- a/apps/docs/src/packages/vitepress-theme.md
+++ b/apps/docs/src/packages/vitepress-theme.md
@@ -88,12 +88,23 @@ Applies CSS only to the given VitePress breakpoint.
 @use "@sit-onyx/vitepress-theme/mixins.scss";
 
 .some-class {
-  @include mixins.breakpoint(m, max) {
+  @include mixins.breakpoint(max, m) {
     // your styles for m breakpoint and smaller
   }
 
-  @include mixins.breakpoint(m, min) {
+  @include mixins.breakpoint(min, m) {
     // your styles for m breakpoint and larger
+  }
+
+  // the breakpoint is inclusive so if you e.g. want to use
+  // min and max for the same breakpoint you should
+  // define an offset for either min or max
+  @include mixins.breakpoint(max, xl) {
+    // your styles for smaller and equal xl breakpoint
+  }
+
+  @include mixins.breakpoint(min, xl, 1) {
+    // your styles for grater than xl breakpoint (exclusive)
   }
 }
 ```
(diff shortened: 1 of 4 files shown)


apps/docs/src/packages/vitepress-theme.md
```diff
@@ -62,3 +62,38 @@ export default OnyxTheme;
...
+```scss
+@use "@sit-onyx/vitepress-theme/mixins.scss";
+
+.some-class {
+  @include mixins.breakpoint(m, max) {
+    // your styles for m breakpoint and smaller
+  }
+
+  @include mixins.breakpoint(m, min) {
+    // your styles for m breakpoint and larger
```
BoppLi, line 96: 1) sure? or did smth get flipped?
```suggestion
  @include mixins.breakpoint(m, max) {
    // your styles for m breakpoint and larger
  }

  @include mixins.breakpoint(m, min) {
    // your styles for m breakpoint and smaller
```

2) is "m" always included? so both would apply when you have exactly 768px?

BoppLi (changes requested):

packages/vitepress-theme/src/mixins.scss
```diff
@@ -43,11 +43,34 @@
...
+
+/**
+ * Applies CSS only to the given VitePress breakpoint.
+ * @param {string} $name - name of the VitePress breakpoint. One of: xs, s, m, l, xl or xxl
+ * @param {string} $min-max - Whether the breakpoint should be considered as min or max width.
+ *
+ * @example breakpoint(m, min)
+ * @example breakpoint(m, max)
+ */
+@mixin breakpoint($name, $min-max) {
```
BoppLi, line 63: imho, in terms of readability, it would go smoother with min/max as first param
```suggestion
@mixin breakpoint($min-max, $name) {
```
so then you read "min m" -> minimally m

apps/docs/src/packages/vitepress-theme.md
larsrickert: Yes because its `min-width: 768px` and `max-width: 768px`.
Should I add an optional third parameter `offset` so you can do this if you want to use both:
```scss
  @include mixins.breakpoint(m, max) {
    // your styles for m breakpoint and smaller
  }

  @include mixins.breakpoint(m, min, -1) {
    // your styles for m breakpoint and larger
  }
```

BoppLi (changes requested):

apps/docs/src/packages/vitepress-theme.md
BoppLi: either that or implicitly do this always on either min or max
+ docs

so that you get a consistent experience

larsrickert: I'd prefer the offset parameter since I can imagine that it should be either applied to min or max, depending on the use case :)

apps/docs/src/packages/vitepress-theme.md
```diff
@@ -62,3 +62,49 @@ export default OnyxTheme;
...
+
+  // the breakpoint is inclusive so if you e.g. want to use
+  // min and max for the same breakpoint you should
+  // define an offset for either min or max
+  @include mixins.breakpoint(max, xl) {
+    // your styles for smaller and equal xl breakpoint
+  }
+
+  @include mixins.breakpoint(min, xl, 1) {
+    // your styles for grater than xl breakpoint (exclusive)
```
BoppLi, line 107:
```suggestion
    // your styles for greater than xl breakpoint (exclusive)
```

BoppLi (dismissed):

BoppLi (approved):

PR #182, opened 2024-01-23
refactor(themes): use `.dark` selector instead of `html.dark`

larsrickert:
## Description

Use `.dark` selector for the dark theme instead of `html.dark`. Similarly we could use e.g. `.onyx-lidl`, `.onyx-kaufland` for our future themes. This will give us more flexibility because it allows us to use a specific theme only for a single component (e.g. by setting `class="onyx-lidl"` on a `<div>`.

Like this we can e.g. implement [this component](https://www.figma.com/file/YfEUBOHk4J4nYrk04geswG/onyx---Design-System-Library?type=design&node-id=408-4561&mode=dev) for our docs by simply changing the class to `.onyx-lidl`, `.onyx-kaufland` etc.
image

## Checklist


jannick-ux (approved):

BoppLi (approved):

JoCa96 (changes requested):
wait for discussion result

PR #203, opened 2024-01-26
feat(grid-layout): Implemented span elements grid

JoCa96:
Implemented and documented classes for configuring the grid elements span sizes.


apps/docs/src/development/grid.md
```diff
@@ -17,6 +17,12 @@ The dynamic grid has the following advantages over a fixed column grid:
 - Ensures reasonable minimal and maximal size for individual elements
   :::
 
+Grid elements span the number of columns that is assigned to them.
+If there are less columns then an element is assigned, it will span all columns of the row.
+
+Additionally an element can be also be configured to span a specific column amount for a minium breakpoint (TODO: add link to UX breakpoint definition).
+Multiple span definition can then be combined to resize an element based on the breakpoint.
```
larsrickert, line 24:
```suggestion
Grid elements span the number of columns that is assigned to them.
If there are less columns available than an element is assigned, it will span all columns of the row.
All components already have a default span defined so you don't need to set anything if you just want to align with the default behavior.

Additionally, an element can also be configured to span a specific amount of columns for a minimum [breakpoint](/basics/breakpoints-grid).
Multiple span definitions can then be combined to resize an element based on the breakpoint.
```

FYI you need to merge the current main branch for the link to work :)

apps/docs/src/development/grid.md
```diff
@@ -28,3 +34,31 @@ Additionally, there are some optional modifier classes available.
 - `onyx-grid-xl-20`: Increases the column count from 16 to 20 for the `xl` breakpoint (useful for 4k screens).
 - `onyx-grid-max-md`: Caps the width before the `lg` breakpoint (`1440px`)
 - `onyx-grid-max-lg`: Caps the width before the `xl` breakpoint (`1920px`)
+
+For configuring the amount of columns an element should be spanning use the class `onyx-grid-span-<number>`, where `<number>` is the number of columns from `1` to `20`.
```
larsrickert, line 38:
```suggestion
For configuring the amount of columns an element should be spanning, use the class `onyx-grid-span-<number>`, where `<number>` is the number of columns from `1` to `20`.
```

apps/docs/src/development/grid.md
```diff
@@ -28,3 +34,31 @@ Additionally, there are some optional modifier classes available.
...
+
+To set the column number for a minium screen width use the class `onyx-grid-<breakpoint>-span-<number>`, where `<breakpoint>` and `<number>` are taken from the following table:
```
larsrickert, line 40:
```suggestion
To set the amount of columns for a minimum screen width, use the class `onyx-grid-<breakpoint>-span-<number>`, where `<breakpoint>` and `<number>` are taken from the following table:
```

larsrickert (changes requested):

apps/docs/src/development/grid.md
JoCa96: I would like to only document things that we already do and have.

JoCa96: Therefore I would like to leave out
> All components already have a default span defined so you don't need to set anything if you just want to align with the default behavior.

for now.

I created a ticket, so we don't forget about it: https://github.com/orgs/SchwarzIT/projects/5?pane=issue&itemId=51255861

larsrickert (approved):

PR #227, opened 2024-01-31
feat(grid-layout): implement grid alignment

JoCa96:
- Added class generation for grid alignment
- Refactored grid.scss to be hopefully be more understandable
- Grid margin is now implemented through padding, but we still call it "margin" for consistency with the UX concept


packages/sit-onyx/src/styles/grid.scss
```diff
@@ -1,102 +1,96 @@
...
+/**
+ * Grid container base class
+ */
 .onyx-grid {
   display: grid;
   grid-template-columns: repeat(var(--onyx-grid-columns), minmax(0, 1fr));
   gap: var(--onyx-grid-gutter);
-  margin: var(--onyx-grid-margin);
+  // Its called margin in the UX concept. So we keep that name for consistency, even when it is implemented through a padding.
+  padding: var(--onyx-grid-margin);
```
larsrickert, line 20: 🤨 
If its a padding, can't we adjust the UX concept accordingly to also call it by what it is?^^

larsrickert: Alternatively implement it as margin, is there a reason why its padding?

larsrickert (changes requested):

packages/sit-onyx/src/styles/grid.scss
JoCa96: We use "margin: 0 auto" for centering and the padding to ensure the Grid-Margin is still applied.
Afaik "padding" is not a concept that is used or exists in UX

jannick-ux (approved):

larsrickert (approved):

PR #231, opened 2024-01-31
feat: add icon library and `OnyxIcon` component

larsrickert:
Add all icons from our current `@schwarz-ux/icons` library (version 1.6.2) as well as a `OnyxIcon` component for using the icons.
All icons were optimized with svgo to:
- reduce their file size
- remove "fill" properties so they can be colored via CSS

Also I manually fixed some SVGs with 25x24px size to 24x24px.

## Checklist


apps/docs/src/.vitepress/config.ts
```diff
@@ -36,6 +36,11 @@ export default defineConfig({
       { text: "Basics", link: "/basics/", activeMatch: "/basics/" },
       { text: "Tokens", link: "/tokens/introduction", activeMatch: "/tokens/" },
       { text: "Development", link: "/development/", activeMatch: "/development/" },
+      {
+        text: "Resources",
```
BoppLi, line 40: fyi with thsi new nav bar entry, we get a bug (too many items overall ^^)
https://github.com/orgs/SchwarzIT/projects/5/views/7?layout=board&pane=issue&itemId=51919555
I created this follow up ticket.

Alternatively, to prevent that bug, we might consider putting the resources as an additional category under the development section ¯\_(ツ)_/¯

apps/docs/src/basics/iconography.md
```diff
@@ -1,3 +1,5 @@
 # Iconography
 
 <!--@include: @/.vitepress/to-be-done.md-->
+
+For a full list of icons, please take a look at thr [icon resources](/resources/icons).
```
BoppLi, line 5:
```suggestion
For a full list of icons, please take a look at the [icon resources](/resources/icons).
```

apps/docs/src/development/packages/changelogs/[name].md
```diff
@@ -35,6 +35,12 @@ const { params } = useData();
 
 </div>
 
+<div v-else-if="params.name === 'icons'">
```
Comment (BoppLi) [inline apps/docs/src/development/packages/changelogs/[name].md:38]: How can this site be reached? The Icons package is not listed under packages as the other ones.

image

\+ http://localhost:5173/development/packages/changelogs/icons.html is empty, is it because it is new and we have not published yet? :)

Comment (BoppLi) [reply apps/docs/src/development/packages/changelogs/[name].md:38]: found it in apps/docs/src/resources/icons.md, but shouldn't there be a packages entry as well, if only to link to that page? if people are coming from the npmjs perspective to the docs. :)

packages/icons/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "@sit-onyx/icons",
+  "description": "SVG icons for the onyx design system",
```
BoppLi, line 3: in the README.md logo, the phrasing is
```suggestion
  "description": "SVG icons for the onyx design systems",
```
should we align?

BoppLi (changes requested):

packages/icons/package.json
larsrickert: If we want to align this we should also do it for all other packages and GitHub documentation.
Imho for now we only have one design system so it would be fine for me

apps/docs/src/development/packages/changelogs/[name].md
Comment (larsrickert) [reply apps/docs/src/development/packages/changelogs/[name].md:38]: Yes its empty because it has not been published yet. I added an entry for the icons and linked to the resources page

apps/docs/src/.vitepress/components/OnyxIconLibrary.vue
```diff
@@ -0,0 +1,47 @@
+<script lang="ts" setup>
```
BoppLi, line 1: [Question] is a search / copy functionality planned for the future (out of scope of this PR) 
if so: I created https://github.com/orgs/SchwarzIT/projects/5/views/7?layout=board&pane=issue&itemId=51923734 as follow up

apps/docs/src/resources/icons.md
```diff
@@ -0,0 +1,39 @@
...
+yarn install -D @sit-onyx/icons@alpha
+```
+
+:::
+
+Afterwards, you can import and use icons as needed. Please see the [OnyxIcon](/development/OnyxIcon) for the technical documentation.
+
+## Available icons
+
+You can hover over an icon to see its name.
```
BoppLi, line 33: hover is not implemented. remove this text then?

apps/docs/src/.vitepress/config.ts
BoppLi: sounds good aswell :)

apps/docs/src/.vitepress/components/OnyxIconLibrary.vue
```diff
@@ -0,0 +1,47 @@
...
+
+const getIconContent = (iconName: string) => {
+  return ALL_ICONS[`../../../node_modules/@sit-onyx/icons/src/assets/${iconName}.svg`];
+};
+</script>
+
+<template>
+  <div>
+    <section v-for="(icons, category) in ICON_CATEGORIES" :key="category" class="category">
+      <h4 class="category__headline">{{ category }}</h4>
```
BoppLi, line 18: image

As the above headline is H2, best practice would be to make this h3 -> don't skip headline levels.
```suggestion
      <h3 class="category__headline">{{ category }}</h3>
```

apps/docs/src/.vitepress/components/OnyxIconLibrary.vue
```diff
@@ -0,0 +1,47 @@
...
+
+      <div class="category__icons">
+        <OnyxIcon
+          v-for="icon in icons"
+          :key="icon.iconName"
+          :icon="getIconContent(icon.iconName)"
+          size="md"
+          :title="`${icon.iconName}.svg`"
```
BoppLi, line 26: title does not exist as prop on OnyxIconProps

BoppLi (changes requested):
16/617

packages/icons/src/metadata.ts
```diff
@@ -0,0 +1,1816 @@
+/**
+ * Metadata for all available onyx icons.
+ */
+export const ICON_METADATA = {
```
BoppLi, line 4: question: how is this list going to be maintained?

packages/icons/src/metadata.ts
```diff
@@ -0,0 +1,1816 @@
...
+  metadata: IconMetadata;
+};
+
+export type IconCategories = Record<string, GroupedIconCategory[]>;
+
+/**
+ * Groups all available icon metadata by category.
+ * Categories and icons will be sorted alphabetically.
+ */
+function groupIconsByCategory() {
```
BoppLi, line 1793: don't we usually use arrow functions? is there a need for the named function here?

```suggestion
const groupIconsByCategory = () => {
```

BoppLi: only thing you need to change then is to put it above ICON_CATEGORIES so you don't get a warning "used before being assigned

packages/icons/src/metadata.ts
```diff
@@ -0,0 +1,1816 @@
+/**
+ * Metadata for all available onyx icons.
+ */
```
BoppLi, line 3: [opinion] I'd prefer to split the file and move the ICON_METADATA to a separate file from the rest of the file below it. for our own DX's sake when we want to work on the things going on below ^^ this section looks kinda auto generated

packages/icons/src/metadata.ts
```diff
@@ -0,0 +1,1816 @@
...
+  Object.keys(categories)
+    .sort()
+    .forEach((category) => {
+      const sortedMetadata = categories[category].slice().sort((a, b) => {
+        return a.iconName.localeCompare(b.iconName);
+      });
+      sortedCategories[category] = sortedMetadata;
+    });
+
+  return sortedCategories;
```
BoppLi, line 1815: in lack of a unit test I can not be sure, but shouldn't this do the same thing?

```suggestion
  return Object.entries(ICON_METADATA).reduce<IconCategories>(
    (categories, [iconName, metadata]) => {
      const icons = categories[metadata.category] ?? [];
      icons.push({ iconName: iconName as GroupedIconCategory["iconName"], metadata });
      icons.sort((a, b) => {
        return a.iconName.localeCompare(b.iconName);
      });
      categories[metadata.category] = icons;
      return categories;
    },
    {},
  );
```

BoppLi: here's the better readable diff for you :)

image

packages/icons/src/utils.ts
```diff
@@ -0,0 +1,47 @@
+import fs from "node:fs/promises";
+import path from "node:path";
+import { fileURLToPath } from "node:url";
+import { optimize } from "svgo";
+
+/**
+ * Optimizes the given SVG content and removes all fills so the color can be set via CSS.
+ */
+export function optimizeSvg(svgContent: string, path: string) {
+  const { data } = optimize(svgContent, {
```
BoppLi, line 10: In the PR description, you wrote that you also change 25x24px files to 24x24 -> where does this happen?

BoppLi (changes requested):
7 files left :)

packages/sit-onyx/src/components/OnyxIcon/OnyxIcon.stories.ts
```diff
@@ -0,0 +1,100 @@
...
+const ALL_ICONS = import.meta.glob("../../../node_modules/@sit-onyx/icons/src/assets/*.svg", {
+  as: "raw",
+  eager: true,
+});
+
+/**
+ * Mapping between icon SVG content (key) and icon name (value).
+ * Needed to display a labelled dropdown list of all available icons.
+ */
+const iconLabels = Object.entries(ALL_ICONS).reduce<Record<string, string>>(
```
BoppLi, line 17: could the function to get iconlabels be something that the icon package exports as a feature? what we learned from CoreUI is that many more components might need an icon selection in the storybook files. this logic also is more focusing into the structure how the icons package is implemented, not sure this is a good place here for this logic (when you work on the internal implementation -> structure of packages/icons, you probably don't expect side effects in the other packages)

BoppLi: -> does not have to happen in this PR! just wanted to ask for your opinion :)

packages/sit-onyx/src/components/OnyxIcon/OnyxIcon.stories.ts
```diff
@@ -0,0 +1,100 @@
...
+    return labels;
+  },
+  {},
+);
+
+/**
+ * Component to display icons. Supports all inline SVG icon libraries.
+ * We recommend using the official icons from `@sit-onyx/icons`.
+ *
+ * When importing icons, make sure to add `?raw` after the file name as shown in the examples to import the SVG content
```
BoppLi, line 29: does this apply to all icons, even from other libs? or rather
```suggestion
 * When importing icon files, make sure to add `?raw` after the file name as shown in the examples to import the SVG content
```
or
```suggestion
 * When importing onyx icons, make sure to add `?raw` after the file name as shown in the examples to import the SVG content
```

packages/sit-onyx/src/components/OnyxIcon/OnyxIcon.vue
```diff
@@ -0,0 +1,93 @@
...
+
+<template>
+  <!-- eslint-disable vue/no-v-html -->
+  <i
+    class="onyx-icon"
+    :class="[
+      props.size !== 'sm' ? `onyx-icon--${props.size}` : '',
+      props.color !== 'currentColor' ? `onyx-icon--${props.color}` : '',
+    ]"
+    aria-hidden="true"
```
BoppLi, line 18: is aria-hidden a good idea? shouldne we rather offer an alt with e.g. 
"Icon of ${iconName}" that can be overwritten or so?

-> Image you have an icon button, then as a screen reader user you would just get an info that it is a button but no idea what it is about?

packages/sit-onyx/src/components/OnyxIcon/OnyxIcon.vue
```diff
@@ -0,0 +1,93 @@
...
+
+const props = withDefaults(defineProps<OnyxIconProps>(), {
+  size: "sm",
+  color: "currentColor",
+});
+</script>
+
+<template>
+  <!-- eslint-disable vue/no-v-html -->
+  <i
```
BoppLi, line 12: why `<i>`?

packages/sit-onyx/src/components/OnyxIcon/types.ts
```diff
@@ -0,0 +1,16 @@
+import type { OnyxColor } from "@/index";
+
+export type OnyxIconProps = {
+  /**
+   * SVG source of the icon. **Important**: Only provide trustworthy content, the SVG content will
+   * not be sanitized.
```
BoppLi, line 6: because we can not sanitize it or because sanitizing would make it too complicated / has side effects?

packages/sit-onyx/src/components/OnyxIcon/OnyxIcon.vue
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  &--2xl {
+    --icon-size: 6rem;
+  }
+
+  //
+  // icon colors:
+  //
+  &--primary {
```
BoppLi, line 65: this looks like it could be done with a mixin but we don't have to of course :)

maybe if we need this behavior more often, also in other components, we could still do it then :)

BoppLi (dismissed):
Done! great work! ✨

apps/docs/src/.vitepress/components/OnyxIconLibrary.vue
larsrickert: Yes that would be very nice :) I did not implement it yet because we don't have a input component yet

apps/docs/src/resources/icons.md
larsrickert: If you hover over it the `title` attribute will be shown :) Thats the browser default

apps/docs/src/.vitepress/components/OnyxIconLibrary.vue
larsrickert: title is a global HTML attribute that exists on every HTML element: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title

packages/icons/src/metadata.ts
larsrickert: If UX adds a new icon, they should also add a metadata entry here.
I documented the contribution flow in the README :)

packages/icons/src/metadata.ts
larsrickert: This only sorts the icons, not the categories itself. Also its less performant since all icons are sorted in every iteration instead of once after all icons are grouped.

packages/icons/src/utils.ts
larsrickert: I had to do this manually so this is a one-time fix where we don't have code for :)

packages/sit-onyx/src/components/OnyxIcon/OnyxIcon.stories.ts
larsrickert: Its needed whenever a `.svg` file is directly imported so also for other icon libraries.

packages/sit-onyx/src/components/OnyxIcon/OnyxIcon.vue
larsrickert: I asked Jannick in our Wednesday daily and he said that icons are only used in 2 use cases:

1. interactively inside another component, e.g. a button. In this case the `<button>` is the relevant aria component so it will have a title
2. non-interactively / standalone only as visual component. Since its only visual we can hide it for aria

So for both cases, the icon itself is not relevant for aria

larsrickert: Also MDN explains that `aria-hidden` is often used on icons: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-hidden

packages/sit-onyx/src/components/OnyxIcon/OnyxIcon.vue
larsrickert: We need a wrapper component due to the `v-html` since we need to put it somewhere :D Google said that `<i>` and `<span>` are often used for icons: https://www.w3schools.com/icons/#:~:text=The%20and%20%3Cspan,widely%20used%20to%20add%20icons.

Should I change it to span or something else? I don't have a preference here

packages/sit-onyx/src/components/OnyxIcon/OnyxIcon.vue
larsrickert: I think this is a more special case because we need a specific color and CSS variable but I used the `@each` feature of SCSS to generate the classes. Thanks for the tip! This is much shorter

packages/sit-onyx/src/components/OnyxIcon/OnyxIcon.stories.ts
larsrickert: Good idea, I do not want to mix up the icon package with Storybook. That's why we have the Storybook utils package where we could move this :)

packages/icons/src/metadata.ts
BoppLi: ok

packages/icons/src/utils.ts
BoppLi: ok, I misunderstood because it was a sub-point of "optimized with svgo" ^^
it's fine, ppl probably won't look up old commit messages anyways :)

apps/docs/src/resources/icons.md
JoCa96: afaik the hover behaviour is OS dependent

packages/icons/src/utils-node.ts
```diff
@@ -0,0 +1,53 @@
...
+import { optimize } from "svgo";
+
+/**
+ * Optimizes the given SVG content and removes all fills so the color can be set via CSS.
+ */
+export const optimizeSvg = (svgContent: string, path: string) => {
+  const { data } = optimize(svgContent, {
+    path,
+    multipass: true,
+    plugins: [
```
JoCa96, line 19: The description says that we reduce the file size using svgo, but here we only use `removeAttrs`, shouldn't we at least use the `default-preset`?

packages/sit-onyx/src/components/OnyxIcon/OnyxIcon.vue
BoppLi: ok, then we just have to remember it once we do the icon-button :)

packages/sit-onyx/src/components/OnyxIcon/OnyxIcon.vue
JoCa96: Maybe `figure`, but it is not an inline element

apps/docs/src/resources/icons.md
larsrickert: Maybe @jannick-ux has a nice idea how to show the icon names on hover. Maybe similar to the design tokens? 
Could be implemented together with https://github.com/orgs/SchwarzIT/projects/5/views/7?layout=board&pane=issue&itemId=51923734

packages/sit-onyx/src/components/OnyxIcon/OnyxIcon.vue
larsrickert: @JoCa96 Should I change it to `<figure>` then and make it `display: inline-block`?
I would also be fine with just keeping `<i>` since it kinda sounds like "icon"

apps/docs/src/.vitepress/components/DesignToken.vue
```diff
@@ -1,6 +1,7 @@
 <script lang="ts" setup>
-import CheckIcon from "./icons/CheckIcon.vue";
-import CopyIcon from "./icons/CopyIcon.vue";
+import checkIcon from "@sit-onyx/icons/check-small.svg?raw";
+import copyIcon from "@sit-onyx/icons/copy.svg?raw";
+import OnyxIcon from "../../../../../packages/sit-onyx/src/components/OnyxIcon/OnyxIcon.vue";
```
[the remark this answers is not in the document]
larsrickert: This does not work unless we add the components folder to the `exports` in `package.json`.
But then everyone could directly import the vue files...I am not sure if we want this.

apps/docs/src/resources/icons.md
BoppLi: ok, I added this info to the ticket. it's out of scope of this PR then :)

packages/sit-onyx/src/components/OnyxIcon/OnyxIcon.vue
BoppLi: the thing is, it does not really fit the [description](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/i?retiredLocale=de) of what the i tag is

image

BoppLi: image
figure sounds more like it :)
but span would also be fine for me ¯\\_(ツ)_/¯

packages/icons/src/utils-node.ts
larsrickert: I though this is included by default^^ I added it, thanks!

BoppLi (approved):

apps/docs/src/.vitepress/components/DesignToken.vue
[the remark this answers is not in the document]
larsrickert: I added a vite + TypeScript alias instead so we can now import components using `~components` (only for our docs), e.g.:
`import OnyxIcon from "~components/OnyxIcon/OnyxIcon.vue";`

packages/sit-onyx/src/components/OnyxIcon/OnyxIcon.vue
larsrickert: I changed it to `<figure>` :)

apps/docs/src/resources/icons.md
jannick-ux: Yes I will create a concept for promoting the icons

BoppLi (approved):

PR #232, opened 2024-01-31
docs(components): remove .DS_Store file from component sidebar

larsrickert:
remove .DS_Store file from component sidebar

## Checklist

### DIFF
--- a/apps/docs/src/.vitepress/utils.ts
+++ b/apps/docs/src/.vitepress/utils.ts
@@ -1,12 +1,22 @@
-import fs from "node:fs";
+import fs from "node:fs/promises";
+import path from "node:path";
 import { fileURLToPath } from "node:url";
 
 /**
  * Gets a list of all available onyx component names.
  */
-export const getComponents = () => {
-  const path = fileURLToPath(
+export const getComponents = async () => {
+  const componentsPath = fileURLToPath(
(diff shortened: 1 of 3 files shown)


apps/docs/src/.vitepress/utils.ts
```diff
@@ -1,12 +1,22 @@
...
+
+  const components: string[] = [];
+
+  // filter out non-directories to prevent files like ".DS_Store" on macOS to be included
+  for (const componentName of await fs.readdir(componentsPath)) {
+    const stat = await fs.stat(path.join(componentsPath, componentName));
+    if (stat.isDirectory()) components.push(componentName);
+  }
+
+  return components;
```
BoppLi, line 21:
```suggestion

  const componentNames = await fs.readdir(componentsPath);

  // filter out non-directories to prevent files like ".DS_Store" on macOS to be included
  return Promise.all(
    componentNames.filter(async (componentName) => {
      const stat = await fs.stat(path.join(componentsPath, componentName));
      return stat.isDirectory();
    }),
  );
```

BoppLi (approved):

apps/docs/src/.vitepress/utils.ts
BoppLi: nevermind, produces more problems :D

BoppLi (approved):

PR #236, opened 2024-02-01
docs: use OnyxHeadline for typography preview

larsrickert:
Replace the typography preview for headlines with the `OnyxHeadline` component instead of a custom class

## Checklist


apps/docs/src/.vitepress/components/OnyxTypography.vue
```diff
@@ -49,7 +55,16 @@ const currentTab = ref<AvailableFontTab>(AVAILABLE_FONT_TABS[0]);
         :wide-name="props.wideName"
       >
         <template #name>
+          <OnyxHeadline
+            v-if="HEADLINE_TYPES.includes(token.htmlTag as HeadlineType)"
```
BoppLi, line 59: Don't we want to prevent that something not part of HEADLINE_TYPES is passed? / Do we even want to support something else?

1. update type in props
image

2. remove v-if

```suggestion
```
3. remove v-else block
image

BoppLi: lint all + build all scripts would be happy with that adjustment :)

BoppLi (approved):

apps/docs/src/.vitepress/components/OnyxTypography.vue
larsrickert: We also need `<p>` and `<a>` :) But I simplified it quiet a lot, can you please take another look?

BoppLi: NICE! 😊

Issue #222, opened 2024-01-30
Implement OnyxRadioGroup


larsrickert:
### Dependencies
- [ ] #244 

### Why?

A radio group is an input control that allows a user to select (only) one value from a number of choices.
The options must only be used inside the radio group and must not be used individually.

### Design

https://www.figma.com/file/YfEUBOHk4J4nYrk04geswG/onyx---Design-System-Figma-Library?type=design&node-id=626%3A15627&mode=design&t=rydWTxp4TqL9ckXI-1

### Acceptance criteria

- [x] multiple radio options can be put into the radio group
- [x] every radio option must have a label
- the following states are implemented:
  - [x] default
  - [x] hover
  - [x] focus-visible
  - [x] disabled
  - [x] invalid
  - [x] checked
- [x] at most one option can be selected
- [x] the radio group supports a vertical and horizontal layout
- [x] optionally a headline/legend can be shown above the radio group
- [x] keyboard interaction is supported as documented in the ARIA pattern below

_Out-of-scope:_
- densities
- loading
- errorMessage

### Definition of Done

[Storybook](https://storybook.onyx.schwarz/?path=/docs/components-onyxradiobuttongroup--docs)

- [x] covered by component tests
- [x] screenshot tests are updated
- [x] updated version + documentation is deployed
  - [x] Approved by dev @larsrickert
- [x] Storybook can show the feature
  - [x] Approved by designer @mj-hof 

### Implementation details

- radio options are passed as property so we do not export the single radio option component so the user can not use it individually
- the value is typed generically

#### Reference implementations

- [Nuxt UI](https://ui.nuxt.com/forms/radio-group)
- [Adobe Spectrum](https://spectrum.adobe.com/page/radio-group/)
- [WaWi](https://dev.azure.com/schwarzit/lidl.wawi-core/_git/lidl-wawi?path=/ui/components/src/components/WuxRadioGroup/WuxRadioGroup.vue)

#### Applicable ARIA Pattern

https://www.w3.org/WAI/ARIA/apg/patterns/radio/

JoCa96:
> - the following value types are supported:
>   - [ ] string
>   - [ ] number
>   - [ ] boolean

Can we just say that generic value is supported, or why exactly these three?

JoCa96:
@larsrickert did you look into implementation options? Like do we implement this using native radiobuttons or custom implementation using the aria pattern? E.g. the nuxt example uses native radiobuttons.

[...]

PR #275, opened 2024-02-12
feat (OnyxRadioButton): add styling for support component

JoCa96:
#222 

Implemented styling for base component.


packages/sit-onyx/src/components/OnyxRadioButton/OnyxRadioButton.vue
```diff
@@ -8,34 +9,128 @@ export type RadioButtonProps<TValue> = SelectionProps<TValue> & {
    * See also: https://html.spec.whatwg.org/multipage/input.html#radio-button-group
    */
   name: string;
+  isRequired?: boolean;
+  errorMessage?: string;
 };
 
 const props = defineProps<RadioButtonProps<TValue>>();
+
+const selector = ref<HTMLInputElement>();
```
larsrickert, line 18:
```suggestion
const selector = ref<HTMLInputElement | null>(null);
```

Template refs can be null if not mounted yet or due to "v-if". We should type it accordingly :)

packages/sit-onyx/src/components/OnyxRadioButton/OnyxRadioButton.vue
```diff
@@ -8,34 +9,128 @@ export type RadioButtonProps<TValue> = SelectionProps<TValue> & {
...
+
+watch(
+  [() => props.errorMessage, selector],
+  () => selector.value?.setCustomValidity(props.errorMessage ?? ""),
+  { immediate: true },
+);
```
larsrickert, line 24:
```suggestion
watchEffect(() => {
  selector.value?.setCustomValidity(props.errorMessage ?? "");
});
```

packages/sit-onyx/src/components/OnyxRadioButton/OnyxRadioButton.ct.tsx
```diff
@@ -64,3 +64,26 @@ test("should display correctly when disabled", async ({ mount, makeAxeBuilder, p
...
+      label="radio label"
+      name="radio form"
+      value="radio-value"
+      errorMessage="radio error"
+      selected
+    />,
+  );
+
+  // ASSERT
+  await expect(page.locator("input:invalid")).toBeAttached();
```
larsrickert, line 82: Is there also something like "expect(component).toBeInvalid"?

packages/sit-onyx/src/components/OnyxRadioButton/OnyxRadioButton.stories.ts
```diff
@@ -15,6 +15,7 @@ const meta: Meta<typeof OnyxRadioButton> = {
       isLoading: { type: "boolean" },
       isDisabled: { type: "boolean" },
       selected: { type: "boolean" },
+      errorMessage: { type: "string" },
```
larsrickert, line 18: The Storybook props table is missing all descriptions and prop types because an union type is used. This is a known Storybook limitation currently. I'd prefer to remove the union type for now

image

packages/sit-onyx/src/components/OnyxRadioButton/OnyxRadioButton.vue
larsrickert: Also is selector a standard name? Imho something like "inputRef" would be clearer to me

packages/sit-onyx/src/components/OnyxRadioButton/OnyxRadioButton.vue
```diff
@@ -8,34 +9,128 @@ export type RadioButtonProps<TValue> = SelectionProps<TValue> & {
...
     />
     <span class="onyx-radio-button__label">{{ props.label }}</span>
   </label>
 </template>
 
 <style lang="scss">
 .onyx-radio-button {
+  display: inline-flex;
+  align-items: center;
+  height: 2.5rem;
```
larsrickert, line 50: Shouldn't the height be determined by the padding?

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxRadioButton/OnyxRadioButton.ct.tsx
JoCa96: unfortunately not, otherwise I would have used it

packages/sit-onyx/src/components/OnyxRadioButton/OnyxRadioButton.vue
JoCa96: @BoppLi brought up the argument, that BEM should be semantical, not technical. To keep the naming in sync I also named the ref after this. I couldn't think of a better name

JoCa96: @larsrickert Actually `null` is incorrect here. The initial value is `undefined` and by not passing one into to the `ref()` it already types it as `HTMLInputElement | undefined`. But I made the same error :D

larsrickert: See [Vue docs](https://vuejs.org/guide/essentials/template-refs#accessing-the-refs).
See the following example where the ref will be `null`:

```vue
<script lang="ts" setup>
const inputRef = ref<HTMLInputElement | null>(null);

onMounted(() => {
  console.log(inputRef.value);
});
</script>

<template>
    <input v-if="false" />
</template>

```

packages/sit-onyx/src/components/OnyxRadioButton/OnyxRadioButton.vue
JoCa96: I tend to prefer `watch` over `watchEffect` as it can have drastic performance impacts. But for a one-liner like this it's actually a great idea!

packages/sit-onyx/src/components/OnyxRadioButton/OnyxRadioButton.stories.ts
JoCa96: That sucks :( Do you have an issue link? Any idea when this will be fixed?
Because if this will be a long running issue, I would prefer having great typing over duplicated docs.

packages/sit-onyx/src/components/OnyxRadioButton/OnyxRadioButton.vue
```diff
@@ -8,34 +9,124 @@ export type RadioButtonProps<TValue> = SelectionProps<TValue> & {
...
       :checked="props.selected"
-      :disabled="props.isDisabled"
+      :disabled="props.isDisabled || props.isReadonly"
     />
     <span class="onyx-radio-button__label">{{ props.label }}</span>
   </label>
 </template>
 
 <style lang="scss">
 .onyx-radio-button {
```
larsrickert, line 43: I'd prefer to define the custom CSS variables and their fallback/default value at the top. Imho its better understandable whats going on. Example:

Currently:
```css
  &:has(&__selector:hover) {
    --onyx-radio-button-selector-border-color: var(--onyx-color-base-primary-300);
  }

    border: {
      style: solid;
      width: 1px;
      color: var(--onyx-radio-button-selector-border-color, var(--onyx-color-base-neutral-400));
    }
```


Suggestion:
```css
--onyx-radio-button-selector-border-color: var(--onyx-color-base-neutral-400);
// other custom CSS variable here

  &:has(&__selector:hover) {
    --onyx-radio-button-selector-border-color: var(--onyx-color-base-primary-300);
  }

    border: {
      style: solid;
      width: 1px;
      color: var(--onyx-radio-button-selector-border-color);
    }
```

packages/sit-onyx/src/components/OnyxRadioButton/OnyxRadioButton.vue
JoCa96: Lets put this discussion off until we have a density concept

packages/sit-onyx/src/components/OnyxRadioButton/OnyxRadioButton.stories.ts
JoCa96: I added a follow-up ticket: https://github.com/orgs/SchwarzIT/projects/5/views/7?pane=issue&itemId=53102909

packages/sit-onyx/src/components/OnyxRadioButton/OnyxRadioButton.vue
JoCa96: done, not sure if I like it tough 🤔  Lets keep it for now :)

larsrickert (approved):

Issue #220, opened 2024-01-30
Implement OnyxCheckbox


larsrickert:
### Dependencies
- [ ] #244 

### Why?

A checkbox is an input control that allows a user to select one or more options from a number of choices.

### Design

https://www.figma.com/file/YfEUBOHk4J4nYrk04geswG/onyx---Design-System-Figma-Library?type=design&node-id=511%3A199&mode=design&t=28ZL5f0BYOVImIvs-1

### Acceptance criteria

- [x] Research how to best implement (check references)
  - using aria pattern?
  - using svg and invisible checkbox?
- [x] a checkbox with an optional label can be shown
- the following states are implemented:
  - [x] default
  - [x] hover
  - [x] focus-visible
  - [x] disabled
  - [x] invalid
  - [x] required
  - [x] optional
  - [x] checked
  - [x] indeterminate
- the following value types are supported:
  - [x] string
  - [x] number
  - [x] boolean
- [x] keyboard interaction is supported as documented in the ARIA pattern below

**Out of scope**

- an optional loading indicator can be shown, #393
- skeleton state, #379 
- truncation, #350 
- multiline text is supported (meaning newlines are rendered accordingly), #350 
- the following densities are supported (#349):
   - compact
   - default
   - cozy
- the following states are implemented (#339):
    - [ ] read-only 

### Definition of Done

- [x] covered by component tests
- [x] screenshot tests are updated
- [x] follow up tickets were created if necessary
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature
- [x] Set task to `In review`

**Review**
 -  [View in Storybook](https://onyx-storybook-prod.apps.01.cf.eu01.stackit.cloud/?path=/docs/support-onyxcheckbox--docs)
 - [x] approved by designer @jannick-ux 
 - [x] approved by dev @JoCa96 

### Implementation details

- Make use of css pseudo selectors to render different states e.g. https://developer.mozilla.org/en-US/docs/Web/CSS/:indeterminate
- default truncation settings is: truncated + top-aligned 

#### Reference implementations

- [Nuxt UI](https://ui.nuxt.com/forms/checkbox)
- [Adobe Spectrum](https://spectrum.adobe.com/page/checkbox/)
- [Atlassian](https://atlassian.design/components/checkbox/examples)
[...]

PR #278, opened 2024-02-12
feat: add basic `OnyxCheckbox` component

larsrickert:
#220

Add the basic `OnyxCheckbox` component with support for checked and indeterminate state.

## Checklist


packages/sit-onyx/src/components/OnyxCheckbox/OnyxCheckbox.vue
```diff
@@ -0,0 +1,110 @@
...
+
+const isChecked = computed({
+  get: () => props.modelValue,
+  set: (value) => emit("update:modelValue", value),
+});
+
+watchEffect(() => {
+  if (!inputRef.value) return;
+  inputRef.value.indeterminate = props.indeterminate;
+});
```
JoCa96, line 26: What is the reason for not setting it in the template?

packages/sit-onyx/src/index.ts
```diff
@@ -1,5 +1,7 @@
 import "@/styles/index.scss";
 
+export { default as OnyxCheckbox } from "@/components/OnyxCheckbox/OnyxCheckbox.vue";
```
JoCa96, line 3: We only ever wanted to use a single checkbox as a support component, so we shouldn't export it.

packages/sit-onyx/src/components/OnyxCheckbox/OnyxCheckbox.stories.ts
```diff
@@ -0,0 +1,37 @@
+import { defineStorybookActionsAndVModels } from "@sit-onyx/storybook-utils";
+import type { Meta, StoryObj } from "@storybook/vue3";
+import OnyxCheckbox from "./OnyxCheckbox.vue";
+
+/**
+ * Checkboxes are a fundamental UI element, that allows users to make a binary selection.
+ * They are commonly used for tasks such as selecting multiple items, opting into services or confirming and agreeing.
+ */
+const meta: Meta<typeof OnyxCheckbox> = {
+  title: "components/OnyxCheckbox",
```
JoCa96, line 10:
```suggestion
  title: "support/OnyxCheckbox",
```

JoCa96 (changes requested):

packages/sit-onyx/src/index.ts
larsrickert: Imho this is only true for the radio button. I can image several use cases where only one checkbox is needed. Examples:
- Custom content inside a table that can be checked
- "I agree to the terms of usage" with only 1 entry
- ...

packages/sit-onyx/src/components/OnyxCheckbox/OnyxCheckbox.vue
larsrickert: Good point, MDN said it must be set via JavaScript^^ It works totally fine in the template so I changed it. Thanks :)

packages/sit-onyx/src/index.ts
JoCa96: Right, but in that case you can also give a radio group with a single entry

JoCa96: I would prefer to only have a single public component that has a name with checkbox.  if it doesn't work out we can still make it public again :)

larsrickert: Alright I changed it :)

JoCa96 (approved):

jannick-ux (approved):

PR #289, opened 2024-02-13
docs: create hover and copy functionality for icon library

### LINKED ISSUE #242: Create Search+Copy+Hover Feature for docs OnyxIconLibrary
We want to create a search + copy feature in apps/docs/src/.vitepress/components/OnyxIconLibrary.vue
\+ Improve the hover behavior, currently we use the standard browser feature which might take 3secs until a hover reacts.

- [x] Search
- [x] Copy
- [x] Hover

Image

Downloading an svg is out of scope of this ticket.

BoppLi:
closes #242 

In the icon library documentation, we can now hover or tab through the icon list to see the item name in a tooltip immediately.
image
When clicking or hitting enter while an icon is hovered/focused, the import statement for the icon will be copied into the clipboard. The import statement will include a camelCase name to prevent naming collisions when importing multiple icons.
```ts
import bellRing from "@sit-onyx/icons/bell-ring.svg?raw";
```
image

## Checklist


apps/docs/src/.vitepress/components/IconLibraryItem.vue
```diff
@@ -0,0 +1,75 @@
...
+  await navigator.clipboard.writeText(
+    `import ${importName} from "@sit-onyx/icons/${iconName}.svg?raw";`,
+  );
+  isCopied.value = true;
+  setTimeout(() => (isCopied.value = false), 3000);
+};
+</script>
+
+<template>
+  <div class="icon" tabindex="0" @click="handleCopy" @keyup.enter="handleCopy">
```
larsrickert, line 23: If its clickable, I'd prefer to use a button instead. So we can also get rid of the tabindex and the keyup.enter

apps/docs/src/.vitepress/components/IconLibraryItem.vue
```diff
@@ -0,0 +1,75 @@
...
+<style lang="scss" scoped>
+.icon {
+  height: 3.5rem;
+  width: 3.5rem;
+  padding: var(--onyx-spacing-md);
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  cursor: pointer;
+  border: 0.0625rem solid transparent;
```
larsrickert, line 44:
```suggestion
  border: 1px solid transparent;
```

I think we should make an exception for 1px borders, what do you think?

larsrickert: Same for the border in &:hover below

apps/docs/src/.vitepress/utils-icons.ts
```diff
@@ -1,6 +1,23 @@
 import { ICON_CATEGORIES } from "@sit-onyx/icons";
 import { capitalize } from "vue";
 
+export type EnrichedIcon = {
+  metadata: {
+    aliases: string[] | undefined;
+    category: string;
+    deprecated?: boolean | undefined;
```
larsrickert, line 8:
```suggestion
    category: string;
    aliases?: string[];
    deprecated?: boolean;
```

larsrickert (changes requested):

apps/docs/src/.vitepress/components/IconLibraryItem.vue
```diff
@@ -0,0 +1,75 @@
...
+  isCopied.value = true;
+  setTimeout(() => (isCopied.value = false), 3000);
+};
+</script>
+
+<template>
+  <div class="icon" tabindex="0" @click="handleCopy" @keyup.enter="handleCopy">
+    <OnyxIcon
+      :icon="props.icon.content"
+      :color="props.icon.metadata.deprecated ? 'secondary' : 'currentColor'"
```
larsrickert, line 26: I know the color was not added in your PR but I don't think we should use secondary for deprecated because secondary is a colored hightlight color. Imho deprecated icons should be greyed out

BoppLi: does neutral count as grey?
image
image

larsrickert: Neutral is the default icon color^^ 
Maybe we could add a class if the icon is deprecated an then add custom CSS to change the color?

larsrickert: But you don't have to do this in this PR :)

BoppLi: nope. 
image
attachment is for test reasons deprecated but neutral is the same as currentcolor in this case

BoppLi: I'll ask jannick

BoppLi: short response: deprecated is no use case for now. no special color treatment

apps/docs/src/.vitepress/components/IconLibraryItem.vue
BoppLi: nice! thanks! done :)

PR #295, opened 2024-02-13
refactor(OnyxIcon)!: rewrite icon size map

### LINKED ISSUE #291: Follow-up: Icon size map with number values
Inside OnyxIcon, we need to update the ICON_SIZES map to reflext the pixel numbers so users will have an easier time to figure out which size is needed based on the figma design.

Image

Relates to #224

BoppLi:
closes #291 

Rewrites the icon size list to reflect the pixel values in Figma in order to make it more intuitiv for devs using OnyxIcon what to choose.

image

## Checklist


packages/sit-onyx/src/components/OnyxIcon/types.ts
```diff
@@ -6,11 +6,14 @@ export type OnyxIconProps = {
...
+   * Icon size in px. Will be translated to the according rem value.
+   * @default 24
+   */
   size?: IconSize;
   /** Icon color. */
   color?: OnyxColor | "currentColor";
 };
 
-export const ICON_SIZES = ["2xs", "xs", "sm", "md", "lg", "xl", "2xl"] as const;
+export const ICON_SIZES = ["12", "16", "24", "32", "48", "64", "96"] as const;
```
larsrickert, line 18: Should we change it to numbers instead of strings?

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxIcon/types.ts
JoCa96, line 18: I am still in favor of adding the "px" unit. A number without unit is meaningless. We have pixels in figma and these are pixel values, why hide that fact?

larsrickert: Fine for me :)

BoppLi: ok ^^ thought the description in the docs is enough ^^ (also I like the ability on project side to globally search for "px" to find out where px is used instead of rem ¯\\_(ツ)_/¯

BoppLi: you know what, I'll offer both: the ones with px and the ones with rem :)
because if (public) users have figma in dev mode set to rem, they'll see
image

larsrickert: Hmm I would not offer both...then you would have 2 options that do the same thing. We should decide on one

BoppLi: see other comment

larsrickert: I am also fine with keeping the t-shirt sizes, I mean we also use them everywhere else like in the spacing tokens, border radius tokens etc

BoppLi: just need to recreate the screenshots now :)

BoppLi: image

with supporting both, we support devs that have now access to figma dev mode aswell as those who use figma dev mode with rem

larsrickert: I highly vote against this :(

BoppLi: ok killed the rem options. people can still use the css variable after all.

larsrickert (approved):

PR #305, opened 2024-02-14
feat(OnyxCheckbox): add property `required`

larsrickert:
#220

Add property `required` to the OnyxCheckbox component.

## Checklist


jannick-ux (approved):

packages/sit-onyx/playwright/snapshots/components/OnyxCheckbox/invalid-unchecked-disabled--chromium-linux.png
BoppLi, line 1: @jannick-ux is this even something we allow? to have disabled + required + unchecked kind of seems illegal right? ^^ because the user can not change it?

or is this a case of "we leave the freedom" because devs could e.g. make the check automatically on some condition that something else happen?

packages/sit-onyx/src/components/OnyxCheckbox/OnyxCheckbox.ct.tsx
```diff
@@ -2,65 +2,77 @@ import { expect, test } from "../../playwright-axe";
...
-      <OnyxCheckbox label="Unchecked" />
-      <OnyxCheckbox label="Hover" />
-      <OnyxCheckbox label="Focus visible" />
-    </div>,
-  );
+  const testCases = [{ name: "default" }, { name: "required", required: true }];
 
-  // ACT
-  const accessibilityScanResults = await makeAxeBuilder().analyze();
+  for (const testCase of testCases) {
```
BoppLi, line 7: I guess this is the equivalent of test.each in the playwright world: https://playwright.dev/docs/test-parameterize#parameterized-tests 

_if_ you want to change the order like in the example :) but you don't have to of course :)

packages/sit-onyx/src/components/OnyxCheckbox/OnyxCheckbox.ct.tsx
```diff
@@ -2,65 +2,77 @@ import { expect, test } from "../../playwright-axe";
...
-  await checkboxes[2].focus();
+    const checkboxes = await component.getByRole("checkbox").all();
 
-  // ASSERT
-  for (const checkbox of checkboxes) {
-    await expect(checkbox).not.toBeChecked();
-  }
+    // ACT
+    await checkboxes[1].hover();
+    await checkboxes[2].focus();
```
BoppLi, line 27: wouldn't it make more sense to swap those around? so we would get a warning if our hover/focus styles break accessibility?

```suggestion
    const checkboxes = await component.getByRole("checkbox").all();
    await checkboxes[1].hover();
    await checkboxes[2].focus();
    
    // ACT
    const accessibilityScanResults = await makeAxeBuilder().analyze();    

    // ASSERT
    expect(accessibilityScanResults.violations).toEqual([]);
```

packages/sit-onyx/src/components/OnyxCheckbox/OnyxCheckbox.ct.tsx
```diff
@@ -2,65 +2,77 @@ import { expect, test } from "../../playwright-axe";
...
+    );
 
-  // ASSERT
-  expect(accessibilityScanResults.violations).toEqual([]);
+    // ACT
+    const accessibilityScanResults = await makeAxeBuilder().analyze();
 
-  const checkboxes = await component.getByRole("checkbox").all();
+    // ASSERT
+    expect(accessibilityScanResults.violations).toEqual([]);
```
BoppLi, line 60: see the other comment

packages/sit-onyx/src/components/OnyxCheckbox/OnyxCheckbox.ct.tsx
```diff
@@ -140,6 +156,62 @@ test("should render disabled", async ({ mount, makeAxeBuilder }) => {
...
+          required
+        />
+      </div>,
+    );
+
+    // ACT
+    const accessibilityScanResults = await makeAxeBuilder().analyze();
+
+    // ASSERT
+    expect(accessibilityScanResults.violations).toEqual([]);
```
BoppLi, line 200: see the other comment (order)

packages/sit-onyx/playwright/snapshots/components/OnyxCheckbox/invalid-unchecked-disabled--chromium-linux.png
BoppLi: also, "invalid unchecked disabled" can never get the red style @larsrickert because the red style only happens after touch, right? and you can not touch if it is disabled :(

image

packages/sit-onyx/src/components/OnyxCheckbox/OnyxCheckbox.vue
```diff
@@ -37,6 +44,33 @@ const isChecked = computed({
 </template>
 
 <style lang="scss">
+@mixin define-hover-border($state, $color) {
+  .onyx-checkbox__input {
+    &#{$state} {
+      border-color: var(--onyx-color-base-#{$color}-300);
+    }
+  }
+}
```
BoppLi, line 53: I guess you could make the selector a oneliner?

```suggestion
  .onyx-checkbox__input #{$state} {
    border-color: var(--onyx-color-base-#{$color}-300);
  }
}
```

packages/sit-onyx/src/components/OnyxCheckbox/OnyxCheckbox.vue
```diff
@@ -37,6 +44,33 @@ const isChecked = computed({
...
+
+@mixin define-focus-ring($state, $color) {
+  .onyx-checkbox__container {
+    &:has(.onyx-checkbox__input#{$state}) {
+      background-color: var(--onyx-color-base-#{$color}-200);
+    }
```
BoppLi, line 59:
```suggestion
  .onyx-checkbox__container :has(.onyx-checkbox__input#{$state}) {
      background-color: var(--onyx-color-base-#{$color}-200);
```

packages/sit-onyx/src/components/OnyxCheckbox/OnyxCheckbox.vue
```diff
@@ -26,9 +30,12 @@ const isChecked = computed({
       <input
         v-model="isChecked"
         class="onyx-checkbox__input"
+        :class="{ 'onyx-checkbox__input--touched': isTouched }"
```
BoppLi, line 33: fyi it's not uncommon with bem to move the modifier on the block level -> line 28
```html
  <label class="onyx-checkbox"
    :class="{ 'onyx-checkbox--touched': isTouched }"
  >
```

this might make you drop a few of the ":has" selectors


(just a suggestion, if it makes things more complicated, leave it as-is 🙂)

BoppLi (changes requested):

packages/sit-onyx/playwright/snapshots/components/OnyxCheckbox/invalid-unchecked-disabled--chromium-linux.png
larsrickert: I would leave it because the checkbox could be programatically disabled, e.g. while a form is loading or some async check is done

packages/sit-onyx/src/components/OnyxCheckbox/OnyxCheckbox.ct.tsx
larsrickert: I had flaky firefox tests with this order so I had to do the accessibility test before

packages/sit-onyx/src/components/OnyxCheckbox/OnyxCheckbox.vue
larsrickert: I had this at first but imho it made things more complicated :D

packages/sit-onyx/src/components/OnyxCheckbox/OnyxCheckbox.ct.tsx
larsrickert: Nice thanks! That looks better in the dev tools :) I also remove redundant test cases for the required indicator.
In don't think we need additional screenshots for unchecked, checked, indeterminate etc. which only include the additional asterisk. I changed it to one screenshot

packages/sit-onyx/src/components/OnyxCheckbox/OnyxCheckbox.ct.tsx
BoppLi: whoah.. then let's just hope that the hover and focus still have enough contrast etc ^^

packages/sit-onyx/src/components/OnyxCheckbox/OnyxCheckbox.vue
BoppLi: ok :)

packages/sit-onyx/playwright/snapshots/components/OnyxCheckbox/invalid-unchecked-disabled--chromium-linux.png
BoppLi: then should we maybe drop all screenshots for invalid+disabled as there look the same as just disabled anyways? 
there is not even a figma reference to compare because they can not be selected at once ^^

or should there be a way to force the red outline on disabled somehow? ^^ 

Let's ask Jannick in the daily pls :)

larsrickert: I added them because I want to ensure that the disabled has higher priority. Depending on how we write the CSS it can happen (and did while I implemented this PR) that invalid disabled checkboxes still had red hover effects etc. although they were disabled

BoppLi: ok then let's merge it but in the review session we should point it out ok? :) because it was not defined in figma so we don't know if this is the intended behavior :)

BoppLi (approved):

packages/sit-onyx/playwright/snapshots/components/OnyxCheckbox/invalid-unchecked-disabled--chromium-linux.png
larsrickert: Yes I will show it in the review :) I also asked Jannick, disabled has higher priority than all other states so even if its invalid, indeterminate etc., as soon as its disabled, it should be greyed out and the user can not interact with

jannick-ux (approved):

PR #319, opened 2024-02-15
Support headline style independent of headline level

oemueller:
There are cases when a headline style should be different from the headline level. 
Here is an example. The level should be "2", but the visual size of the headline should be level 3.

## Checklist

- [ ] All changes are documented in the documentation app (folder `apps/docs`)

### DIFF
--- a/packages/sit-onyx/src/components/OnyxHeadline/OnyxHeadline.vue
+++ b/packages/sit-onyx/src/components/OnyxHeadline/OnyxHeadline.vue
@@ -12,7 +12,7 @@ defineSlots<{
   default(): unknown;
 }>();
 
-const headlineSize = computed( () => (props.style || props.level)  );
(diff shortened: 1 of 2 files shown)


larsrickert:
Hey @oemueller, thanks for your contribution!
Your screenshot / use case is from another design system, right?
From UX perspective for onyx, we do not plan to have such inconsistencies for headlines because every headline has its dedicated use case:

- h1 = page headline
- h2 = section headline
- h3 = element headline
- h4 = label headline
 
Therefore, we do not plan to provide an "official" property for this. In exceptional cases, the developer can still override the style with custom CSS or by manually setting the corresponding class like so (although we discourage from doing so):

```html
<OnyxHeadline is="h2">
  <span class="onyx-headline--h3"> Content </span>
</OnyxHeadline>
```

PR #320, opened 2024-02-15
feat(OnyxButton): Add icon property to button

### LINKED ISSUE #190: Implement OnyxButton
### Why?

We want to provide a button component that the user can interact with to perform/trigger actions.

### Design

https://www.figma.com/file/YfEUBOHk4J4nYrk04geswG/onyx---Design-System-Figma-Library?type=design&node-id=633%3A20349&mode=dev&t=FFcU24NwNonvL6XX-1

### Acceptance criteria

- [x] a label can be shown
- the following states are implemented:
  - [x] default
  - [x] hover
  - [x] active
  - [x] focus-visible
  - [x] disabled
- the following colors are supported for all states mentioned above:
  - [x] primary (default)
  - [x] secondary
  - [x] danger
- the following variants are supported for all colors above:
  - [x] solid (default)
  - [x] outline
  - [x] plain
- [ ] an optional icon can be displayed on the left side of the label
- [ ] keyboard interaction is supported as documented in the ARIA pattern below
- [ ] Create follow-up tickets with dependencies for `out of scope` below

**Out of scope**
- an optional loading indicator can be shown
- skeleton state
- active animation
- truncation
- the following densities are supported:
  - compact
  - default
  - cozy
- a "icon-only" button is supported that only shows an icon and no label

### Definition of Done

- [ ] covered by component tests
- [ ] screenshot tests are updated
- [ ] follow up tickets were created if necessary
- [ ] Storybook can show the feature
- [ ] Set task to `In review`

**Review**
    -  _add link to deployed storybook_
    - _add link to deployed docs_
    - [ ] approved by designer _add designer tag_
    - [ ] approved by dev _add dev tag_

### Implementation details

```ts
type ButtonProps = {
        label?: string;
        isLoading?: boolean;
        isDisabled?: boolean;
        /**
         * Standard html attribute, we want it to defaults to 'button'
         */
        type?: "button" | "submit" | "reset";
        variant?: "primary" | "secondary" | "danger";
        icon?: string;
    }
```

- Consider fallthrough props https://vuejs.org/guide/components/attrs.html#disabling-attribute-inheritance
- [MDN link](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button)
- the following standard HTML properties are supported:
  - [ ] [disabled](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button?retiredLocale=de#disabled)
  - [ ] [type](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button?retiredLocale=de#type)
  - [ ] [autofocus](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autofocus)
  - [ ] Even if we offer boolean property in Figma instead we use a slot as implementation
 
#### Reference implementations

- [Element Plus](https://element-plus.org/en-US/component/button.html)
- [WaWi](https://dev.azure.com/schwarzit/lidl.wawi-core/_git/lidl-wawi?path=/ui/components/src/components/WuxButton/WuxButton.vue)

#### Applicable ARIA Pattern

https://www.w3.org/WAI/ARIA/apg/patterns/button/

MajaZarkova:
closes #190

Add icon property to button

## Checklist

- [X] The added / edited code has been documented with [JSDoc](https://jsdoc.app/about-getting-started)
- [X] All changes are documented in the documentation app (folder `apps/docs`)


packages/sit-onyx/src/components/OnyxButton/OnyxButton.stories.ts
```diff
@@ -1,8 +1,23 @@
 import { defineStorybookActionsAndVModels } from "@sit-onyx/storybook-utils";
+import checkSmall from "@sit-onyx/icons/check-small.svg?raw";
 import type { Meta, StoryObj } from "@storybook/vue3";
 import OnyxButton from "./OnyxButton.vue";
 import { BUTTON_VARIATIONS, BUTTON_TYPES, BUTTON_MODES } from "./types";
 
+const ALL_ICONS = import.meta.glob("../../../node_modules/@sit-onyx/icons/src/assets/*.svg", {
```
larsrickert, line 7: We also have this for the OnyxIcon, right? Can we extract this to a utility so we can re-use for future icon props as well?

packages/sit-onyx/src/components/OnyxButton/OnyxButton.vue
```diff
@@ -147,5 +149,9 @@ const props = withDefaults(defineProps<ButtonProps>(), {
     font-weight: 600;
     line-height: 1.5rem;
   }
+
+  &__icon {
+    color: var(--onyx-button-text-color);
```
larsrickert, line 154: I'd prefer to set the color directly on the "onyx-button" class instead of on the label only. Then we don't need to set it for the icon here because it will just inherit the font color

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxButton/OnyxButton.stories.ts
larsrickert: You can use #337 once its merged :)

larsrickert (dismissed):

mj-hof (approved):

PR #333, opened 2024-02-16
chore(OnyxRadioButtonGroup): Use OnyxHeadline in OnyxRadioButtonGroup

JoCa96:
#222 

Use Headline in RadioButtonGroup

### DIFF
--- a/packages/sit-onyx/src/components/OnyxRadioButtonGroup/OnyxRadioButtonGroup.vue
+++ b/packages/sit-onyx/src/components/OnyxRadioButtonGroup/OnyxRadioButtonGroup.vue
@@ -2,6 +2,7 @@
 import type { TargetEvent } from "@/types/dom";
 import OnyxRadioButton from "../OnyxRadioButton/OnyxRadioButton.vue";
 import type { SelectionOption } from "../OnyxRadioButton/types";
+import OnyxHeadline from "../OnyxHeadline/OnyxHeadline.vue";
 
 type ChangeEvent = TargetEvent<HTMLInputElement>;
 
@@ -12,7 +13,10 @@ const props = defineProps<{
    */
   name: string;
   modelValue?: SelectionOption<T>;
-  label?: string;
+  /**
+   * Headline and legend shown above the radio button group
+   */
+  headline?: string;
   required?: boolean;
   disabled?: boolean;
   readonly?: boolean;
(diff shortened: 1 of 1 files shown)


packages/sit-onyx/src/components/OnyxRadioButtonGroup/OnyxRadioButtonGroup.vue
```diff
@@ -37,7 +41,9 @@ const handleChange = (event: ChangeEvent) =>
     :disabled="props.disabled"
     @change="handleChange($event as ChangeEvent)"
   >
-    <legend v-if="props.label" class="onyx-radio-button-group__label">{{ props.label }}</legend>
+    <legend v-if="props.headline" class="onyx-radio-button-group__label">
```
larsrickert, line 44:
```suggestion
    <legend v-if="props.headline" class="onyx-radio-button-group__headline">
```

Would prefer to also rename the CSS class

packages/sit-onyx/src/components/OnyxRadioButtonGroup/OnyxRadioButtonGroup.vue
```diff
@@ -12,7 +13,10 @@ const props = defineProps<{
    */
   name: string;
   modelValue?: SelectionOption<T>;
-  label?: string;
+  /**
+   * Headline and legend shown above the radio button group
```
larsrickert, line 17: Its only one headline, right? Not both a headline and legend
```suggestion
   * Headline to show above the radio button group
```

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxRadioButtonGroup/OnyxRadioButtonGroup.vue
JoCa96: Yes, I clarified it. Fine like that? :)

packages/sit-onyx/src/components/OnyxRadioButtonGroup/OnyxRadioButtonGroup.vue
larsrickert: I don't have a preference on that^^ I guess UX would call it headline

larsrickert (approved):

PR #342, opened 2024-02-16
chore: Implement State Screenshot testing utility

### LINKED ISSUE #341: Implement Matrix Screenshot Util

JoCa96:
Implemented a playwright utility that allwos us to create a screenshot of a matrix with all relevant component states and variations. This will make it easier for UX to review the screenshots and reduce test execution time and screenshot storage.

Added screenshot tests for #222 

closes #341


packages/sit-onyx/src/components/OnyxButton/OnyxButton.ct.tsx
```diff
@@ -48,3 +49,23 @@ test("should render button with icon", async ({ mount, makeAxeBuilder }) => {
...
+
+test(
+  "Screenshot matrix",
+  createMatrixScreenshot(STATES, "matrix.png", ({ variation, state, mode }) => (
+    <OnyxButton
+      label={"label"}
+      variation={variation}
+      mode={mode}
+      disabled={state === "disabled"}
+      icon={state === "icon" ? happyIcon : ""}
```
larsrickert, line 68:
```suggestion
      label="label"
      variation={variation}
      mode={mode}
      disabled={state === "disabled"}
      icon={state === "icon" ? happyIcon : undefined}
```

packages/sit-onyx/src/utils/playwright.tsx
```diff
@@ -0,0 +1,177 @@
...
+ * "colHeadline1 case3 case4 case5"
+ * "colHeadline2 case6 case7 case8"
+ * "colHeadline3 case9 case10 case11"
+ * ```
+ */
+const buildGridTemplate = (rows: number, columns: number) =>
+  Array.from({ length: rows + 1 })
+    .map((_, ri) =>
+      Array.from({ length: columns + 1 })
+        .map((_, ci) => {
```
larsrickert, line 58: Would prefer to call the variables "row" and "column" so its easier understandable what they are

packages/sit-onyx/src/utils/playwright.tsx
```diff
@@ -0,0 +1,177 @@
...
+type CaseBuilder<S extends ComponentStates> = (c: Permutation<S>, i: number) => JSX.Element;
+
+/**
+ * Generates a test case in the grid and places it accordingly.
+ */
+const createTestCases = <S extends ComponentStates>(
+  testCases: Permutation<S>[],
+  caseBuilder: CaseBuilder<S>,
+) =>
+  testCases.map((tC, i) => (
```
larsrickert, line 79: Would prefer "testCase" instead of "tc"

packages/sit-onyx/src/utils/playwright.tsx
```diff
@@ -0,0 +1,177 @@
...
+    <div
+      style={{ gridArea: `case${i}` }}
+      key={i}
+      // Handle common cases
+      class={{
+        [`pw-${tC["focusState"]}`]: tC["focusState"],
```
larsrickert, line 85: Would prefer a data attribute like `data-test-${tc.focusState}` so its clear that it is only used for testing

packages/sit-onyx/src/utils/playwright.tsx
```diff
@@ -0,0 +1,177 @@
...
+/**
+ * Generates the jsx for the screenshot matrix test.
+ * It uses a grid with named areas to place all test cases, the row and column headlines.
+ */
+const generateScreenshotMatrix = <S extends ComponentStates>(
+  states: S,
+  caseBuilder: CaseBuilder<S>,
+) => {
+  const [rowStates] = Object.values(states);
+  const colStates = generatePermutations(Object.fromEntries(Object.entries(states).slice(1))).map(
```
larsrickert, line 102: Imho this line is a little hard to understand, maybe a short explanatory comment or extract to a variable might help?

packages/sit-onyx/src/utils/playwright.tsx
```diff
@@ -0,0 +1,177 @@
...
+        ))
+      }
+      {createTestCases(testCases, caseBuilder)}
+    </main>
+  );
+};
+
+type TestArg = Parameters<Parameters<typeof test>[1]>[0];
+
+export const createMatrixScreenshot =
```
larsrickert, line 157: Some JSDoc would be nice so its easier when other devs use this in the future

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxCheckbox/OnyxCheckbox.ct.tsx
```diff
@@ -213,3 +214,23 @@ for (const testCase of invalidTestCases) {
     await expect(component).toHaveScreenshot(`invalid-${testCase.name}.png`);
   });
 }
+
+const STATES = {
+  state: ["default", "disabled", "invalid", "required", "optional"],
```
larsrickert, line 219: Invalid screenshots does not seem to work, they should be read.
Invalid only triggers when the checkbox is touched so it needs to be blurred first

larsrickert:
In general I like the matrix screenshots but I am concerned that the reliability is no longer guaranteed then...because then we  are basically only testing custom CSS selectors instead of actually hovering etc.

packages/sit-onyx/src/utils/playwright.tsx
JoCa96: done

packages/sit-onyx/src/utils/playwright.tsx
JoCa96: done

larsrickert (approved):

Issue #191, opened 2024-01-24
Implement OnyxSwitch


JoCa96:
### Dependencies
- [ ] #244
- [ ] #368 

### Why?
A switch is visually and semantically distinct from a checkbox.
The switch allows switching something *on* or *off*, rather than selecting something.

### Design
[Figma Design](https://www.figma.com/file/YfEUBOHk4J4nYrk04geswG/onyx---Design-System-Figma-Library?node-id=711%3A12812&mode=dev)

### Acceptance criteria
- [x] can be toggled on click and updates modelValue
- [x] the toggling is animated
- [x] the following styles are implemented:
  - [x] :hover
  - [x] :focus-visible
  - [x] :disabled
  - [x] :checked
- [x] can be used with or without label
- [x] must accessible even without label (using aria-label)

#### Out of scope
- Support number and string values
- Ability to change on/off icon

### Definition of Done
- [x] covered by component tests
- [x] screenshot tests are updated
- [x] follow-up tickets were created if necessary
- [x] updated version + documentation is deployed
  - [x] _approved by dev @larsrickert 
- [x] Storybook can show the feature
- [View in StoryBook](https://onyx-storybook-prod.apps.01.cf.eu01.stackit.cloud/?path=/docs/components-onyxswitch--docs)
  - [x] _approved by designer @jannick-ux 

### Implementation details
- Implement using a checkbox (See reference implementations:  [ARIA APG](https://www.w3.org/WAI/ARIA/apg/patterns/switch/examples/switch-checkbox/))
- `hideLabel` uses [aria-label](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label) instead of a visual label

```ts
type OnyxSwitchProps = {
  modelValue?: boolean;
  label: string;
  hideLabel?: boolean;
  disabled?: boolean;
  required?: boolean;
}
```

#### Reference implementations
- [ARIA APG](https://www.w3.org/WAI/ARIA/apg/patterns/switch/examples/switch-checkbox/)
- [Vuetify](https://vuetifyjs.com/en/components/switches/)
- [Nuxt](https://ui.nuxt.com/components/toggle)
- [KendoUI](https://www.telerik.com/kendo-vue-ui/components/inputs/switch/)

#### Applicable ARIA Pattern
- https://www.w3.org/WAI/ARIA/apg/patterns/switch/

MajaZarkova:
There is no UX design for errorMessage, or am I missing something?

[...]

Issue #339, opened 2024-02-16
Remove readonly state 


JoCa96:
### Why?

The HTML standard only supports readonly for text inputs and textareas, so we should remove any readonly states from other components.

### Acceptance criteria

- in Figma, all readonly designs/states are removed from all components except the input and textarea

### Definition of Done

- [ ] covered by component tests
- [ ] screenshot tests are updated
- [ ] follow up tickets were created if necessary
- [ ] updated version + documentation is deployed
- [ ] Storybook can show the feature
- [ ] Set task to `In review`

**Review**

- _add link to deployed storybook_
- _add link to deployed docs_
- [ ] approved by designer @jannick-ux
- [ ] approved by dev _add dev tag_

mj-hof:
Re-refine ticket again for removing readonly

PR #419, opened 2024-02-26
feat(OnyxSwitch): Implement Base OnyxSwitch component

MajaZarkova:
relates to #191 

Implement basic OnyxSwitch

## Checklist

- [X] The added / edited code has been documented with [JSDoc](https://jsdoc.app/about-getting-started)
- [X] All changes are documented in the documentation app (folder `apps/docs`)
- [X] If a new component is added, at least one [Playwright screenshot test](https://github.com/SchwarzIT/onyx/actions/workflows/playwright-screenshots.yml) is added


packages/sit-onyx/src/components/OnyxSwitch/types.ts
```diff
@@ -0,0 +1,31 @@
...
+   */
+  label?: string;
+  /**
+   * Whether to disable the switch and prevent user interaction.
+   */
+  disabled?: boolean;
+  /**
+   * Whether the switch should be readonly or not.
+   */
+  readonly?: boolean;
```
JoCa96, line 17: We don't have a concept for the read-only mode for now, so please remove it for now

packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.vue
```diff
@@ -0,0 +1,274 @@
+<script lang="ts" setup>
+import { OnyxIcon } from "@/index";
+import checkSmall from "@sit-onyx/icons/check-small.svg?raw";
+import xSmall from "@sit-onyx/icons/x-small.svg?raw";
+import type { OnyxSwitchProps } from "./types";
+import { computed } from "vue";
+const props = withDefaults(defineProps<OnyxSwitchProps>(), {
+  modelValue: false,
+  label: "Onyx Switch",
```
JoCa96, line 9: Please remove the default value

packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.vue
```diff
@@ -0,0 +1,274 @@
...
+
+        .onyx-switch__icon {
+          color: var(--onyx-color-text-icons-primary-soft);
+        }
+      }
+    }
+  }
+
+  &__container {
+    position: relative;
```
JoCa96, line 176: Please try to avoid `position: relative` and `position: absolute` if they are not absolutely necessary

packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.vue
```diff
@@ -0,0 +1,274 @@
...
+      position: absolute;
+      display: flex;
+      flex-wrap: wrap;
+      align-content: normal;
+      left: 0.125rem;
+      top: 0.125rem;
+      right: 0.125rem;
+      bottom: 0.125rem;
+      width: 1.25rem;
+      height: 1.25rem;
```
JoCa96, line 201: Please make use of flex `justify-content` and `align-items`, as well as `padding`. Absolute positioning must be avoided.

packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.vue
```diff
@@ -0,0 +1,274 @@
...
+      :disabled="props.disabled || props.readonly"
+      :required="props.required"
+    />
+    <span class="onyx-switch__container">
+      <span class="onyx-switch__track"></span>
+      <span class="onyx-switch__position">
+        <OnyxIcon class="onyx-switch__icon" :icon="isChecked ? checkSmall : xSmall" size="24px" />
+      </span>
+    </span>
+    <span v-if="props.label && !props.hideLabel" class="onyx-switch__label">{{ props.label }}</span>
```
JoCa96, line 56: When label is mandatory:

```suggestion
    <span v-if="!props.hideLabel" class="onyx-switch__label">{{ props.label }}</span>
```

packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.vue
```diff
@@ -0,0 +1,274 @@
...
+
+      .onyx-switch__icon {
+        color: var(--onyx-color-text-icons-neutral-soft);
+      }
+    }
+  }
+
+  &__label {
+    padding: var(--onyx-spacing-2xs) 0;
+    align-items: flex-end;
```
JoCa96, line 217: Shouldn't be necessary and doesn't have an effect
```suggestion
```

packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.vue
```diff
@@ -0,0 +1,274 @@
...
+  <label
+    tabindex="0"
+    class="onyx-switch"
+    :class="{ 'onyx-required-marker': props.required, 'onyx-optional-marker': !props.required }"
+  >
+    <input
+      v-model="isChecked"
+      class="onyx-switch__input"
+      :class="{
+        'onyx-switch__input--invalid': props.invalid,
```
JoCa96, line 41: Please use the [`setCustomValidity`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/setCustomValidity) function to set an error. Then you can use the `:invalid` pseudo-class.

packages/sit-onyx/src/components/OnyxSwitch/types.ts
```diff
@@ -0,0 +1,31 @@
...
+  /**
+   * Whether the switch is required / has to be checked.
+   */
+  required?: boolean;
+  /**
+   * Whether the switch should be invalid or not.
+   */
+  invalid?: boolean;
```
JoCa96, line 25: Please name the prop `errorMessage` instead.
```suggestion
  errorMessage?: string;
```

JoCa96 (changes requested):

packages/sit-onyx/src/components/OnyxSwitch/types.ts
MajaZarkova: image

What is the fourth column for?

packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.vue
```diff
@@ -0,0 +1,274 @@
...
+    &:has(.onyx-switch__input--invalid:checked:enabled) .onyx-switch__track {
+      outline: 0.25rem solid var(--onyx-color-base-danger-200);
+    }
+  }
+
+  &:has(.onyx-switch__input:disabled) {
+    cursor: default;
+  }
+
+  &:has(.onyx-switch__input--disabled) .onyx-switch__label {
```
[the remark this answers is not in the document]
larsrickert: The question is if we even want to support readonly...we have the same issue for radio and checkbox where readonly is natively not supported also other other component libraries don't support it.

Jannick already said he does not have a strong opinion on that. 
So my suggestion would be to not support readonly for checkboxes, radio and switch. But lets talk about it in the daily :)

packages/sit-onyx/src/components/OnyxSwitch/types.ts
JoCa96: Sorry for my imprecision, I meant to say: We don't have a implementation concept for the `readonly` mode yet and didn't decide on how and if we implement it. Relevant ticket: [#339](https://github.com/SchwarzIT/onyx/issues/339)

packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.vue
MajaZarkova: @JoCa96 I implemented the setCustomValidity function, but I'm note sure if I did it right?

packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.vue
```diff
@@ -0,0 +1,229 @@
...
+  // as long as it is not an empty string
+  inputElement.value.setCustomValidity(props.errorMessage || "");
+});
+
+watch(
+  [inputElement, isChecked, errorMessage],
+  () => {
+    if (!inputElement.value) return;
+
+    const newValidityState = transformValidityStateToObject(inputElement.value!.validity);
```
larsrickert, line 45:
```suggestion
    const newValidityState = transformValidityStateToObject(inputElement.value.validity);
```

packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.vue
```diff
@@ -0,0 +1,229 @@
...
+
+const emit = defineEmits<{
+  /** Emitted when the checked state changes. */
+  "update:modelValue": [value: boolean];
+  /** Emitted whenever the validity state of the input changes */
+  validityChange: [state: ValidityState];
+}>();
+
+const { errorMessage } = toRefs(props);
+const inputElement = ref<HTMLInputElement | null>(null);
```
larsrickert, line 23:
```suggestion
const inputElement = ref<HTMLInputElement>();
```

packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.vue
```diff
@@ -0,0 +1,229 @@
...
+      emit("validityChange", validityState.value);
+    }
+  },
+  { immediate: true },
+);
+</script>
+
+<template>
+  <label
+    tabindex="0"
```
larsrickert, line 58: Why is tabindex needed?

packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.vue
```diff
@@ -0,0 +1,229 @@
...
+<style lang="scss">
+.onyx-switch {
+  display: inline-flex;
+  align-items: center;
+  cursor: pointer;
+  margin: var(--onyx-spacing-2xs);
+  height: 1.5rem;
+
+  &__input {
+    position: absolute;
```
larsrickert, line 89: Is absolute needed? Does it even have an effect without defining top/left? Also the parent should be relative then, right?

packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.vue
```diff
@@ -0,0 +1,229 @@
...
+  }
+
+  &__container {
+    display: inline-flex;
+    margin: var(--onyx-spacing-2xs);
+    width: 2.375rem;
+    height: 1.5rem;
+    background-color: var(--onyx-color-base-neutral-300);
+    border-radius: var(--onyx-radius-full);
+    transition: background-color 0.2s ease;
```
larsrickert, line 150:
```suggestion
    transition: background-color var(--onyx-duration-sm) ease;
```

packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.vue
```diff
@@ -0,0 +1,229 @@
...
+
+    .onyx-switch__icon {
+      display: flex;
+      flex-wrap: wrap;
+      align-content: baseline;
+      align-self: center;
+      justify-content: flex-end;
+      margin: 0.125rem;
```
larsrickert, line 158:
```suggestion
      margin: var(--onyx-spacing-5xs);
```

packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.vue
```diff
@@ -0,0 +1,229 @@
...
+      width: 1.25rem;
+      height: 1.25rem;
+      background-color: var(--onyx-color-themed-neutral-100);
+      border-radius: 50%;
```
larsrickert, line 162:
```suggestion
      border-radius: var(--onyx-radius-full);
```

packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.vue
```diff
@@ -0,0 +1,229 @@
...
+      transition:
+        transform 0.2s ease,
+        background-color 0.2s ease;
```
larsrickert, line 165:
```suggestion
      transition:
        transform var(--onyx-duration-sm) ease,
        background-color var(--onyx-duration-sm) ease;
```

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.vue
larsrickert: But <input /> elements are already focussable so we don't need to set the tabindex manually :)

MajaZarkova: When I remove the tabindex I cannot focus it. Does this only happen in storybook?

packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.vue
MajaZarkova: When I remove the position it makes the label wider

larsrickert: Ah this is the native checkbox, right :) Maybe we can add a small comment that position is needed to hide the native checkbox?

packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.vue
larsrickert: Thats just styling because the focus-visible is then on the input instead of the label (which should be the right behavior):

You can change the selector in line 200:
```css
&:focus-visible // old
&:has(&__input:focus-visible) // new
```

packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.vue
MajaZarkova: Yup, I'll do that

packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.ct.tsx
```diff
@@ -0,0 +1,47 @@
...
+  // ASSERT
+  expect(accessibilityScanResults.violations).toEqual([]);
+});
+
+test("should render disabled OnyxSwitch", async ({ mount, makeAxeBuilder }) => {
+  // ARRANGE
+  const component = await mount(<OnyxSwitch label="Switch" disabled />);
+
+  // ASSERT
+  await expect(component).toContainText("Switch");
```
larsrickert, line 24:
```suggestion
  await expect(component).toContainText("Switch");
  await expect(component).toBeDisabled(); 
```

packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.stories.ts
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * This example shows the invalid state of the switch.
+ */
+export const Invalid = {
+  args: {
+    ...DefaultChecked.args,
+    errorMessage: "Error message",
+  },
+} satisfies Story;
```
larsrickert, line 48:
```suggestion
/**
 * This example shows the default switch.
 */
export const Default = {
  args: {
    label: "Switch",
  },
} satisfies Story;

/**
 * This example shows the switch that is initially checked.
 */
export const Checked = {
  args: {
    label: "Switch",
    modelValue: true,
  },
} satisfies Story;

/**
 * This example shows the invalid state of the switch.
 */
export const Invalid = {
  args: {
    ...Checked.args,
    errorMessage: "Error message",
  },
} satisfies Story;
```

I think its more intuitive if the switch is unchecked in the default example

packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.vue
```diff
@@ -0,0 +1,230 @@
+<script lang="ts" setup>
+import { OnyxIcon } from "@/index";
+import checkSmall from "@sit-onyx/icons/check-small.svg?raw";
+import xSmall from "@sit-onyx/icons/x-small.svg?raw";
+import type { OnyxSwitchProps } from "./types";
+import { areObjectsFlatEqual } from "@/utils/comparator";
+import { transformValidityStateToObject } from "@/utils/forms";
+import { computed, ref, toRefs, watch } from "vue";
+const props = withDefaults(defineProps<OnyxSwitchProps>(), {
```
larsrickert, line 9:
```suggestion

const props = withDefaults(defineProps<OnyxSwitchProps>(), {
```

I'd like to add a new line :)

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.vue
MajaZarkova: image


I don't think this is the desired result 😄

packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.ct.tsx
MajaZarkova: I'll add this one, but I have prepared more tests for the second PR

larsrickert (approved):

packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.vue
```diff
@@ -0,0 +1,230 @@
...
+        color: var(--onyx-color-text-icons-neutral-inverted);
+      }
+    }
+
+    &:invalid:checked + .onyx-switch__container {
+      background-color: var(--onyx-color-base-danger-500);
+
+      .onyx-switch__icon {
+        background-color: var(--onyx-color-base-background-blank);
+        transform: translateX(12px);
```
JoCa96, line 138: All the sizing are given in `rem`, so we should probably give a rem value here, too.

JoCa96 (approved):

Issue #237, opened 2024-02-01
Implement basic Link component + use it in the docs


larsrickert:
Replace the [docs typography: link section](https://onyx.schwarz/basics/typography.html#links) with a OnyxLink component.

### Why?

The typography preview currently only shows style without having a `OnyxLink` component in use here which was added in #211.
To make sure that we won't diverge with the typography docs and a future link component, we want to create the basic link already now.

### Design

[Figma](https://www.figma.com/file/YfEUBOHk4J4nYrk04geswG/onyx---Design-System-Figma-Library?type=design&node-id=1213-34638&mode=dev)

### Acceptance criteria

- The link component
  - [x] has a label (which can be marked), and it is a slot.
  - [x] can be clicked [via mouse or keyboard](https://www.w3.org/WAI/ARIA/apg/patterns/link/)
  - [x] make href required
  - [x] for now: does not support file type, or icon examples in storybook
- [x] OnyxLink is used in [docs typography: link section](https://onyx.schwarz/basics/typography.html#links)

### Definition of Done

- [x] covered by component tests
- [x] screenshot tests are updated
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature
- [x] Set task to `In review`

**Review**

- [Storybook link](https://storybook.onyx.schwarz/?path=/docs/components-onyxlink--docs)
- [x] approved by designer @mj-hof 
- [x] approved by dev @BoppLi 

### Implementation details

- OnyxLink
  - stay close to the `<a>`-tag: [MDN Link](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a)
  - offer a property [`target`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#target). if `blank` is selected, preset `rel="noreferrer"` on the `<a>`-tag, [see MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/noreferrer) and [this article](https://medium.com/@halilatilla/target-blank-rel-noopener-noreferrer-b43e9ec05ade) for background info
- relevant component in the docs is: `apps/docs/src/.vitepress/components/OnyxTypography.vue`

#### Reference implementations

- [CoreUI](https://dev.azure.com/schwarzit/schwarzit.core-ui/_git/schwarz.core-ui?path=/packages/core-ui/src/components/scu-link)

#### Applicable ARIA Pattern

[Aria Link](https://www.w3.org/WAI/ARIA/apg/patterns/link/)

[...]

Issue #294, opened 2024-02-13
Implement new component feature in documentation


mj-hof:
### Depends on

- [x] Depends on [#237](https://github.com/SchwarzIT/onyx/issues/237)

### Why?

Due to a top level decision, our components shouldn't be structured into different sets on our documentation start page, so we have to change the visualization of this feature.

### Design

[Figma Link](https://www.figma.com/file/yPj9gRAHzqY1NPky9k8Voz/onyx-docu?type=design&node-id=0%3A1&mode=design&t=4PE7fmUc97CxGw9a-1)

### Acceptance criteria

- [x]  Remove the tab layout
- Initially, 12 component tiles are shown
  - [x] 4 Tiles per row in the default layout. Make it adjust responsively
  - [x] Try out how 12 tiles "feel" on a mobile device. Reduce number of initial tiles for mobile if needed.
- [x] With a "more" option, the user can switch between showing the 12 or all component tiles
- [x] Order the tiles on the implementation status: `Implemented` first, then `In Progress`, then `Planned`
- [x] Update the intro text above the tiles
- [x] Provide a legend for that status badges

### Definition of Done

- [x] updated version + documentation is deployed
- [x] Set task to `In review`

**Review**

- [Documentation link](https://onyx.schwarz)
- [x] approved by designer @mj-hof 
- [x] approved by dev @BoppLi 

### Implementation details

- stay close to the current implementation
- the relevant component for the tiles is `apps/docs/src/.vitepress/components/ComponentCard.vue`
- the relevant components for the layout are `apps/docs/src/.vitepress/components/OnyxRoadmap.vue` and `apps/docs/src/.vitepress/components/ComponentGrid.vue`

mj-hof:
Please also change or add some other status chips, i.e. like planned, implemented, work in progress, partly implemented

BoppLi:
Note from the Daily:

- [ ] use plain button instead of the OnyxLink for `show less` / `show more`

PR #426, opened 2024-02-27
feat: add `OnyxLink` component

larsrickert:
relates to #237

Add a default `OnyxLink` component. States like hover, active etc. will be implemented in a follow-up PR once they are clarified with UX.

## Checklist


packages/sit-onyx/src/components/OnyxLink/types.ts
```diff
@@ -0,0 +1,24 @@
...
+   */
+  target?: LinkTarget;
+  /**
+   * Text size.
+   */
+  size?: TextSize;
+  /**
+   * If `true`, the monospace font family will be used instead of the default one.
+   */
+  monospace?: boolean;
```
JoCa96, line 20: We shouldn't use properties for styling. The component will mainly be used inline in running text, where the font styling should come from the surrounding text. With this approach, the user would need to always manually sync text block style and link style.

JoCa96 (changes requested):

packages/sit-onyx/src/components/OnyxLink/types.ts
larsrickert: Done

JoCa96 (approved):

packages/sit-onyx/src/components/OnyxLink/OnyxLink.vue
```diff
@@ -0,0 +1,43 @@
+<script lang="ts" setup>
+import type { OnyxLinkProps } from "./types";
+
+const props = withDefaults(defineProps<OnyxLinkProps>(), {
```
BoppLi, line 4: didn't we want to set href to empty string as default?
image

e.g. in https://github.com/SchwarzIT/onyx/issues/294 the link is only used as click handler without navigating thus the href does not have to be set

BoppLi (changes requested):

packages/sit-onyx/src/components/OnyxLink/OnyxLink.vue
BoppLi: Ah, it's conform with the ticket AC of #237 . guess we'll then need to set the empty href when implementing #294 on that instance.

BoppLi (dismissed):

packages/sit-onyx/src/components/OnyxLink/OnyxLink.vue
BoppLi: question goes to @mj-hof

BoppLi (approved):

Issue #374, opened 2024-02-20
Implement layout component for z-index demo


BoppLi:
### Q&A

- do we want to use Teleport in the layout component or try to work without it?
  - [x] as long as we can not use the popover api, we will stick to Teleport for popups.
- should the temporary overlay be behind or above the top-bar?
  - [x] it is a variation of the side-bar, thus it is lower than the nav-bar.
- Do the toasts need to cover the page content? or can the page content also "shrink" above it?
  - [x] it will be decoupled from the bottom bar, floating in the middle of the page. it will overlay the page content.
- Should the POC Page be available similar to the Storybook Grid Playground?
  - maybe later, when all components are there in onyx

### Depends on

- [x] Depends on #108

### Why?

We want to proide a reusable App Layout component that covers the structure of a standard onyx website.   
During #108, it got clear that z-indices alone will not provide a stable structure and can not prevent a clash of multiple overlay components.

### Acceptance criteria

- [x] create app layout component(s)
   - [x] -> storybook should categorize it as "layout component"
- [x] use in `apps/alpha-test-app/src/views/LayoutDemoView.vue`
  - [x] switch out e.g. custom checkbox with OnyxCheckbox, button with OnyxButton
- [x] create documentation

### Definition of Done

- [ ] ~covered by component tests~ -> [#485](https://github.com/SchwarzIT/onyx/issues/485)
- [ ] ~screenshot tests are updated~ -> [#485](https://github.com/SchwarzIT/onyx/issues/485)
- [x] follow-up tickets were created if necessary
  - [#455](https://github.com/SchwarzIT/onyx/issues/455)
  - [#472](https://github.com/SchwarzIT/onyx/issues/472)
  - [#485](https://github.com/SchwarzIT/onyx/issues/485)
  - [#492](https://github.com/SchwarzIT/onyx/issues/492)
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature
- [x] Set task to `In Approval`

**Review**

- [storybook app layout](https://storybook.onyx.schwarz/?path=/docs/layout-onyxapplayout--docs)
- [storybook page layout](https://storybook.onyx.schwarz/?path=/docs/layout-onyxpagelayout--docs)
- [x] approved by designer @jannick-ux
- [x] approved by dev @larsrickert 

### Implementation details

- [x] offer slots for
  - [x] top nav
  - [x] side bar
  - [x] footer + toast
  - [x] full page layovers
  - [x] page layovers below top nav

[...]

Issue #376, opened 2024-02-20
Align Selection Control API


JoCa96:
### Why?

We want to make the UI feel cohesive and should use unified APIs and types 

### Acceptance criteria

- [x] Align API for options property of all selection components
- [x] Align API for modeValue property for radio button and checkbox

#### Follow up tickets:

- #951 

### Definition of Done

- [x] covered by component tests
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature
- [x] Set task to `In approval`

**Review**

- [x] approved by dev @JoCa96

PR #427, opened 2024-02-27
feat(demo-app): use onyx components for settings

BoppLi:
relates to #374

Replaces the settings of the layer demo page with OnyxRadioButtonGroup and OnyxCheckboxGroup.
Replaces the headlines with OnyxHeadline.
Replaces buttons with OnyxButton.
Removes the full page overlay, this will not be part of onyx.

## Checklist


apps/alpha-test-app/src/components/layout-demo/LayoutSettings.vue
```diff
@@ -0,0 +1,165 @@
...
+  showTempOverlay?: boolean;
+
+  detailFooter?: boolean;
+  fullFooter?: boolean;
+
+  showPopover?: boolean;
+  showMobileFlyIn?: boolean;
+  showPageLoader?: boolean;
+  topBarFlyout?: boolean;
+};
```
larsrickert, line 28:
```suggestion
export type Settings = Partial<{
  longPageContent: boolean;
  showFlyout: boolean;
  forceTooltip: boolean;
  showStickyContent: boolean;
  showToast: boolean;

  showSideBar: boolean;
  showSideBarCollapse: boolean;
  showTempOverlayTransparent: boolean;
  showTempOverlay: boolean;

  detailFooter: boolean;
  fullFooter: boolean;

  showPopover: boolean;
  showMobileFlyIn: boolean;
  showPageLoader: boolean;
  topBarFlyout: boolean;
}>;
```

(optional) we can wrap the type in `Partial` so we don't need `?` for every key

larsrickert: Is there a reason why some keys start with "show" and others don't?

apps/alpha-test-app/src/components/layout-demo/LayoutSettings.vue
```diff
@@ -0,0 +1,165 @@
...
+  },
+);
+
+const emit = defineEmits<{
+  "update:modelValue": [value: SettingsSections];
+}>();
+
+const noneOption = { id: "none", label: "None" };
+
+const contentOptions: SelectionOption<undefined>[] = [
```
larsrickert, line 51: We could give the generic a default value so we don't need `<undefined>` everywhere
```ts
export type SelectionOption<T = unknown>
```

apps/alpha-test-app/src/components/layout-demo/LayoutSettings.vue
```diff
@@ -0,0 +1,165 @@
...
+const activeFooterSetting = computed({
+  get: () => props.modelValue.footer || {},
+  set: (value) => emit("update:modelValue", { ...props.modelValue, footer: value }),
+});
+const activeOverlaySetting = computed({
+  get: () => props.modelValue.overlay || {},
+  set: (value) => emit("update:modelValue", { ...props.modelValue, overlay: value }),
+});
+
+const showSections = computed(() => {
```
larsrickert, line 95:
```suggestion
const enabledSections = computed(() => {
```

showSections sound more like a function that can be called to me :)

apps/alpha-test-app/src/components/layout-demo/LayoutSettings.vue
```diff
@@ -0,0 +1,165 @@
...
+  activeSidebarSetting,
+  (sidebarSetting) => {
+    if (!sidebarSetting.showSideBar && !sidebarSetting.showSideBarCollapse) {
+      footerOptions.value = baseFooterOptions;
+    } else {
+      footerOptions.value = sidebarFooterOptions;
+    }
+  },
+  { immediate: true },
+);
```
larsrickert, line 124: The footerOptions could be a computed right instead of ref + watcher. It is not used anywhere else

apps/alpha-test-app/src/components/layout-demo/LayoutSettings.vue
```diff
@@ -0,0 +1,165 @@
...
+</script>
+
+<template>
+  <div :key="modelValue.toString()">
```
larsrickert, line 128: Why is the key needed?

apps/alpha-test-app/src/components/layout-demo/LayoutSettings.vue
```diff
@@ -0,0 +1,165 @@
...
+      v-if="showSections.overlay"
+      v-model="activeOverlaySetting"
+      headline="Overlay Options"
+      :options="overlayOptions"
+      :horizontal="horizontal"
+    />
+  </div>
+</template>
+
+<style lang="scss" scoped></style>
```
larsrickert, line 165: unused styles

apps/alpha-test-app/src/components/layout-demo/LayoutSettings.vue
```diff
@@ -0,0 +1,165 @@
+<script lang="ts" setup>
+import { computed, ref } from "vue";
+import SingleSettingsGroup from "./SingleSettingsGroup.vue";
+import MultiSettingsGroup from "./MultiSettingsGroup.vue";
+import { watch } from "vue";
```
larsrickert, line 5: import can be combined with the first one from "vue"

apps/alpha-test-app/src/components/layout-demo/MultiSettingsGroup.vue
```diff
@@ -0,0 +1,45 @@
...
+
+const props = defineProps<{
+  headline: string;
+  options: SelectionOption<undefined>[];
+  horizontal?: boolean;
+}>();
+
+const options = computed<SelectionOption<Settings>[]>(() =>
+  props.options.map(
+    (option) => ({ ...option, value: { [option.id]: true } }) as SelectionOption<Settings>,
```
larsrickert, line 16: Why is the value added here? Is it used somewhere? The option id is used for the modeValue, `value` is only to attach custom data to the option.

Also, the checkbox currently does not support the value at all until #376

apps/alpha-test-app/src/components/layout-demo/MultiSettingsGroup.vue
```diff
@@ -0,0 +1,45 @@
...
+<template>
+  <OnyxCheckboxGroup
+    v-model="selectedOptions"
+    :headline="headline"
+    :options="options"
+    :direction="horizontal ? 'horizontal' : 'vertical'"
+  />
+</template>
+
+<style lang="scss" scoped></style>
```
larsrickert, line 45: unused styles

apps/alpha-test-app/src/components/layout-demo/MultiSettingsGroup.vue
```diff
@@ -0,0 +1,45 @@
...
+}>();
+
+const options = computed<SelectionOption<Settings>[]>(() =>
+  props.options.map(
+    (option) => ({ ...option, value: { [option.id]: true } }) as SelectionOption<Settings>,
+  ),
+);
+
+const settingsToSelection = (settings?: Settings): (keyof Settings)[] => {
+  return settings ? (Object.keys(settings) as (keyof Settings)[]) : [];
```
larsrickert, line 21: Shouldn't only true values be included?

apps/alpha-test-app/src/components/layout-demo/MultiSettingsGroup.vue
```diff
@@ -0,0 +1,45 @@
...
+};
+const selectionToSettings = (selection: (keyof Settings)[]): Settings => {
+  return selection.reduce((settings: Settings, selectedKey) => {
```
larsrickert, line 24:
```suggestion
  return selection.reduce<Settings>((settings, selectedKey) => {
```

apps/alpha-test-app/src/components/layout-demo/MultiSettingsGroup.vue
larsrickert: Same for the SingleSettingsGroup.vue

apps/alpha-test-app/src/components/layout-demo/SingleSettingsGroup.vue
```diff
@@ -0,0 +1,44 @@
...
+  ),
+);
+
+const settingsToSelection = (settings?: Settings): SelectionOption<Settings> | undefined => {
+  return settings
+    ? options.value.find((option) => option.id === Object.keys(settings)[0])
+    : undefined;
+};
+const selectionToSettings = (selection?: SelectionOption<Settings>): Settings => {
+  return selection && selection.value ? selection.value : {};
```
larsrickert, line 26:
```suggestion
  return selection?.value ?? {};
```

larsrickert: Maybe not even worth a dedicated function then^^ We could inline it

apps/alpha-test-app/src/components/layout-demo/SingleSettingsGroup.vue
```diff
@@ -0,0 +1,44 @@
...
+
+const options = computed<SelectionOption<Settings>[]>(() =>
+  props.options.map(
+    (option) => ({ ...option, value: { [option.id]: true } }) as SelectionOption<Settings>,
+  ),
+);
+
+const settingsToSelection = (settings?: Settings): SelectionOption<Settings> | undefined => {
+  return settings
+    ? options.value.find((option) => option.id === Object.keys(settings)[0])
```
larsrickert, line 22: This doesn't work if the value for the first key is false, right? Shouldn't only true values be included?

apps/alpha-test-app/src/components/layout-demo/MultiSettingsGroup.vue
larsrickert: Same for the SingleSettingsGroup.vue

apps/alpha-test-app/src/views/LayoutDemoView.vue
```diff
@@ -1,32 +1,17 @@
...
+} from "../components/layout-demo/LayoutSettings.vue";
+
+const settings = ref<SettingsSections>({
+  content: { longPageContent: true },
+  sideBar: { showSideBar: true },
+  footer: { detailFooter: true },
+  overlay: {},
 });
 
 const tempOverlayOpen = ref(true);
```
larsrickert, line 17: Alternative without .fill():
```ts
const muchContent = Array.from({ length: 100 }, (_, index) => `Lorem ipsum dolor ${index}`);
```

larsrickert (changes requested):

apps/alpha-test-app/src/components/layout-demo/LayoutSettings.vue
BoppLi: good point, I'll adjust :)

apps/alpha-test-app/src/components/layout-demo/LayoutSettings.vue
BoppLi: not anymore since I switched to getter/setter, will remove it :)

apps/alpha-test-app/src/components/layout-demo/LayoutSettings.vue
BoppLi: weird, in the past this happened automatically :O

apps/alpha-test-app/src/components/layout-demo/MultiSettingsGroup.vue
BoppLi: somehow still works. but technically sounds like it should not ^^ I'll filter it

apps/alpha-test-app/src/components/layout-demo/LayoutSettings.vue
BoppLi: I left the forceTooltip without "show" because it will always show on hover :) the force will just overrule the hover behavior :)

apps/alpha-test-app/src/components/layout-demo/MultiSettingsGroup.vue
BoppLi: I previously used value instead of the ID. will check

BoppLi: rewrote it so it only happens in the setter now in SingleSettingsGroup. Removed it from MultiSettingsGroup

apps/alpha-test-app/src/components/layout-demo/SingleSettingsGroup.vue
BoppLi: done

apps/alpha-test-app/src/components/layout-demo/SingleSettingsGroup.vue
BoppLi: kept it after the latest change where the "value" was removed (got longer again)

larsrickert (approved):

PR #429, opened 2024-02-27
docs: fix OnyxButton code snippet when using an icon

larsrickert:
Fix OnyxButton code snippet when using an icon.

The issues was that the icon property is the SVG content which also includes double quotes. But since the property itself already has double quotes, it would generate invalid HTML, e.g.
`<OnyxButton icon="<svg some-prop="test"" />`


packages/storybook-utils/src/preview.ts
```diff
@@ -134,6 +119,29 @@ export const createPreview = <T extends Preview = Preview>(overrides?: T) => {
...
+    // remove empty v-binds, e.g. v-bind="{}" will be removed
+    { searchValue: / v-bind=['"]{}['"]/g, replaceValue: "" },
+    // replace boolean shortcuts for true, e.g. disabled="true" will be changed to just disabled
+    { searchValue: /:(.*)=['"]true['"]/g, replaceValue: "$1" },
+  ];
+
+  return replacements.reduce((code, replacement) => {
+    return replaceAll(code, replacement.searchValue, replacement.replaceValue);
+  }, sourceCode);
+};
```
JoCa96, line 143: Regex are tricky, would be nice to have a test for this function

JoCa96 (approved):

packages/storybook-utils/src/preview.ts
larsrickert: done

JoCa96 (approved):

PR #431, opened 2024-02-27
feat: storybook mandatory mode toggle

JoCa96:
Enable toggling of mandatory mode between `optional` and `required` indicator.


BoppLi:
relates to #433 ? :) found it as draft in the backlog.
maybe move it to the sprint so we remember it in the sprint-end meeting for presentation :)

packages/storybook-utils/src/preview.ts
```diff
@@ -39,6 +51,28 @@ const themes = {
  */
 export const createPreview = <T extends Preview = Preview>(overrides?: T) => {
   const defaultPreview = {
+    globalTypes: {
```
BoppLi, line 54: it works perfectly in the `Docs` tabs of a component, but in the other tabs e.g. `Default`, a  reload is needed. But we probably can't change that? ¯\\_(ツ)\_/¯

packages/storybook-utils/src/preview.ts
```diff
@@ -39,6 +51,28 @@ const themes = {
...
+      mandatoryMode: {
+        name: "Mandatory mode",
+        description: "Switch between 'required' and 'optional' indicator",
+        defaultValue: "required",
+        toolbar: {
+          icon: "flag",
+          items: [
+            { value: "required", right: "*", title: "Required indicator" },
+            { value: "optional", right: "(optional)", title: "Optional indicator" },
```
BoppLi, line 63: [opinion]
```suggestion
            { value: "required", right: "*", title: "Required indicator:" },
            { value: "optional", right: "(optional)", title: "Optional indicator:" },
```

BoppLi (approved):

packages/storybook-utils/src/preview.ts
JoCa96: fixed and ready for review :)

packages/sit-onyx/src/styles/required.scss
```diff
@@ -16,7 +16,12 @@
 
 .onyx-optional-marker {
   .onyx-use-optional &::after {
-    font-family: var(--onyx-font-family);
+    font: {
+      family: var(--onyx-font-family);
+      size: 1rem;
```
larsrickert, line 21: Shouldn't the font size be inherited? So that it aligns if we use a different density

packages/storybook-utils/src/mandatory.ts
```diff
@@ -0,0 +1,44 @@
+import { type Decorator } from "@storybook/vue3";
+import { ref, watch } from "vue";
+
+type MandatoryIndicator = "required" | "optional";
+type MandatoryGlobalType = {
```
larsrickert, line 5: This is a general Storybook type, right? We could make it more generic by adding a generic for the items value so we can re-use this type int he future for additional globals

packages/storybook-utils/src/mandatory.ts
```diff
@@ -0,0 +1,44 @@
+import { type Decorator } from "@storybook/vue3";
+import { ref, watch } from "vue";
+
+type MandatoryIndicator = "required" | "optional";
```
larsrickert, line 4: We agreed to call it "required" instead of mandatory :)

larsrickert (changes requested):

packages/storybook-utils/src/mandatory.ts
larsrickert: Also for the file name

packages/storybook-utils/src/mandatory.ts
JoCa96: done

packages/storybook-utils/src/mandatory.ts
JoCa96: done

larsrickert (approved):

PR #442, opened 2024-02-28
feat(OnyxLink): add CSS states and external link icon

larsrickert:
relates to #237

Add hover, active, focus-visible and visited state to the OnyxLink.
Also added an icon for external links which can be enabled/disabled via the `withExternalIcon` property. By default it will show automatically when the link is external (`http://` or `https://`)

## Checklist

- [x] All changes are documented in the documentation app (folder `apps/docs`


packages/sit-onyx/src/utils/index.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * Checks whether the given link is external (starts with http:// or https://).
+ */
+export const isExternalLink = (href: string): href is `http${"s" | ""}://${string}` => {
```
JoCa96, line 4: Based on the comment below: 
```suggestion
export const hasCurrentOrigin = (href: string): href is `http${"s" | ""}://${string}` => {
```

packages/sit-onyx/src/components/OnyxLink/OnyxLink.vue
```diff
@@ -29,15 +33,49 @@ defineSlots<{
     @click="emit('click')"
   >
     <slot></slot>
+    <OnyxIcon
+      v-if="!props.hideExternalIcon && isExternalLink(props.href)"
```
JoCa96, line 37: We need to clarify (and/or research) if the icon means "opens external page in the same tab" (when `target`) or "opens external page in the same or new tab" (by checking the origin). What is the common/recommended pattern? I have seen both in the past.

JoCa96: We need to document in the description what an external link means.

JoCa96 (changes requested):

packages/sit-onyx/src/components/OnyxLink/types.ts
```diff
@@ -8,6 +8,11 @@ export type OnyxLinkProps = {
    * For `_blank`, the `rel="noopener noreferrer"` will be set automatically.
    */
   target?: LinkTarget;
+  /**
+   * Whether to hide the external link icon if `href` points to an external website.
+   * For links not starting with "http://" or "https://" it will be hidden automatically.
+   */
+  hideExternalIcon?: boolean;
```
JoCa96, line 15: It might be a nicer API to offer the option to set it directly:
```suggestion
hasExternalIcon?: boolean;
```

And per default, when it is `undefined`, we use our logic to determine the external icon

packages/sit-onyx/src/components/OnyxLink/OnyxLink.vue
```diff
@@ -29,15 +33,49 @@ defineSlots<{
...
+
+  &:hover {
+    color: var(--onyx-color-text-icons-primary-medium);
+  }
+
+  &:focus-visible {
+    color: var(--onyx-color-text-icons-primary-intense);
+    border-radius: var(--onyx-radius-sm);
+    border-color: var(--onyx-color-base-primary-300);
+    outline: none;
```
JoCa96, line 62: Semantically and technical (doesn't take any space) `outline` is preferable here. `outline` will adhere to the `border-radius`

packages/sit-onyx/src/utils/index.ts
```diff
@@ -0,0 +1,6 @@
...
+  return /^http(s?):\/\//.test(href);
```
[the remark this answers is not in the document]
larsrickert: I already thought about this, I see 2 issues here:

1. window.location is not defined in SSR like in VitePress or Nuxt
2. Links to the same origin are project internal links which will most likely be relative URLs, like `/home`, not full URLs because you don't know where your site is hosted. Also for internal links, you should use the Vue router because the `OnyxLink` or `<a>` will lead to a full page reload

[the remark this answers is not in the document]
larsrickert: I already have an idea for integrating the Vue Router but I'd like to do a POC in a follow up PR

packages/sit-onyx/src/components/OnyxLink/types.ts
larsrickert: Good idea! I changed to type to `boolean | "auto"` so its clearer to the user that we use our logic

packages/sit-onyx/src/components/OnyxLink/types.ts
```diff
@@ -8,6 +8,12 @@ export type OnyxLinkProps = {
    * For `_blank`, the `rel="noopener noreferrer"` will be set automatically.
    */
   target?: LinkTarget;
+  /**
+   * Whether to show the external link icon.
+   * If set to `auto`, it will be shown when the `href` leads to another website
+   * and will be hidden otherwise.
+   */
+  externalIcon?: boolean | "auto";
```
JoCa96, line 16: The naming suggestions to me that an icon can be chosen to use this prop, maybe we can rename it?

```suggestion
  withExternalIcon?: boolean | "auto";
```

packages/sit-onyx/src/utils/index.ts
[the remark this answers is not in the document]
larsrickert: The problem is that we don't have the vue-router as dependency so we can not use the `<router-link>` component.
Lets discuss this in a POC / follow up PR.

@JoCa96 Is it fine for you to keep the simple check for external links for now?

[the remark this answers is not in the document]
larsrickert: #447

packages/sit-onyx/src/components/OnyxLink/OnyxLink.vue
larsrickert: image

Do you know how to prevent the outline to not have this break at the icon?

BoppLi:
> * [x]  All changes are documented in the documentation app (folder `apps/docs`)(https://github.com/SchwarzIT/onyx/actions/workflows/playwright-screenshots.yml) is added

I think the playwright screenshot url line in the pr description is a leftover of the template? ^^

packages/sit-onyx/src/components/OnyxLink/OnyxLink.vue
JoCa96: Is it already fixed? I am unable to reproduce the issue

JoCa96: this works on all browsers for me
```scss
  text-decoration: underline;
  color: var(--onyx-color-text-icons-primary-intense);
  border-radius: var(--onyx-radius-sm);
  outline: 1px solid transparent;

  // other styles like size, font family etc. should be inherited
  // so we don't define them here

  &:hover {
    color: var(--onyx-color-text-icons-primary-medium);
  }

  &:focus-visible {
    color: var(--onyx-color-text-icons-primary-intense);
    outline-color: var(--onyx-color-base-primary-300);
  }
```

larsrickert: I needed to add `display: inline-block;`, It works now with the outline

packages/sit-onyx/src/utils/index.ts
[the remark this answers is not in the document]
JoCa96: > The problem is that we don't have the vue-router as dependency so we can not use the `<router-link>` component. Lets discuss this in a POC / follow up PR.
> 
> @JoCa96 Is it fine for you to keep the simple check for external links for now?

Of course! :)

packages/sit-onyx/src/components/OnyxLink/OnyxLink.vue
larsrickert: Done

JoCa96 (approved):

PR #462, opened 2024-03-04
docs(home): update component roadmap

larsrickert:
relates to #294

Update the component roadmap on the VitePress home page.

## Checklist


apps/docs/src/.vitepress/components/ComponentCard.vue
```diff
@@ -1,14 +1,33 @@
...
 };
 
 const props = defineProps<ComponentCardProps>();
+
+const estimationValue = computed(() => {
+  if (!props.estimation) return "n/a";
+
+  const date = new Date(props.estimation);
+  const month = (date.getMonth() + 1).toString().padStart(2, "0");
+  return `${month}/${date.getFullYear()}`;
```
JoCa96, line 29: Prefer the [`Intl.DateTimeFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat) for formatting of dates

JoCa96 (approved):

apps/docs/src/.vitepress/components/ComponentCard.vue
larsrickert: Done :)

JoCa96 (approved):

BoppLi:
Oh no.. In the figma link of the ticket, the lower variant was planned to be implemented. It is the one with this layout:
image
You implemented the other variant of the tile design..

That's the reason why the refined ticket included a screenshot but somehow the screenshot is gone from the ticket now....

apps/docs/src/.vitepress/components/ComponentCard.vue
```diff
@@ -18,26 +39,32 @@ const props = defineProps<ComponentCardProps>();
...
+    <ComponentStatusBadge :status="props.status" class="card__status" />
+    <OnyxHeadline is="h2">{{ props.name }}</OnyxHeadline>
+    <p class="estimation">
+      Estimation:
+      <span class="estimation__value">{{ estimationValue }}</span>
+    </p>
   </component>
 </template>
 
 <style lang="scss" scoped>
```
BoppLi, line 51: Oh no.. In the figma link of the ticket, the lower variant was planned to be implemented. It is the one with this layout:
image
You implemented the other variant of the tile design..

That's the reason why the refined ticket included a screenshot but somehow the screenshot is gone from the ticket now....

BoppLi (changes requested):

apps/docs/src/.vitepress/components/ComponentCard.vue
larsrickert: Done :) Please review again

apps/docs/src/index.data.ts
```diff
@@ -63,78 +62,90 @@ export default defineLoader({
...
     };
 
-    const componentTabs: HomePageData["componentTabs"] = [
+    const basicComponentsDueDate = new Date(2024, 4);
+
+    const components: HomePageData["components"] = [
+      {
+        name: "Button",
+        dueDate: basicComponentsDueDate,
+        ...getImplementedStatus("OnyxButton"),
```
BoppLi, line 78: question: for the third status, instead of using the function `getImplementedStatus`, would we write `implemented` by hand?

apps/docs/src/.vitepress/components/ComponentCard.stories.ts
```diff
@@ -0,0 +1,43 @@
...
+      (story) => ({
+        components: { story },
+        template: `<div style="width: 16rem;"> <story /> </div>`,
+      }),
+    ],
+  }),
+};
+
+export default meta;
+type Story = StoryObj<typeof ComponentCard>;
```
BoppLi, line 20: please add an example with n/a due date 
image
in the storybook controls I can't find out how to "unset" the date control

BoppLi (approved):

apps/docs/src/index.data.ts
larsrickert: Thats how I intended it at least since we agreed that we set them manually to "implemented"

apps/docs/src/.vitepress/components/ComponentCard.stories.ts
larsrickert: Good point :D Looks like Storybook does not provide a way to do this. Should I change the control to string?

larsrickert: I added a story without a due date

BoppLi: it's handy like the control is now, the extra story does the trick :p thx!

BoppLi (approved):

PR #482, opened 2024-03-05
fix: add padding-right for checkbox and radio button label

larsrickert:
Add padding-right to checkbox and radio button label (based on Feedback from Jannick) to align with Figma.
image

## Checklist


packages/sit-onyx/playwright/snapshots/components/OnyxCheckboxGroup/horizontal-chromium-linux.png
BoppLi, line 1: shouldn't the distance between the checkboxes in the group be balanced out by the new padding?

larsrickert: According to Figma it is still var(--onyx-spacing-xl)

MajaZarkova (approved):

Issue #498, opened 2024-03-06
Fix default of loading indicator


mj-hof:
Change default in storybook from loading spinner to dots

### Definition of Done
**Review**

- _add link to deployed storybook_
- _add link to deployed docs_
- [ ] approved by designer @jannick-ux
- [ ] approved by dev _add dev tag_

PR #522, opened 2024-03-07
changed Loading Indicator default to dots

Y4nnik:
relates to https://github.com/SchwarzIT/onyx/issues/498

I changed the default in storybook from loading spinner to dots


packages/sit-onyx/src/components/OnyxLoadingIndicator/OnyxLoadingIndicator.stories.ts
```diff
@@ -21,10 +21,14 @@ const meta: Meta<typeof OnyxLoadingIndicator> = {
 export default meta;
 type Story = StoryObj<typeof OnyxLoadingIndicator>;
 
-export const Circle = { args: {} } satisfies Story;
-
 export const Dots = {
   args: {
     type: "dots",
   },
 } satisfies Story;
```
larsrickert, line 28:
```suggestion
export const Dots = { args: {} } satisfies Story;
```

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxLoadingIndicator/OnyxLoadingIndicator.stories.ts
Y4nnik: Yes, but when you for example choose "circle" in the controls and then reset the controls it does not switch back to "dots" and when you first load the page, the dots are not automatically selected in the controls. Thats why I changed it or is it suppose to work that way ?

larsrickert: Ah okay then thats a Storybook bug :) Its fine like that

Issue #189, opened 2024-01-24
Implement OnyxTextInput


JoCa96:

BoppLi:
Note: the label is not animated! it always is just on top.

larsrickert:
Follow up tickets:
- #570
- #571
- #572
- #573
- #574

jannick-ux:
Findings:

- [ ] Make label hideable
- [ ] implement "clear-all-x-icon" while typing
- [ ] implement "copy-input-icon" and function in general
- [ ] implement optional-state
- [ ] would be a default value for min and max width helpful?
- [ ] would be a default value for placeholder text helpful?
- [ ] by clicking into a pre-filled input, the whole text should be marked. Or at least there should be the option to do so. (What is the web standard here?)

larsrickert:
> Findings:

1. Hide label is implemented in #612
2. Clear and copy was out of scope, follow up ticket is #573 
3. Optional state is already implemented, you can use the flag toggle in the Storybook toolbar to test it
4. I wouldn't define a min/max width, imho this is responsibility of the grid
5. The placeholder is typically dependant on the context so I would leave this up the the project
6. The default behavior is whats currently implemented so I'd like to keep it. When clicking into the input it will just focus it. To mark the text you need to double click it or select it via mouse/keyboard

jannick-ux:
approved

PR #537, opened 2024-03-08
feat(OnyxInput): add property `maxlength`, `minLength`, `withCounter` and `message` 

larsrickert:
relates to #189

Enhance the OnyxInput to support min/max length, counter and message

## Checklist


packages/sit-onyx/src/components/OnyxInput/OnyxInput.vue
```diff
@@ -46,38 +46,51 @@ const patternSource = computed(() => {
...
+        />
+        <!-- eslint-enable vuejs-accessibility/no-autofocus -->
+      </div>
+    </label>
+
+    <div v-if="props.message || shouldShowCounter" class="onyx-input__footer onyx-text--small">
+      <span v-if="props.message" class="onyx-truncation-ellipsis">{{ props.message }}</span>
+      <span v-if="shouldShowCounter" class="onyx-input__counter">
+        {{ value.length }}/{{ props.maxlength }}
+      </span>
```
JoCa96, line 91: You could consider to use `div` when you need `block` elements

packages/sit-onyx/src/components/OnyxInput/OnyxInput.vue
```diff
@@ -176,5 +193,19 @@ const patternSource = computed(() => {
...
+    align-items: center;
+    gap: var(--onyx-spacing-2xs);
+    color: var(--onyx-color-text-icons-neutral-soft);
+  }
+
+  &__counter {
+    margin-left: auto;
+    display: block;
+    width: max-content;
+  }
```
JoCa96, line 209: I would consider `justify-content: space-between`, `flex-grow: 1` (both require removing the `v-if="props.message"`) or using `grid` a cleaner solution.

packages/sit-onyx/src/components/OnyxInput/types.ts
```diff
@@ -57,6 +57,26 @@ export type OnyxInputProps = {
    * Whether the input should be disabled.
    */
   disabled?: boolean;
+  /**
+   * Maximum number of characters that are allowed to be entered.
+   * For validation: If the value is set by code, the browser can not detect the validity
+   * so it will only get invalid when the user interacts with the component.
```
JoCa96, line 63: What about submitting a form?

JoCa96 (dismissed):
minor comments

packages/sit-onyx/src/components/OnyxInput/OnyxInput.vue
larsrickert: But this would mean that there is always an empty `<span>` if there is no message, right?

JoCa96: Yes, which would be fine IMHO. I like to avoid margin for placing content, if it can be avoided:

```css
text-align: right;
flex-grow: 1;
```

larsrickert: Nice, done :)

JoCa96 (approved):

JoCa96 (approved):

PR #576, opened 2024-03-11
chore: Add Axe testing to 'createScreenshotsForAllStates'

JoCa96:
We want to check each test case for accessibility violations.
This allows us to quickly catch issues and address them.


packages/sit-onyx/src/components/OnyxButton/OnyxButton.ct.tsx
```diff
@@ -97,5 +97,6 @@ test(
       if (focusState === "active") await page.mouse.down();
       return component;
     },
+    { disableAxeCheck: true },
```
larsrickert, line 100:
```suggestion
    { disableAxeCheck: true }, // TODO: must be addressed by UX, see [#577](https://github.com/orgs/SchwarzIT/projects/5?pane=issue&itemId=56062129)
```

packages/sit-onyx/src/utils/playwright.tsx
```diff
@@ -103,35 +111,54 @@ export const createScreenshotsForAllStates =
...
+        await page.mouse.up(); // reset mouse
+
+        // Setup component test case
+        const component = await caseBuilder(testCase, wrappedMount, page);
+
+        // ASSERT
+        const screenshotResult = expect(component).toHaveScreenshot(`${screenshotName}.png`);
+        const accessibilityScanResult = performAxeScan();
+
+        await Promise.all([screenshotResult, accessibilityScanResult]);
```
larsrickert, line 160: Is it safe to run them in parallel? I remember some firefox issues in the past where the axe tests would cause a page reload

packages/sit-onyx/src/utils/playwright.tsx
```diff
@@ -103,35 +111,54 @@ export const createScreenshotsForAllStates =
...
+
+    const axeBuilder = makeAxeBuilder();
+    const performAxeScan = options?.disableAxeCheck
+      ? () => Promise.resolve()
+      : async () => axeBuilder.analyze().then((r) => expect(r.violations).toEqual([]));
+
     const permutations = generatePermutations(states);
 
+    // give a maximum of 1 second per test case
+    test.setTimeout(permutations.length * 1000);
```
larsrickert, line 127: We can remove our global timeout in the playwright config then, right?

larsrickert (changes requested):

packages/sit-onyx/src/utils/playwright.tsx
JoCa96: Locally I don't have any issues, lets see what the CI says... A page reload would be destroy any component state and browser state, sounds like something that shouldn't happen. But yeah could very well be that Playwright isn't intended to be used like this.

larsrickert:
Closing this in favor of #942

Issue #564, opened 2024-03-11
Implement OnyxListbox


mj-hof:
### Why?

The flyout is a fundamental element utilized across various components such as dropdowns, navigation bars, paginations, tables, etc. It provides the users with the ability to open a small modal window, facilitating single or multisselection based on the context in which it is employed.

### Design

[Figma link](https://www.figma.com/file/YfEUBOHk4J4nYrk04geswG/onyx---Design-System-Figma-Library?type=design&node-id=1130-19227&mode=dev)

### Acceptance criteria

- [x] single select flyout can be shown
- [x] keyboard interaction is supported as described in the ARIA pattern
- [x] an optional label can be shown at the bottom
- [x] at max. 8 options are shown without scrolling (can be changed with a property)

#### Out of scope

- #732
- #733
- #734
- #742
- #743
- #745

### Definition of Done

- [x] example usage added to alpha-test-app
- [x] covered by component tests
- [x] screenshot tests are updated
- [x] follow-up tickets were created if necessary
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature
- [x] Set task to `In Approval`

**Review**

- [Storybook link](https://storybook.onyx.schwarz/?path=/docs/support-onyxlistbox--docs)
- [x] approved by designer @jannick-ux
- [x] approved by dev @JoCa96 

#### Applicable ARIA Pattern

[Listbox pattern](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/)

larsrickert:
FYI, I removed multiselect, truncation and grouping in follow up tickets so I can close this before my vacation and another dev can continue working on it

jannick-ux:
By default, the label should be hidden.
Rest looks fine.

larsrickert:
> By default, the label should be hidden.

To be unify with the other components like input, checkbox, radio button and switch I would prefer to show it by default with a property to hide it. The label is required anyways for screen readers

jannick-ux:
This label, that is used here, is not that kind of label, that has an important role to describe the purpose of the whole component. 
In the end this label here is just ment to give the possibility to write something like "items only referr to the date xx.xx.xxxx". 
So it is just a flexible hinttext/labeltext without any dependecy or importance to the whole component. 

That's the reason why I would like to hide it in first place.

Was the description of the purpose understandable?

PR #621, opened 2024-03-14
feat: add basic single select listbox

larsrickert:
relates to #564

Add a basic single select listbox

## Checklist


.changeset/plenty-queens-matter.md
JoCa96, line 1: Can we add a changeset for the headless package, or is the composable not ready yet?

packages/sit-onyx/src/components/OnyxListboxOption/OnyxListboxOption.vue
```diff
@@ -0,0 +1,85 @@
...
+
+const emit = defineEmits<{
+  "update:modelValue": [value: boolean];
+}>();
+
+const {
+  elements: { option },
+} = createListbox({
+  label: "",
+  onSelect: () => emit("update:modelValue", !props.modelValue),
```
JoCa96, line 17: Can we pass in the data via props? I don't think it's a good idea to have a composable that needs to be used in two places to work.
There is an implicit dependency between the usage in the parent and in the option.
As soon as the composable has some kind of state, it will become an issue.

My suggestions from most preferable to the least preferable
1. Use the composable only in the parent and pass the attributes down via props
1. Inline this component into the parent
1. Create a separate composable for the options.

packages/sit-onyx/src/components/OnyxListboxOption/OnyxListboxOption.vue
```diff
@@ -0,0 +1,85 @@
...
+      option({
+        id: props.label,
+        label: props.label,
+        selected: props.modelValue,
+        disabled: props.disabled,
+      })
+    "
+    class="onyx-listbox-option"
+    :class="{ 'onyx-listbox-option--selected': props.modelValue }"
+    :disabled="props.disabled"
```
JoCa96, line 37: AFAIK `disabled` doesn't have an effect here? 

```suggestion
```

packages/headless/src/composables/listbox/createListbox.ts
JoCa96, line 1: Please don't forget about testing :) Doesn't need to be done in this PR

JoCa96 (changes requested):

packages/sit-onyx/src/components/OnyxListboxOption/OnyxListboxOption.vue
larsrickert: Nope it is a leftover from my previous button approach. I removed it

packages/headless/src/composables/listbox/createListbox.ts
larsrickert: Sure, I will do the testing in a follow up PR

packages/sit-onyx/src/components/OnyxListboxOption/OnyxListboxOption.vue
larsrickert: I removed the composable from the option, can you check it again please?
I am not sure if there is a cleaner way to implement, if you find one, please let me know :)

packages/sit-onyx/src/components/OnyxListbox/OnyxListbox.vue
```diff
@@ -0,0 +1,102 @@
...
+        v-for="option in props.options"
+        :key="option.id.toString()"
+        class="onyx-listbox__option"
+        :headless-option="
+          headlessOption({
+            id: option.id,
+            label: option.label,
+            disabled: option.disabled,
+            selected: option.id === props.modelValue,
+          })
```
JoCa96, line 43: Can we flatten the API?
Not really a fan of a property called `headlessOption` .

```suggestion
        v-bind="
          headlessOption({
            id: option.id,
            label: option.label,
            disabled: option.disabled,
            selected: option.id === props.modelValue,
          })
```

JoCa96 (approved):

packages/sit-onyx/src/components/OnyxListbox/OnyxListbox.vue
larsrickert: I wanted to do this first but then the Vue compiler crashes because it can not resolve the type...

packages/headless/src/composables/listbox/createListbox.ts
```diff
@@ -0,0 +1,66 @@
...
+            "aria-label": data.label,
+            "aria-checked": isMultiselect.value ? isSelected : undefined,
+            "aria-selected": !isMultiselect.value ? isSelected : undefined,
+            "aria-disabled": data.disabled,
+            tabindex: "0",
+            onKeydown: (event) => handleKeydown(event, data.id),
+            onClick: () => options.onSelect?.(data.id),
+          } as const;
+        };
+      }),
```
JoCa96, line 62: According to https://www.w3.org/WAI/ARIA/apg/patterns/listbox/examples/listbox-scrollable/ the tabindex must be on the listbox, not on each option:
```suggestion
        role: "listbox",
        "aria-multiselectable": isMultiselect.value,
        "aria-label": unref(options.label),
        tabindex: "0",
      })),
      group: computed(() => {
        return (options: { label: string }) => ({
          role: "group",
          "aria-label": options.label,
        });
      }),
      option: computed(() => {
        return (data: {
          label: string;
          id: ListboxValue;
          selected?: boolean;
          disabled?: boolean;
        }) => {
          const isSelected = data.selected ?? false;
          return {
            role: "option",
            "aria-label": data.label,
            "aria-checked": isMultiselect.value ? isSelected : undefined,
            "aria-selected": !isMultiselect.value ? isSelected : undefined,
            "aria-disabled": data.disabled,
            onKeydown: (event) => handleKeydown(event, data.id),
            onClick: () => options.onSelect?.(data.id),
          } as const;
        };
      }),
```

packages/sit-onyx/src/components/OnyxListbox/OnyxListbox.vue
JoCa96: What about defining the Props explicitly or/and using fallthrough props?

larsrickert: I used fallthrough attributes :)

packages/headless/src/composables/listbox/createListbox.ts
larsrickert: I removed it for and will add it accordingly when implementing the keyboard shortcuts. I also wrote Jannick that we need a design for the focus-visible state of the listbox

packages/sit-onyx/src/components/OnyxListboxOption/OnyxListboxOption.vue
```diff
@@ -0,0 +1,50 @@
+<template>
+  <li class="onyx-listbox-option">
+    <span class="onyx-truncation-ellipsis"> {{ $attrs["aria-label"] }}</span>
```
JoCa96, line 3: I think a slot would be nice here to not rely on the undocumented `aria-label`
```suggestion
    <span class="onyx-truncation-ellipsis"><slot /></span>
```

JoCa96 (approved):

JoCa96 (approved):

Issue #644, opened 2024-03-15
Try to self hosted GitHub runner


larsrickert:
We want to try out a self hosted GitHub runner that runs on STACKIT so see if we can get improved workflow speeds (esspecially for the Playwright screenshot tests).

PR #661, opened 2024-03-15
[DRAFT] perf: use self hosted runner for Playwright tests

larsrickert:
Relates to #644

I set up a self hosted GitHub runner in STACKIT with 16 CPU cores and 16GB RAM which will be used for the Playwright screenshot tests. Performance improvement > 250%, see #644


larsrickert:
Note: I removed the runner for now until we have a decision if and which server we will use

JoCa96 (approved):

PR #672, opened 2024-03-18
chore: add composables and dedicated prop types for  required-marker and density  

JoCa96:
Relates to [#673](https://github.com/SchwarzIT/onyx/issues/673)

- Created dedicated `DensityProp` type for `density`
- Created `useDensity` composable, so that its class naming logic is encapsulated in one place
- Created dedicated `RequiredMarkerProp` type for `required` and `requiredTypes`
- Created `useRequired` composable, so that its class naming logic is encapsulated in one place


packages/sit-onyx/src/components/OnyxCheckbox/OnyxCheckbox.vue
```diff
@@ -37,7 +35,7 @@ const requiredMarkerClass = computed(() => {
     <OnyxSkeleton v-if="!props.hideLabel" class="onyx-checkbox-skeleton__label" />
   </div>
 
-  <label v-else class="onyx-checkbox" :class="[requiredMarkerClass]">
+  <label v-else class="onyx-checkbox" :class="[requiredMarkerClass, requiredTypeClass]">
```
larsrickert, line 38: Didn't jannick say the marker type should only be set on application level?

packages/sit-onyx/src/types/required.ts
larsrickert, line 1: Since its a composable, we should place it inside the "composables" folder

packages/sit-onyx/src/types/required.ts
```diff
@@ -0,0 +1,29 @@
...
+export type RequiredMarkerProp = {
+  /**
+   * Whether the form input is required.
+   */
+  required?: boolean;
+  /**
+   * Required mode: `optional` will show an `(optional)` text after the label for optional inputs.
+   * `required` will show an `*` indicator for required inputs after the label instead.
+   */
+  requiredMarker?: RequiredMarkerType;
```
larsrickert, line 15: Hmm I am not sure about this property...

packages/sit-onyx/src/types/required.ts
```diff
@@ -0,0 +1,29 @@
+import { computed } from "vue";
+
+const REQUIRED_MARKER = ["optional", "required"] as const;
+export type RequiredMarkerType = (typeof REQUIRED_MARKER)[number];
+
+export type RequiredMarkerProp = {
+  /**
+   * Whether the form input is required.
```
larsrickert, line 8:
```suggestion
   * Whether the form value is required.
```

The prop is also used for non-text-input components like switches etc.

packages/sit-onyx/src/styles/required.scss
```diff
@@ -1,28 +1,39 @@
...
+  --onyx-use-optional: var(--OFF);
+}
+
+.onyx-use-optional {
+  --onyx-use-required: var(--OFF);
+  --onyx-use-optional: var(--ON);
+}
+
+.onyx-required-marker::after {
+  font-family: var(--onyx-use-required, var(--onyx-font-family));
```
larsrickert, line 24: Why does this work? When required is ON, the fallback value would not be used because its value is "initial".

larsrickert (changes requested):

packages/sit-onyx/src/types/required.ts
larsrickert: If we keep it, we should define a default value of "required" in all components, otherwise it might be confusing for the users

packages/sit-onyx/src/styles/required.scss
JoCa96: The trick is that the value of `initial` for a custom property will trigger the fallback, while a whitespace will not.

JoCa96: I added a link to an explanation above

packages/sit-onyx/src/types/required.ts
JoCa96: you are toooooo faaaast

packages/sit-onyx/src/types/required.ts
JoCa96: I would argue that conceptually these are all kind of inputs. In HTML they are also, just different types of inputs.

larsrickert: Sure, I just think that "input" might confuse some users because they are just thinking about a text input and not a checkbox, switch, dropdown etc.

packages/sit-onyx/src/types/required.ts
JoCa96: No, that would break the inheritance.

packages/sit-onyx/src/types/required.ts
JoCa96: done

packages/sit-onyx/src/components/OnyxCheckbox/OnyxCheckbox.vue
JoCa96: Yes. This property allows overriding the application default per component.

packages/sit-onyx/src/types/required.ts
JoCa96: changed it

larsrickert (approved):

larsrickert (approved):

Issue #567, opened 2024-03-11
Implement tooltip


mj-hof:
### Open questions

- [x] Can we remove the Figma requirement to truncate the tooltip text after 3 rows?: Yes, don't truncate after 3 lines

### Why?

Tooltips in ui design and web development offer contextual information or additional details to a parent element.

### Design

[Figma link](https://www.figma.com/file/YfEUBOHk4J4nYrk04geswG/onyx---Design-System-Figma-Library?type=design&node-id=957-897&mode=dev)

### Acceptance criteria

- a tooltip can be shown with one of the following colors
  - [x] default
  - [x] danger
- [x] an optional icon can be shown on the left side of the label
- [x] the text is truncated with mutliline
- [x] a the wedge is displayed and can be positioned manually via a property
- the tooltip can be triggered via:
    - [x] hover
    - [x] click
    - [x] manually set boolean

#### Out of scope:

- #730

### Definition of Done

- [x] example usage added to alpha-test-app
- [x] covered by component tests
- [x] screenshot tests are updated
- [x] follow-up tickets were created if necessary
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature
- [x] Set task to `In Approval`

**Review**

- [Storybook link](https://storybook.onyx.schwarz/?path=/docs/components-onyxtooltip--docs)
- [x] approved by designer @jannick-ux
- [x] approved by dev @JoCa96 

### Implementation details

- Use type `OnyxColor` for color property
- Use class `onyx-truncation-multiline` for the truncation
- Use z-index `var(--onyx-z-index-flyout)`

#### Applicable ARIA Pattern

[Tooltip pattern](https://www.w3.org/WAI/ARIA/apg/patterns/tooltip/)

BoppLi:
Question: should the tooltip be auto-applied to all components where a "title" property is set for consistency?

larsrickert:
> Question: should the tooltip be auto-applied to all components where a "title" property is set for consistency?

We currently have no components with a title property, right?

BoppLi:
> > Question: should the tooltip be auto-applied to all components where a "title" property is set for consistency?
> 
> We currently don't have any components with a title property, right?

As Jonathan said, title can be set on every component and it will trigger the auto-tooltip of the browser on hover. If we set it explicitly as new property I think it would disable the browser auto-tooltip so we could place our own.. maybe we can discuss this in the next dev exchange :)

[...]

PR #685, opened 2024-03-19
feat: add basic `OnyxTooltip`

larsrickert:
Relates to #567

Add a basic OnyxTooltip component.

## Checklist


packages/sit-onyx/playwright/snapshots/components/OnyxTooltip/tooltip-variant--bottom-text--fitParent-chromium-linux.png
BoppLi, line 1: please add     box-sizing: border-box;
to actually match the parent.
the paddings make it bigger currently.

image

BoppLi (changes requested):

packages/sit-onyx/src/components/OnyxTooltip/OnyxTooltip.vue
```diff
@@ -0,0 +1,115 @@
...
+  border-radius: var(--onyx-radius-sm);
+  padding: var(--onyx-spacing-4xs) var(--onyx-spacing-sm);
+  box-shadow: 0px 2px 8px 0px rgba(0, 0, 0, 0.1);
+  z-index: var(--onyx-z-index-flyout);
+
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  gap: var(--onyx-spacing-4xs);
+  position: relative;
```
JoCa96, line 59:
```suggestion
```

BoppLi (changes requested):
44 / 54 reviewed

packages/sit-onyx/playwright/snapshots/components/OnyxTooltip/tooltip-variant--icon-text--long-chromium-linux.png
BoppLi, line 1: chromium and webkit overflow here. should the example text be shortened a bit?

packages/sit-onyx/src/components/OnyxIcon/OnyxIcon.vue
```diff
@@ -28,6 +28,7 @@ const props = withDefaults(defineProps<OnyxIconProps>(), {
 
   display: inline-block;
   width: var(--icon-size);
+  min-width: var(--icon-size);
```
BoppLi, line 31: for clarification: I suppose there is a reason to add this, but is it not needed for min-height as well?

packages/sit-onyx/src/components/OnyxTooltip/types.ts
```diff
@@ -0,0 +1,24 @@
...
+   * Optional icon to show on the left of the text.
+   */
+  icon?: string;
+  color?: Extract<OnyxColor, "neutral" | "danger">;
+  /**
+   * How to position the tooltip relative to the parent element.
+   */
+  position?: TooltipPosition;
+  /**
+   * If `true`, the tooltip will use the full width of the parent/slot element.
```
BoppLi, line 18:
```suggestion
   * If `true`, the tooltip will match the width of the parent/slot element.
```
because it goes both ways, right?: even when the parent is less wide, the tooltip will be restricted to that width

BoppLi (changes requested):
51/54 reviewed

packages/sit-onyx/playwright/snapshots/components/OnyxTooltip/tooltip-variant--icon-text--long-chromium-linux.png
larsrickert: I updated it to calculate the needed space

packages/sit-onyx/src/components/OnyxIcon/OnyxIcon.vue
larsrickert: Yes its needed so the icon does not shrink/get squeezed when used in a flex container.
Its not needed for the height because the icon is squared so setting either width or height is enough

BoppLi: ok I didn't see aspect-ratio so I was not sure if it is enough that the file itself is squared.

packages/sit-onyx/src/components/OnyxTooltip/OnyxTooltip.vue
```diff
@@ -0,0 +1,115 @@
...
+const {
+  elements: { trigger, tooltip },
+} = createTooltip({});
+</script>
+
+<template>
+  <div class="onyx-tooltip-wrapper">
+    <div
+      v-bind="tooltip"
+      class="onyx-tooltip onyx-text--small onyx-truncation-multiline"
```
BoppLi, line 28: I can not find the implementation for this class anywhere, where does it come from?

also, wasn't this feature removed from the ticket?
```suggestion
      class="onyx-tooltip onyx-text--small"
```

packages/sit-onyx/src/components/OnyxTooltip/OnyxTooltip.vue
```diff
@@ -0,0 +1,115 @@
...
+
+<style lang="scss">
+.onyx-tooltip {
+  --background-color: var(--onyx-color-base-neutral-900);
+  $wedge-size: 0.5rem;
+
+  border-radius: var(--onyx-radius-sm);
+  padding: var(--onyx-spacing-4xs) var(--onyx-spacing-sm);
+  box-sizing: border-box;
+  box-shadow: 0px 2px 8px 0px rgba(0, 0, 0, 0.1);
```
BoppLi, line 53: px values -> rem conversion?

also, "0" does not need a unit

packages/sit-onyx/src/components/OnyxTooltip/OnyxTooltip.vue
```diff
@@ -0,0 +1,115 @@
...
+
+    <div v-bind="trigger">
+      <slot></slot>
+    </div>
+  </div>
+</template>
+
+<style lang="scss">
+.onyx-tooltip {
+  --background-color: var(--onyx-color-base-neutral-900);
```
BoppLi, line 47: if the background-color can be easily overwritten with this local css variable, shouldn't we do the same for the "color"?

packages/sit-onyx/src/components/OnyxTooltip/OnyxTooltip.vue
```diff
@@ -0,0 +1,115 @@
...
+
+  &::after {
+    content: " ";
+    position: absolute;
+    top: 100%; /* At the bottom of the tooltip */
+    left: 50%;
+    margin-left: -$wedge-size;
+    border-width: $wedge-size;
+    border-style: solid;
+    border-color: var(--background-color) transparent transparent transparent;
```
BoppLi, line 95: the last one is redundant, it will default to the second :)
```suggestion
    border-color: var(--background-color) transparent transparent;
```

packages/sit-onyx/src/components/OnyxTooltip/OnyxTooltip.vue
```diff
@@ -0,0 +1,115 @@
...
+
+  &--bottom {
+    margin-bottom: 0;
+    margin-top: $wedge-size;
+    bottom: 0;
+    top: 100%;
+
+    &::after {
+      top: -2 * $wedge-size;
+      border-color: transparent transparent var(--background-color) transparent;
```
BoppLi, line 106: the last one is redundant, it will default to the second :)
```suggestion
      border-color: transparent transparent var(--background-color);
```

BoppLi (changes requested):
Done :)

packages/sit-onyx/src/components/OnyxTooltip/OnyxTooltip.vue
larsrickert: Its in the `styles/fonts.scss` file :) No truncation is still part of the ticket, only the wedge alignment is out of scope

BoppLi: ah right first I thought this means "truncation even if it is multiline".. sry ^^

BoppLi (approved):

Issue #565, opened 2024-03-11
Implement OnyxSelect


mj-hof:
### Depends on

- #564 
- #568 
- #706 
- #762

### Why?

Allow the user to select one or multiple entries from a list of options.

### Design

[Figma](https://www.figma.com/file/YfEUBOHk4J4nYrk04geswG/onyx---Design-System-Figma-Library?type=design&node-id=1569-34084&mode=design&t=lgLzv7i9PKKCmZad-4)

### Acceptance criteria

- [x] Figma design is implemented 
- [x] label can be shown or hidden (then `aria-label` is used instead)
- [x] All densities are implemented
- [x] Can be set to skeleton mode
- [x] Can be set to loading

**Out-of-scope**

- valid
- invalid

### Definition of Done

- [x] example usage added to alpha-test-app
- [x] covered by component tests
- [x] screenshot tests are updated
- [x] follow-up tickets were created if necessary
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature
- [x] Set task to `In Approval`

**Approval**

- [Storybook link](https://storybook.onyx.schwarz/?path=/docs/components-select--docs)
- [x] approved by designer @jannick-ux
- [x] approved by dev @BoppLi 

### Implementation details

```typescript
type OnyxSelectProps<TValue> = {
  label: string;
  hideLabel?: boolean;
  skeleton?: boolean;
  loading?: boolean;
  hideSearch?: boolean;
  searchTerm?: string;
  open?: boolean;
  // add searchFunc callback
  modelValue?: ComboboxOption<TValue>;
  options: ComboboxOption<TValue>[];
}

type OnyxSelectEmits<TValue> = {
  (event: "update:modelValue", value: ComboboxOption<TValue>): void;
  (event: "update:searchTerm", value: string): void;
  (event: "update:open", value: boolean): void;
};
```

- The visible placeholder area is not an actual input!
- Make use of `packages/headless/src/composables/comboBox/createComboBox.ts`
- Extend with the listbox implementation in `packages/headless/src/composables/listbox/createListbox.ts`

#### Reference implementations

- [KendoUI](https://www.telerik.com/kendo-vue-ui/components/dropdowns/combobox/)
- [Vuetify](https://vuetifyjs.com/en/components/combobox/)

#### Applicable ARIA Pattern

[Combobox](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/)

BoppLi:
[...]

PR #693, opened 2024-03-19
feat(OnyxListbox): support keyboard shortcuts

larsrickert:
Relates to #565

Support keyboard shortcuts as described by the [ARIA pattern](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/examples/listbox-scrollable/).

## Checklist


packages/headless/src/utils/id.ts
```diff
@@ -3,7 +3,7 @@
  */
 // ⚠️ we make use of an IIFE to encapsulate the globalCounter so it can never accidentally be used somewhere else.
 const nextId = (() => {
-  let globalCounter = 0;
+  let globalCounter = 1;
```
JoCa96, line 6: What's the difference? 😄

packages/headless/src/composables/listbox/createListbox.ts
```diff
@@ -1,34 +1,146 @@
...
+   */
+  onFocusNext?: (currentValue: ListboxValue) => void;
+  /**
+   * Hook when the previous option should be focused.
+   */
+  onFocusPrevious?: (currentValue: ListboxValue) => void;
+  /**
+   * Hook when the first option starting with the given label should be focused.
+   */
+  onFocusByLabel?: (label: string) => void;
```
JoCa96, line 49: “Focus” is an imprecise term here. The real focus always stays on the listbox itself. So IMHO would be preferable to either drop the term “focus” or use the term “active”, as that is what is used in the spec: https://w3c.github.io/aria/#managingfocus

JoCa96 (commented):
test

JoCa96 (commented):
test

packages/headless/src/composables/listbox/createListbox.ts
```diff
@@ -1,34 +1,146 @@
...
+  /**
+   * Map for option IDs. key = option value, key = ID for the HTML element
+   */
+  const descendantKeyIdMap = new Map<ListboxValue, string>();
+
+  const getOptionId = (value: ListboxValue) => {
+    if (!descendantKeyIdMap.has(value)) {
+      descendantKeyIdMap.set(value, createId("listbox-option"));
+    }
+    return descendantKeyIdMap.get(value) as NonNullable<ReturnType<typeof descendantKeyIdMap.get>>;
```
JoCa96, line 70:
```suggestion
    return descendantKeyIdMap.get(value)!;
```

packages/headless/src/composables/listbox/TestListbox.vue
```diff
@@ -0,0 +1,90 @@
...
+    if (currentIndex < options.length - 1) {
+      focusedOption.value = options[currentIndex + 1];
+    }
+  },
+  onFocusPrevious: (currentValue) => {
+    const currentIndex = options.findIndex((i) => i === currentValue);
+    if (currentIndex > 0) focusedOption.value = options[currentIndex - 1];
+  },
+  onScrollIntoView: (id) => {
+    const option = listboxRef.value?.querySelector(`#${id}`);
```
JoCa96, line 50: You don't need a listboxRef :)
```suggestion
    const option = document.getElementById(id);
```

JoCa96 (changes requested):

JoCa96:
test

packages/headless/src/composables/listbox/TestListbox.vue
JoCa96: We don't even need the function now :)

JoCa96 (changes requested):

packages/headless/src/utils/id.ts
larsrickert: Its less confusing if you take a look at the DOM in the dev tools, because with initial value 0 e.g. the active-descendant is 1 if you have focused the second option.

Functionally it does not make a difference

packages/headless/src/composables/listbox/createListbox.ts
larsrickert: Done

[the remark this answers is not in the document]
larsrickert: #714

packages/headless/src/composables/listbox/TestListbox.vue
```diff
@@ -0,0 +1,90 @@
+<script lang="ts" setup>
+import { ref } from "vue";
+import { createListbox, type ListboxValue } from "./createListbox";
+
+const listboxRef = ref<HTMLUListElement>();
```
JoCa96, line 5:
```suggestion
```

JoCa96 (approved):

Issue #682, opened 2024-03-19
Fix OnyxButton alignment


BoppLi:
The HTML button behaves differently than the OnyxButton: 

- <button> has display: inline-block
- <OnyxButton> has display: flex

We could align it by using `inline-flex`.

### Open Questions/To-dos

- [x] Was it intentional to **not** make it inline? -> no

### Why?

The HTML button behaves differently than the OnyxButton.

### Acceptance criteria

- The button is inline aligned.
- Outer text is aligned with button text.

### Definition of Done

- [x] examples updated in alpha-test-app
- [x] updated version + documentation is deployed
- [x] Set task to `In Approval`

**Approval**
- [x] approved by dev @BoppLi 

### Implementation details

- use `inline-flex`

BoppLi:
Maybe noteworthy:
https://stackoverflow.com/questions/27418104/whats-the-difference-between-displayinline-flex-and-displayflex
![Image](https://github.com/SchwarzIT/onyx/assets/151019977/28484bb4-dae3-41e0-9c37-7ac1f9c43bd2)

PR #709, opened 2024-03-20
fix(OnyxButton): change to inline button

JoCa96:
Relates to #682 

## Checklist

- [ ] All changes are documented in the documentation app (folder `apps/docs`)

### DIFF
--- a/packages/sit-onyx/src/components/OnyxButton/OnyxButton.vue
+++ b/packages/sit-onyx/src/components/OnyxButton/OnyxButton.vue
@@ -48,7 +48,7 @@ $button-height: 2.5rem;
   --onyx-button-text-color: var(--onyx-color-text-icons-primary-intense);
   --onyx-button-outline-color: var(--onyx-color-base-primary-200);
 
-  display: flex;
+  display: inline-flex;
   height: $button-height;
   max-width: 100%;
   box-sizing: border-box;


packages/sit-onyx/playwright/snapshots/components/OnyxButton/inline-aligned-chromium-linux.png
larsrickert, line 1: Can we limit the screenshot width to the actual content width so its better visible?

packages/sit-onyx/src/styles/fonts.scss
```diff
@@ -1,6 +1,9 @@
 .onyx-text {
-  font-weight: 400;
-  font-size: 1rem;
+  font: {
+    weight: 400;
+    size: 1rem;
+    family: var(--onyx-font-family);
```
larsrickert, line 5: I didn't add the font family intentionally because it breaks the inheritance...you can take a look at the typography page in the VitePress docs which now will no longer work with the monospace font

larsrickert (changes requested):

packages/sit-onyx/src/styles/fonts.scss
JoCa96: Removed my changes and created follow-up ticket

larsrickert (approved):

PR #717, opened 2024-03-21
feat(OnyxTooltip): add property `open`

larsrickert:
Relates to #567

Add trigger property to trigger the tooltip on hover, click or boolean.

## Checklist


packages/headless/src/composables/tooltip/createTooltip.ts
```diff
@@ -1,19 +1,116 @@
...
 import { createBuilder } from "../../utils/builder";
 
-export const createTooltip = createBuilder(() => {
+export type CreateTooltipOptions = {
+  trigger: MaybeRef<TooltipTrigger | boolean>;
+  /**
+   * Number of milliseconds to use as debounce when showing/hiding the tooltip
+   * with openMode "hover".
+   *
+   * @default 0
```
JoCa96, line 11: I would prefer an API that doesn't allow a delay to be set when open type is not "hover"

E.g.
```ts
type CreateTooltipOptions = {
    open: {
    type: 'hover';
    delay: number;
  } | "hover" | "click" | boolean;
}
```
or some using Generics

packages/headless/src/composables/tooltip/createTooltip.ts
```diff
@@ -1,19 +1,116 @@
...
 import { createId } from "../..";
 import { createBuilder } from "../../utils/builder";
 
-export const createTooltip = createBuilder(() => {
+export type CreateTooltipOptions = {
+  trigger: MaybeRef<TooltipTrigger | boolean>;
+  /**
+   * Number of milliseconds to use as debounce when showing/hiding the tooltip
+   * with openMode "hover".
+   *
```
JoCa96, line 10: A delay of `0` can be annoying, so I would prefer something like `1000`

packages/headless/src/composables/tooltip/createTooltip.ts
```diff
@@ -1,19 +1,116 @@
...
     elements: {
-      trigger: {
+      trigger: computed(() => ({
         "aria-describedby": tooltipId,
-      },
-      tooltip: {
+        onClick: unref(options.trigger) === "click" ? handleClick : undefined,
+        ...hoverEvents.value,
+      })),
+      tooltip: computed(() => ({
```
JoCa96, line 106: The tooltip content should not be able to receive focus, to be safe we should add `tabindex: -1`.
```suggestion
      tooltip: computed(() => ({
        tabindex: "-1",
```

packages/sit-onyx/src/components/OnyxTooltip/OnyxTooltip.vue
```diff
@@ -18,11 +20,15 @@ defineSlots<{
...
+  state: { isVisible },
+} = createTooltip({
+  trigger: computed(() => props.trigger),
+  debounce: 200,
+});
 </script>
 
 <template>
-  <div class="onyx-tooltip-wrapper">
+  <div class="onyx-tooltip-wrapper" :class="{ 'onyx-tooltip-wrapper--hidden': !isVisible }">
```
JoCa96, line 31: Might be nicer to set the class on the tooltip element, where it is actually applied.

JoCa96 (changes requested):

packages/headless/src/composables/tooltip/createTooltip.ts
```diff
@@ -1,19 +1,116 @@
...
-export const createTooltip = createBuilder(() => {
+export type CreateTooltipOptions = {
+  trigger: MaybeRef<TooltipTrigger | boolean>;
+  /**
+   * Number of milliseconds to use as debounce when showing/hiding the tooltip
+   * with openMode "hover".
+   *
+   * @default 0
+   */
+  debounce?: number;
```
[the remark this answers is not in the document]
larsrickert: - delay: if multiple events are fired, of all them get executed
- debounce: if multiple events are fired, only the last one gets executed

So we currently have (and need) a debounce, not a delay :)

packages/headless/src/composables/tooltip/createTooltip.ts
larsrickert: I set a default value of 200 which can be overridden if needed

packages/headless/src/composables/tooltip/createTooltip.ts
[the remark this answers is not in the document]
BoppLi: from our 2022 sitcon talk, if helpful:
- throttle: Reduces trigger rate - recommended for scroll/resize events
- debounce: Waits until firing stops - recommended for input events

JoCa96 (approved):

PR #722, opened 2024-03-21
chore: add title for invisible labels

JoCa96:
Add native tooltips when there is no visible label


packages/sit-onyx/src/components/OnyxCheckbox/OnyxCheckbox.vue
```diff
@@ -44,6 +44,7 @@ const isTouched = ref(false);
         v-else
         v-model="isChecked"
         :aria-label="props.hideLabel ? props.label : undefined"
+        :title="props.hideLabel ? props.label : undefined"
```
BoppLi, line 47: I'd suggest to enhance the description text inside the jsdoc of the `label` property to inform the devs that this will happen, so they understand that all users will be able to see the text even when hidden (and are not tempted to take a short-cut with an " " string e.g. ^^)

packages/sit-onyx/src/components/OnyxIconButton/OnyxIconButton.vue
```diff
@@ -24,6 +24,7 @@ const emit = defineEmits<{
   <button
     class="onyx-icon-button"
     :aria-label="props.label"
+    :title="props.label"
```
BoppLi, line 27: question: shouldn't the icon itself also have a description label? 
could be a follow-up task how to approach it, also if we want to provide e.g. default labels for the onyx icons if that is even possible ^^

packages/sit-onyx/src/components/OnyxInput/OnyxInput.vue
```diff
@@ -89,6 +89,7 @@ const shouldShowCounter = computed(() => props.withCounter && props.maxlength);
           :minlength="props.minlength"
           :maxlength="props.maxlength"
           :aria-label="props.hideLabel ? props.label : undefined"
+          :title="props.hideLabel ? props.label : undefined"
```
BoppLi, line 92: could you maybe do a mini-adjustment to align the storybook names of the "without/hidden label" examples?
image

packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.vue
```diff
@@ -74,6 +74,7 @@ watch(
       :class="{ 'onyx-switch__input': true, 'onyx-switch__loading': props.loading }"
       type="checkbox"
       :aria-label="props.hideLabel ? props.label : undefined"
+      :title="props.hideLabel ? props.label : undefined"
```
BoppLi, line 77: did you try it out? somehow (only) the switch does not show the label when I hover on it in the 
http://localhost:6006/?path=/story/components-onyxswitch--default&args=hideLabel:!true

BoppLi: also, we don't have a storybook pre-defined example here like on the other components for the "hidden label" example. not sure if we always need it but it might be better for consistency ¯\_(ツ)_/¯

BoppLi (changes requested):

BoppLi:
Please also create a changeset because it affects the lib usage

packages/sit-onyx/src/components/OnyxIconButton/OnyxIconButton.vue
JoCa96: The issue is that icons not necessarily have a single use-case or meaning. The meaning might even be different based on the context it's used in. Therefore, it's preferably to force the user to set a context specific label and to always aria-hide the icon.

packages/sit-onyx/src/components/OnyxInput/OnyxInput.vue
JoCa96: done

packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.vue
JoCa96: fixed

packages/sit-onyx/src/components/OnyxCheckbox/OnyxCheckbox.vue
JoCa96: done

packages/sit-onyx/src/components/OnyxIconButton/OnyxIconButton.vue
BoppLi: sounds good :) 
should we then add a ticket to add `label` to OnyxIconProps?

packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.stories.ts
```diff
@@ -58,6 +58,17 @@ export const CheckedLoading = {
   },
 } satisfies Story;
 
+/**
+ * This example shows an switch without a visual label.
```
BoppLi, line 62:
```suggestion
 * This example shows a switch without a visual label.
```

packages/sit-onyx/src/components/OnyxListbox/types.ts
```diff
@@ -15,7 +15,7 @@ export type OnyxListboxProps<TValue extends SelectionOptionValue = SelectionOpti
    */
   modelValue?: TValue;
   /**
-   * If true, the label will be visually hidden.
+   * If `true`, the label will be visually hidden and the `title` attribute will be set.
```
BoppLi, line 18: the listbox does not have the title feature

BoppLi (changes requested):

packages/sit-onyx/src/components/OnyxListbox/types.ts
JoCa96: done

packages/sit-onyx/src/components/OnyxIconButton/OnyxIconButton.vue
JoCa96: No, I think the icon should always rather be used in a context where there is some label or description on the parent or accompanying element

BoppLi: ok :) hope UX will consider it ^^

BoppLi: in the past there were examples e.g. in the context of a headline status icon or the table, but then the projects devs can still wrap it in a tooltip ^^

BoppLi (approved):

PR #740, opened 2024-03-22
feat: apply tooltip feedback

larsrickert:
Relates to https://github.com/SchwarzIT/onyx/issues/567#issuecomment-2014799803

Implement feedback for the OnyxTooltip.

- Use button for click example
- document that trigger slot should be focussble
- add slot `tooltip` for custom content
- close tooltip with "Escape" for both hover and click mode

## Checklist


packages/sit-onyx/src/components/OnyxTooltip/OnyxTooltip.vue
```diff
@@ -14,8 +14,17 @@ const props = withDefaults(defineProps<OnyxTooltipProps>(), {
 defineSlots<{
   /**
    * Default slot where the parent content is placed that should open/close the tooltip.
+   *
+   * **Accessibility**: Please ensure that your content is focusable (e.g. the component itself or by adding a tabindex)
```
JoCa96, line 18:
```suggestion
   * **Accessibility**: Please ensure that your content includes at least one focusable element (by using a button or input element)
```
It's better to not promote `tabindex` to users, otherwise they might not use semantic HTML and creating more issues by making random elements focusable.

packages/sit-onyx/src/components/OnyxTooltip/OnyxTooltip.stories.ts
```diff
@@ -43,7 +48,7 @@ type Story = StoryObj<typeof OnyxTooltip>;
 export const Default = {
   args: {
     text: "Tooltip text",
-    default: () => "Slot content goes here",
+    default: () => h("span", { tabindex: "0" }, "Slot content goes here"),
```
JoCa96, line 51: See comment before, we should avoid promoting setting `tabindex: 0`.

packages/headless/src/composables/tooltip/createTooltip.ts
```diff
@@ -96,16 +96,15 @@ export const createTooltip = createBuilder((options: CreateTooltipOptions) => {
     if (isOutsideClick) _isVisible.value = false;
   };
 
+  document.addEventListener("keydown", handleDocumentKeydown);
```
JoCa96, line 99: I just noticed we don't have any way to clean up this global event-listener on unmount.

packages/headless/src/composables/tooltip/createTooltip.ts
```diff
@@ -96,16 +96,15 @@ export const createTooltip = createBuilder((options: CreateTooltipOptions) => {
...
    * the tooltip.
    */
   watchEffect(() => {
-    document.removeEventListener("keydown", handleDocumentKeydown);
     document.removeEventListener("click", handleDocumentClick);
-
     if (openType.value === "click") {
-      document.addEventListener("keydown", handleDocumentKeydown);
       document.addEventListener("click", handleDocumentClick);
     }
```
JoCa96, line 109: We can optimize the code a bit by only removing and adding when necessary.
`addEventListener` won't do anything when the same event and callback combo already exists.
`removeEventListener` won't do anything when the event-listener doesn't exist. 

```vue
  watchEffect(() => {
    if (openType.value === "click") {
      document.addEventListener("click", handleDocumentClick);
    } else {
      document.removeEventListener("click", handleDocumentClick);
    }
```

packages/headless/src/composables/tooltip/createTooltip.ts
larsrickert: We already have a onBeforeUnmount hook to clean up:
https://github.com/SchwarzIT/onyx/blob/cb3a72b4359cad361fe89f61c01fdc18d014c005/packages/headless/src/composables/tooltip/createTooltip.ts#L116

packages/sit-onyx/src/components/OnyxTooltip/OnyxTooltip.stories.ts
larsrickert: I used a button for all examples now

Issue #422, opened 2024-02-26
Define css layer concept


mj-hof:
POC for the usage of CSS layers with `reset.css` as a concept

### Why?

Layers allow us to explicitly define the Style cascade.
This makes it easier for us to handle styling and trivial for users to overwrite styles.

### Outcome of POC

Definition and how layers are used:
- Amount of layers
- Naming of layers
- Usage of sublayers

### Acceptance criteria

- [x] dev colleagues are onboarded
- [x] reset.css is considered

### Implementation details

See also:
- #318 (discussion)
- #338  

- consider defining component state styles via sublayer

- https://css-tricks.com/css-cascade-layers/#introduction-what-are-cascade-layers
- https://gist.github.com/Asjas/4b0736108d56197fce0ec9068145b421
- https://developer.mozilla.org/en-US/docs/Web/CSS/@layer

Issue #338, opened 2024-02-16
Implement css layer concept


JoCa96:
### Depends on

- #318 (discussion)
- #422 

### Why?

Layers allow us to explicitly define the Style cascade.
This makes it easier for us to handle styling and trivial for users to overwrite styles.

### Acceptance criteria

- [x] layers and potential sub-layers are defined in `layers.scss` and used in the global SCSS
- [x] existing components are updated to use these layers
- [x] dev colleagues are onboarded before
- [x] reset.css is considered

### Definition of Done

- [x] follow-up tickets were created if necessary
- [x] updated version + documentation is deployed
- [x] Set task to `In review`

**Review**
- [x] approved by dev @larsrickert 
  

### Implementation details

- consider defining component state styles via sublayer

- https://css-tricks.com/css-cascade-layers/#introduction-what-are-cascade-layers
- https://gist.github.com/Asjas/4b0736108d56197fce0ec906- 
- https://developer.mozilla.org/en-US/docs/Web/CSS/@layer

JoCa96:
See #318

PR #768, opened 2024-03-26
feat: Make use of CSS layers

JoCa96:
Relates to #422, #338

- Created component mixin that applies component layer and normalize CSS
- Restructured styles and split them into global and mixins. This avoid repeated includes of global styles.
- Applied mixin for all components and removed redundant styles
- Updated screenshots as the fieldset had an unintentional margin

Added developer documentation in https://github.com/SchwarzIT/onyx/wiki/Contribution-Coding-Guidelines#css-layers


packages/sit-onyx/src/components/OnyxAppLayout/OnyxAppLayout.vue
```diff
@@ -33,55 +33,57 @@ const slots = defineSlots<{
 </template>
 
 <style lang="scss">
+@use "../../styles/mixins/layers.scss";
```
larsrickert, line 36:
```suggestion
@use "@/styles/mixins/layers.scss";
```

(optional) Its maybe easier to import it like so :) Same for all the other components then

packages/sit-onyx/src/components/OnyxIconButton/OnyxIconButton.vue
```diff
@@ -25,7 +25,7 @@ const emit = defineEmits<{
 
 <template>
   <button
-    class="onyx-icon-button"
+    class="onyx-icon-button onyx-reset"
```
larsrickert, line 28: Why is this needed?

packages/sit-onyx/src/styles/globals/density.scss
larsrickert, line 1: Would prefer to remove the `globals` styles since all our styles are global anyways so we can keep the folder structure flat

packages/sit-onyx/src/styles/globals/layers.scss
```diff
@@ -0,0 +1 @@
+@layer onyx.reset, onyx.utility, onyx.component, onyx.density;
```
larsrickert, line 1: What does this do / why is this needed? Maybe you can add a small comment

packages/sit-onyx/src/styles/globals/root.scss
```diff
@@ -0,0 +1,30 @@
+@layer onyx.utility {
```
larsrickert, line 1: Instead of putting all these general styles in a `utility` layer, can't we define a general `onyx` layer instead?

packages/sit-onyx/src/styles/mixins/layers.scss
```diff
@@ -0,0 +1,11 @@
+@use "normalize.scss";
+
+@mixin component() {
+  @layer onyx.reset {
+    @include normalize.apply;
```
larsrickert, line 5: Alternatively we could place the normalize styles itself in `onyx.reset` layer so we could remove this block and just define the below layers as `@layer onyx.reset onyx.component`

packages/sit-onyx/src/styles/mixins/normalize.scss
```diff
@@ -0,0 +1,221 @@
...
+  }
+
+  &,
+  * {
+    &,
+    &::before,
+    &::after {
+      margin: 0;
+
+      box-sizing: border-box;
```
larsrickert, line 13: So we can remove this from all our existing components, right?

packages/sit-onyx/src/styles/mixins/normalize.scss
```diff
@@ -0,0 +1,221 @@
...
+
+      -moz-text-size-adjust: none;
+      -webkit-text-size-adjust: none;
+      text-size-adjust: none;
+    }
+
+    &:is(ul, ol) {
+      list-style: none;
```
larsrickert, line 21: Not sure if this is intended...what if I want to place a simple list in my project inside the app layout?

packages/sit-onyx/src/styles/mixins/normalize.scss
```diff
@@ -0,0 +1,221 @@
...
+    ================
+    */
+
+    /**
+    1. Add the correct height in Firefox.
+    2. Correct the inheritance of border color in Firefox. (https://bugzilla.mozilla.org/show_bug.cgi?id=190655)
+    */
+
+    &:is(hr) {
+      height: 0; /* 1 */
```
larsrickert, line 41: This will hide the element, right? Why would we want this?

packages/sit-onyx/src/styles/mixins/normalize.scss
```diff
@@ -0,0 +1,221 @@
...
+      border-style: none;
+      padding: 0;
+    }
+
+    /**
+    Restore the focus styles unset by the previous rule.
+    */
+
+    &:-moz-focusring {
+      outline: 1px dotted ButtonText;
```
larsrickert, line 143:
```suggestion
      outline: var(--onyx-1px-in-rem) dotted ButtonText;
```

Also what is this magic `ButtonText` value?

larsrickert (changes requested):

larsrickert:
The bundle sizes is increasing by around 100kb. Is it possible that all the layers styles are redundantly copied to every component?

packages/sit-onyx/src/components/OnyxIconButton/OnyxIconButton.vue
JoCa96: Removed it. My first approach was to set a class.

packages/sit-onyx/src/components/OnyxAppLayout/OnyxAppLayout.vue
JoCa96: Done

packages/sit-onyx/src/styles/globals/density.scss
JoCa96: If we don't separate them all globals will be duplicated with each include

packages/sit-onyx/src/styles/globals/layers.scss
JoCa96: done

packages/sit-onyx/src/styles/globals/root.scss
JoCa96: See in [packages/sit-onyx/src/styles/globals/layers.scss](https://github.com/SchwarzIT/onyx/pull/768/files/20d7fb6c28eb4bdb8a515bd85c4f7a98ff6fe7c5#diff-7d8e57f01d19f241269bebb9d98720661ce8bdb62b45d21d46e28c6dff96cf28) we only define sub-layers.

packages/sit-onyx/src/styles/mixins/normalize.scss
JoCa96: correct!

packages/sit-onyx/src/styles/mixins/normalize.scss
JoCa96: updated comment

packages/sit-onyx/src/styles/mixins/layers.scss
JoCa96: thats not how layers work, layers only define the cascade not any rules

packages/sit-onyx/src/styles/mixins/normalize.scss
JoCa96: A firefox specifc style value, as described in the comment we are setting the original default values here.

packages/sit-onyx/src/styles/globals/density.scss
larsrickert: I mean we can still separate the files but move them 1 folder up so we don't need the globals folder

JoCa96:
> The bundle sizes is increasing by around 100kb. Is it possible that all the layers styles are redundantly copied to every component?

Yes, the include happens for every component. The 100kb increase is only for uncompressed build outputs. We could optimize this with some build scripts, but the payoff is not worth it right now.

larsrickert:
> Yes, the include happens for every component. The 100kb increase is only for uncompressed build outputs. We could optimize this with some build scripts, but the payoff is not worth it right now.

Hmm I think we should find a way to prevent this. 100kb is a lot in comparisson for the benefits we get when using the layers

JoCa96:
@larsrickert As discussed, added follow-up issue to reduce CSS duplication: https://github.com/SchwarzIT/onyx/issues/812

larsrickert (dismissed):

packages/sit-onyx/src/components/OnyxBadge/OnyxBadge.vue
```diff
@@ -38,7 +38,7 @@ defineSlots<{
 </template>
 
 <style lang="scss">
-@use "../../styles/density.scss";
+@use "../../styles/mixins/density.scss";
```
larsrickert, line 41: We also need to use the component layer here, right?

larsrickert (approved):

packages/sit-onyx/src/components/OnyxBadge/OnyxBadge.vue
JoCa96: true, done :)

larsrickert (dismissed):

larsrickert (approved):

larsrickert (approved):

Issue #368, opened 2024-02-20
Implement density


MajaZarkova:
### Depends on:

- [x] #246 

### Acceptance criteria

- [x] For all existing component, that support density:
  - [x] Extended component properties with `DensityProp`
  - [x] Implemented all three densities types
- [x] Add feature to Alpha-Test-App to toggle density for the whole page
- [x] Updated AC in the [Dev Reference Ticket](https://github.com/orgs/SchwarzIT/projects/5/views/7?filterQuery=refer&pane=issue&itemId=49798142) to include density as a new baseline requirement

### Design

- [Figma Button](https://www.figma.com/file/YfEUBOHk4J4nYrk04geswG/onyx---Design-System-Figma-Library?node-id=896%3A5709&mode=dev)
- [Figma Checkbox](https://www.figma.com/file/YfEUBOHk4J4nYrk04geswG/onyx---Design-System-Figma-Library?node-id=511%3A199&mode=dev)
- [Figma RadioGroup](https://www.figma.com/file/YfEUBOHk4J4nYrk04geswG/onyx---Design-System-Figma-Library?type=design&node-id=626%3A15811&mode=dev)
- [Figma Switch](https://www.figma.com/file/YfEUBOHk4J4nYrk04geswG/onyx---Design-System-Figma-Library?node-id=711%3A12776&mode=dev)

### Definition of Done

- [x] covered by component tests
- [x] screenshot tests are updated
- [x] follow-up tickets were created if necessary
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature
- [x] Set task to `In final approval`

**Review**
- [Storybook Button](https://storybook.onyx.schwarz/?path=/docs/components-onyxbutton--docs)
- [Storybook Checkbox Group](https://storybook.onyx.schwarz/?path=/docs/components-onyxcheckboxgroup--docs)
- [Storybook Icon Button](https://storybook.onyx.schwarz/?path=/docs/components-onyxiconbutton--docs)
- [Storybook Input](https://storybook.onyx.schwarz/?path=/docs/components-onyxinput--docs)
- [Storybook RadioButton Group](https://storybook.onyx.schwarz/?path=/docs/components-onyxradiobuttongroup--docs)
- [Storybook Swith](https://storybook.onyx.schwarz/?path=/docs/components-onyxswitch--docs)
- [x] approved by designer @mj-hof 
- [x] approved by dev @larsrickert 
  

### Implementation details

[...]

PR #780, opened 2024-03-28
feat(alpha-test-app): Add density to alpha test app

MajaZarkova:
Relates to #368 

Added density radio button which automatically changes the density to all components that have this property.

## Checklist


apps/alpha-test-app/src/views/HomeView.vue
```diff
@@ -51,6 +51,21 @@ const show = computed(() => {
   return (componentName: (typeof COMPONENTS)[number]) => activeConfig.value.includes(componentName);
 });
 
+const DENSITIES = ["Compact", "Default", "Cozy"];
+type DensityType = "default" | "cozy" | "compact" | undefined;
+
+const densityOptions = DENSITIES.map((id) => ({
+  id,
+  label: id,
+  value: id.toLowerCase() as DensityType,
```
larsrickert, line 60:
```suggestion
```

This is not needed, right? The id is used as value for the radio group

apps/alpha-test-app/src/views/HomeView.vue
```diff
@@ -51,6 +51,21 @@ const show = computed(() => {
   return (componentName: (typeof COMPONENTS)[number]) => activeConfig.value.includes(componentName);
 });
 
+const DENSITIES = ["Compact", "Default", "Cozy"];
+type DensityType = "default" | "cozy" | "compact" | undefined;
```
larsrickert, line 55: We already have this inside /composables/density.ts, we just need to export it :)

apps/alpha-test-app/src/views/HomeView.vue
```diff
@@ -51,6 +51,21 @@ const show = computed(() => {
...
+  id,
+  label: id,
+  value: id.toLowerCase() as DensityType,
+}));
+
+const activeDensityOption = ref({
+  id: "Default",
+  label: "Default",
+  value: "default" as DensityType,
+});
```
larsrickert, line 67:
```suggestion
const activeDensityOption = ref({ ...densityOptions[0] })
```

apps/alpha-test-app/src/views/HomeView.vue
```diff
@@ -111,13 +132,18 @@ const listboxOptions = [
...
+            label="Button"
+            :density="activeDensityOption.value"
+          />
 
           <template v-if="show('OnyxCheckboxGroup')">
             <OnyxCheckboxGroup
               v-model="checkboxState"
               headline="Checkbox Group"
               :options="dummyOptions"
+              :density="activeDensityOption.value"
```
larsrickert, line 146: We don't need to set it on every component. We can just set the [density class](https://onyx.schwarz/development/density.html) on the parent div

larsrickert (changes requested):

apps/alpha-test-app/src/views/HomeView.vue
```diff
@@ -21,6 +21,7 @@ import {
 } from "sit-onyx";
 import { computed, ref } from "vue";
 import { useRouter } from "vue-router";
+import type { DensityType } from "../../../../packages/sit-onyx/dist/composables/density";
```
larsrickert, line 24: You can also import the `DENSITY` array from here so we don't need to re-define it here :)
To capitalize the first letter you can import `capitalize` from vue

larsrickert (changes requested):

apps/alpha-test-app/src/views/HomeView.vue
MajaZarkova: currently we are not exporting the array, I'm trying to export it but it gets removed for some reason and I'm getting errors

larsrickert: You probably need to re-start the demo app so it re-builds and recognized the new exported type.
Are you actually using the new array? Otherwise prettier will remove the unused import

MajaZarkova: image

MajaZarkova: Maybe the path is not correct?

apps/alpha-test-app/src/views/HomeView.vue
```diff
@@ -51,6 +52,13 @@ const show = computed(() => {
   return (componentName: (typeof COMPONENTS)[number]) => activeConfig.value.includes(componentName);
 });
 
+const densityOptions = DENSITY.map((id: string) => ({
+  id: id as DensityType,
+  label: capitalize(id),
+})) satisfies SelectionOption<string>[];
```
larsrickert, line 58:
```suggestion
const densityOptions = DENSITY.map((id) => ({
  id,
  label: capitalize(id),
})) satisfies SelectionOption<string>[];
```

larsrickert (changes requested):

larsrickert (approved):

PR #787, opened 2024-03-28
refactor(OnyxListbox): remove property label in favor of message

larsrickert:
Relates to https://github.com/SchwarzIT/onyx/issues/564#issuecomment-2015089322

As described by Jannick in the comment above, the text is more a message/description than a (aria-)label.

## Checklist


packages/sit-onyx/src/components/OnyxListbox/types.ts
```diff
@@ -1,24 +1,22 @@
 import type { SelectionOptionValue } from "../OnyxRadioButton/types";
 
 export type OnyxListboxProps<TValue extends SelectionOptionValue = SelectionOptionValue> = {
+  /**
+   * Aria label. Must be set for accessibility reasons.
+   */
+  label: string;
```
BoppLi, line 7: will the listbox ever be used standalone without a select?
if not, the label won't be needed here I think because from a11y perspective the select above will provide all context you need. might get redundant else.
but you can keep it for now and we'll see how it works out when baking the Listbox into the Select.

BoppLi (approved):

packages/sit-onyx/src/components/OnyxListbox/types.ts
larsrickert: Yes in general the listbox can be used standalone :)
Also accessibility tests will fail if it does not have a aria label

Issue #386, opened 2024-02-21
Implement OnyxIconButton


mj-hof:
### Depends on

- [x] #190

### Why?

- In addition to the OnyxButton we want to implement the OnyxIconButton as a standalone button.
- It's different enough to be its own component

### Design

[Figma link](https://www.figma.com/file/YfEUBOHk4J4nYrk04geswG/onyx---Design-System-Figma-Library?type=design&node-id=836-6659&mode=design)

### Acceptance criteria

- Supports mode:
  - [x] primary
  - [x] secondary
  - [x] danger
- States:
  - [x] default
  - [x] disabled
  - [x] loading
- Focus:
  - [x] none
  - [x] hover
  - [x] focus-visible
- [ ] ~For the LayoutDemoPage in the alpha-test-app: 
        replace the OnyxButton inside the "flotingbutton" component with `OnyxIconButton`~

### Definition of Done

- [x] covered by component tests
- [x] screenshot tests are updated
- [x] follow-up tickets were created if necessary
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature
- [x] Set task to `In review`

**Review**

- [Storybook](https://storybook.onyx.schwarz/?path=/docs/components-onyxiconbutton--docs)
- [x] approved by designer @jannick-ux
- [x] approved by dev @larsrickert 

### Implementation details

- must always require label for ARIA purposes

#### Reference implementations

- [CoreUI](https://core-ui.schwarz/components/scu-icon-button)

#### Applicable ARIA Pattern

- [Button](https://www.w3.org/WAI/ARIA/apg/patterns/button/)

JoCa96:
> For the LayoutDemoPage in the alpha-test-app: 
>   - replace the OnyxButton inside the "flotingbutton" component with `OnyxIconButton`

Removed this AC, as the button not intended to be used here from a design perspective

jannick-ux:
The icon size is restricted to 24px. It cannot be larger or smaller. 
Also please use the icon "trash" as default icon.

PR #794, opened 2024-04-02
feat: implement invalid state for existing form components

larsrickert:
Relates to #386

Implement invalid states for checkbox, input, radio button and switch.
Screenshots and async custom errors will be done in a follow up PR.

## Checklist


apps/docs/src/.vitepress/components/OnyxIconLibrary.vue
```diff
@@ -48,7 +48,9 @@ const filteredCategories = computed(() => {
       <OnyxHeadline is="h3" class="category__headline">{{ category.name }}</OnyxHeadline>
 
       <div class="category__icons">
-        <IconLibraryItem v-for="icon in category.icons" :key="icon.iconName" :icon="icon" />
+        <ClientOnly>
```
JoCa96, line 51: Please add a commen to explain why this is necessary

apps/alpha-test-app/src/components/form-demo/FormDemo.vue
```diff
@@ -97,16 +102,14 @@ watch(
...
-      opacity: 0.4;
-    }
-  }
 
   &__actions {
     display: flex;
     gap: var(--onyx-spacing-xs);
   }
+
+  :deep(.onyx-input) {
```
JoCa96, line 111: Why do we need `:deep` here?

JoCa96 (changes requested):

apps/alpha-test-app/src/components/form-demo/FormDemo.vue
larsrickert: We don't need it, I removed it

apps/docs/src/.vitepress/components/OnyxIconLibrary.vue
larsrickert: Because the build failed because the `createTooltip` composable which is used in the OnyxTooltip called `document.addEventListener()` directly but the document is not available in SSR.

I fixed it to only add the listeners onBeforeMount

apps/alpha-test-app/src/components/form-demo/FormDemo.vue
```diff
@@ -1,26 +1,33 @@
...
+  type OnyxCheckboxGroupProps,
+  type OnyxRadioButtonGroupProps,
+} from "sit-onyx";
+import { ref, watchEffect } from "vue";
+
+export type FormData = Partial<{
   defaultInput: string;
   requiredInput: string;
   minlengthInput: string;
-  maxInput: number;
```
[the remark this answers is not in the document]
larsrickert: > you removed maxInput because we don't have a number input yet, right? but do we still keep the translations for now?

Good point, I removed the translations for now, is that fine for you?

> was it intentional not to move the TestInput into the alpha-test-app for the number example as discussed before easter?

Yes, I don't like having this TestInput lying around without any real purpose :) But if you insist, I can re-add it

> is there a kind of follow up ticket that points out including a number input here when we get it?

Probably the OnyxStepper component, right? Or should be support `type="number"` for the OnyxInput until we have a stepper?

larsrickert:
> Showing the error messages below the text-input is a later ticket right?

Yes :)

apps/alpha-test-app/src/components/form-demo/FormDemo.vue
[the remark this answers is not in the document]
BoppLi: > Yes, I don't like having this TestInput lying around without any real purpose :) But if you insist, I can re-add it

would prefer that over removing the translations, because e.g. the korean translations are already there.
of course, if onyxStepper comes in the next sprint already, that would also be fine. I just don't know when we planned to implement it, if it will be some time, use the TestInput in the meantime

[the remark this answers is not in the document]
BoppLi: > Probably the OnyxStepper component, right? Or should be support type="number" for the OnyxInput until we have a stepper?

should Jannick/Jonathan answer this maybe?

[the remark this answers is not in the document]
BoppLi: https://github.com/orgs/SchwarzIT/projects/5/views/7?pane=issue&itemId=59042435

JoCa96 (dismissed):

packages/sit-onyx/src/composables/useCustomValidity.ts
```diff
@@ -0,0 +1,96 @@
...
+  /**
+   * Sync isTouched state.
+   */
+  const stopWatch = watch(
+    () => options.props.modelValue,
+    () => {
+      isTouched.value = true;
+      nextTick(stopWatch);
+    },
+  );
```
JoCa96, line 57:
```suggestion

  watch(
    () => options.props.modelValue,
    () => isTouched.value = true,
    { once: true }
  );
```

JoCa96: Wouldn't this work, too?

JoCa96 (approved):

Issue #744, opened 2024-03-22
Implement OnyxEmpty component


larsrickert:
### Why?

A empty component can be used to indicate the user that there is currently no data available.

### Design

[Figma link](https://www.figma.com/file/YfEUBOHk4J4nYrk04geswG/onyx---Design-System-Figma-Library?node-id=1253%3A58552&mode=dev)

### Acceptance criteria

- [x] a required text is displayed and can be changed
- [x] an icon is display and can be changed from the default value

### Definition of Done

- [x] example usage added to `apps/alpha-test-app/src/views/HomeView.vue`
- [x] covered by component tests
- [x] screenshot tests are updated
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature
- [x] Set task to `In Approval`

**Approval**

- [Storybook link](https://storybook.onyx.schwarz/?path=/docs/components-onyxempty--docs)
- [x] approved by designer @jannick-ux
- [x] approved by dev @BoppLi 

### Implementation details

- background color should be inherit by the parent

PR #830, opened 2024-04-08
feat: add `OnyxEmpty` component

larsrickert:
Relates to #744

Add empty component.

## Checklist


packages/sit-onyx/src/components/OnyxEmpty/OnyxEmpty.stories.ts
```diff
@@ -0,0 +1,61 @@
+import { OnyxLink } from "@/index";
+import { createTruncationDecorator } from "@/utils/storybook";
+import emojiSad from "@sit-onyx/icons/emoji-sad.svg?raw";
+import { defineStorybookActionsAndVModels } from "@sit-onyx/storybook-utils";
+import type { Meta, StoryObj } from "@storybook/vue3";
+import { h } from "vue";
+import OnyxIcon from "../OnyxIcon/OnyxIcon.vue";
+import OnyxEmpty from "./OnyxEmpty.vue";
+
```
BoppLi, line 9: should there be an example with a different icon size? in Figma it is a property, I understand that that is not how it works with the slot but a Dev that inspects the component in the project might search for how it is done.
image

on the other hand teh dev might see in the example that OnyxIcon is used and look into the docs of OnyxIcon ^^

it's fine either way :)

image
when I saw this I first thought the size was about density

packages/sit-onyx/src/components/OnyxEmpty/OnyxEmpty.vue
```diff
@@ -0,0 +1,53 @@
+<script lang="ts" setup>
+import circleX from "@sit-onyx/icons/circle-x.svg?raw";
+import OnyxIcon from "../OnyxIcon/OnyxIcon.vue";
+
```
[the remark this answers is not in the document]
larsrickert: #833

packages/sit-onyx/src/components/OnyxEmpty/OnyxEmpty.vue
```diff
@@ -0,0 +1,53 @@
...
+</script>
+
+<template>
+  <div class="onyx-empty">
+    <slot name="icon">
+      <OnyxIcon :icon="circleX" size="48px" />
+    </slot>
+
+    <div class="onyx-empty__label onyx-text onyx-truncation-multiline">
+      <slot></slot>
```
BoppLi, line 24: question: was a default text desired from UX?
if not, should / do we want to add a default empty text to our translations that is also used inside the storybook example? could be quite handy to spare the projects from translating that all the time ^^

BoppLi (changes requested):

packages/sit-onyx/src/components/OnyxEmpty/OnyxEmpty.ct.tsx
```diff
@@ -0,0 +1,39 @@
+import { mockPlaywrightIcon } from "@/utils/playwright";
+import { expect, test } from "../../playwright-axe";
+import OnyxIcon from "../OnyxIcon/OnyxIcon.vue";
+import OnyxEmpty from "./OnyxEmpty.vue";
+
+test("should render", async ({ mount }) => {
```
BoppLi, line 6: isn't the a11y part missing?

```
  // ACT
  const accessibilityScanResults = await makeAxeBuilder().analyze();

  // ASSERT
  expect(accessibilityScanResults.violations).toEqual([]);
```

BoppLi (approved):

BoppLi (dismissed):

packages/sit-onyx/src/components/OnyxEmpty/OnyxEmpty.vue
BoppLi: we should ask jannick if we want to align it with figma, then the figma component should also have a default text :)
Inside components like the table would also be totally fine ^^

BoppLi (approved):

PR #908, opened 2024-04-15
feat(OnyxListbox): Improved typeAhead with support for matching multiple characters

JoCa96:
Relates to #565 

Improved type-ahead to support matching multiple characters typed in quick succession.
Also, [Named Keys](https://www.w3.org/TR/uievents-key/#named-key-attribute-values) are filtered out.


BoppLi:
> ### ⚠️ No Changeset found

this is a new feature for the listbox etc, I'd suggest to add a changeset

packages/headless/src/composables/typeAhead.ts
```diff
@@ -0,0 +1,26 @@
...
+ * @example
+ * ```ts
+ * const typeAhead = useTypeAhead((inputString) => onTypeAhead(inputString));
+ * // ...
+ * const onKeydown = (key: KeyboardEvent) => typeAhead(key.event);
+ * ```
+ */
+export const useTypeAhead = (callback: (input: string) => void) => {
+  let inputString = "";
+  const debouncedReset = debounce(() => (inputString = ""), 500);
```
BoppLi, line 16: maybe we should export the "500" as const and use it in the typeAhead.spec.ts and then add 100 in there, so if we change the debounce timeout for any reason, it will still work. (on the other hand the test will fail and we can adjust it then if needed.)

Or are we sure that 500 is the optimum and will not be touched?

packages/headless/src/composables/typeAhead.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  return (event: KeyboardEvent) => {
+    if (!isPrintableCharacter(event.key)) {
+      return;
+    }
+    debouncedReset();
```
BoppLi, line 22: each letter will trigger a new timeout, is that correct? should we cancel a previous timeout if there is one before starting a new one?

BoppLi: we might need a longer word in the listbox to test this

BoppLi (changes requested):

packages/headless/src/composables/typeAhead.ts
larsrickert: I changed it to `vi.runAllTimers();` so we do not rely on timeouts

BoppLi (dismissed):

larsrickert (approved):

BoppLi (approved):

Issue #926, opened 2024-04-16
Fix switch required marker when truncated


larsrickert:
The required and optional marker are not shown when truncated with ellipsis

![Image](https://github.com/SchwarzIT/onyx/assets/67898185/c78e93e6-d843-41a9-9a72-aaba333b44fc)

We can take a look at the checkbox group

- [x] Approved by UX

Issue #904, opened 2024-04-15
Fix switch label not aligned


larsrickert:
The switch label is not centerred vertically which can e.g. be seen if using the cozy density.

![Image](https://github.com/SchwarzIT/onyx/assets/67898185/f529bc7b-4930-4e90-bb0d-b878bab1c777)

When fixing this, make sure to check the the input is still top aligned if multiline truncation is used.

- [x] approved by UX

Issue #871, opened 2024-04-11
Optimize Playwright screenshots


larsrickert:
With our current Playwright matrix screenshot utility, a lot of single screenshot files are created. E.g. by adding one element to the matrix, several hundred images can be created because every combination is used (although only some might be relevant).

We should optimize this and e.g. combine multiple screenshots into one.

PR #928, opened 2024-04-16
refactor(OnyxSwitch): use new matrix screenshot utility

larsrickert:
Relates to #871

I found two bugs which should be fixed in follow up PRs:
- #904
- #926


packages/sit-onyx/playwright/snapshots/components/OnyxSwitch/Switch-densities--chromium-linux.png
MajaZarkova, line 1: The label for Cozy should be center aligned right?
image

MajaZarkova: oh, we have a follow up ticket for that. Okay :)

MajaZarkova (approved):

PR #935, opened 2024-04-16
refactor: clean up old playwright utilities

larsrickert:
Relates to #871

- delete old `executeScreenshotsForAllStates` utility
- move all Playwright utils to `playwright` folder so we have everything in one place

Finally the last PR for the screenshot refactoring :D


packages/eslint-plugin/src/rules/import-playwright-a11y.cjs
```diff
@@ -15,13 +15,18 @@ module.exports = {
...
+      if (context.filename.endsWith("/playwright/a11y.ts")) return;
 
       const hasFixture = ["expect", "test"].includes(node.imported.name);
       if (!hasFixture) return;
 
       // type check that node.parent.source.value exists
-      if (!("source" in node.parent) || !("value" in node.parent.source)) {
+      if (
+        !("source" in node.parent) ||
+        !node.parent.source ||
```
BoppLi, line 26: this one is included in the next line, as "value" could not exist in an undefined source
```suggestion
```

packages/eslint-plugin/src/rules/import-playwright-a11y.cjs
```diff
@@ -32,8 +37,8 @@ module.exports = {
       if (isDisallowedImport) {
         context.report({
           node,
-          loc: node.loc,
-          message: `Import "${node.imported.name}" from "../../playwright-axe" instead because Onyx uses custom Playwright fixtures for providing a global configuration for accessability testing.`,
+          loc: node.loc ?? undefined,
+          message: `Import "${node.imported.name}" from "../../playwright/a11y" instead because Onyx uses custom Playwright fixtures for providing a global configuration for accessability testing.`,
```
BoppLi, line 41:
```suggestion
          message: `Import "${node.imported.name}" from "../../playwright/a11y" instead because onyx uses custom Playwright fixtures for providing a global configuration for accessability testing.`,
```

BoppLi (changes requested):

BoppLi (approved):

Issue #855, opened 2024-04-10
Implement matrix accessibility testing


larsrickert:
- run accessibility tests for every permutation of our matrix screenshot tests

larsrickert:
POC: https://github.com/SchwarzIT/onyx/tree/screenshot-util-axe-tests

Issue #410, opened 2024-02-26
Research color contrast for themes


mj-hof:
- Regarding our accessibility topic, check the color contrast especially for the Digits theme (maybe Kaufland…)
- [ ] table hover effects in dark mode barely visible
- [x] #859

### Teams chat to this topic
https://teams.microsoft.com/l/message/19:KGEIFTqojUuOMMggXfOhvc772DhwrQWRJNMqrnvL6ak1@thread.tacv2/1707918850979?tenantId=d04f4717-5a6e-4b98-b3f9-6918e0385f4c&groupId=969ca3b3-9c78-411f-8629-eb770bbfdc7b&parentMessageId=1707918850979&teamName=onyx&channelName=General&createdTime=1707918850979

ChristianBusshoff:
color-contrast Rule got disabled for the Projekt in 
packages/sit-onyx/src/playwright/a11y.ts
Should be removed.

PR #942, opened 2024-04-16
test: run accessibility tests for every screenshot

larsrickert:
Relates to #855

Run the Playwright accessibility tests for every screenshot / component variation in our matrix screenshot utility.
For some components I disabled the contrast checking for now since they are failing until #410 is fixed by UX.

The workflow to update screenshots seems to take ~1m 30s longer (~7m 15s instead of 5m 45s) which is acceptable I guess based on the fact that we are performing ~400 accessibility tests.

## Checklist


packages/sit-onyx/src/components/OnyxButton/OnyxButton.vue
```diff
@@ -35,6 +35,7 @@ const emit = defineEmits<{
     ]"
     :disabled="props.disabled || props.loading"
     :type="props.type"
+    :aria-label="props.label"
```
BoppLi, line 38: are you sure that the label will not be read out by the screen reader twice?
maybe aria-labelledby would be better?

BoppLi (changes requested):
6/14

packages/sit-onyx/src/components/OnyxButton/OnyxButton.vue
larsrickert: Good point, we can not use `aria-labelledby` because there is no label when loading.
I changed it to only add the aria label when loading

packages/sit-onyx/src/components/OnyxListboxOption/OnyxListboxOption.ct.tsx
```diff
@@ -8,12 +8,17 @@ test.describe("Screenshot tests", () => {
...
+      // aria-required-parent is ignored here because this component is only a single option which is internally always
+      // used together with a parent so we disable the failing rule here
+      disabledAccessibilityRules: ["aria-required-parent", "color-contrast"],
       component: (column, row) => (
         <OnyxListboxOption
           aria-label="Label"
           aria-selected={column === "selected"}
           active={row === "focus-visible"}
           aria-disabled={state === "disabled"}
+          {...{ role: "option" }}
```
BoppLi, line 21: does this syntax not work? :O
```suggestion
          role="option"
```

BoppLi (changes requested):

packages/sit-onyx/src/components/OnyxListboxOption/OnyxListboxOption.ct.tsx
larsrickert: Unfortunately no, it throws a type error :( Thats why I did it in this "ungly" way

BoppLi: oki

BoppLi (approved):

Issue #949, opened 2024-04-17
Implement table header style


BoppLi:
### Depends on

- #948

### Why?

We need the styling for the table header

### Design

[Figma](https://www.figma.com/file/YfEUBOHk4J4nYrk04geswG/onyx---Design-System-Figma-Library?type=design&node-id=1967%3A33309&mode=dev)

### Acceptance criteria

- [x] `OnyxTable` has the expected styles for "th" matching figma

#### Out of scope

- hover state
- all features except the styling (e.g. sorting)
- scrolling behavior

### Definition of Done

- The following component requirements are implemented:
  - [x] `density`
- [x] covered by component tests
- [x] screenshot tests are updated
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature
- [x] Set task to `In Approval`

**Approval**

- [Storybook link](https://storybook.onyx.schwarz/?path=/docs/components-table--docs)
- [x] approved by designer @jannick-ux
- [x] approved by dev @BoppLi 

#### Reference implementations

- [Kendo UI](https://docs.telerik.com/kendo-ui/controls/grid/overview)
- [AG Grid](https://www.ag-grid.com/vue-data-grid/deep-dive/#create-a-basic-grid) 

#### Applicable ARIA Pattern

[Table](https://www.w3.org/WAI/ARIA/apg/patterns/table/)

Issue #837, opened 2024-04-08
Implement table row styles


JoCa96:
### Depends on

- #948

### Why?

We need a baseline of basic table elements, so that we are able to build simple, static tables.

### Design

[Figma](https://www.figma.com/file/YfEUBOHk4J4nYrk04geswG/onyx---Design-System-Figma-Library?type=design&node-id=1967%3A33309&mode=dev)

### Acceptance criteria

- [x] There is a `OnyxTable` which can be used to render simple, static tables.
- [x] The table can optionally be shown with striped rows
- [x] The borders can be shown as `horizontal` or `grid` (default: horizontal)

#### Out of scope

- #949
- #838
- The table must be implemented using `display: grid`
- Additional semantic elements like `thead`, `tbody`, `tfooter`, `tr` can be made transparent to the grid using `display: contents`

### Definition of Done

- The following component requirements are implemented:
  - [x] `density`
- [x] covered by component tests
- [x] screenshot tests are updated
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature
- [x] Set task to `In Approval`

**Approval**

- [Storybook link](https://storybook.onyx.schwarz/?path=/docs/components-table--docs)
- [x] approved by designer @jannick-ux
- [x] approved by dev @BoppLi 

### Implementation details
- The `OnyxTable` API, so far, must only be implemented using slots and the `strokeStyle` and `rowStyle` property.
- The `strokeStyle` and `rowStyle` properties must only set classes on the table itself, and use these to style the table.
- For now, `th` and `td` elements should be treated equally
- `thead` is used to identify the header columns

```vue
<OnyxTable strokeStyle="grid" rowStyle="crosswalk" >
   <thead>
      <tr>
        <th scope="col">Items</th>
        <th scope="col">Expenditure</th>
      </tr>
  </thead>
 <tbody>
    <tr>
      <td>Donuts</td>
      <td>3,000</td>
    </tr>
    <tr>
      <td>Stationery</td>
      <td>18,000</td>
    </tr>
  </tbody>
</OnyxTable>
```

#### Reference implementations

- [Kendo UI](https://docs.telerik.com/kendo-ui/controls/grid/overview)
- [AG Grid](https://www.ag-grid.com/vue-data-grid/deep-dive/#create-a-basic-grid) 

#### Applicable ARIA Pattern

[Table](https://www.w3.org/WAI/ARIA/apg/patterns/table/)

PR #959, opened 2024-04-18
feat: add table row and header styles

larsrickert:
Relates to #837, #949

Add table row and header styles

## Checklist


packages/sit-onyx/src/components/OnyxTable/types.ts
```diff
@@ -1,5 +1,12 @@
 import type { DensityProp } from "../../composables/density";
 
 export type OnyxTableProps = DensityProp & {
-  // TODO: add props as needed
+  /**
+   * Whether to use striped row background colors.
+   */
+  striped?: boolean;
+  /**
+   * If `true`, both horizontal and vertical borders will be displayed.
```
BoppLi, line 9: maybe this could help to make clear that we always have horizontal borders
```suggestion
   * If `true`, vertical borders will be displayed in addition to the default horizontal borders.
```

packages/sit-onyx/src/components/OnyxTable/OnyxTable.vue
```diff
@@ -1,11 +1,31 @@
...
 
-defineProps<OnyxTableProps>();
+const props = withDefaults(defineProps<OnyxTableProps>(), {
+  striped: false,
+  grid: false,
+});
+
+defineSlots<{
+  /**
+   * Table content (columns + rows).
```
BoppLi, line 12: should we recommend here to use th/tr/td elements?
if something else is passed, the whole style would not work after all.

packages/sit-onyx/src/components/OnyxTable/OnyxTable.vue
```diff
@@ -14,19 +34,121 @@ defineProps<OnyxTableProps>();
...
+  // border radius
+  tr:first-child th:first-child {
+    border-top-left-radius: $border-radius;
+  }
+
+  tr:first-child th:last-child {
+    border-top-right-radius: $border-radius;
+  }
+
+  tr:last-child td:first-child {
```
BoppLi, line 74: do the border bottoms still work if we have no td rows? I mean only the header, when e.g. the table content is still loading / is empty initially. 

well, probably we'll look after that when implementing loading/empty behavior 😋

BoppLi (approved):

packages/sit-onyx/src/components/OnyxTable/OnyxTable.vue
larsrickert: image

No the bottom borders will not be rounded then but I assumed that the table always has either content or an "empty/loading" state that we will implement in the future.

But if you want I can take care of this edge case

BoppLi: let's merge it for now and ask jannick tomorrow if we need a follow up ticket for that :>

packages/sit-onyx/src/components/OnyxTable/OnyxTable.vue
```diff
@@ -9,7 +9,7 @@ const props = withDefaults(defineProps<OnyxTableProps>(), {
 
 defineSlots<{
   /**
-   * Table content (columns + rows).
+   * Table content. Must only contain valid HTML `<table>` children like `<tr>`, `<th>` and `<td>`.
```
BoppLi, line 12: thead and tbody for completionism? ^^

BoppLi (approved):

BoppLi (approved):

Issue #566, opened 2024-03-11
Implement tag


mj-hof:
### Why?

Tags are succinct textual labels that provide single-worded information or hints to their related parent element.

### Design

[Figma link](https://www.figma.com/file/YfEUBOHk4J4nYrk04geswG/onyx---Design-System-Figma-Library?type=design&node-id=1110-1647&mode=dev)

### Acceptance criteria

- a tag can be shown with one of the available onyx colors
  - [x] primary
  - [x] secondary
  - [x] danger
  - [x] warning
  - [x] success
  - [x] info
- [x] an optional icon can be shown on the left side
- [x] the label is truncated with ellipsis (no multiline)

### Definition of Done

- [x] example usage added to alpha-test-app
- [x] covered by component tests
- [x] screenshot tests are updated
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature
- [x] Set task to `In Approval`

**Review**

- [Storybook link](https://storybook.onyx.schwarz/?path=/docs/components-tag--docs)
- [x] approved by designer @jannick-ux
- [x] approved by dev @MajaZarkova 

### Implementation details

- Use type `OnyxColor` for color property
- Use class `onyx-truncation-ellipsis` for the truncation

nerdrun:
Seems like not all the components have their usages in `alpha-test-app` folder.
Could you let me know the process for `example usage added to alpha-test-app` checkbox?

larsrickert:
> Seems like not all the components have their usages in `alpha-test-app` folder. Could you let me know the process for `example usage added to alpha-test-app` checkbox?

Usually all components should be used there. Which ones are missing? We will add them then :)

PR #962, opened 2024-04-19
docs: add tag example to alpha demo app

larsrickert:
Relates to #566

Add tag example to alpha demo app

### DIFF
--- a/apps/alpha-test-app/src/views/HomeView.vue
+++ b/apps/alpha-test-app/src/views/HomeView.vue
@@ -316,7 +316,7 @@ const singleSelectState = ref("Apple");
             </tbody>
           </OnyxTable>
 
-          <OnyxTag label="Example tag" :icon="emojiHappy2" />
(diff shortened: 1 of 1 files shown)


apps/alpha-test-app/src/views/HomeView.vue
```diff
@@ -314,6 +316,8 @@ const singleSelectState = ref("Apple");
             </tbody>
           </OnyxTable>
 
+          <OnyxTag label="Example tag" :icon="emojiHappy2" />
```
BoppLi, line 319: v-if missing

BoppLi (changes requested):

apps/alpha-test-app/src/views/HomeView.vue
larsrickert: fixed

BoppLi (approved):

Issue #841, opened 2024-04-09
Implement dot variation of OnyxBadge


BoppLi:
- #564

### Open Questions/To-dos

- [x] _Should Aria pattern be added and which one? -> No_

### Why?

_Info badges in UI design serve as valuable tools for conveying various types of information to users._

### Design

[Figma Link](https://www.figma.com/file/YfEUBOHk4J4nYrk04geswG/onyx---Design-System-Figma-Library?node-id=1009%3A321975&mode=dev)

### Acceptance criteria

- Supports mode:
  - [x] primary
  - [x] secondary
  - [x] danger
  - [x] warning
  - [x] success
  - [x] info
- Supports type:
  - [x] text
  - [x] number
  - [x] icon
- States:
  - [x] truncated
- [x] Add feature to Alpha-Test-App
- [x] Add badge to OnyxSelect and remove temporary styles

#### Out of scope

- dot

### Definition of Done

- [x] covered by component tests
- [x] screenshot tests are updated
- [x] follow-up tickets were created if necessary
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature
- [x] Set task to `In Approval`

**Review**

- [StoryBook Link](https://storybook.onyx.schwarz/?path=/docs/components-onyxbadge--docs)
- [OnyxBadge inside OnyxSelect](https://storybook.onyx.schwarz/?path=/story/components-onyxselect--filled-multi-select-preview)
- [ ] approved by designer @jannick-ux
- [x] approved by dev @BoppLi 

### Implementation details

- The badge should have the same design like in FIgma

#### Reference implementations

- [CoreUI](https://core-ui.schwarz/components/scu-badge)

PR #975, opened 2024-04-19
feat(OnyxBadge): add property dot

larsrickert:
Relates to #841

Add dot variant of the badge.

## Checklist


packages/sit-onyx/src/components/OnyxBadge/OnyxBadge.vue
```diff
@@ -24,16 +25,19 @@ defineSlots<{
       'onyx-truncation-ellipsis',
       'onyx-text',
       `onyx-badge--${props.variation}`,
+      props.dot ? 'onyx-badge--dot' : '',
```
MajaZarkova, line 28:
```suggestion
      { 'onyx-badge--dot': props.dot },
```

larsrickert: This does not work here, because we use an array of classes instead of an object :)
You can read the difference here between array and object binding: https://vuejs.org/guide/essentials/class-and-style#binding-html-classes

MajaZarkova (approved):

Issue #955, opened 2024-04-18
Documentation: Update component overview


larsrickert:
- remove components from sidebar
- add overview tiles with link to Storybook (we can reuse the component from the home page)

PR #980, opened 2024-04-19
docs: link components to Storybook

larsrickert:
Relates to #955

- remove duplicating the Storybook sidebar structure to VitePress and instead link to Storybook directly

## Checklist


apps/docs/src/.vitepress/components/ComponentRoadmap.vue
```diff
@@ -61,6 +61,10 @@ const shouldShowAllButton = computed(() => {
       regularly, we will keep it up to date with our progress.
     </p>
 
+    <p class="components__description">
+      <b>Tip:</b> You can click on a component to see its implementation and documentation.
```
BoppLi, line 65: should we specifiy that this only works on implemented/started components? or is this common sense

apps/docs/src/.vitepress/components/OnyxHomePage.vue
```diff
@@ -16,6 +16,8 @@ const kpiTimestamp = Intl.DateTimeFormat("en-US", {
   hour: "2-digit",
   minute: "2-digit",
 }).format(new Date(props.data.timestamp));
+
+const storybookHost = "https://storybook.onyx.schwarz" as const;
```
BoppLi, line 20: should we refer to the env variable for this?

BoppLi: ah, in the vitepress config you removed it intentionally. guess that's ok ^^

apps/docs/src/.vitepress/config.ts
```diff
@@ -1,12 +1,8 @@
...
 import { fileURLToPath } from "node:url";
-import { defineConfig, loadEnv, type DefaultTheme } from "vitepress";
+import { defineConfig } from "vitepress";
 import packageJson from "../../../../packages/sit-onyx/package.json";
-import { getStorybookHost } from "./env";
-import { getStorybookSidebarFolders } from "./utils";
 
-const env = loadEnv("", path.join(process.cwd(), "src"));
-const STORYBOOK_HOST = getStorybookHost(env);
+const storybookLink = "https://storybook.onyx.schwarz" as const;
```
BoppLi, line 5: if you want you could export that URL in a central file so if it changes we only have to adjust it once.

as soon as we got staging with the v1 release, this might be relevant?

BoppLi (approved):

apps/docs/src/.vitepress/config.ts
larsrickert: I guess we also would have to change it in all markdown files then :)
For now I wanted to keep it simple so we don't need to use env variables and vue bindings in markdown files

Issue #881, opened 2024-04-11
Implement OnyxUserMenu


MajaZarkova:
### Open Questions/To-dos

- [x] Do we create separate popUp component? -> Separate Component
- [x] Do we have avatar pictures ready in the project, maybe use the ones from CoreUI?

### Why?

This component is needed and will be used inside the navigationBar. This is just a support component.

### Design

[Figma Link (The User Menu section)](https://www.figma.com/file/YfEUBOHk4J4nYrk04geswG/onyx---Design-System-Figma-Library?node-id=1709%3A12334&mode=dev)

### Out of scope

- Appearance pop up
- Mobile behavior
- Avatar component
- The component supports two states:
  - [ ] Logged In
  - [ ] Logged out

### Acceptance criteria

- [x] Avatar slot for the avatar component or custom user image
- [x] Can display settings in the flyout
- [x] Displays the name of the User
- [x] Displays a description under the name
- [x] Be able to show text footer inside the flyout (ex. App version)

### Definition of Done

- [x] covered by component tests
- [x] screenshot tests are updated
- [x] follow-up tickets were created if necessary, #883 
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature
- [x] Set task to `In Approval`

**Approval**

- [Storybook](https://storybook.onyx.schwarz/?path=/docs/components-usermenu--docs)
- [x] approved by designer @jannick-ux
- [x] approved by dev @MajaZarkova 

#### Reference implementations

[CoreUI Reference](https://core-ui.schwarz/components/scu-top-bar-user)

#### Applicable ARIA Pattern

[ARIA Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/)

PR #1015, opened 2024-04-25
feat: add `OnyxUserMenu` component

larsrickert:
Relates to #881

Add basic `OnyxUserMenu` component

## Checklist


packages/sit-onyx/src/components/OnyxListboxOption/OnyxListboxOption.vue
```diff
@@ -39,8 +52,13 @@ defineSlots<{
   @include checkbox.variables();
 
   @include layers.component() {
+    --onyx-listbox-option-color: var(--onyx-color-text-icons-neutral-intense);
+    --onyx-listbox-option-color-active: var(--onyx-color-text-icons-primary-intense);
```
MajaZarkova, line 56: Should we rename this to --onyx-listbox-options-color-selected, so it's aligned with the background-selected?

MajaZarkova:
On the first click the listbox opens, but on second click on the button the listbox stays open. It closes only on outside click. Is this intentional?

MajaZarkova (changes requested):

MajaZarkova:
Also, should we close the listbox on option click as well?

packages/sit-onyx/src/components/OnyxListboxOption/OnyxListboxOption.vue
larsrickert: The color is only applied when the option is active, right? Not when its only selected?

larsrickert:
> On the first click the listbox opens, but on second click on the button the listbox stays open. It closes only on outside click. Is this intentional?

I checked with Jannick, the menu should only close when clicking an option.
Is it fine for you when we do this in a follow up PR?

packages/sit-onyx/src/components/OnyxListboxOption/OnyxListboxOption.vue
MajaZarkova: I think it's applied when it's selected and active

MajaZarkova:
> > On the first click the listbox opens, but on second click on the button the listbox stays open. It closes only on outside click. Is this intentional?
> 
> I checked with Jannick, the menu should only close when clicking an option. Is it fine for you when we do this in a follow up PR?

Sure, no problem

MajaZarkova (dismissed):

MajaZarkova (approved):

Issue #838, opened 2024-04-08
Implement table hover effects


JoCa96:
### Depends on
- #837

### Design

[Figma](https://www.figma.com/file/YfEUBOHk4J4nYrk04geswG/onyx---Design-System-Figma-Library?type=design&node-id=1979%3A34579&mode=dev)

### Acceptance criteria

- [x] Hovering a row will show the row hover effect
- [x] Hovering a column header should show a hover effect on the column
- [x] Cursor should stay the default

#### Out of scope

- cell hover effect

### Definition of Done

- [x] covered by component tests
- [x] screenshot tests are updated
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature
- [x] Set task to `In Approval`

**Approval**

- [Storybook link](https://storybook.onyx.schwarz/?path=/docs/components-table--docs)
- [x] approved by designer @jannick-ux
- [x] approved by dev @BoppLi 

### Implementation details
- Must be implemented only using CSS
- Example: https://css-tricks.com/simple-css-row-column-highlighting/
  - use `contain: paint;` instead of `overflow: hidden`

#### Reference implementations

- [WaWi-UI](https://dev.int.wawi.lidl/components/?path=/docs/others-wuxtable--docs)
- [Kendo UI](https://docs.telerik.com/kendo-ui/controls/grid/overview)
- [AG Grid](https://www.ag-grid.com/vue-data-grid/deep-dive/#create-a-basic-grid) 

#### Applicable ARIA Pattern

[Table](https://www.w3.org/WAI/ARIA/apg/patterns/table/)

PR #1018, opened 2024-04-25
feat(OnyxTable): add row and column hover styles

larsrickert:
Relates to #838

Add table row and column hover styles (plain CSS implementation).

## Checklist


packages/sit-onyx/src/components/OnyxTable/OnyxTable.vue
```diff
@@ -124,10 +124,12 @@ const { densityClass } = useDensity(props);
     font-family: var(--onyx-font-family);
     color: var(--onyx-color-text-icons-neutral-intense);
     text-align: left;
+    contain: paint;
```
BoppLi, line 127: will this make trouble in the future e.g. with a dropdown opening from within a cell?

packages/sit-onyx/src/components/OnyxTable/OnyxTable.ct.tsx
```diff
@@ -55,4 +55,22 @@ test.describe("Screenshot tests", () => {
       </OnyxTable>
     ),
   });
+
+  executeMatrixScreenshotTest({
+    name: "Table (hover styles)",
+    columns: ["default", "striped"],
+    rows: ["row-hover", "column-hover"],
+    // TODO: remove when contrast issues are fixed in https://github.com/SchwarzIT/onyx/issues/410
+    disabledAccessibilityRules: ["color-contrast"],
```
BoppLi, line 64: fyi in dark mode the hover color is nearly invisible. maybe point that out to Jannick before we go beta?

BoppLi (approved):

packages/sit-onyx/src/components/OnyxTable/OnyxTable.ct.tsx
BoppLi: would be fine to add this info e.g. as checkbox to #410 I think :)

packages/sit-onyx/src/components/OnyxTable/OnyxTable.vue
larsrickert: I think we can only prevent this if we know the exact table height so we can size the column hover effect without `contain: paint` or `overflow: hidden`...but this means we propably need JavaScript...

larsrickert: @BoppLi I am not sure if I should take care of this in this PR or if we check this later or we might even place the dropdown absolute idk

BoppLi: it's fine for me. let's try to remember it when it comes to it. 👍

Issue #1043, opened 2024-05-07
Fix css position on OnyxNavItem options


MajaZarkova:
Currently the options are always visible / taking up height which does not work as intended when used inside the nav bar because the whole nav bar adjusts its height then.

We probably should position it absolute to the parent

![Image](https://github.com/SchwarzIT/onyx/assets/60856270/89370068-9e9c-4b16-b18a-46504668bba2)

PR #1049, opened 2024-05-07
fix(OnyxNavItem): Fix listbox position

MajaZarkova:
Relates to #1043 

Set position: absolute to the listbox options of the nav item so that the options are not taking up height.

## Checklist


packages/sit-onyx/playwright/snapshots/components/OnyxNavItem/NavItem-with-nestedItems-chromium-linux.png
larsrickert, line 1: Tip: Since the listbox is positioned absolute it does not count into the component size so it is cut off here, you can do this in the Playwright `executeMatrixScreenshotTest` test to display the screenshot correctly:

```ts
   beforeScreenshot: async (component, page, column, row) => {
      if (row === "focus-visible") {
        await page.keyboard.press("Tab");

        // since the listbox is positioned absolute, we need to set the component size accordingly
        // so the screenshot contains the whole component
        await component.evaluate((element) => {
          element.style.height = `${element.scrollHeight}px`;
          element.style.width = `${element.scrollWidth}px`;
        });
      }
    },
```

larsrickert, line 1: Somehow the "Nested Item 1" is no longer active

larsrickert (changes requested):

larsrickert (approved):

Issue #867, opened 2024-04-10
Implement simple navigation bar


mj-hof:
### Open Questions/To-dos

- [x] Decide if the separator should be just a class or a separate component? Seperator for user menu on the right is a slot component
- [x] Do we include mobile behavior in this ticket? No, not part of that ticket

### Why?

This component is crucial for our library

### Design

[Figma Link](https://www.figma.com/file/YfEUBOHk4J4nYrk04geswG/onyx---Design-System-Figma-Library?node-id=1709%3A12388&mode=dev)

### Acceptance criteria

- [x] Appname area supports both icon and text (optional)
- [x] A slot for icon or custom svg image is available
- [x] Appname area is clickable
- [x] Show back button prop
- [x] It should follow the grid

#### Out of scope

- mobile/responsive behaviour, #1073

### Definition of Done

- [x] example usage added to `apps/alpha-test-app/src/views/HomeView.vue`
- [x] covered by component tests
- [x] screenshot tests are updated
- [x] follow-up tickets were created if necessary
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature
- [ ] Set task to `In Approval`

### Implementation Details

- [x] The app Name respects line breaks
- [x] We don't restrict max-width by default

**Approval**

- [Storybook](https://storybook.onyx.schwarz/?path=/docs/components-navbar--docs)
- [x] approved by designer @jannick-ux
- [x] approved by dev @MajaZarkova 

#### Reference implementations

[CoreUI Reference Link](https://core-ui.schwarz/components/scu-top-bar)

#### Applicable ARIA Pattern

[Disclosure navigation](https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/examples/disclosure-navigation/)

jannick-ux:
This is now with center alignment. Is left alignment available as well? With this component there is a high dependency on the overall page grid. Left aligned should be the fault here. 
Or will it be part of another ticket?

larsrickert:
> This is now with center alignment. Is left alignment available as well? With this component there is a high dependency on the overall page grid. Left aligned should be the fault here. Or will it be part of another ticket?

Currently the default is left aligned but it also adjust to the grid so if you set the grid to be max md and centered, the nav bar will also adjust accordingly

jannick-ux:
Ah okay, becaus with the examples I saw, there was a unused white gap on the left, that looked like it would exist because of center alignment

larsrickert:
[...]

PR #1070, opened 2024-05-13
feat: add `OnyxNavBar` component

larsrickert:
Relates to #867

- Add basic navigation bar component without mobile behaviour
- add navigation bar to alpha test app

Note on the "2xs" screenshots: For now we accept the "broken" visuals until we implement the mobile behaviour of the nav bar in #1073

## Checklist


packages/storybook-utils/src/theme.ts
```diff
@@ -90,50 +91,22 @@ const defineTheme = (colors: {
...
-  xl: {
-    name: "xl",
-    styles: {
-      width: "1920px",
-      height: "100%",
-    },
-  },
-} as const;
+  {} as Record<OnyxBreakpoint, StorybookBreakpoint>,
+);
```
JoCa96, line 101: I prefer to avoid `.reduce` where possible as it requires a higher cognitive overhead.
In this case we can also easily reuse the entries syntax.
```suggestion
export const ONYX_BREAKPOINTS = Object.fromEntries(
  Object.entries(RAW_ONYX_BREAKPOINTS).map(([breakpoint, width]) => [
    breakpoint,
    { name: breakpoint, styles: { width: `${width}px`, height: "100%" } },
  ]),
) as Record<OnyxBreakpoint, StorybookBreakpoint>;
```

packages/sit-onyx/src/components/OnyxSeparator/OnyxSeparator.vue
```diff
@@ -0,0 +1,18 @@
+<template>
+  <div class="onyx-separator"></div>
```
JoCa96, line 2:
```suggestion
  <!-- TODO: Add aria context: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/roles/separator_role -->
  <div class="onyx-separator"></div>
```

packages/sit-onyx/src/i18n/locales/de-DE.json
```diff
@@ -22,5 +22,9 @@
     "currentSelection": "{n} ausgewählt"
   },
   "empty": "Keine Daten verfügbar",
-  "navItemOptionsLabel": "Unterseiten von {label}"
+  "navItemOptionsLabel": "Unterseiten von {label}",
+  "navigation": {
+    "appLogo": "App logo",
+    "goBack": "Zurück"
```
JoCa96, line 28: `userMenuLabel` seems to be missing, can you please add it?

packages/sit-onyx/src/components/OnyxSeparator/OnyxSeparator.vue
JoCa96, line 1: This is intended for the use in the `NavBar`, we should follow the pattern of the other components and call it `OnyxNavSeparator`

packages/sit-onyx/src/components/OnyxSeparator/OnyxSeparator.vue
larsrickert: Nice I didn't know about that role :) I added it so we don't need the TODO

packages/sit-onyx/src/components/OnyxSeparator/OnyxSeparator.vue
larsrickert: I wasn't sure if the separator might also be used standalone or in other components (maybe even with a horizontal direction).

Do you think thats a valid use case? If not I will rename it to `OnyxNavSeparator` :)

packages/storybook-utils/src/theme.ts
larsrickert: To be honest I think your suggested syntax requires even more cognitive complexity^^
I'd say `.reduce()` is way more common and known to devs than `Object.fromEntries()`

larsrickert: And the `.reduce()` is more type safe since it ensures that the value of each breakpoints actually is of type `StorybookBreakpoint` while `as Record<OnyxBreakpoint, StorybookBreakpoint>;` just forces TypeScript to use this type for `ONYX_BREAKPOINTS `

packages/sit-onyx/src/styles/grid.scss
```diff
@@ -24,13 +24,16 @@ $gridVariants: (
     // often web projects override the default box-sizing, but it is important for the max-width and alignment to work correctly
     box-sizing: content-box;
     max-width: var(--onyx-grid-max-width, none);
+
+    margin-left: var(--onyx-grid-center-margin);
+    margin-right: var(--onyx-grid-center-margin);
```
JoCa96, line 29: Can be simplified by using `margin-inline`:
```suggestion
    margin-inline: var(--onyx-grid-margin-inline);
```

packages/sit-onyx/src/styles/breakpoints.scss
```diff
@@ -23,3 +23,12 @@ $breakpoints: (
     @content;
   }
 }
+
+/**
+ * Same as `breakpoint` mixin but using container queries instead of media queries.
+ */
+@mixin container($min-max, $name, $offset: 0) {
```
JoCa96, line 30: If we call this `container`, then the `breakpoint` mixin would make more sense to be called `media` or `screen`?

packages/sit-onyx/src/components/OnyxSeparator/OnyxSeparator.vue
JoCa96: IMHO it's preferable to focus on the use-cases we have defined and know about. So I would prefer to rename it :)

packages/sit-onyx/src/components/OnyxNavBar/types.ts
```diff
@@ -0,0 +1,8 @@
+import type { OnyxNavAppAreaProps } from "../OnyxNavAppArea/types";
+
+export type OnyxNavBarProps = OnyxNavAppAreaProps & {
+  /**
+   * Whether to show a back button.
+   */
+  showBackButton?: boolean;
```
JoCa96, line 7:
```suggestion
  withBackButton?: boolean;
```

packages/sit-onyx/src/styles/breakpoints.scss
larsrickert: I like it :) Also it reads very nice then:

```sass
breakpoints.screen(max, md)
breakpoints.container(max, md)
```

packages/sit-onyx/src/components/OnyxNavBar/OnyxNavBar.vue
```diff
@@ -0,0 +1,122 @@
...
+      gap: var(--onyx-spacing-4xs);
+      padding: 0;
+    }
+
+    &__context {
+      display: flex;
+      align-items: center;
+      gap: var(--onyx-spacing-4xs);
+      margin-left: auto;
+    }
```
JoCa96, line 119: Prefer flex properties over margins to place elements:
```suggestion
      flex: 1;
      justify-content: flex-end;
```

This will make it easier when we need to implement the `More` button.

packages/sit-onyx/src/components/OnyxNavBar/OnyxNavBar.vue
```diff
@@ -0,0 +1,122 @@
...
+      display: flex;
+      align-items: center;
+      gap: var(--onyx-spacing-md);
+      height: 100%;
+      padding: 0 var(--onyx-spacing-3xl);
+
+      // sync with grid
+      max-width: var(--onyx-grid-max-width);
+      margin-inline: var(--onyx-grid-margin-inline);
+
```
JoCa96, line 97:
```suggestion
      margin-inline: var(--onyx-grid-margin-inline);
```

packages/storybook-utils/src/theme.ts
JoCa96: I thought so, too. But now (https://betterprogramming.pub/think-again-before-you-use-array-reduce-28f785b5aea9) I pretty convinced that `fromEntries` is preferably syntax-wise. But you are right, that the type-support of `fromEntries` needs to be improved

Issue #706, opened 2024-03-20
Implement OnyxComboBox


JoCa96:
### Why?

The flyout is a fundamental element utilized across various components such as dropdowns, navigation bars, paginations, tables, etc. It provides the users with the ability to open a small modal window, facilitating single or multisselection based on the context in which it is employed.

Get in contact with @jannick-ux.

### Design

[Figma link](https://www.figma.com/file/YfEUBOHk4J4nYrk04geswG/onyx---Design-System-Figma-Library?type=design&node-id=1130-19227&mode=dev)

### Acceptance criteria

- search/filtering

#### Out of scope

- badge
- tag
- loading
- fit size to the parent
- sorting behavior for multiselect
- empty state

### Definition of Done

- [x] example usage added to alpha-test-app
- [x] covered by component tests
- [x] screenshot tests are updated
- [x] follow-up tickets were created if necessary
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature
- [x] Set task to `In Approval`

**Review**

- [Storybook link](https://storybook.onyx.schwarz/?path=/story/components-select--with-search)
- [x] approved by designer @jannick-ux
- [x] approved by dev @BoppLi 

### Implementation details

- During search "Check All" is hidden
- User can provided custom search function 
- Implement using headless composable, build up on listbox
- The `modelValue` prop and `update:modelValue` should be generically typed based on whether its multiselect or single select

- With a search field:
  - It's a [combobox](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/)
- Without a searchfield:
  - It's a [listbox](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/)

#### Applicable ARIA Pattern

[Listbox pattern](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/)

PR #1094, opened 2024-05-15
feat(OnyxListbox): Implement search functionality

JoCa96:
Relates to #706 

- Updated `OnyxListboxProps` to use discriminating unions instead of conditional generics for better type support
- Added `OnyxMiniSearch` component, which is used as part of the `OnyxListbox` to search the list
- Added `withSearch` property to `OnyxListbox` 
- renamed `ObjectToDottedStrings` to more appropriate `FlattenedKeysOf`

## Checklist


packages/sit-onyx/src/i18n/locales/en-US.json
```diff
@@ -21,7 +21,12 @@
     "selectAll": "Select all",
     "currentSelection": "{n} selected"
   },
-  "empty": "No data available",
+  "listbox": {
+    "empty": "No data available",
+    "noMatch": "No entry matches your search",
+    "searchInputLabel": "Filter the list entries",
```
larsrickert, line 27:
```suggestion
    "noMatch": "No items found matching your search",
    "searchInputLabel": "Filter the list items",
```

packages/sit-onyx/src/i18n/locales/de-DE.json
```diff
@@ -21,7 +21,12 @@
     "selectAll": "Alle auswählen",
     "currentSelection": "{n} ausgewählt"
   },
-  "empty": "Keine Daten verfügbar",
+  "listbox": {
+    "empty": "Keine Daten verfügbar",
+    "noMatch": "Kein Element passt zum Suchtext",
+    "searchInputLabel": "Durchsuche die Liste",
+    "clearSearch": "Lösche Suchtext"
```
larsrickert, line 28:
```suggestion
    "noMatch": "Kein Element passt zum Suchtext",
    "searchInputLabel": "Liste durchsuchen",
    "clearSearch": "Suchtext löschen"
```

packages/sit-onyx/src/i18n/index.ts
```diff
@@ -16,6 +16,8 @@ type GetTypeOfTranslations<T> = T extends object
 /** Available translations that are used by onyx components. */
 export type OnyxTranslations = GetTypeOfTranslations<typeof enUS>;
 
+export type OnyxTranslationKeys = FlattenedKeysOf<OnyxTranslations>;
```
larsrickert, line 19:
```suggestion
export type OnyxTranslationKey = FlattenedKeysOf<OnyxTranslations>;
```

Its a union type, no array so I'd say its a single key type

packages/sit-onyx/src/components/OnyxListbox/OnyxMiniSearch.vue
```diff
@@ -0,0 +1,101 @@
...
+</script>
+<template>
+  <div class="onyx-mini-search" v-bind="rootAttrs">
+    <input
+      v-model="value"
+      :aria-label="props.label"
+      v-bind="inputAttrs"
+      class="onyx-mini-search__input"
+      placeholder="Search"
+      type="text"
```
larsrickert, line 50:
```suggestion
      type="search"
```

Type search also already has a clear icon that we can maybe style more easily and we might get right of the rotoAttrs and inputAttrs above

packages/sit-onyx/src/components/OnyxListbox/OnyxMiniSearch.vue
```diff
@@ -0,0 +1,101 @@
...
+    />
+    <!-- We use `@mousedown.prevent` here to not lose the input focus when the button is clicked  -->
+    <button
```
larsrickert, line 53: button should be hidden when the search value is empty, right?

packages/sit-onyx/src/components/OnyxListbox/OnyxMiniSearch.vue
```diff
@@ -0,0 +1,101 @@
...
+import { computed, useAttrs, type HtmlHTMLAttributes } from "vue";
+import OnyxIcon from "../OnyxIcon/OnyxIcon.vue";
+import xSmall from "@sit-onyx/icons/x-small.svg?raw";
+import { injectI18n } from "../../i18n";
+
+defineOptions({ inheritAttrs: false });
+
+const attrs = useAttrs();
+
+const props = defineProps<{ modelValue: string; label: string }>();
```
larsrickert, line 11: Should we align with other components and create a `types.ts` file with the props?

packages/sit-onyx/src/components/OnyxListbox/OnyxMiniSearch.vue
larsrickert, line 1: Minisearch is a fulltext search library, should we maybe rename this component so something like `OnyxListboxSearch`?:
https://github.com/lucaong/minisearch

packages/sit-onyx/src/components/OnyxListbox/OnyxListbox.vue
```diff
@@ -130,7 +140,12 @@ const { vScrollEnd, isScrollEnd } = useScrollEnd({
   offset: computed(() => props.lazyLoading?.scrollOffset),
 });
 
-const isEmpty = computed(() => props.options.length === 0);
+const isEmptyMessage = computed<FlattenedKeysOf<OnyxTranslations> | undefined>(
+  () =>
+    (props.options.length === 0 && "listbox.empty") ||
+    (props.withSearch && props.searchTerm && props.options.length === 0 && "listbox.noMatch") ||
```
larsrickert, line 146: listbox.noMatch can never be reached because it checks for `props.options.length === 0` which is always returned earlier with "listbox.empty". Also imho I think the `&&` syntax is a bit hard to understand. What do you think about this?

```ts
const isEmptyMessage = computed(() => {
  if (props.options.length === 0) return t.value("listbox.empty");
  if (props.withSearch && props.searchTerm && props.options.length === 0) {
    return t.value("listbox.noMatch");
  }
  return undefined;
});
```

larsrickert: Also I'd ususally prefer to directly return the translated message instead of the key

packages/sit-onyx/playwright/snapshots/components/OnyxListbox/with-search-empty-chromium-linux.png
larsrickert, line 1: Would prefer to use our matrix screenshot util and also add test cases for the empty states (no options and no search matches)

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxListbox/OnyxMiniSearch.vue
JoCa96: The clear icon is User-Agent dependent and non-standard, see also: https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-search-cancel-button
Therefore I would like to stay with type `text`

packages/sit-onyx/src/components/OnyxListbox/OnyxMiniSearch.vue
JoCa96: The name is aligned with Figma, so I would prefer to keep the name.

packages/sit-onyx/src/components/OnyxListbox/OnyxMiniSearch.vue
JoCa96: sounds good

packages/sit-onyx/src/i18n/locales/en-US.json
JoCa96: updated

packages/sit-onyx/src/i18n/index.ts
JoCa96: sounds good

packages/sit-onyx/src/components/OnyxListbox/OnyxListbox.vue
JoCa96: fixed

packages/sit-onyx/src/components/OnyxListbox/OnyxMiniSearch.vue
larsrickert: Then I'd at least move it to a dedicated type and export it from the .vue file so it can be easily used :)

packages/sit-onyx/playwright/snapshots/components/OnyxListbox/with-search-empty-chromium-linux.png
JoCa96: done

packages/sit-onyx/src/components/OnyxListbox/OnyxMiniSearch.vue
JoCa96: done

packages/sit-onyx/playwright/snapshots/components/OnyxListbox/with-search-empty-chromium-linux.png
JoCa96: done

packages/sit-onyx/src/components/OnyxListbox/OnyxMiniSearch.vue
```diff
@@ -0,0 +1,109 @@
...
+    <input
+      v-model="value"
+      :aria-label="props.label"
+      v-bind="inputAttrs"
+      class="onyx-mini-search__input"
+      placeholder="Search"
+      type="text"
+    />
+    <!-- We use `@mousedown.prevent` here to not lose the input focus when the button is clicked  -->
+    <button
```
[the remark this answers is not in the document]
JoCa96: the width gets jumpy if we do that

packages/sit-onyx/src/components/OnyxListbox/OnyxMiniSearch.vue
```diff
@@ -0,0 +1,109 @@
...
+      "class" | "style"
+    >,
+);
+
+const inputAttrs = computed<Omit<HtmlHTMLAttributes, "class" | "style">>(() => {
+  const rest = { ...attrs };
+  delete rest.class;
+  delete rest.style;
+  return rest;
+});
```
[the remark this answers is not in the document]
JoCa96: Want to avoid a wrapper, so all attributes that effect style are applied to the root, everything else applies to the input.

larsrickert (approved):

PR #1098, opened 2024-05-15
feat: include color-scheme to include scrollbars in the dark theme

### LINKED ISSUE #618: Implement browser alignment to onyx dark/light mode
### Why?

Currently we are using a `.dark` CSS class for the dark mode which has 2 downsides:

1. It does not work out of the box / the project explicitly has to implement it
2. Even if the class is set, browsers default components like scrollbars etc. will still be in light

We should use media queries and the `color-schema: light` or `color-schema: dark` CSS property

### Acceptance criteria

- [ ] ~the light/dark CSS variables use a media query for the user preferred color schema instead of classes~
  - [x] the `color-schema` should also be set to dark or light based on that
- [ ] ~Our `storybook-utils` and `vitepress-theme` are checked and updated accordingly~
- [x] A "Theming" page in our docs is added that explains the usage of light/dark mode
  - [x] can reference [`useDark`](https://vueuse.org/core/useDark/#usedark) from `vueuse` for switching
- [ ] ~The [Playground](https://playground.onyx.schwarz) works with the new implementation~
- [ ] ~The `chartjs-plugin` package is updated accordingly~
- [ ] ~update figma variable import pipeline~

### Definition of Done

- [ ] ~covered by component tests~
- [ ] ~screenshot tests are updated~
- [ ] ~follow-up tickets were created if necessary~
- [x] Storybook can show the feature
- [x] Set task to `In Approval`

**Approval**

- ~Storybook~ -> only visible in alpha test app
- [x] approved by designer @jannick-ux
- [ ] approved by dev 

### Implementation details

- localStorage key should have a default value but can be changed

#### Reference implementations

- [`useDark`](https://vueuse.org/core/useDark/#usedark) from `vueuse`

BoppLi:
Closes #618 

- Include the color-scheme in the dark/light classes / theme setup
- Add a button to toggle dark/light in the alpha demo app
image
image

## Checklist


apps/docs/src/development/theming.md
```diff
@@ -0,0 +1,20 @@
+# Theming Setup
+
+Onyx supports a dark and a light theme.  
+To learn more about the theming concept of onyx, take a look at our [colors documentation](/basics/colors.html)
+
+In order to let the user choose the preferred theme, we recommend [useDark of vueuse](https://vueuse.org/core/useDark/#usedark).
```
larsrickert, line 6: I would also document how the apply the dark mode without vueuse so something like:

"To apply the dark mode, just set the `.dark` class on your application (e.g. on the `<html>` element. For allowing the user to switch between light/dark mode, we recommend to use the `useDark` composable from vueuse"

apps/docs/src/development/theming.md
```diff
@@ -0,0 +1,20 @@
+# Theming Setup
+
+Onyx supports a dark and a light theme.  
+To learn more about the theming concept of onyx, take a look at our [colors documentation](/basics/colors.html)
+
+In order to let the user choose the preferred theme, we recommend [useDark of vueuse](https://vueuse.org/core/useDark/#usedark).  
+Combine it as follows with an onyx button to let the users set the theme of their choice:
```
larsrickert, line 7: Jannicks intended way is to have an option in the OnyxUserMenu inside the nav bar so we should recommend to use a button here

apps/docs/src/development/theming.md
```diff
@@ -0,0 +1,20 @@
...
+
+```vue
+<script lang="ts" setup>
+import { useDark, useToggle } from "@vueuse/core";
+import { OnyxButton } from "sit-onyx";
+
+const toggleDark = useToggle(useDark());
```
larsrickert, line 14: Usually you also need a ref whether its currently dark or not because Jannick intended to show the current theme in the nav bar user menu

apps/docs/src/development/theming.md
larsrickert: https://www.figma.com/design/YfEUBOHk4J4nYrk04geswG/onyx---Design-System-Figma-Library?node-id=1709-12347&m=dev

packages/sit-onyx/src/styles/root.scss
```diff
@@ -18,6 +18,15 @@
     --onyx-shadow-soft-right: 0.5rem 0 1.5rem 0 rgba(0, 0, 0, 0.1);
   }
 
+  :root,
+  .onyx-light {
```
larsrickert, line 22: Imho we can drop support for the `.onyx-light` and `.onyx-dark` class since I don't see the benefit of them

larsrickert (changes requested):

apps/alpha-test-app/src/views/HomeView.vue
```diff
@@ -118,12 +119,16 @@ const groupedListboxOptions: ListboxOption[] = [
...
 const singleSelectState = ref(selectOptions[0]);
+
+const toggleDark = useToggle(useDark());
 </script>
 
 <template>
   <OnyxPageLayout>
     <template #sidebar>
       <div class="sidebar">
+        <OnyxButton label="Toggle Dark Mode" @click="toggleDark" />
```
larsrickert, line 130: I'd prefer to put a switch in the global nav bar inside App.vue so the theme can be toggled on every page

apps/docs/src/development/theming.md
BoppLi: we don't support that feature yet, do we?

packages/sit-onyx/src/styles/root.scss
BoppLi: there is no use case where inside a dark page, a light section is needed, right?

apps/alpha-test-app/src/views/HomeView.vue
BoppLi: do we have the user menu component yet?

apps/docs/src/development/theming.md
larsrickert: No not yet :/ Should we check with Jannick how the handle dark mode toggle until we have this? Maybe as a switch/icon button in the context area of the nav bar

apps/alpha-test-app/src/views/HomeView.vue
larsrickert: Yes: https://storybook.onyx.schwarz/?path=/story/components-navbar--with-context-area

apps/docs/src/development/theming.md
BoppLi: ok did it like that for now ^^

larsrickert (approved):

Issue #223, opened 2024-01-30
Implement OnyxTextarea


larsrickert:
### Why?

A textarea is an input component for entering multiline text. Note that the spelling is textarea, not text area or textArea.

### Design

[Figma link](https://www.figma.com/file/YfEUBOHk4J4nYrk04geswG/onyx---Design-System-Figma-Library?type=design&node-id=1255-61268&mode=dev)

### Acceptance criteria

- the following states are implemented:
  - [x] default
  - [x] hover
  - [x] focus
  - [x] disabled
  - [x] read-only
  - [x] required
  - [x] optional
  - [x] placeholder-shown
- the textarea supports auto-sizing
  - [x] the height increases/decreases to the value
  - [x] a max/min height can be set
- [x] optionally the textarea can be resized by dragging the bottom right corner
- [x] an optional label can be shown
- [x] an optional placeholder can be shown
- [x] optionally minlength can be set to define the min value length
- [x] optionally maxlength can be set to define the max value length
- [x] multiline text is supported
- [x] an optional message can be displayed
- [x] an optional counter can be displayed if maxlength if set
- the following standard HTML attributes are supported
  - [x] [autocapitalize](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea#autocapitalize)
  - [x] [autofocus](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea#autofocus)

#### Out of scope

- [x] validation/invalid is supported for required, minlength and maxlength as well as a custom error message

### Definition of Done

- The following component requirements are implemented:
  - [x] `skeleton`
  - [x] `density`
  - [x] example usage added to `apps/alpha-test-app/src/views/HomeView.vue`
- [x] covered by component tests
- [x] screenshot tests are updated
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature
- [x] Set task to `In Approval`

**Approval**

- [Storybook link](https://storybook.onyx.schwarz/?path=/docs/components-textarea--docs)
- [x] approved by designer @jannick-ux
- [ ] approved by dev 

### Implementation details

- [MDN link](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea)
- an event is emitted whenever the validity changes
- POC for resize functionality:
  - https://github.com/SchwarzIT/onyx/tree/poc-textarea
  - also consider: https://css-tricks.com/the-cleanest-trick-for-autogrowing-textareas/
[...]

PR #1125, opened 2024-05-17
feat(OnyxTextarea): support autosize

larsrickert:
Relates to #223

- support autosize for the textarea

## Checklist

### DIFF
--- a/.changeset/pre.json
+++ b/.changeset/pre.json
@@ -148,6 +148,7 @@
     "three-icons-leave",
     "tiny-bananas-design",
     "tiny-fireants-wonder",
+    "tough-vans-punch",
     "tricky-kids-bake",
     "twenty-books-shop",
     "twenty-colts-film",
--- a/apps/alpha-test-app/CHANGELOG.md
+++ b/apps/alpha-test-app/CHANGELOG.md
@@ -1,5 +1,12 @@
 # alpha-test-app
 
+## 0.1.0-alpha.103
+
+### Patch Changes
+
+- Updated dependencies [34547be]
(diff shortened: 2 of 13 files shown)


packages/sit-onyx/src/components/OnyxTextarea/types.ts
```diff
@@ -22,4 +22,26 @@ export type OnyxTextareaProps = DensityProp &
...
     | "skeleton"
-  >;
+  > & {
+    /**
+     * Override the default autosize behavior (height adjusts based on the current value).
+     * By default, the textarea will autosize while maintaining at least 3 and at most 6 rows.
+     * If the user resizes the textarea manually, the autosize will no longer work and the height
+     * set by the user is used.
+     */
+    autosize?: TextareaAutosize;
```
JoCa96, line 32: IMHO resizing and autosizing should be opt-in and be disabled by default.
- resizing: Currently there is no option for the Dev to disable resizing without overwriting our internal CSS
- autosizing: API wise it seems unintuitive to have to set `min` and `max` to the same number to disable it. Additionally it would probably nice to have a prop for setting a static height based on lines. E.g.:


```suggestion
    /**
     * Enable vertical resizing of the textarea.
     */
    resizable?: boolean;
    /**
     * Set the textarea height based on linecount or enable autosizing by setting `min` or `max`.
     * By default, the textarea will autosize while maintaining at least 3 and at most 6 rows.
     * If the user resizes the textarea manually, the autosize will no longer work and the height
     * set by the user is used.
     */
    height?: number | TextareaAutosize;
```

JoCa96 (approved):

Issue #1032, opened 2024-04-26
Implement external link icon for nested NavItem


MajaZarkova:
### Why?

We should separate external links visually.

### Design

- [Figma](https://www.figma.com/file/YfEUBOHk4J4nYrk04geswG/onyx---Design-System-Figma-Library?node-id=1625%3A41198&mode=dev)
- Looks the same as in the other places

### Acceptance criteria

- [x] Arrow icon should be visible next to the label

### Definition of Done

- [x] covered by component tests
- [x] screenshot tests are updated
- [x] follow-up tickets were created if necessary
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature
- [x] Set task to `In Approval`

**Approval**

- [Storybook](https://storybook.onyx.schwarz/?path=/story/support-navitem--with-options)
- [x] approved by designer @jannick-ux
- [x] approved by dev @larsrickert 

#### Reference implementations

OnyxNavItem, OnyxLink

jannick-ux:
Please set the padding between the text and the arrow-icon to 0. We have it the same way in the link-component.

MajaZarkova:
There is a bug ticket for that -> #1148

PR #1131, opened 2024-05-20
feat(OnyxNavItem): Implement external link icon for nested nav items

MajaZarkova:
Relates to #1032 

Implement external link icon for nested nav items

## Checklist


packages/sit-onyx/src/components/OnyxNavItem/OnyxNavItem.vue
```diff
@@ -29,10 +28,10 @@ defineSlots<{
...
 const { t } = injectI18n();
 
-const shouldShowExternalIcon = computed(() => {
-  if (props.withExternalIcon !== "auto") return props.withExternalIcon;
-  return isExternalLink(props.href ?? "");
-});
+const shouldShowExternalIcon = (args: OnyxNavItemProps) => {
+  if (args.withExternalIcon !== "auto") return args.withExternalIcon ?? false;
+  return isExternalLink(args.href ?? "");
+};
```
larsrickert, line 34:
```suggestion
const shouldShowExternalIcon = (args: OnyxNavItemProps) => {
  const withExternalIcon = args.withExternalIcon ?? "auto";
  if (withExternalIcon !== "auto") return args.withExternalIcon;
  return isExternalLink(args.href ?? "");
};
```

The default value defined by `withDefaults(defineProps<>())` does not work now for the children so we need to explicitly set it here

packages/sit-onyx/src/components/OnyxNavItem/OnyxNavItem.stories.ts
```diff
@@ -51,7 +51,7 @@ export const WithOptions = {
     ...Default.args,
     options: [
       { label: "Nested Item 1", href: "#", active: true },
-      { label: "Nested Item 2", href: "#2" },
+      { label: "Nested Item 2", href: "#2", withExternalIcon: true },
```
larsrickert, line 54:
```suggestion
      { label: "Nested Item 2", href: "https://onyx.schwarz" },
```

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxNavItem/OnyxNavItem.stories.ts
larsrickert: Can we also add an example for this inside the screenshot tests?

larsrickert (approved):

PR #1147, opened 2024-05-22
feat: implement `OnyxVisuallyHidden` component

JoCa96:
Relates to #1083 

- Implemented `OnyxVisuallyHidden` component to be able to add assistive-technology-only text
- Used new component to add `opens in new tab` information to `OnyxLink` component
- Added label to `OnyxNavAppArea` button for improved accessibility


packages/sit-onyx/src/components/OnyxVisuallyHidden/OnyxVisuallyHidden.vue
larsrickert, line 1: Would prefer to implement it as class instead of component so its easier to apply it with conditions like:

`<span :class="{'onyx-visually-hidden': props.test === true}">Test</span>`

packages/sit-onyx/src/i18n/locales/de-DE.json
```diff
@@ -27,9 +27,13 @@
...
     "clearSearch": "Suchtext löschen"
   },
+  "link": {
+    "opensExternally": "Wird in einem neuen Tab geöffnet."
+  },
   "navItemOptionsLabel": "Unterseiten von {label}",
   "navigation": {
-    "appLogo": "App logo",
+    "appLogo": "App logo von {appName}",
+    "goToHome": "Gehe zur Startseite",
```
larsrickert, line 36: Technically the app area click is up to the project so it does not have to go to the home page

packages/sit-onyx/src/components/OnyxLink/OnyxLink.vue
```diff
@@ -24,21 +26,20 @@ defineSlots<{
...
 
 <template>
-  <a
-    class="onyx-link"
-    :href="props.href"
-    :target="props.target"
-    :rel="props.target === '_blank' ? 'noreferrer' : undefined"
-    @click="emit('click')"
-  >
+  <a class="onyx-link" :href="props.href" :target="props.target" @click="emit('click')">
```
larsrickert, line 38: Why did you remove the `rel`? The browser only infers `noopener`but not `noreferrer`

packages/sit-onyx/src/components/OnyxNavAppArea/types.ts
```diff
@@ -7,4 +7,8 @@ export type OnyxNavAppAreaProps = {
    * Application name to show on the left.
    */
   appName?: string;
+  /**
+   * Describes the action that is taken on click.
```
larsrickert, line 11:
```suggestion
   * Describes the action that is taken on click.
   * By default, a translated "Go to home" will be used depending on the current locale.
```

packages/sit-onyx/src/components/OnyxNavBar/types.ts
```diff
@@ -5,4 +5,9 @@ export type OnyxNavBarProps = OnyxNavAppAreaProps & {
    * Whether to show a back button.
    */
   withBackButton?: boolean;
+  /**
+   * Custom label for the app click area.
+   * @default "Go to Home"
```
larsrickert, line 10:
```suggestion
   * @default Translated "Go to home" depending on the current locale.
```

packages/sit-onyx/src/components/OnyxVisuallyHidden/OnyxVisuallyHidden.ct.tsx
```diff
@@ -0,0 +1,20 @@
+import { expect, test } from "../../playwright/a11y";
+import OnyxVisuallyHidden from "./OnyxVisuallyHidden.vue";
+
+test("should not be visible", async ({ mount }) => {
+  const component = await mount(
+    <div>
```
larsrickert, line 6:
```suggestion
    <div style={{ width: "max-content" }}>
```

So the screenshots are smaller and easier to read

packages/sit-onyx/src/components/OnyxVisuallyHidden/types.ts
```diff
@@ -0,0 +1,3 @@
+export type OnyxVisuallyHiddenProps = {
+  is?: string;
```
larsrickert, line 2: Would prefer to make it required so its more explicit

packages/sit-onyx/src/i18n/locales/en-US.json
```diff
@@ -27,9 +27,13 @@
...
     "clearSearch": "Clear search filter"
   },
+  "link": {
+    "opensExternally": "Opens in a new tab."
+  },
   "navItemOptionsLabel": "Subpages of {label}",
   "navigation": {
-    "appLogo": "App logo",
+    "appLogo": "App logo of {appName}",
+    "goToHome": "Go to Home",
```
larsrickert, line 36:
```suggestion
    "goToHome": "Go to home page",
```

larsrickert (changes requested):

packages/sit-onyx/src/i18n/locales/de-DE.json
JoCa96: Yes, that's why the label is configurable. But we could also make the label required, if you prefer

packages/sit-onyx/src/components/OnyxVisuallyHidden/types.ts
JoCa96: Usually there should not be a need to change the element.

packages/sit-onyx/src/components/OnyxVisuallyHidden/OnyxVisuallyHidden.vue
JoCa96: This component is not something that should be toggled.

packages/sit-onyx/src/components/OnyxNavBar/types.ts
```diff
@@ -5,4 +5,9 @@ export type OnyxNavBarProps = OnyxNavAppAreaProps & {
    * Whether to show a back button.
    */
   withBackButton?: boolean;
+  /**
+   * Custom label for the app click area.
```
larsrickert, line 9:
```suggestion
   * Custom (aria) label for the app click area.
```

packages/sit-onyx/src/components/OnyxVisuallyHidden/OnyxVisuallyHidden.vue
```diff
@@ -0,0 +1,27 @@
...
+.onyx-visually-hidden {
+  @include layers.component() {
+    border: 0;
+    clip: rect(0, 0, 0, 0);
+    height: 0;
+    margin: 0;
+    overflow: hidden;
+    padding: 0;
+    position: absolute;
+    width: 1px;
```
larsrickert, line 23:
```suggestion
    width: var(--onyx-1px-in-rem);
```

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxVisuallyHidden/OnyxVisuallyHidden.vue
JoCa96: The component should have the smallest possible width, so that it is still recognized as a "visible" element, but it shouldn't scale with the root element font-size.

larsrickert (approved):

Issue #877, opened 2024-04-11
Update implementation for OnyxGrid


mj-hof:
### Open Questions/To-dos

- [x] Double check if and how we need to implement "sub grids" and how they should work @jannick-ux. Reply from @jannick-ux: sub grid column count should depend on the width of the component aligned with the breakpoints. So using container queries is perfectly fine

### Why?

We want to update the grid implementation because:

- separate actual grid from outer page spacing, max-width and center styles to allow full-width page layouts (e.g. hero image, striped section backgrounds etc.)
- align with how the grid works in Figma
- support nested / subgrids (e.g. place a tile in the grid but also use a new grid inside the tile itself)

Current grid idea: Place grid at the very top of the application, everything below must define grid span on every single element
New idea: Only apply grid to certain child components where its actually used so by default no grid is used so all components have the default behaviour (full width)

### Acceptance criteria

- [x] the `onyx-grid` class only applies the grid + gutters
- [x] a new `onyx-page__content` class is added that applies the outer page spacing + max-width + centered styles of the previous `onyx-grid`
- [x] [Documentation](https://onyx.schwarz/development/grid.html) is updated and a code example is added on how to use the grid
- [x] [Grid playground](https://storybook.onyx.schwarz/?path=/docs/utilities-gridplayground--docs) is updated
- [ ] ~~container queries are used for the breakpoints instead of media queries~~

### Definition of Done

- The following component requirements are implemented:
  - [x] example usage added to `apps/alpha-test-app/src/views/HomeView.vue`
- [x] covered by component tests
- [x] screenshot tests are updated
- [x] follow-up tickets were created if necessary
- [x] updated version + documentation is deployed
- [x] Set task to `In Approval`

**Approval**

- [Grid docs](https://onyx.schwarz/development/grid.html)
- [x] approved by dev @MajaZarkova 

### Implementation details

Example usage of the grid:

```html
<!-- this is e.g. the very top app root element -->
<div class="onyx-grid-max-md onyx-grid-center">
  <OnyxNavBar></OnyxNavBar>

  <div class="onyx-page__content">
    <OnyxHeadline>I am outside the grid</OnyxHeadline>

    <div class="onyx-grid">
      I am inside the grid, e.g. a tile layout

      <div class="onyx-grid onyx-grid-span-6">
        I am a nested grid with my own grid that spans 6 columns of the parent
        grid
      </div>
    </div>
  </div>
</div>
```

PR #1159, opened 2024-05-23
refactor: update grid implementation

larsrickert:
Relates to #877

- remove padding, max-width and center from "onyx-grid" class and move it to a new "onyx-page__content"
- update grid docs / examples
- update grid demo in alpha test app
- fix default button width to max-content so it does not stretch e.g. when applying `onyx-grid-span-16` to it

## Checklist


JoCa96 (changes requested):

packages/sit-onyx/src/styles/grid.scss
```diff
@@ -2,100 +2,108 @@
...
     grid-template-columns: repeat(var(--onyx-grid-columns), minmax(0, 1fr));
     gap: var(--onyx-grid-gutter);
-    // Its called margin in the UX concept. So we keep that name for consistency, even when it is implemented through a padding.
+  }
+
+  /**
+  * Page class to apply padding, max-width and optional centered.
+  * Is usually placed at the root element of every page.
+  */
+  .onyx-page__content {
```
JoCa96, line 32: As this BEM class is defined outside of the context of the `AppLayout` component, we should consider to rename to fit the grid context, e.g.:
- `onyx-grid-container`
- `onyx-grid-page-content`

larsrickert: done

JoCa96 (approved):

Issue #1247, opened 2024-06-05
Implement Nuxt module


larsrickert:
### Why?

We want to offer a [Nuxt module](https://nuxt.com/docs/guide/concepts/modules) that simplifies how onyx can be used/installed in Nuxt projects.

### Acceptance criteria

- [x] the Nuxt module is implemented and published as `@sit-onyx/nuxt` npm package
- [x] the [Getting Started docs](https://onyx.schwarz/development/) are updated with instructions for Nuxt users
- [x] the onyx Nuxt module is showcased in the official [Nuxt module list](https://nuxt.com/modules)
- [x] auto imports are supported for onyx components

#### Out of scope

- #1345

### Definition of Done

- [x] covered by component tests
- [x] follow-up tickets were created if necessary
- [x] updated version + documentation is deployed
- [x] Set task to `In Approval`

**Approval**

- [Deployed docs](https://onyx.schwarz/development/packages/nuxt.html)
- [Listed Nuxt module](https://nuxt.com/modules/onyx)
- [x] approved by dev @JoCa96 

### Implementation details

- [Nuxt module dev guide](https://nuxt.com/docs/guide/going-further/modules)

larsrickert:
PR to add the onyx module to the Nuxt module list: https://github.com/nuxt/modules/pull/983

PR #1249, opened 2024-06-05
feat: implement Nuxt module

markbrockhoff:
Relates to #1247

This PR adds a basic nuxt module to integrate onyx into nuxt projects. It just adds the global styles and auto imports all onyx components, more features may be added later.

## Checklist

### DIFF
--- a/apps/docs/src/development/index.md
+++ b/apps/docs/src/development/index.md
@@ -53,19 +53,3 @@ import { OnyxButton } from "sit-onyx";
   <OnyxButton label="Click me" />
 </template>
 ```
-
-## Nuxt
-
-To setup Onyx in a Nuxt Js project you can simply install the pre-configured Nuxt module for it:
-
-```sh [npm]
-npx nuxi add @sit-onyx/nuxt
-```
-
-Afterwards you're able to just use all onyx components inside your app and the global styles will automatically be set up for you. E.g.:
-
-```vue
-<template>
-  <OnyxButton label="Click me" />
-</template>
-```
--- a/apps/docs/src/development/index.md
+++ b/apps/docs/src/development/index.md
@@ -53,3 +53,19 @@ import { OnyxButton } from "sit-onyx";
   <OnyxButton label="Click me" />
 </template>
 ```
+
+## Nuxt
+
+To setup Onyx in a Nuxt Js project you can simply install the pre-configured Nuxt module for it:
+
+```sh [npm]
+npx nuxi add @sit-onyx/nuxt
+```
+
+Afterwards you're able to just use all onyx components inside your app and the global styles will automatically be set up for you. E.g.:
+
+```vue
+<template>
+  <OnyxButton label="Click me" />
+</template>
+```
--- a/packages/nuxt/README.md
+++ b/packages/nuxt/README.md
@@ -1,20 +1,26 @@
-# Onyx Nuxt
+<div align="center">
+  <picture>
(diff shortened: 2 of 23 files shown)


markbrockhoff:
I hope this works with your current release flow, I tried to adapt to the other packages. But please double check that e.g. all required scripts are in place and so on :)

apps/docs/src/development/index.md
```diff
@@ -53,3 +53,19 @@ import { OnyxButton } from "sit-onyx";
   <OnyxButton label="Click me" />
 </template>
 ```
+
+## Nuxt
+
+To setup Onyx in a Nuxt Js project you can simply install the pre-configured Nuxt module for it:
```
BoppLi, line 59:
```suggestion
To setup Onyx in a [Nuxt.js](https://nuxt.com/) project you can simply install the pre-configured Nuxt module for it:
```
or
```suggestion
To setup Onyx in a [Nuxt JS](https://nuxt.com/) project you can simply install the pre-configured Nuxt module for it:
```
? :)

packages/nuxt/README.md
```diff
@@ -0,0 +1,20 @@
+# Onyx Nuxt
```
BoppLi, line 1: please place the description in the docs app, so it appears here in the docs
image
like http://localhost:5173/development/packages/figma-utils.html

copy the content of this file into 
image

the readme of the actual packages are pointing to the docs with a logo on top, compare to
 `packages/figma-utils/README.md`

packages/nuxt/src/module.ts
```diff
@@ -0,0 +1,25 @@
+import { addComponent, defineNuxtModule } from "@nuxt/kit";
+import * as onyx from "sit-onyx";
```
larsrickert, line 2: Do you know if this is performant and stree-shakeable for bundle size? Or would we need a dynamic import for this?

packages/nuxt/src/module.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export interface ModuleOptions {}
```
larsrickert, line 4: There are `OnyxPluginOptions` that we could use here and also call the `createOnyx()` plugin.

But we can also do this in a follow up PR

packages/nuxt/test/fixtures/basic/app.vue
```diff
@@ -0,0 +1,3 @@
+<template>
+  <OnyxHeadline is="h1">Hello Onyx</OnyxHeadline>
```
larsrickert, line 2:
```suggestion
  <OnyxHeadline is="h1">Hello onyx</OnyxHeadline>
```

onyx is written in lowercase :D

larsrickert (changes requested):
Thanks a lot for your contribution!

packages/nuxt/src/module.ts
markbrockhoff: Everything you see in this file will only run at build time, that's how nuxt modules work. It only registers the modules so they can be auto imported by nuxt. The import here is only used to get the names of the components for that. If one of the components is used in e.g. a page Nuxt will add `import { OnyxComponentName } from "sit-onyx"` to the page at build time. So all of this has nothing to do with the final bundle or the runtime.

packages/nuxt/src/module.ts
markbrockhoff: That's something to do once i18n and so on is integrated into this module. Like written in the description, this PR is just to bring the very basics of onyx into nuxt projects.

larsrickert (approved):
Great work!

Issue #570, opened 2024-03-11
Implement input/textarea invalid state


larsrickert:
### Depends on

- #189
- #567
- #565 
- Blocked by #574  

### Why?

The input, textarea ~and select~ should be able to show an invalid state to show the user that the entered value is invalid and he has to perform further actions.

### Design

- [Figma - input](https://www.figma.com/file/YfEUBOHk4J4nYrk04geswG/onyx---Design-System-Figma-Library?type=design&node-id=1475-259920&mode=dev)
- [Figma - textarea](https://www.figma.com/design/YfEUBOHk4J4nYrk04geswG/onyx---Design-System-Figma-Library?node-id=1259-76009&m=dev)
- ~[Figma - select](https://www.figma.com/design/YfEUBOHk4J4nYrk04geswG/onyx---Design-System-Figma-Library?node-id=1504-31388&m=dev)~

### Acceptance criteria

- the input is shown invalid with a default translated error message if the validation for one of the following properties fails:
  - [x] required
  - [x] minlength
  - [x] maxlength
  - [x] pattern
- For each error type (min/max length, pattern missmatch..., 
  - [x] a short text summary is displayed below the input 
  - [x] with a more detailed text in the tooltip).
- [x] a custom error message can be passed
- [x] the `TestInput` component is replaced with the `OnyxInput` inside the alpha demo app.
  - [ ] ~The TestInput is moved to the alpha test app for number inputs.~
- Implement translated error message like in OnyxInput for all other existing form components that we have.
  - [x] OnyxTextarea
  - [ ] ~OnyxSelect~ -> #1302
- [x] inside the alpha-test-app, inside the FormDemoView, include all modified form components from this ticket as examples
- [x] verify: the translation for the preview texts don't need data object? (e.g. value, n, max, ..) -> then shorten the useCustomValidity errorMessages.
- [x] for customError: support `{longError string, shortError: string}` aswell as the current `string`

### Definition of Done

- [x] example usage added to alpha-test-app
- [x] covered by component tests
- [x] screenshot tests are updated
- [x] follow-up tickets were created if necessary
  - #1302 
  - #1312
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature
- [x] Set task to `In Approval`

**Review**

- [Storybook - Input required example](https://storybook.onyx.schwarz/?path=/story/components-input--required)
- [Storybook - Input minlength example](https://storybook.onyx.schwarz/?path=/story/components-input--maxlength)
[...]

PR #1269, opened 2024-06-10
feat(OnyxInput,OnyxTextarea): show invalid messages

BoppLi:
Relates to #570 

Show translated error messages for OnyxInput and OnyxTextarea
Info: OnyxSelect will follow in a separate PR

## Checklist


packages/sit-onyx/src/components/OnyxInfoTooltip/types.ts
```diff
@@ -1,4 +1,4 @@
 import type { OnyxTooltipProps } from "../OnyxTooltip/types";
 
 export type OnyxInfoTooltipProps = Required<Pick<OnyxTooltipProps, "text">> &
-  Pick<OnyxTooltipProps, "position">;
+  Pick<OnyxTooltipProps, "position"> & { label?: string };
```
larsrickert, line 4: Can we add JSdocs here? :)

packages/sit-onyx/src/components/OnyxInput/OnyxInput.vue
```diff
@@ -129,8 +129,24 @@ const { t } = injectI18n();
...
+      class="onyx-input__footer onyx-text--small"
+    >
+      <span v-if="errorMessages.shortMessage" class="onyx-input__error-message">
+        <span class="onyx-truncation-ellipsis">{{ errorMessages.shortMessage }}</span>
+
+        <OnyxInfoTooltip
+          v-if="errorMessages.longMessage"
+          class="onyx-input__message-tooltip"
+          color="danger"
+          style="color: red"
```
larsrickert, line 143: Is this a leftover?

packages/sit-onyx/src/components/OnyxInput/OnyxInput.vue
```diff
@@ -129,8 +129,24 @@ const { t } = injectI18n();
...
+          position="bottom"
+          label="Show error tooltip"
```
larsrickert, line 145: Would prefer to translate this

packages/sit-onyx/src/components/OnyxInput/types.ts
```diff
@@ -75,10 +75,16 @@ export type OnyxInputProps = DensityProp &
     loading?: boolean;
     /**
      * Maximum number of characters that are allowed to be entered.
+     * Warning: when the value is (pre)set by code,
```
larsrickert, line 78: This is true for every validation (also required, custom error etc.) so I don't think we need to document this for every property :)

packages/sit-onyx/src/components/OnyxSelectInput/OnyxSelectInput.vue
```diff
@@ -142,6 +142,7 @@ defineExpose({ focus: () => input.value?.focus() });
       </div>
     </label>
 
+    <!-- TODO: #570 error message inclusion -->
```
larsrickert, line 145: If we have an issue for this, we don't need a TODO here

packages/sit-onyx/src/composables/useCustomValidity.spec.ts
```diff
@@ -6,8 +6,16 @@ import {
...
 
+vi.mock("../i18n", () => ({
+  injectI18n: () => ({
+    t: () => "Test",
+  }),
+}));
+
 describe("useCustomValidity", () => {
   test("should set custom error", async () => {
+    // provideI18n(app, { locale: "test" });
```
larsrickert, line 17: Can this be removed?

packages/sit-onyx/src/composables/useCustomValidity.spec.ts
```diff
@@ -6,8 +6,16 @@ import {
   type UseCustomValidityOptions,
 } from "./useCustomValidity";
 
+vi.mock("../i18n", () => ({
```
larsrickert, line 9: Do you also want to add a test for the error messages?

packages/sit-onyx/src/composables/useCustomValidity.ts
```diff
@@ -13,7 +15,16 @@ export type UseCustomValidityOptions = {
...
-  props: CustomValidityProp & { modelValue?: unknown };
+  props: CustomValidityProp & {
+    modelValue?: unknown;
+    type?: (typeof TRANSLATED_INPUT_TYPES)[number] | string;
+    pattern?: string | RegExp;
+    maxlength?: number;
+    minlength?: number;
+    min?: number;
+    max?: number;
+    step?: number;
```
larsrickert, line 26: We don't have this currently, right because thats only for number inputs?

packages/sit-onyx/src/composables/useCustomValidity.ts
```diff
@@ -42,15 +61,18 @@ export type InputValidationElement = Pick<HTMLInputElement, "validity" | "setCus
  * ```
  */
 export const useCustomValidity = (options: UseCustomValidityOptions) => {
+  // TODO: #570 usage of "t" breaks the vitest test!
```
larsrickert, line 64: This is fixed right?

packages/sit-onyx/src/composables/useCustomValidity.ts
```diff
@@ -61,34 +83,88 @@ export const useCustomValidity = (options: UseCustomValidityOptions) => {
...
+          if (!isDirty.value || !validityState.value) return;
+
           options.emit("validityChange", validityState.value);
         },
         { immediate: true },
       );
     },
   } satisfies Directive<InputValidationElement, undefined>;
 
+  const errorMessages = computed<{ longMessage?: string; shortMessage: string }>(() => {
```
larsrickert, line 120:
```suggestion
  const errorMessages = computed<{ longMessage?: string; shortMessage?: string }>(() => {
```

Would prefer to not use an empty message if there is no error

packages/sit-onyx/src/composables/useCustomValidity.ts
```diff
@@ -61,34 +83,88 @@ export const useCustomValidity = (options: UseCustomValidityOptions) => {
...
+        }),
+        shortMessage: t.value(`validations.preview.typeMismatch.${type}`, {
+          value: options.props.modelValue?.toString(),
+        }),
+      };
+    }
+
+    const validationData = {
+      value: options.props.modelValue?.toString(),
+      n: options.props.modelValue?.toString().length ?? 0,
```
larsrickert, line 148: `.toString()` works for input and textarea but not for selects because its an array there...maybe we should change the type of modelValue to `string` for now?

packages/sit-onyx/src/composables/useCustomValidity.ts
```diff
@@ -61,34 +83,88 @@ export const useCustomValidity = (options: UseCustomValidityOptions) => {
...
+      shortMessage: t.value(`validations.preview.${errorType}`, validationData),
+    };
+  });
+
   return {
     /**
      * Directive to set the custom error message and emit validityChange event.
      */
     vCustomValidity,
+    // TODO: add tests
```
larsrickert, line 167: open TODO

packages/sit-onyx/src/i18n/locales/de-DE.json
```diff
@@ -1,16 +1,33 @@
 {
   "optional": "(optional)",
   "validations": {
```
larsrickert, line 3: What do you think about grouping them by validation type instead? So something like;

```json
  "validations": {
    "badInput": {
      "preview": "Invalid input",
      "long": "\"{value}\" does not match the expected type."
    }
  },
```

Imho its easier to compare them or add new validations in the future

packages/sit-onyx/src/i18n/locales/de-DE.json
```diff
@@ -1,16 +1,33 @@
...
+    "preview": {
+      "tooShort": "Zu kurz",
+      "tooLong": "Zu lang",
+      "rangeUnderflow": "Zu niedrig",
+      "rangeOverflow": "Zu hoch",
+      "patternMismatch": "Ungültiges Format",
+      "valueMissing": "Ist leer",
```
larsrickert, line 10:
```suggestion
      "valueMissing": "Pflichtfeld",
```

packages/sit-onyx/src/i18n/locales/de-DE.json
```diff
@@ -1,16 +1,33 @@
...
+      "stepMismatch": "Ungültige Zahl",
+      "badInput": "Ungültige Eingabe",
+      "typeMismatch": {
+        "generic": "Ungültiges Eingabeformat",
+        "email": "Ungültiges Mail Format",
```
larsrickert, line 15:
```suggestion
        "email": "Ungültige E-Mail",
```

packages/sit-onyx/src/i18n/locales/en-US.json
```diff
@@ -1,6 +1,24 @@
 {
   "optional": "(optional)",
   "validations": {
+    "preview": {
+      "tooShort": "Too short",
+      "tooLong": "Too long",
+      "rangeUnderflow": "Too low",
+      "rangeOverflow": "Too high",
+      "patternMismatch": "Invalid format",
+      "valueMissing": "Is empty",
```
larsrickert, line 10:
```suggestion
      "valueMissing": "Required",
```

packages/sit-onyx/src/i18n/locales/en-US.json
```diff
@@ -1,6 +1,24 @@
...
+      "stepMismatch": "Invalid number",
+      "badInput": "Invalid input",
+      "typeMismatch": {
+        "generic": "Invalid input type",
+        "email": "Invalid mail format",
```
larsrickert, line 15:
```suggestion
        "email": "Invalid mail",
```

packages/sit-onyx/src/i18n/locales/ko-KR.json
```diff
@@ -1,6 +1,23 @@
 {
   "optional": "(선택사항)",
   "validations": {
+    "preview": {
```
larsrickert, line 4: If they are missing, just omit them. Otherwise English fallback values will not work

packages/sit-onyx/src/styles/mixins/input.scss
```diff
@@ -96,6 +96,17 @@
     }
   }
 
+  &__error-message {
+    display: none;
+  }
+  &:has(#{$base-selector}__native:user-invalid) {
+    #{$base-selector}__error-message {
+      display: flex;
+      flex-flow: row nowrap;
```
larsrickert, line 105:
```suggestion
```

Thats the default for flex :)

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxInput/OnyxInput.vue
BoppLi: which part are you referring to? the extra v-if is needed, you could have only the short error message

BoppLi: aye 🤦🏻‍♀️
```suggestion
```

packages/sit-onyx/src/composables/useCustomValidity.ts
BoppLi: then we need to convert to string when we call the composable, right?

packages/sit-onyx/src/styles/mixins/input.scss
BoppLi: oops I thought wrap was default 🤦🏻‍♀️ thx

packages/sit-onyx/src/composables/useCustomValidity.ts
BoppLi:
```suggestion
```

packages/sit-onyx/src/composables/useCustomValidity.ts
BoppLi: correct. the question is, will we remember to include it when building the number input if I remove it now?

packages/sit-onyx/src/composables/useCustomValidity.spec.ts
BoppLi:
```suggestion
```

packages/sit-onyx/src/components/OnyxSelectInput/OnyxSelectInput.vue
BoppLi:
```suggestion
```

packages/sit-onyx/src/components/OnyxInput/types.ts
BoppLi: hmm but it also is not detected when you try to send a form.. while the other errors will block sending
image

packages/sit-onyx/src/composables/useCustomValidity.ts
BoppLi: if we already have a ticket for it I can move the code to the ticket. what do you think?

packages/sit-onyx/src/composables/useCustomValidity.ts
larsrickert: Not for now since the type is only string for input and textarea, when supporting the select in the future, we can maybe pass in the displayed string value instead

packages/sit-onyx/src/composables/useCustomValidity.ts
larsrickert: Sounds good

packages/sit-onyx/src/composables/useCustomValidity.ts
BoppLi: the composable is also used in checkbox, switch, ... where the value can be boolean.
select is the next step and part of this ticket #570

packages/sit-onyx/src/composables/useCustomValidity.ts
BoppLi: done https://github.com/SchwarzIT/onyx/issues/1273

packages/sit-onyx/src/components/OnyxInput/types.ts
larsrickert: Even if you click submit?

packages/sit-onyx/src/composables/useCustomValidity.ts
larsrickert: Ah okay, I guess its fine for now then

packages/sit-onyx/src/components/OnyxInput/types.ts
BoppLi: exactly, that's the issue.. :(
https://github.com/SchwarzIT/onyx/assets/151019977/f898802d-e211-43c1-86b9-c006feaaf702

larsrickert: Thats the expected behaviour, if only setting minlength and not required, the form is valid when its empty
image

BoppLi: in the video you can see that the value is not empty (it's preset by "A") which is what this comment is warning about

packages/sit-onyx/src/components/OnyxInput/types.ts
```diff
@@ -75,10 +75,16 @@ export type OnyxInputProps = DensityProp &
...
+     * the input invalidity will not be detected by the browser, it will only show as invalid
+     * as soon as a user interacts with the input (types something).
      */
     maxlength?: number;
     /**
      * Minimum number of characters that have to to be entered.
+     * Warning: when the value is (pre)set by code,
+     * the input invalidity will not be detected by the browser, it will only show as invalid
+     * as soon as a user interacts with the input (types something).
```
BoppLi, line 87:
```suggestion
     * Warning: when the value is (pre)set programatically,
     * the input invalidity will not be detected by the browser, it will only turn invalid
     * as soon as a user interacts with the input (types something).
```

packages/sit-onyx/src/components/OnyxInput/types.ts
```diff
@@ -75,10 +75,16 @@ export type OnyxInputProps = DensityProp &
     loading?: boolean;
     /**
      * Maximum number of characters that are allowed to be entered.
+     * Warning: when the value is (pre)set by code,
+     * the input invalidity will not be detected by the browser, it will only show as invalid
+     * as soon as a user interacts with the input (types something).
```
BoppLi, line 80:
```suggestion
     * Warning: when the value is (pre)set programatically,
     * the input invalidity will not be detected by the browser, it will only turn invalid
     * as soon as a user interacts with the input (types something).
```

packages/sit-onyx/src/components/OnyxTextarea/OnyxTextarea.vue
```diff
@@ -144,7 +144,23 @@ const handleInput = (event: Event) => {
...
+      class="onyx-textarea__footer onyx-text--small"
+    >
+      <span v-if="errorMessages" class="onyx-textarea__error-message">
+        <span class="onyx-truncation-ellipsis">{{ errorMessages.shortMessage }}</span>
+
+        <OnyxInfoTooltip
+          v-if="errorMessages.longMessage"
+          class="onyx-textarea__message-tooltip"
+          color="danger"
+          style="color: red"
```
larsrickert, line 158:
```suggestion
```

packages/sit-onyx/src/composables/useCustomValidity.ts
```diff
@@ -13,7 +15,12 @@ export type UseCustomValidityOptions = {
   /**
    * Component props as defined with `const props = defineProps()`
    */
-  props: CustomValidityProp & { modelValue?: unknown };
+  props: CustomValidityProp & {
+    modelValue?: unknown;
+    type?: (typeof TRANSLATED_INPUT_TYPES)[number] | string;
```
larsrickert, line 20:
```suggestion
    type?: InputType;
```

`| string` is more generic than the TRANSLATED_INPUT_TYPES so the resulting type will just be `string`.
We can use `InputType` instead

packages/sit-onyx/src/composables/useCustomValidity.ts
```diff
@@ -22,6 +29,31 @@ export type UseCustomValidityOptions = {
...
+       * A short error message for giving a quick info to the user about the cause of the error
+       */
+      shortMessage: string;
+      /**
+       * An extended informative error message to provide more info
+       * how the error cause can be resolved
+       */
+      longMessage?: string;
+    }
+  | undefined;
```
larsrickert, line 55: For the type itself I wouldn't add `| undefined` and instead add this where the type is used.
Otherwise the type could not be used "required"

packages/sit-onyx/src/composables/useCustomValidity.ts
```diff
@@ -61,34 +95,89 @@ export const useCustomValidity = (options: UseCustomValidityOptions) => {
...
+
           options.emit("validityChange", validityState.value);
         },
         { immediate: true },
       );
     },
   } satisfies Directive<InputValidationElement, undefined>;
 
+  const errorMessages = computed<FormErrorMessages>(() => {
+    if (!validityState.value || validityState.value.valid) return undefined;
```
larsrickert, line 133:
```suggestion
    if (!validityState.value || validityState.value.valid) return;
```

returning "nothing" is equal to undefined

packages/sit-onyx/src/i18n/locales/de-DE.json
```diff
@@ -38,5 +77,8 @@
     "goBack": "Zurück",
     "userMenuLabel": "Benutzeroptionen"
   },
-  "showInfoTooltip": "Info-Tooltip anzeigen"
+  "showTooltip": {
+    "info": "Info-Tooltip anzeigen",
+    "error": "Error-Tooltip anzeigen"
```
larsrickert, line 82:
```suggestion
    "error": "Fehler-Tooltip anzeigen"
```

Is "Error" a German word?

packages/sit-onyx/src/i18n/locales/en-US.json
```diff
@@ -38,5 +77,8 @@
     "goBack": "Go back",
     "userMenuLabel": "User options"
   },
-  "showInfoTooltip": "Show info Tooltip"
+  "showTooltip": {
+    "info": "Show info Tooltip",
```
larsrickert, line 81:
```suggestion
    "info": "Show info tooltip",
```

packages/sit-onyx/src/i18n/locales/ko-KR.json
```diff
@@ -1,20 +1,24 @@
...
-    "tooShort": "{minLength}자 이상으로 입력하세요. (현재 1글자 입력 중) | {minLength}자 이상으로 입력하세요. (현재 {n}글자 입력 중)",
-    "tooLong": "{maxLength}자 이하로 입력하세요. (현재 1글자 입력 중) | {maxLength}자 이하로 입력하세요. (현재 {n}글자 입력 중)",
-    "rangeUnderflow": "입력값이 {min}보다 크거나 동일해야 합니다",
-    "rangeOverflow": "입력값이 {max}보다 작거나 동일해야 합니다",
-    "patternMismatch": "올바른 형식을 입력해주세요.",
-    "valueMissing": "입력란에 내용을 입력해주세요.",
-    "stepMismatch": "{step}의 배수에 해당하는 값을 입력해주세요.",
-    "badInput": "\"{value}\"는 올바른 타입이 아닙니다.",
+    "tooShort": {
+      "fullError": "{minLength}자 이상으로 입력하세요. (현재 1글자 입력 중) | {minLength}자 이상으로 입력하세요. (현재 {n}글자 입력 중)"
```
larsrickert, line 5: If only having fullErrors it will only show the tooltip, right and no error message?
I'd prefer to either remove the translations then or ask @nerdrun for updates ones :)

larsrickert (changes requested):

packages/sit-onyx/src/i18n/locales/de-DE.json
BoppLi: so much denglish, I forgot how to german 🤦🏻‍♀️ I felt sure it is but nope.. ^^'

packages/sit-onyx/src/i18n/locales/ko-KR.json
BoppLi: you said the preview error will fall back to english

BoppLi: I tried it in the form demo. that's how it behaves when korean is set:
image

larsrickert: Ah true sorry^^

larsrickert (approved):

Issue #1142, opened 2024-05-22
Fix Textarea resizing not working correctly on Firefox


JoCa96:
Repro steps:

1. Open https://storybook.onyx.schwarz/?path=/docs/components-textarea--docs in Firefox.
2. Try to resize the textarea using the resize drag button
3. => Only works on the second try

larsrickert:
Hmm thats weird since it works in our Playwright tests 🤔

PR #1277, opened 2024-06-11
fix(OnyxTextarea): Fix first manual resizing being canceled on Firefox

JoCa96:
Relates to #1142 

The issue seems to be specific to Firefox:
Changing CSS properties that effect if `onyx-textarea__wrapper::after` is rendered abort the user's resizing action.
As a workaround, we instead set the height to `0px`, which also disables the auto-sizing of the textarea.


packages/sit-onyx/src/components/OnyxTextarea/OnyxTextarea.vue
```diff
@@ -178,12 +178,14 @@ const handleInput = (event: Event) => {
...
-  // remove max height if user resizes the textarea manually
+  // remove max height and disable auto-sizing if user resizes the textarea manually
   &:has(.onyx-textarea__native[style*="height"]) {
     --max-height: unset;
 
     .onyx-textarea__wrapper::after {
-      display: none;
+      // workaround for [#1142](https://github.com/SchwarzIT/onyx/issues/1142)
+      // `display: none` or changing "content" causes user resizing to be interrupted
+      height: 0px;
```
larsrickert, line 188:
```suggestion
      height: 0;
```

larsrickert (changes requested):

larsrickert (approved):

Issue #883, opened 2024-04-11
Implement OnyxFlyoutMenu


MajaZarkova:
### Open Questions/To-dos

- [x] How to interpret the blank context area? → Not relevant to this ticket and is also only intended for designers 
- [x] Should we separate the icon button and label button from the flyout? → Not relevant to this ticket

### Why?

This component is needed and will be used inside the NavMenuItem. This is just a support component.
It's responsible for handling keyboard navigation, rendering and opening of the flyout.

![Image](https://github.com/SchwarzIT/onyx/assets/18483883/cc42f86f-ff38-48a9-bfae-d165b58b04b4)

### Design

[Figma Link (Context Area section)](https://www.figma.com/file/YfEUBOHk4J4nYrk04geswG/onyx---Design-System-Figma-Library?node-id=1709%3A12388&mode=dev)

### Acceptance criteria

- [x] Has default slot for select items
- [x] has keyboard support
- [x] make sure keyboard navigation works correctly in OnyxNavItem
- [x] make sure aria-selected styles are correctly applied on the OnyxNavItem nested options
- [x] considers the relevant aria roles and patterns
- [x] Used in the OnyxNavItem and the OnyxUserMenu

### Definition of Done

- [x] covered by component tests
- [x] screenshot tests are updated
- [x] follow-up tickets were created if necessary -> #1415 
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature
- [x] Set task to `In Approval`

**Approval**

- [OnyxFlyoutMenu](https://storybook.onyx.schwarz/?path=/docs/future-flyoutmenu--docs)
- [OnyxNavButton](https://storybook.onyx.schwarz/?path=/docs/navigation-modules-navbutton--docs)
      - OnyxNavButton will replace OnyxNavItem. OnyxNavItem is re-written and it's actually used inside OnyxNavButton
- [x] approved by designer @jannick-ux
- [x] approved by dev @larsrickert 

#### Reference implementations

[CoreUI Reference](https://core-ui.schwarz/components/scu-top-bar-context-dropdown)

![Image](https://github.com/SchwarzIT/onyx/assets/18483883/3960e595-58f6-4aee-b282-0cb99c482949)

#### ARIA
https://www.w3.org/WAI/ARIA/apg/patterns/menubar/examples/menubar-navigation/
https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/
[...]

PR #1281, opened 2024-06-12
feat(OnyxNavButton): Implement onyx nav button

MajaZarkova:
Relates to #883 

Implement onyx nav button

## Checklist


packages/sit-onyx/playwright/snapshots/components/OnyxNavButton/NavButton-chromium-linux.png
MajaZarkova, line 1: Open question here, in Storybook we can see the external icon, but I'm not sure why it's not rendered in the screenshots?

packages/sit-onyx/src/components/OnyxFlyoutMenu/future/OnyxFlyoutMenu.stories.ts
```diff
@@ -0,0 +1,45 @@
...
+  { label: "Falcon" },
+  { label: "Owl" },
+];
+
+/**
+ * This example shows a basic OnyxFlyoutMenu
+ */
+export const Default = {
+  args: {
+    default: () => h("button", "Click Me"),
```
larsrickert, line 42:
```suggestion
    default: () => h(OnyxButton, { label: "Hover me" }),
```

packages/sit-onyx/src/components/OnyxFlyoutMenu/future/OnyxFlyoutMenu.vue
```diff
@@ -0,0 +1,123 @@
...
+
+const activeItem = ref<string>();
+
+const {
+  elements: { button, menu, menuItem, listItem, parentComponent },
+  state: { isExpanded },
+} = createMenuButton({
+  onSelect: (href) => {
+    activeItem.value = href;
+  },
```
larsrickert, line 35:
```suggestion
  onSelect: (value) => {
    activeItem.value = value;
  },
```

It could also be something other than a href right?

packages/sit-onyx/src/components/OnyxFlyoutMenu/future/OnyxFlyoutMenu.vue
```diff
@@ -0,0 +1,123 @@
...
+});
+
+const filterVNodesByComponent = (vnodes: VNode[]): VNode[] => {
```
larsrickert, line 38:
```suggestion
const getSlotComponents = (vnodes: VNode[]): VNode[] => {
```

I think something like this is a better suiting name for what it does :)

packages/sit-onyx/src/components/OnyxFlyoutMenu/future/OnyxFlyoutMenu.vue
```diff
@@ -0,0 +1,123 @@
...
+const options = computed(() => {
+  return filterVNodesByComponent(slots.options?.() ?? []);
+});
+</script>
+
+<template>
+  <div class="onyx-flyout-menu" v-bind="parentComponent">
+    <component :is="slots.default?.()?.[0]" v-bind="button" />
+    <div
+      v-if="slots.options || slots.header || slots.footer"
```
larsrickert, line 59:
```suggestion
      v-if="slots.options.length || slots.header || slots.footer"
```

packages/sit-onyx/src/components/OnyxFlyoutMenu/future/OnyxFlyoutMenu.vue
```diff
@@ -0,0 +1,123 @@
...
+      v-show="isExpanded"
+      :class="{
+        'onyx-flyout-menu__list--with-header': !!slots.header,
+        'onyx-flyout-menu__list--with-footer': !!slots.footer,
+        'onyx-flyout-menu__list': true,
+      }"
+    >
+      <slot name="header"></slot>
+      <ul
+        v-if="slots.options"
```
larsrickert, line 69:
```suggestion
        v-if="slots.options.length"
```

packages/sit-onyx/src/components/OnyxFlyoutMenu/future/types.ts
```diff
@@ -0,0 +1,6 @@
+export type OnyxFlyoutMenuProps<TValue> = {
```
larsrickert, line 1: Since the new flyout uses slots instead if props, I think we can remove this

packages/sit-onyx/src/components/OnyxNavButton/OnyxNavButton.stories.ts
```diff
@@ -0,0 +1,84 @@
...
+ * The nav button is used internally to build the main navigation bar component and is not intended to be used individually.
+ */
+const meta: Meta<typeof OnyxNavButton> = {
+  title: "future/NavButton",
+  ...defineStorybookActionsAndVModels({
+    component: OnyxNavButton,
+    events: ["click"],
+    argTypes: {
+      default: {
+        control: { disable: true },
```
larsrickert, line 18:
```suggestion
        control: { type: "text" },
```

Like this, the user can change the slot content as text.
But you also need to remove the `() =>` in the stories below for the default slot for it to work correctly

packages/sit-onyx/src/index.ts
```diff
@@ -101,6 +104,8 @@ export * from "./components/OnyxUserMenu/types";
 export { default as OnyxVisuallyHidden } from "./components/OnyxVisuallyHidden/OnyxVisuallyHidden.vue";
 export * from "./components/OnyxVisuallyHidden/types";
 
+export { default as OnyxFlyoutMenu } from "./components/OnyxFlyoutMenu/OnyxFlyoutMenu.vue";
```
larsrickert, line 107: The FlyoutMenu is already exported. Did you want to export the future instead?

packages/sit-onyx/src/components/OnyxNavButton/OnyxNavButton.stories.ts
```diff
@@ -0,0 +1,84 @@
...
+import { h } from "vue";
+import OnyxNavButton from "./OnyxNavButton.vue";
+import OnyxListItem from "../OnyxListItem/OnyxListItem.vue";
+import OnyxBadge from "../OnyxBadge/OnyxBadge.vue";
+
+/**
+ * The nav button is used internally to build the main navigation bar component and is not intended to be used individually.
+ */
+const meta: Meta<typeof OnyxNavButton> = {
+  title: "future/NavButton",
```
larsrickert, line 12:
```suggestion
  title: "support/NavButton",
```

The nav button is a completely new component, right? :)

packages/sit-onyx/src/components/OnyxFlyoutMenu/future/OnyxFlyoutMenu.vue
```diff
@@ -0,0 +1,123 @@
...
+
+<style lang="scss">
+@use "../../../styles/mixins/layers";
+@use "../../../styles/mixins/list";
+
+.onyx-flyout-menu {
+  @include layers.component() {
+    display: inline-block;
+
+    &__list {
```
larsrickert, line 102: We need to position this absolute because currently the component "jumps" when opening/closing the flyout

packages/sit-onyx/src/components/OnyxUserMenu/future/OnyxUserMenu.vue
```diff
@@ -0,0 +1,173 @@
...
+});
+</script>
+
+<template>
+  <div class="onyx-user-menu">
+    <OnyxFlyoutMenu class="onyx-user-menu__flyout" :aria-label="t('navigation.userMenuLabel')">
+      <button class="onyx-user-menu__trigger onyx-text">
+        <OnyxAvatar v-bind="avatar" size="24px" />
+        <span class="onyx-truncation-ellipsis"> {{ props.username }}</span>
+        <OnyxIcon class="onyx-user-menu__chevron" :icon="chevronLeftSmall" />
```
larsrickert, line 42: The chevron does not update correctly when closing the menu by clicking the button:
image

packages/sit-onyx/src/components/OnyxUserMenu/future/OnyxUserMenu.vue
```diff
@@ -0,0 +1,173 @@
...
+const avatar = computed(() => {
+  if (typeof props.avatar === "object") return { ...props.avatar, label: props.username };
+  return { src: props.avatar, label: props.username };
+});
+</script>
+
+<template>
+  <div class="onyx-user-menu">
+    <OnyxFlyoutMenu class="onyx-user-menu__flyout" :aria-label="t('navigation.userMenuLabel')">
+      <button class="onyx-user-menu__trigger onyx-text">
```
larsrickert, line 39: When hovering from the button down to the flyout, it always closes so there is no way to click an option

larsrickert (changes requested):

packages/headless/src/composables/menuButton/createMenuButton.ts
```diff
@@ -0,0 +1,58 @@
+import { createId } from "../../utils/id";
+import { computed, ref } from "vue";
+
+import { createBuilder } from "../../utils/builder";
+
+export type MenuButtonValue = string | number | boolean;
+export type CreateMenuButtonOptions = { onSelect: (value: string) => void };
```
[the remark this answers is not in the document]
MajaZarkova: @larsrickert I've added one, is ti okay?

packages/sit-onyx/src/components/OnyxFlyoutMenu/future/OnyxFlyoutMenu.vue
MajaZarkova: Checking for length is messing up hover behavior.

packages/sit-onyx/src/components/OnyxFlyoutMenu/future/OnyxFlyoutMenu.vue
MajaZarkova: Checking for length is messing up hover behavior.

packages/sit-onyx/src/components/OnyxFlyoutMenu/future/OnyxFlyoutMenu.vue
```diff
@@ -0,0 +1,123 @@
...
+      <slot name="footer"></slot>
+    </div>
+  </div>
+</template>
+
+<style lang="scss">
+@use "../../../styles/mixins/layers";
+@use "../../../styles/mixins/list";
+
+.onyx-flyout-menu {
```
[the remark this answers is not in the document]
MajaZarkova: Changing the className creates even more side effects with the nav button

packages/sit-onyx/src/index.ts
MajaZarkova: Yes, it was an error, I removed it.

packages/headless/src/composables/menuButton/createMenuButton.ts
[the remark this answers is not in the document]
larsrickert: Yes but we should add the docs only to the `onSelect` instead of the whole type, so e.g.:
```ts
export type CreateMenuButtonOptions = {
  /**
   * Called when a menu item is selected (via mouse or keyboard).
   */
  onSelect: (value: string) => void;
};
```

packages/sit-onyx/playwright/snapshots/components/OnyxNavButton/NavButton-chromium-linux.png
larsrickert, line 1: The external link icon is missing here, right?

packages/sit-onyx/src/components/OnyxFlyoutMenu/future/OnyxFlyoutMenu.vue
```diff
@@ -0,0 +1,123 @@
...
+  return allNodes;
+};
+
+const options = computed(() => {
+  return getSlotComponents(slots.options?.() ?? []);
+});
+</script>
+
+<template>
+  <div class="onyx-flyout-menu" v-bind="parentComponent">
```
larsrickert, line 54: The flyout also openes when hovering outside of the button

https://github.com/SchwarzIT/onyx/assets/67898185/df6800b0-a5a7-454e-9410-3914fbd6798e

packages/sit-onyx/src/components/OnyxFlyoutMenu/future/OnyxFlyoutMenu.vue
```diff
@@ -0,0 +1,123 @@
...
+const {
+  elements: { button, menu, menuItems, listItem, parentComponent },
+  state: { isExpanded },
+} = createMenuButton({
+  onSelect: (value) => {
+    activeItem.value = value;
+  },
+});
+
+const getSlotComponents = (vnodes: VNode[]): VNode[] => {
```
larsrickert, line 36: Jonathan commented on one of my PRs that he would like to prevent iteration over VNodes in slots.
Its fine for me to keep this for now so the PR can be merged but maybe we should get in touch with Jonathan to check if he finds an alternative here?

packages/sit-onyx/src/components/OnyxNavButton/OnyxNavButton.stories.ts
```diff
@@ -0,0 +1,84 @@
...
+  args: {
+    ...Default.args,
+    active: true,
+  },
+} satisfies Story;
+
+/**
+ * This example shows the nav button with nested children.
+ */
+export const WithChildren = {
```
larsrickert, line 58: The example is cut off when opening the flyout. We propably need to add a decorator here to set height/padding

image

packages/sit-onyx/src/components/OnyxFlyoutMenu/future/OnyxFlyoutMenu.vue
```diff
@@ -0,0 +1,123 @@
...
+@use "../../../styles/mixins/list";
+
+.onyx-flyout-menu {
+  @include layers.component() {
+    // TODO: remove comment after replacing this component with the official OnyxFlyoutMenu
+    // display: inline-block;
+
+    &__list {
+      @include list.styles();
+      position: absolute;
```
larsrickert, line 103: I sometimes have this weird behaviour that the flyout is overlapping the button.
Might be caused by the missing left/bottom and left/right properties for the `position: absolute`

image

larsrickert (changes requested):

packages/sit-onyx/playwright/snapshots/components/OnyxNavButton/NavButton-chromium-linux.png
MajaZarkova: Yes, I left an open question at the beginning of the PR about that. I can't understand why it's like that. I can see the external link icon in Storybook.

larsrickert: I think the issue is related to the fact that you also pass slot content in your test in addition to the labe property.
When passing custom slot content, it will override all default content which includes the external link icon

image

larsrickert: You can fix it by removing the default slot with "Nav Button"

packages/headless/src/composables/menuButton/createMenuButton.ts
```diff
@@ -0,0 +1,72 @@
...
+      },
+      flyout: {
+        ...hoverEvents.value,
+      },
+      menu: {
+        id: menuId,
+        role: "menu",
+        "aria-labelledby": buttonId,
+      },
+      menuItems: (data: { active?: boolean; value: string }) => ({
```
larsrickert, line 62:
```suggestion
      menuItem: (data: { active?: boolean; value: string }) => ({
```

This is only for a single menu item :)

packages/headless/src/composables/menuButton/createMenuButton.ts
```diff
@@ -0,0 +1,72 @@
...
+  });
+
+  const hoverEvents = computed(() => {
+    return {
+      onMouseover: () => (debouncedHidden.value = true),
+      onMouseout: () => (debouncedHidden.value = false),
+      onFocusin: () => (isExpanded.value = true),
+      onFocusout: () => (isExpanded.value = false),
+    };
+  });
```
larsrickert, line 36:
```suggestion
import { debounce } from "../../utils/timer";

  /**
   * Debounced expanded state that will only be toggled after a given timeout.
   */
  const updateDebouncedExpanded = debounce(
    (expanded: boolean) => (isExpanded.value = expanded),
    400,
  );

  const hoverEvents = computed(() => {
    return {
      onMouseover: () => updateDebouncedExpanded(true),
      onMouseout: () => updateDebouncedExpanded(false),
      onFocusin: () => (isExpanded.value = true),
      onFocusout: () => (isExpanded.value = false),
    };
  });
```

We already have a debounce utility that we can make use of here :)

packages/headless/src/composables/menuButton/createMenuButton.ts
```diff
@@ -0,0 +1,72 @@
...
+            "aria-expanded": isExpanded.value,
+            "aria-haspopup": true,
+            id: buttonId,
+            ...hoverEvents.value,
+          }) as const,
+      ),
+      listItem: {
+        role: "none",
+      },
+      flyout: {
```
larsrickert, line 54: Isn't the flyout the same as the menu? So we could get rid of the flyout and bind it to the menu directly

packages/sit-onyx/src/components/OnyxNavButton/OnyxNavButton.vue
```diff
@@ -0,0 +1,136 @@
...
+</template>
+
+<style lang="scss">
+@use "../../styles/mixins/layers.scss";
+
+.onyx-nav-button {
+  @include layers.component() {
+    $gap: var(--onyx-spacing-2xs);
+
+    .onyx-future-flyout-menu__list {
```
larsrickert, line 72: Shouldn't these styles be applied by the OnyxFlyoutMenu itself? Because they are the same for all flyouts (nav button, user menu etc.)

packages/sit-onyx/src/components/OnyxFlyoutMenu/future/OnyxFlyoutMenu.vue
MajaZarkova: You still have this behavior?

packages/sit-onyx/src/components/OnyxUserMenu/future/OnyxUserMenu.vue
```diff
@@ -0,0 +1,177 @@
...
+
+const avatar = computed(() => {
+  if (typeof props.avatar === "object") return { ...props.avatar, label: props.username };
+  return { src: props.avatar, label: props.username };
+});
+</script>
+
+<template>
+  <div class="onyx-user-menu">
+    <OnyxFlyoutMenu class="onyx-user-menu__flyout" :aria-label="t('navigation.userMenuLabel')">
```
larsrickert, line 38: The header/footer background looks a little off
image

packages/headless/src/composables/menuButton/createMenuButton.ts
larsrickert: Also, should we only use the debounce for closing? Otherwise it might feel weird if it does not open when hovering.

Or maybe change the timeout to 200 or so

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxFlyoutMenu/future/OnyxFlyoutMenu.vue
larsrickert: Does not seem so...I'm just wondering where the flyout gets its position from? We haven't set any left/top property

MajaZarkova: When I tried to set the left/top properties, I couldn't align the properly.

packages/headless/src/composables/menuButton/createMenuButton.ts
MajaZarkova: We are binding the menu only to the <ul> element, and in the userMenu component where we have header and footer the hovering effects doesn't work. This is why I added the flyout.

packages/headless/src/composables/menuButton/createMenuButton.ts
MajaZarkova: Now I'm using it only on closing

larsrickert (approved):

Issue #1241, opened 2024-06-04
Implement basic modal


BoppLi:

BoppLi:
@larsrickert is this a bug? or does the project dev have to take care of the overflow behavior
![image](https://github.com/SchwarzIT/onyx/assets/151019977/0dc0b7ef-0d57-4053-bbf8-c76f8c797eef)

larsrickert:
> [@larsrickert](https://github.com/larsrickert) is this a bug? or does the project dev have to take care of the overflow behavior

Thanks a bug :) I fixed it in #1318

BoppLi:
another bug or out-of-scope?

![Image](https://github.com/SchwarzIT/onyx/assets/151019977/e9b5b0f3-42a1-479a-ad0d-6f072f34feef)

you can try this out on the branch `feat/alpha-test-update`

larsrickert:
> maybe you need to set the `--onyx-z-index-app-overlay` z-index, at least for non-modal dialogs

Good catch, I also fixed this in #1318

PR #1285, opened 2024-06-12
feat: add `OnyxDialog` component

larsrickert:
Relates to #1241

Add `OnyxDialog` component

## Checklist


packages/sit-onyx/src/components/OnyxDialog/OnyxDialog.vue
```diff
@@ -0,0 +1,125 @@
...
+
+// sync open state
+watch(
+  [dialogRef, () => props.open],
+  () => {
+    if (props.open) openDialog();
+    else dialogRef.value?.close();
+  },
+  { immediate: true },
+);
```
JoCa96, line 47: `immediate` doesn't have an effect here, as dialogRef will always be undefined initially.

```suggestion
watch(
  [dialogRef, () => props.open],
  () => {
    if (props.open) openDialog();
    else dialogRef.value?.close();
  }
);
```

packages/sit-onyx/src/components/OnyxDialog/OnyxDialog.vue
```diff
@@ -0,0 +1,125 @@
...
+
+    @include density.compact {
+      --onyx-dialog-padding: var(--onyx-spacing-md);
+    }
+    @include density.default {
+      --onyx-dialog-padding: var(--onyx-spacing-lg);
+    }
+    @include density.cozy {
+      --onyx-dialog-padding: var(--onyx-spacing-xl);
+    }
```
JoCa96, line 99:
```suggestion
.onyx-dialog {
  @include density.compact {
    --onyx-dialog-padding: var(--onyx-spacing-md);
  }

  @include density.default {
    --onyx-dialog-padding: var(--onyx-spacing-lg);
  }

  @include density.cozy {
    --onyx-dialog-padding: var(--onyx-spacing-xl);
  }

  @include layers.component() {
    --backdrop-opacity: 0.15;
```

JoCa96 (changes requested):

packages/sit-onyx/src/components/OnyxDialog/OnyxDialog.vue
larsrickert: I am not sure if this is also true for server side rendering / static site generation in Nuxt

packages/sit-onyx/src/components/OnyxDialog/OnyxDialog.vue
```diff
@@ -0,0 +1,125 @@
...
+
+    outline: none;
+    border: var(--onyx-1px-in-rem) solid var(--onyx-color-base-neutral-300);
+    border-radius: var(--onyx-radius-md);
+    font-family: var(--onyx-font-family);
+    color: var(--onyx-color-text-icons-neutral-intense);
+    padding: var(--onyx-dialog-padding);
+    background-color: var(--onyx-color-base-background-blank);
+
+    &:where(&) {
```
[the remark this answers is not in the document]
larsrickert: I removed it, my idea was that the size can easily be overridden but since we use layers this should already be the case

packages/sit-onyx/src/components/OnyxDialog/OnyxDialog.vue
```diff
@@ -0,0 +1,125 @@
...
+    @include density.compact {
+      --onyx-dialog-padding: var(--onyx-spacing-md);
+    }
+    @include density.default {
+      --onyx-dialog-padding: var(--onyx-spacing-lg);
+    }
+    @include density.cozy {
+      --onyx-dialog-padding: var(--onyx-spacing-xl);
+    }
+
```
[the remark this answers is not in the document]
larsrickert: Whoa why does this work 🤔  The .dark is nested inside .onyx-dialog but the dark class is set on the `<html>` element

packages/sit-onyx/src/components/OnyxDialog/OnyxDialog.vue
JoCa96: Template refs will only be available after the component is mounted. So on the server it will also be undefined

JoCa96 (approved):

PR #1310, opened 2024-06-14
feat(validity): Allow short/long message distinguishing for custom errors

BoppLi:
Relates to #570 

feat(validity): Allow short/long message distinguishing for custom errors

## Checklist


packages/sit-onyx/src/composables/useCustomValidity.ts
```diff
@@ -5,11 +5,13 @@ import enUS from "../i18n/locales/en-US.json";
 import { injectI18n } from "../i18n";
 import type { InputType } from "../components/OnyxInput/types";
 
+export type CustomErrorType = string | FormErrorMessages;
```
larsrickert, line 8: Should we just change it to always be an object with required shortMessage and optional longMessage? Then we don't have to always check for both cases

packages/sit-onyx/src/composables/useCustomValidity.ts
```diff
@@ -92,7 +108,9 @@ export const useCustomValidity = (options: UseCustomValidityOptions) => {
       /**
        * Sync custom error with the native input validity.
        */
-      watchEffect(() => el.setCustomValidity(options.props.customError ?? ""));
+      watchEffect(() =>
+        el.setCustomValidity(getCustomErrors(options.props.customError)?.shortMessage ?? ""),
```
larsrickert, line 112: Should we use both error messages in `setCustomValidity()` so no information is lost for screen readers?`

larsrickert (changes requested):

packages/sit-onyx/src/composables/useCustomValidity.ts
BoppLi: it also is used for radio button, there it only ever appears in a tooltip so it would be wasteful effort to expect from the devs to provide an object

packages/sit-onyx/src/composables/useCustomValidity.ts
BoppLi: like `${shortMessage}: ${longMessage}` ?

BoppLi: done

BoppLi: image

packages/sit-onyx/src/composables/useCustomValidity.ts
```diff
@@ -53,6 +55,29 @@ export type FormErrorMessages = {
...
+    return { shortMessage: customError, longMessage: customError };
+  }
+  return customError;
+};
+
+export const getCustomErrorText = (customError?: CustomErrorType): string | undefined => {
+  if (!customError) return;
+  if (typeof customError === "string") {
+    return customError;
+  }
```
larsrickert, line 76: Can't we also use `getCustomErrors()` here?

packages/sit-onyx/src/composables/useCustomValidity.ts
```diff
@@ -5,11 +5,13 @@ import enUS from "../i18n/locales/en-US.json";
...
 import type { InputType } from "../components/OnyxInput/types";
 
+export type CustomErrorType = string | FormErrorMessages;
+
 export type CustomValidityProp = {
   /**
    * Custom error message to show. Will only show up after the user has interacted with the input.
    */
-  customError?: string;
+  customError?: CustomErrorType;
```
larsrickert, line 14: Can be update the Storybook controls to be of type text so the user can play around with it?
image

larsrickert (changes requested):

packages/sit-onyx/src/composables/useCustomValidity.ts
BoppLi: then we'll have to check if shortMessage + longMessage are the same else you would have

`getCustomErrorText('foo')` return `"foo: foo"`

packages/sit-onyx/src/composables/useCustomValidity.ts
BoppLi: if I set text, storybook can't interpret the example
image

if I leave the standard, it works for object + string aswell
image
image

when you edit raw, you can just put in a string

larsrickert (approved):

PR #1318, opened 2024-06-14
fix(OnyxDialog): prevent overflow and add z-index

larsrickert:
Relates to [#1241](https://github.com/SchwarzIT/onyx/issues/1241#issuecomment-2167831241)

Prevent overflow of the dialog content.

## Checklist


packages/sit-onyx/src/components/OnyxDialog/OnyxDialog.ct.tsx
```diff
@@ -34,6 +34,23 @@ test("should render in modal", async ({ mount, makeAxeBuilder, page }) => {
...
+  await mount(
+    <OnyxDialog label="Label" open>
+      {"Test".repeat(64)}
+    </OnyxDialog>,
+  );
+
+  // accessibility tests
+  const accessibilityScanResults = await makeAxeBuilder()
+    // the interactive/focusable content is provided by the project
+    .disableRules(["scrollable-region-focusable"])
```
BoppLi, line 47: as I understand this rule, the idea is that a scroll container must be keyboard navigatable so you can use the arrow keys to scroll. so even if the project has a focusable button or so, if there is a long text, if you don't provide focus on the scroll area, the user won't be able to read all content.

so I'd not disable this rule here and instead make the scroll container focusable.

BoppLi (changes requested):

packages/sit-onyx/src/components/OnyxDialog/OnyxDialog.ct.tsx
BoppLi: [further info](https://dequeuniversity.com/rules/axe/4.8/scrollable-region-focusable)

> The key to getting a scrollable region to be accessible with the keyboard is to ensure that a keyboard-only user **can focus the scrollable region itself or a static text item within the scrollable region**. The first example below shows the first technique and the second one shows the second technique.

BoppLi: uff that's a conflict... let's discuss it monday with the team how to proceed?

larsrickert: Sure, why is this a conflict?

BoppLi (approved):

Issue #1324, opened 2024-06-17
Implement OnyxColorSchemeDialog


larsrickert:
### Why?

We want to offer a component where the user can select which appearance he wants to have the application in (light/dark/auto).

### Design

[Figma](https://www.figma.com/design/YfEUBOHk4J4nYrk04geswG/Onyx-Component-Library?node-id=4499-144775&t=4dTsRVGCH9ztrAqS-4)

### Acceptance criteria

- [x] the design is implemented
- [x] the API integrates nicely with [vueuse](https://vueuse.org)
- [x] the current [docs](https://onyx.schwarz/development/theming.html) are updated
- [x] when the dialog opens, the currently selected radio button is focused

### Definition of Done

- The following component requirements are implemented:
  - [x] example usage added to `apps/alpha-test-app/src/App.vue`
- [x] covered by component tests
- [x] screenshot tests are updated
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature
- [x] Set task to `In Approval`

**Approval**

- [Storybook](https://storybook.onyx.schwarz/?path=/docs/support-colorschemedialog--docs)
- [Docs](https://onyx.schwarz/development/theming.html)
- [x] approved by designer @jannick-ux
- [x] approved by dev @BoppLi

PR #1326, opened 2024-06-17
feat: implement `OnyxColorSchemeDialog` component

larsrickert:
Relates to #1324

- add new `OnyxColorSchemeDialog` component
- add `autofocus` property to some components  so it can be used to autofocus the component when used inside a dialog

## Checklist


apps/docs/src/development/theming.md
```diff
@@ -5,26 +5,49 @@ To learn more about the theming concept of onyx, take a look at our [colors docu
...
-const toggleDark = useToggle(isDark);
+<script setup lang="ts">
+import circleContrast from "@sit-onyx/icons/circle-contrast.svg?raw";
+import { useColorMode } from "@vueuse/core";
+import { OnyxColorSchemeDialog, OnyxNavBar, OnyxUserMenu, type SelectOption } from "sit-onyx";
+import { ref } from "vue";
+
+const userMenuOptions = [
+  { value: "color-scheme", label: "Appearance", icon: circleContrast },
+  // your option user enu options...
```
BoppLi, line 23:
```suggestion
  // your user menu options...
```

packages/sit-onyx/src/types/components.ts
```diff
@@ -46,3 +46,14 @@ export type BaseSelectOption<TValue extends SelectOptionValue = SelectOptionValu
 };
 
 export type SelectOptionValue = string | number | boolean;
+
+export type AutofocusProp = {
+  /**
+   * Whether to focus the input on page load or when dialog/popover become shown if used inside a dialog/popover.
```
BoppLi, line 52:
```suggestion
   * Whether to focus an interactive component on page load or inside a dialog/popover once it opens.
```

packages/sit-onyx/src/types/components.ts
```diff
@@ -46,3 +46,14 @@ export type BaseSelectOption<TValue extends SelectOptionValue = SelectOptionValu
 };
 
 export type SelectOptionValue = string | number | boolean;
+
+export type AutofocusProp = {
```
BoppLi, line 50: set it on TextArea + select too?

packages/sit-onyx/src/components/OnyxRadioButton/types.ts
```diff
@@ -1,13 +1,14 @@
...
+import type { AutofocusProp, BaseSelectOption, SelectOptionValue } from "../../types";
 
 export type OnyxRadioButtonProps<TValue extends SelectOptionValue = SelectOptionValue> = Omit<
   BaseSelectOption<TValue>,
   "hideLabel"
 > &
   DensityProp &
-  CustomValidityProp & {
+  CustomValidityProp &
+  AutofocusProp & {
```
BoppLi, line 11: don't you also need to updatet the `RadioButtonOption` type in 
`packages/sit-onyx/src/components/OnyxRadioButtonGroup/types.ts`
like in `packages/sit-onyx/src/components/OnyxCheckboxGroup/types.ts`?

BoppLi (changes requested):
31/36 reviewed

packages/sit-onyx/src/i18n/locales/en-US.json
```diff
@@ -84,5 +86,21 @@
   "showTooltip": {
     "info": "Show info tooltip",
     "error": "Show error tooltip"
+  },
+  "colorScheme": {
+    "headline": "Change appearance",
```
BoppLi, line 91: we could add a key just for the word "Appearance" for the user-menu-flyout as well, if that's a common feature in all projects so they don't have to take care of translating it everytime.

packages/sit-onyx/src/i18n/locales/de-DE.json
```diff
@@ -84,5 +86,21 @@
...
+  "colorScheme": {
+    "headline": "Erscheinungsbild ändern",
+    "subtitle": "Bitte wählen das Erscheinungsbild der Anwendung",
+    "auto": {
+      "label": "Auto",
+      "description": "Die Anwendung passt ihr Aussehen automatisch an deine Systemeinstellungen an."
+    },
+    "light": {
+      "label": "Hell",
+      "description": "Der Modus \"Hell\" bietet einen optimalen Kontrast für helle Umgebungen."
```
BoppLi, line 99:
```suggestion
      "description": "Der helle Modus bietet einen optimalen Kontrast für helle Umgebungen."
```

in english we also don't have extra "", so this would be closer to the OG text?

packages/sit-onyx/src/components/OnyxColorSchemeDialog/OnyxColorSchemeDialog.vue
```diff
@@ -0,0 +1,166 @@
...
+  close: [];
+}>();
+
+const currentValue = ref(props.modelValue);
+watchEffect(() => (currentValue.value = props.modelValue));
+
+const { t } = injectI18n();
+
+const options = computed<
+  { value: ColorSchemeValue; description: string; image: string; label: string }[]
```
BoppLi, line 34: should the type definition also be moved to the `packages/sit-onyx/src/components/OnyxColorSchemeDialog/types.ts` like we do on many other components like checkbox/radio ? or not in this case because the devs don't need to import that type

packages/sit-onyx/src/i18n/locales/en-US.json
```diff
@@ -84,5 +86,21 @@
...
+    "subtitle": "Please select the appearance of the application",
```
BoppLi, line 92: imho this sentence sounds a bit weird in that context. might discuss with jannick about rephrasing?
```suggestion
    "subtitle": "Here you can choose your preferred appearance of the application",
```

asking with "Please" sounds like there is a problem if you don't select something ¯\\\_(ツ)_\/¯

packages/sit-onyx/src/components/OnyxColorSchemeDialog/OnyxColorSchemeDialog.stories.ts
```diff
@@ -0,0 +1,32 @@
...
+        template: `<div style="height: 48rem;"> <story /> </div>`,
+      }),
+    ],
+  }),
+};
+
+export default meta;
+type Story = StoryObj<typeof OnyxColorSchemeDialog>;
+
+export const Default = {
```
BoppLi, line 27: the docs page for this example is problematic.
http://localhost:6006/?path=/docs/support-colorschemedialog--docs

I'd recommend to use a "show dialog" button in the example and connect the cancel/apply to closing the example?
like we did in the examples of https://core-ui.schwarz/components/scu-modal-dialogue

packages/sit-onyx/src/components/OnyxColorSchemeDialog/OnyxColorSchemeDialog.vue
```diff
@@ -0,0 +1,166 @@
...
+          {{ t("colorScheme.subtitle") }}
+        </span>
+      </div>
+
+      <fieldset class="onyx-color-scheme-dialog__list" @change="handleChange">
+        <div v-for="option in options" :key="option.value" class="onyx-color-scheme-dialog__option">
+          <!-- eslint-disable-next-line vue/no-v-html -->
+          <figure class="onyx-color-scheme-dialog__image" v-html="option.image"></figure>
+
+          <div>
```
BoppLi, line 84: [optional ][cherry on top] if you want to flatten the dom a bit, instead of div-wrapping radio + p, you could use a grid layout on `onyx-color-scheme-dialog__option` and assign areas

packages/sit-onyx/src/components/OnyxColorSchemeDialog/OnyxColorSchemeDialog.vue
```diff
@@ -0,0 +1,166 @@
...
+      align-items: flex-start;
+      gap: var(--onyx-spacing-md);
+      align-self: stretch;
+      color: var(--onyx-color-text-icons-neutral-medium);
+
+      &:not(:last-child) {
+        border-bottom: var(--onyx-1px-in-rem) solid var(--onyx-color-base-neutral-300);
+      }
+
+      @container (max-width: 24rem) {
```
BoppLi, line 143: awesome use for container-query! 🤩 here's a 🍪 😋

BoppLi (changes requested):
31/36 reviewed

packages/sit-onyx/src/components/OnyxColorSchemeDialog/OnyxColorSchemeDialog.vue
```diff
@@ -0,0 +1,166 @@
...
+    <form class="onyx-color-scheme-dialog" @submit.prevent="handleApply">
+      <div>
+        <OnyxHeadline is="h2"> {{ t("colorScheme.headline") }}</OnyxHeadline>
+        <span class="onyx-color-scheme-dialog__subtitle onyx-text">
+          {{ t("colorScheme.subtitle") }}
+        </span>
+      </div>
+
+      <fieldset class="onyx-color-scheme-dialog__list" @change="handleChange">
+        <div v-for="option in options" :key="option.value" class="onyx-color-scheme-dialog__option">
```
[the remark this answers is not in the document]
larsrickert: The `OnyxRadioButton` already uses a label, so we can not use a <label> here

packages/sit-onyx/src/components/OnyxRadioButton/types.ts
larsrickert: Good catch, thanks :)

packages/sit-onyx/src/components/OnyxColorSchemeDialog/OnyxColorSchemeDialog.vue
larsrickert: Thats only a local type so I don't want to export it :)

packages/sit-onyx/src/i18n/locales/en-US.json
larsrickert: Currently the translations are only used inside onyx, there is no way for the project to access it.
But I plan to also implement a pre-build user menu option anyways which would inlcude the translation
image


But we need to wait for #883

apps/alpha-test-app/src/App.vue
```diff
@@ -81,5 +89,11 @@ const toggleDark = useToggle(isDark);
     </template>
 
     <RouterView />
+
+    <OnyxColorSchemeDialog
+      v-model="colorScheme"
+      :open="isColorSchemeDialogOpen"
```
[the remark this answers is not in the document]
larsrickert: We could do this, but it would differ from the `OnyxDialog` API then 🤔

packages/sit-onyx/src/components/OnyxColorSchemeDialog/OnyxColorSchemeDialog.vue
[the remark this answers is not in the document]
BoppLi: then is there another way, e.g. to emit a change on click of the div?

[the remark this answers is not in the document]
BoppLi: also, seems to be acceptible to have more than one label:
> The form control that a label is labeling is called the labeled control of the label element. Multiple labels can be associated with the same form control

see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label?retiredLocale=de

apps/alpha-test-app/src/App.vue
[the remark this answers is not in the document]
BoppLi: maybe do the same for OnyxDialog? when the prop and emit control the same ref in the end, it very much feels like a perfect use-case for vmodel?

packages/sit-onyx/src/types/components.ts
BoppLi: maybe a question for our dev lead @JoCa96  :)

packages/sit-onyx/src/i18n/locales/en-US.json
BoppLi: aah alright, sounds good 😊

packages/sit-onyx/src/components/OnyxDialog/OnyxDialog.stories.ts
```diff
@@ -33,3 +35,27 @@ export const Default = {
     default: "Dialog content...",
   },
 } satisfies Story;
+
+export const Modal: Story = {
```
BoppLi, line 39: you could add a description where you recommend to close it with Escape after opening, if someone thinks they are getting stuck on the docs page ^^

BoppLi (changes requested):

apps/alpha-test-app/src/App.vue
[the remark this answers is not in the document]
larsrickert: But the v-model will always only emit false since the dialog does not open itself :/ Also I think not forcing it to be a boolean is quiet helpful for cases like this which might be very common:

```ts
const itemIdToEdit = ref(3);

<OnyxDialog :open="itemIdToEdit != undefined" @close="itemIdToEdit = undefined" />
```

packages/sit-onyx/src/components/OnyxColorSchemeDialog/OnyxColorSchemeDialog.stories.ts
```diff
@@ -13,9 +15,19 @@ const meta: Meta<typeof OnyxColorSchemeDialog> = {
...
+      (story, ctx) => ({
+        components: { story, OnyxButton },
+        setup: () => {
+          const isOpen = ref(false);
+          watchEffect(() => {
+            ctx.args.open = isOpen.value;
+          });
+          return { isOpen };
+        },
+        template: `<div style="height: 48rem;">
```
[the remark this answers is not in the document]
larsrickert: True, its a leftover because it was a non-modal dialog before :)

apps/alpha-test-app/src/App.vue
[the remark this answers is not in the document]
BoppLi: ah right, oki doki :)

packages/sit-onyx/src/components/OnyxColorSchemeDialog/OnyxColorSchemeDialog.vue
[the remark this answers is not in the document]
larsrickert: As discussed, Jannick will adjust the design to remove the radio button.
Should we merge this as it is for now or wait for the updated design?

packages/sit-onyx/src/types/components.ts
larsrickert: Done

BoppLi (dismissed):

BoppLi (approved):

BoppLi (approved):

Issue #1236, opened 2024-06-04
Implement internal component for input/textarea/select header/footer


larsrickert:
### Why?

Currently we have a lot of duplicated/copied code between the `OnyxInput`, `OnyxTextarea` and `OnyxSelect` because they almost share
identical functionality (label, message, tooltips etc.) except the main input component.

We want to create an internal component (something like `OnyxFormElement`) which implements the shared functionality like header/footer and has a slot to place the individual component (input, textarea, select etc.).

### Acceptance criteria

- a shared component is created
- components are updated internally to use the new internal component
  - [x] `OnyxInput`
  - [x] `OnyxTextarea`
  - [x] `OnyxSelect` 

### Definition of Done

- [x] covered by component tests
- [x] screenshot tests are updated
- [x] follow-up tickets were created if necessary
  - #1302 
  - #1355 
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature
- [x] Storybook code snippet of new/changed examples are checked that they are generated correctly
- [x] Set task to `In Approval`

**Approval**

- [Form Element in Storybook](https://storybook.onyx.schwarz/?path=/docs/support-formelement--docs)
- [Input in Storybook](https://storybook.onyx.schwarz/?path=/docs/components-input--docs)
- [Textarea in Storybook](https://storybook.onyx.schwarz/?path=/docs/components-textarea--docs)
- [Select in Storybook](https://storybook.onyx.schwarz/?path=/docs/components-select--docs)
- [x] approved by dev @larsrickert 
- [x] approved by @jannick-ux 

### Implementation details

Example usage:

```html
<OnyxFormElement v-bind="props">
  <textarea />
</OnyxFormElement>
```

PR #1327, opened 2024-06-17
feat(OnyxFormElement): Unify template & style of header/footer form elements

BoppLi:
Relates to #1236 

Unifies template + style of header/footer sections from OnyxInput and OnyxTextarea

## Checklist


packages/sit-onyx/src/components/OnyxFormElement/types.ts
```diff
@@ -0,0 +1,31 @@
+import type { FormErrorMessages } from "../../composables/useCustomValidity";
+import type { RequiredMarkerProp } from "../../composables/required";
+
+export type OnyxFormElementProps = RequiredMarkerProp & {
+  /**
+   * Label to show above the form element.
+   */
+  label?: string;
```
larsrickert, line 8: Label should be required for aria reasons, we offer a `hideLabel` property usually to hide it visually

packages/sit-onyx/src/components/OnyxFormElement/types.ts
```diff
@@ -0,0 +1,31 @@
...
+   */
+  messageTooltip?: string;
+  /**
+   * Error messages that inform about causes for invalidity of form components
+   */
+  errorMessages?: FormErrorMessages;
+  /**
+   * Text that will be displayed on the right of the footer
+   */
+  footerRightText?: string;
```
larsrickert, line 30: This is always a value counter, right? Would prefer to replace this with a `modelValue` and `maxlength` property

packages/sit-onyx/src/styles/mixins/input.scss
larsrickert, line 1: We could also move all these styles to the OnyxFormElement

larsrickert (changes requested):

packages/sit-onyx/src/styles/mixins/input.scss
BoppLi: then how would we access the `$base-selector` that is different for each component?

or do you mean move the mixin to OnyxFormElement and change the import?

packages/sit-onyx/src/components/OnyxFormElement/types.ts
BoppLi: but only as props, not the update:model-value emitter right? the formElement never emits

packages/sit-onyx/src/components/OnyxFormElement/types.ts
BoppLi: ok 👍

packages/sit-onyx/src/components/OnyxFormElement/types.ts
larsrickert: Yes :)

packages/sit-onyx/src/styles/mixins/input.scss
larsrickert: True, nevermind 👍

packages/sit-onyx/src/components/OnyxFormElement/types.ts
```diff
@@ -0,0 +1,46 @@
+import type { FormErrorMessages } from "../../composables/useCustomValidity";
+import type { RequiredMarkerProp } from "../../composables/required";
+
+export type OnyxFormElementProps = RequiredMarkerProp & {
```
larsrickert, line 4: Should we also update the prop types for input/textarea to re-use this so we don't need to duplicate the descriptions? Somehting like

Pick<OnyxFormElementProps, "label" | "hideLabel"> etc.

larsrickert (dismissed):

packages/sit-onyx/src/components/OnyxFormElement/types.ts
BoppLi: I also thought this might be nice but wasn't sure :) I'll do it where the components don't have additional docs 👍

larsrickert (approved):

PR #1330, opened 2024-06-17
fix(OnyxSelect): fix item hidden beneath sticky search when using keyboard

JoCa96:
* added `scroll-padding-top` when search is enabled, so the scrollTo function considers the height of the sticky search

### DIFF
--- a/packages/sit-onyx/src/components/OnyxSelect/OnyxSelect.vue
+++ b/packages/sit-onyx/src/components/OnyxSelect/OnyxSelect.vue
@@ -420,7 +420,7 @@ const selectInputProps = computed(() => {
 
     &__wrapper:has(.onyx-mini-search) {
       // Add scroll padding, so items are not hidden beneath the search input
-      scroll-padding-top: 40px;
+      scroll-padding-top: var(--option-height);
     }
 
     &__slot {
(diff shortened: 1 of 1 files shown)


packages/sit-onyx/src/components/OnyxSelect/OnyxSelect.vue
```diff
@@ -418,6 +418,11 @@ const selectInputProps = computed(() => {
       outline: $outline-size solid var(--onyx-color-base-primary-200);
     }
 
+    &__wrapper:has(.onyx-mini-search) {
+      // Add scroll padding, so items are not hidden beneath the search input
+      scroll-padding-top: 40px;
```
larsrickert, line 423: Can this value be calculated with CSS variables? Because the height changes based on the density, right?

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxSelect/OnyxSelect.vue
JoCa96: Right, we can just use `--option-height`

larsrickert (approved):

Issue #1266, opened 2024-06-10
Implement feedback from alpha testers


larsrickert:
We received some feedback from our alpha testers / developers:

- [ ] create optional `.css` file that can be imported to reset/apply global styles (e.g. rested body margin, apply background color etc.)
- [ ] JSDocs are missing for some properties when hovering over them in the IDE
- [ ] OnxyTable: does not fir to the grid because it has `width: fit-content;`
- [ ] Docs: example for how to override onyx styles (`:deep()` selector)
- [ ] Nuxt: When importing icons, a type error is shown in the editor. Maybe we can include a `.d.ts` file or something in our `@sit-onyx/icons` for this
- [ ] Icons: No intellisense when importing icons. Would be nice if icons are suggested / autocompleted
- [ ] OnyxTable: When using OnyxButton inside the header, the button is not clickable
- [ ] Not all components are exported (mainly support components)
- [ ] Docs: Mention layout components in getting started guide
- [ ] Docs: Source code for `OnyxNavBar` shows `<component />` instead of `<OnyxNavSeparator />`
- [ ] OnyxNavSeparator: `Onyx` prefix is missing for source code
- [ ] OnyxSelect: make more clear that search logic must be implemented on project side
- [ ] OnyxTable: rename property `grid` to not mix it up with the "real" layout grid

JoCa96:
Moved done stuff to #1354

PR #1352, opened 2024-06-18
fix(OnyxNavAppArea): max-height not working correctly

JoCa96:
Relates to #1266


BoppLi (dismissed):
doesn't this need a `patch` changeset?

JoCa96:
> doesn't this need a `patch` changeset?
@BoppLi done

packages/sit-onyx/src/components/OnyxNavAppArea/OnyxNavAppArea.vue
```diff
@@ -61,7 +61,7 @@ const buttonLabel = computed(() => props.label ?? t.value("navigation.goToHome")
     padding: var(--onyx-spacing-md);
     font-weight: 600;
     white-space: pre-line;
-    max-height: 100%;
+    max-height: 100cqh;
```
larsrickert, line 64: Fancy^^ Never saw this unit before :D
A comment might help if we take a look at this in the future or we just set the logo max width also to 1.5rem

BoppLi (dismissed):

packages/sit-onyx/src/components/OnyxNavAppArea/OnyxNavAppArea.vue
JoCa96: added comment

BoppLi (approved):

Issue #1299, opened 2024-06-13
Implement toast progress bar


BoppLi:
### Depends on

- #1286  

### Why?

OnyxToast in the (default) autoclose behavior needs a colored bar on the bottom that represents the time remaining until the toast closes.

### Design

- [Figma](https://www.figma.com/design/YfEUBOHk4J4nYrk04geswG/Onyx-Component-Library?node-id=4922-23696&t=bsuwI8wAhjcT7zdW-4)

### Acceptance criteria

- [x] show (support component) a colored [meter](https://www.w3.org/WAI/ARIA/apg/patterns/meter/) that decreases.
- [x] when hovering, the timer stops => if problematic for implementation, we can drop this feature.

### Definition of Done

- [x] covered by component tests
- [x] screenshot tests are updated (?)
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature
- [x] Storybook code snippet of new/changed examples are checked that they are generated correctly
- [x] Set task to `In Approval`

**Approval**

- [Storybook link](https://storybook.onyx.schwarz/?path=/story/support-toast--success)
- [x] approved by designer @mj-hof 
- [x] approved by dev @BoppLi 

### Reference implementations 

- https://ui.nuxt.com/components/notification#usage

### Aria Pattern

- [Meter](https://www.w3.org/WAI/ARIA/apg/patterns/meter/)

Issue #1286, opened 2024-06-12
Implement OnyxToast


BoppLi:
### Why?

A component for immediate feedback on user actions is needed.  
Info: A Notification Center will be implemented at a later stage.

### Design

- [Figma](https://www.figma.com/design/YfEUBOHk4J4nYrk04geswG/Onyx-Component-Library?node-id=4922-23696&t=bsuwI8wAhjcT7zdW-4)

### Acceptance criteria

- [x] storybook shows different `color` settings: neutral, danger, warning, success
- [x] a toast supports `icon` (optional), `title` (required!), `description` (optional)
- [x] truncation is implemented
  - truncate title: 1 line
  - truncate description: max 3 lines, after that the text gets scrollable
- [x] default: autoclose is implemented
  - [x] timer duration can be overwritten (default: 5sec)
- [x] manual close with "x" at the top right is provided
- [x] Interactivity: optional mode to make the toast clickable -> must be accessibly by keyboard too!
  - [x] when interactive, the cursor becomes a pointer

### Out of scope

- How a toast works in combination with bottom bars
- #1299
- #1300
- #1301 

### Definition of Done

- The following component requirements are implemented:
  - [x] `density`
- [x] covered by component tests
- [x] screenshot tests are updated
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature
- [x] Storybook code snippet of new/changed examples are checked that they are generated correctly
- [x] Set task to `In Approval`

**Approval**

- [Storybook](https://storybook.onyx.schwarz/?path=/docs/support-toast--docs)
- [x] approved by designer @mj-hof 
- [x] approved by dev @BoppLi 

### Implementation Details

- uses the [popover api](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API) for the Toaster
- emits click event if `clickable` is `true`
- toast properties:
```ts
{ 
  title: string; 
  description?: string;
  color: 'neutral' | 'danger' | 'warning' | 'success';
  durationMs?: number /* defaults 5000 */;
  clickable?: boolean
  icon?: string }
```

### Reference implementations 

- https://ui.nuxt.com/components/notification#usage
- https://core-ui.schwarz/components/scu-toast-controller

### Aria Pattern

- [Alert](https://www.w3.org/WAI/ARIA/apg/patterns/alert/examples/alert/)

BoppLi:
will aprove once the icon-less variant is in Storybook 😊

PR #1361, opened 2024-06-19
feat: implement `OnyxToast`

larsrickert:
Relates to #1286, #1299

Implement single toast support component

## Checklist


packages/sit-onyx/src/components/OnyxToastProgressBar/OnyxToastProgressBar.vue
```diff
@@ -0,0 +1,53 @@
...
+  <time
+    class="onyx-toast-progress-bar"
+    role="progressbar"
+    :aria-valuemin="0"
+    :aria-valuemax="props.duration"
+    :aria-valuenow="percentage"
+    :aria-label="t('toast.progressBarLabel')"
+    :style="{ width: `${percentage}%` }"
+  ></time>
+</template>
```
JoCa96, line 41: Great implementation! Unfortunately, this seems to be an impractical pattern for accessibility devices.
Instead, the notification text should be announced via a live-region. The screen-reader device should not be aware of actual toast and it's duration.

packages/sit-onyx/src/components/OnyxToastProgressBar/OnyxToastProgressBar.vue
```diff
@@ -0,0 +1,53 @@
...
+  useAnimationFrame: true,
+});
+
+const percentage = computed(() => {
+  if (props.duration <= 0) return 0;
+  return Math.min(100, (timeLeft.value / props.duration) * 100);
+});
+
+watchEffect(() => isEnded.value && emit("timerEnded"));
+</script>
```
JoCa96, line 29: This can be done with an CSS Animation/Transition

JoCa96 (changes requested):

packages/sit-onyx/src/components/OnyxToast/types.ts
```diff
@@ -0,0 +1,34 @@
...
+  description?: string;
+  /**
+   * Toast color.
+   */
+  color?: Extract<OnyxColor, "neutral" | "danger" | "warning" | "success">;
+  /**
+   * Duration in milliseconds for the toast to close automatically.
+   * Timer will be paused when hovering the toast.
+   *
+   * Can be set to `0` to disable the auto closing.
```
JoCa96, line 21:
```suggestion
   * Can be set to `0` to disable the auto closing.
   *
   * @default 5000
```

packages/sit-onyx/src/components/OnyxToast/OnyxToast.vue
```diff
@@ -0,0 +1,253 @@
...
+            @click="emit('close')"
+          >
+            <OnyxIcon :icon="xSmall" />
+          </button>
+        </div>
+
+        <p
+          v-if="props.description"
+          class="onyx-toast__description onyx-text--small onyx-truncation-multiline"
+          tabindex="0"
```
JoCa96, line 78: Why does the description need to be tabbable? Only interactive elements should be tabbable.
```suggestion
```

packages/sit-onyx/src/components/OnyxToast/OnyxToast.vue
```diff
@@ -0,0 +1,253 @@
...
+
+    border-radius: var(--onyx-radius-md);
+    overflow: hidden;
+    box-shadow: var(--onyx-shadow-soft-bottom);
+    color: var(--onyx-toast-color);
+    background-color: var(--onyx-toast-background-color);
+    position: relative;
+    z-index: var(--onyx-z-index-notification);
+
+    &:hover {
```
JoCa96, line 141:
```suggestion
    &:hover,
    &:focus {
```

packages/sit-onyx/src/components/OnyxToast/OnyxToast.vue
```diff
@@ -0,0 +1,253 @@
...
+  return props.icon || DEFAULT_ICONS[props.color];
+});
+</script>
+
+<template>
+  <component
+    :is="props.clickable ? 'button' : 'div'"
+    class="onyx-toast"
+    :class="[`onyx-toast--${props.color}`, densityClass]"
+    role="alert"
```
JoCa96, line 54: Based on the color, we should switch the role: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions#roles_with_implicit_live_region_attributes

alert should only be used for important messages.

JoCa96 (changes requested):

packages/sit-onyx/src/components/OnyxToast/types.ts
larsrickert: The default value is already documented by Storybook, should I still add this?

image

packages/sit-onyx/src/components/OnyxToast/OnyxToast.vue
larsrickert: Because axe tests failed because the description gets scrollable after 3 lines and scrollable elements must be focussable

packages/sit-onyx/src/components/OnyxToast/OnyxToast.vue
larsrickert: I changed it to only have alert role for danger and warning color.
Should we use a special role for neutral and success colors?

packages/sit-onyx/src/components/OnyxToast/OnyxToast.vue
JoCa96: Good point! Then we should also pause the timer on focus-within and also show the focus-visible effect on focus-within

packages/sit-onyx/src/components/OnyxToast/OnyxToast.vue
JoCa96: I would use `log` or `status` otherwise

JoCa96 (dismissed):

JoCa96 (approved):

PR #1367, opened 2024-06-20
feat(OnyxNavItem): Implement OnyxNavItem

MajaZarkova:
Relates to #883 

Implemented new OnyxNavItem component which is similar to OnyxListItem, but will be used in OnyxNavButton only.

## Checklist


packages/sit-onyx/src/components/OnyxNavItem/future/OnyxNavItem.vue
```diff
@@ -0,0 +1,75 @@
+<script setup lang="ts">
```
BoppLi, line 1: question: is it intentional that the navItems do not yet emit any click event?

packages/sit-onyx/src/components/OnyxNavItem/future/OnyxNavItem.vue
```diff
@@ -0,0 +1,75 @@
...
+const props = defineProps<{ active?: boolean }>();
+</script>
+<template>
+  <li
```
BoppLi, line 5: should it really be a li?
in the DOM, we now have li inside li
image

packages/sit-onyx/src/components/OnyxNavItem/future/OnyxNavItem.vue
```diff
@@ -0,0 +1,75 @@
...
+    :class="{
+      'onyx-future-nav-item': true,
+      'onyx-future-nav-item--active': props.active,
+    }"
```
BoppLi, line 9:
```suggestion
    class="onyx-future-nav-item" 
    :class="{ 'onyx-future-nav-item--active': props.active }"
```

BoppLi (changes requested):

packages/sit-onyx/src/components/OnyxNavButton/OnyxNavButton.stories.ts
```diff
@@ -59,7 +59,10 @@ export const WithChildren: Story = {
   args: {
     ...Default.args,
     active: true,
-    children: () => nestedChildren.map(({ label }) => h(OnyxListItem, label)),
+    children: () =>
+      nestedChildren.map(({ label, active }) =>
+        h(OnyxNavItem, { "aria-selected": active }, [label]),
```
larsrickert, line 64: We should only set the aria attribute internally and create a "real" selected/active property :)

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxNavItem/future/OnyxNavItem.vue
larsrickert, line 5: Don't we want to use the list item here internally?

packages/sit-onyx/src/components/OnyxNavItem/future/OnyxNavItem.vue
MajaZarkova: added a click event

packages/sit-onyx/src/components/OnyxNavItem/future/OnyxNavItem.vue
MajaZarkova: I refactored the component and it's using the list item internally now. We don't have the classes any more :)

packages/sit-onyx/src/components/OnyxNavItem/future/OnyxNavItem.vue
MajaZarkova: I refactored the nav item and we are using the list item internally which means this is not the structure any more :)

packages/headless/src/composables/menuButton/createMenuButton.ts
```diff
@@ -3,14 +3,7 @@ import { createBuilder } from "../../utils/builder";
...
 
-export type CreateMenuButtonOptions = {
-  /**
-   * Called when a menu item is selected (via mouse or keyboard).
-   */
-  onSelect: (value: string) => void;
-};
-
-export const createMenuButton = createBuilder((options: CreateMenuButtonOptions) => {
+export const createMenuButton = createBuilder(() => {
```
JoCa96, line 6:
```suggestion
/**
 * Based on https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/examples/disclosure-navigation/
 */
export const createMenuButton = createBuilder(() => {
```

packages/headless/src/composables/menuButton/createMenuButton.ts
```diff
@@ -32,6 +25,68 @@ export const createMenuButton = createBuilder((options: CreateMenuButtonOptions)
...
   });
 
+  const getMenu = (el: HTMLElement) => {
+    const id = el.getAttribute("aria-controls") || "";
+    return document.getElementById(id);
+  };
+
+  const focusRelativeItem = (next: "next" | "prev" | "first" | "last") => {
+    const currentMenuItem = document.activeElement as HTMLElement;
+    const currentMenu = currentMenuItem?.closest('[role="menu"]') || getMenu(currentMenuItem);
```
JoCa96, line 35:
```suggestion
    // Either the current focus is on a "menuitem", then we can just get the parent menu.
    // Or the current focus is on the button, then we can get the connected menu using the menuId
    const currentMenu = currentMenuItem?.closest('[role="menu"]') || getMenu(currentMenuItem);
```

packages/headless/src/composables/menuButton/createMenuButton.ts
```diff
@@ -32,6 +25,68 @@ export const createMenuButton = createBuilder((options: CreateMenuButtonOptions)
...
+    if (!currentMenu) return;
```
JoCa96, line 36: We don't actually need to get the "aria-controls" value, as we already define the id ourselves:
```suggestion
  const focusRelativeItem = (next: "next" | "prev" | "first" | "last") => {
    const currentMenuItem = document.activeElement as HTMLElement;
    const currentMenu = currentMenuItem?.closest('[role="menu"]') || document.getElementById(menuId);
    if (!currentMenu) return;
```

packages/headless/src/composables/menuButton/createMenuButton.ts
```diff
@@ -3,14 +3,7 @@ import { createBuilder } from "../../utils/builder";
...
-   * Called when a menu item is selected (via mouse or keyboard).
-   */
-  onSelect: (value: string) => void;
-};
-
-export const createMenuButton = createBuilder((options: CreateMenuButtonOptions) => {
+export const createMenuButton = createBuilder(() => {
   const menuId = createId("menu");
   const buttonId = createId("menu-button");
   const isExpanded = ref<boolean>(false);
```
JoCa96, line 9: redundant generic:
```suggestion
  const isExpanded = ref(false);
```

packages/sit-onyx/src/components/OnyxNavItem/future/types.ts
```diff
@@ -0,0 +1,17 @@
...
+  href?: string;
+  active?: boolean;
+};
+
+/**
+ * [Vue injection key](https://vuejs.org/guide/components/provide-inject) that is provided by the nav bar
+ * to communicate child components whether they should render in mobile or desktop mode.
+ *
+ * @returns `true` if mobile, `false` otherwise
+ */
```
JoCa96, line 14: Please update the copied comment, to correctly reflect the purpose of this injection key

packages/sit-onyx/src/components/OnyxNavItem/future/types.ts
```diff
@@ -0,0 +1,17 @@
...
+import type { InjectionKey } from "vue";
+
+export type OnyxNavItemProps = {
+  href?: string;
+  active?: boolean;
+};
+
+/**
+ * [Vue injection key](https://vuejs.org/guide/components/provide-inject) that is provided by the flyout menu
+ * to provide child components with correct aria-patterns.
```
JoCa96, line 11:
```suggestion
 * to provide child components with correct aria attributes.
```

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxNavItem/future/OnyxNavItem.vue
```diff
@@ -0,0 +1,43 @@
...
+}>();
+
+const menuButton = inject(MENU_BUTTON_ITEM_INJECTION_KEY);
+</script>
+<template>
+  <OnyxListItem :selected="props.active" v-bind="menuButton?.listItem" class="onyx-future-nav-item">
+    <a
+      class="onyx-future-nav-item__anchor"
+      v-bind="menuButton?.menuItem({ active: props.active })"
+      :href="props.href ?? 'javascript:void(0)'"
```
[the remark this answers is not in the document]
MajaZarkova: If there is no href, the element is not focusable. This is why Jonathan added this

packages/sit-onyx/src/components/OnyxNavItem/future/types.ts
```diff
@@ -0,0 +1,17 @@
+import type { createMenuButton } from "@sit-onyx/headless";
+import type { InjectionKey } from "vue";
+
+export type OnyxNavItemProps = {
+  href?: string;
+  active?: boolean;
+};
```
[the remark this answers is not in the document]
MajaZarkova: sure

larsrickert (approved):

Issue #1355, opened 2024-06-19
Fix invalid state bugs


BoppLi:
### Related to
- #570 

### ACs

- [x] german: Short error messages are not translatet properly
- [x] hide the message when the error message is shown:
![Image](https://github.com/SchwarzIT/onyx/assets/151019977/8bb388c2-81a0-4ce2-8188-c10dcf851da8)

PR #1368, opened 2024-06-20
fix(OnyxFormElement): hide message while error is shown + prevent error overflow

BoppLi:
Relates to #1355 

- hide a standard message below form elements when an error is shown
- fix german translations for error previews
- add component tests

## Checklist

### DIFF
--- a/.changeset/curvy-hornets-jam.md
+++ b/.changeset/curvy-hornets-jam.md
@@ -1,0 +1,5 @@
+---
+"sit-onyx": patch
+---
+
+fix(OnyxFormElement): hide message while error is shown + prevent error overflow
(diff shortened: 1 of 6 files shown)


MajaZarkova (commented):
Label tooltip is little bit cut off:

image

MajaZarkova (changes requested):
I think I can still reproduce the bug

image

MajaZarkova (changes requested):
I think I can still reproduce the bug

image

BoppLi:
> I think I can still reproduce the bug
> 
> image

yes, the core component does not hide the error message, this can only be done in the "real" components which have access to an actual input/textarea where :user-invalid exists

MajaZarkova (approved):

Issue #1393, opened 2024-06-24
Fix OnyxTable in Storybook not working when deployed


JoCa96:
1. Go to https://storybook.onyx.schwarz/?path=/docs/data-table--docs

larsrickert:
Seems to affect all components where slots are used

PR #1395, opened 2024-06-24
docs: fix Storybook examples not working with slots

larsrickert:
Relates to #1393

This fixes the bug now now. I will try to investigate this further


JoCa96:
Any idea why this fixes the issue?

JoCa96 (approved):

larsrickert:
But the cause why Storybook does not show the examples is the following:

- Storybook uses Volar to extract the component slots so it knows which Story args are slots and which ones are props
- With Volar >= `2.3.1` it seems like the component meta can no longer be extracted at all so thats why all docs for props, events and slots are missing

PR #1399, opened 2024-06-24
feat: Implement CI Reporting for Playwright Tests

JoCa96:
- CI now generates HTML report, which can be downloaded and then shown using `npx playwright show-report`
- Merge blob-reports and screenshots for cleaner artifact overview


.github/workflows/playwright.yml
```diff
@@ -58,6 +58,17 @@ jobs:
...
 
+      - name: Upload Reports
+        uses: actions/upload-artifact@v4
+        if: ${{ !cancelled() }}
+        with:
+          name: blob-report-${{ matrix.shard }}
+          path: |
+            ./packages/**/blob-report
+            ./apps/**/blob-report
+            !./**/node_modules/**
```
larsrickert, line 69:
```suggestion
          path: |
            ./packages/*/blob-report
            ./apps/*/blob-report
```

Isn't this the same?

.github/workflows/playwright.yml
```diff
@@ -58,6 +58,17 @@ jobs:
           # prevent warnings if shard does not update any screenshots
           if-no-files-found: ignore
 
+      - name: Upload Reports
+        uses: actions/upload-artifact@v4
+        if: ${{ !cancelled() }}
```
larsrickert, line 63: Should we only upload the reports if the pipeline failed?

.github/workflows/playwright.yml
```diff
@@ -67,6 +78,56 @@ jobs:
...
+        run: find . -type d -name blob-report | xargs -n1 -I {} sh -c 'npx playwright merge-reports --reporter html {} && mv playwright-report {}/../playwright-report'
+
+      - name: Upload HTML report
+        uses: actions/upload-artifact@v4
+        with:
+          name: html-report--attempt-${{ github.run_attempt }}
+          path: |
+            ./packages/**/playwright-report
+            ./apps/**/playwright-report
+            !./**/node_modules/**
```
larsrickert, line 128:
```suggestion
            ./packages/*/playwright-report
            ./apps/*/playwright-report
```

packages/chartjs-plugin/playwright.config.ts
```diff
@@ -18,7 +18,8 @@ export default defineConfig({
   fullyParallel: true,
   forbidOnly: !!process.env.CI, // fail build on CI if we left test.only in the source code
   retries: process.env.CI ? 2 : 0, // retry on CI only
-  reporter: [["html", { open: "never" }]],
+  /* In the CI pipeline it generates github and blob reports, locally only a html report is generated */
+  reporter: process.env.CI ? [["github"], ["blob"]] : [["html", { open: "never" }]],
```
larsrickert, line 22: Do we need both blob and github or is one of them enough?

larsrickert (changes requested):

packages/chartjs-plugin/playwright.config.ts
larsrickert, line 22: the CI run is spammed with Playwright comments, is this caused by the github reporter?
image

.github/workflows/playwright.yml
JoCa96: Is there a benefit of not uploading the reports for successful runs?

larsrickert: I guess only saving disk space and keep the log minimal.
But I am also fine with keeping it like this

packages/chartjs-plugin/playwright.config.ts
JoCa96: Yes, I replaced it with the dot reporter, when running in the CI now.

larsrickert: I think you changed it for the headless package only, chartjs-plugin and sit-onyx still have the github reporter

JoCa96: done

larsrickert:
Is it possible to not show the error when no screenshots are updated?
image

larsrickert (approved):

JoCa96:
> Is it possible to not show the error when no screenshots are updated? image

No, AFAIK after 5 minutes of googling.

Issue #690, opened 2024-03-19
Implement useControlable composable pattern


JoCa96:
### Why?

We want to give the user the option to control the internal state of the components.
But this should be made possible without overloading the developer with too much delegated state.
We propose the following pattern: 
- `modelValue` state must always be managed by the developer
- other `v-model`s are controlled internally if their value is undefined and otherwise controlled by the developer
- to make this easier to implement for us maintainers and have a consistent pattern we implement a composable `useControllable`

### Acceptance criteria

- [x] `useControllable` composable is implemented
 - [x] has generic type-support

### Definition of Done

- [x] covered by component tests
- [x] Set task to `In Approval`

**Approval**

- [ ] approved by dev _add dev tag_

### Implementation details

```typescript
// Spike
export const useControllable = <
  // eslint-disable-next-line @typescript-eslint/no-explicit-any
  Props extends Record<string, any>,
  Key extends keyof Props,
  TValue extends Props[Key],
  Return extends Exclude<TValue, undefined>,
>(
  props: Props,
  key: Key,
  defaultValue: TValue,  // Open Bug: if `Props[Key]` is `string | number` then `Return` is only based on `defaultValue`, e.g. only `string` but it must still be  `string | number`
  emit: (value: TValue) => void,
): Ref<Return> => {
  const propRef = toRef(() => props[key]);
  const isControlled = computed(() => propRef.value !== undefined);
  const internalState = ref(defaultValue);

  watch(propRef, (newValue) => (internalState.value = newValue));

  return computed({
    get: () => (isControlled.value ? propRef.value : internalState.value),
    set: (value) => {
      internalState.value = value;
      emit(value);
    },
  });
};
```

PR #1405, opened 2024-06-25
feat(OnyxSelect): optionally manage searchTerm and open internally

JoCa96:
Relates to #690 

- feat(OnyxSelect): optionally manage searchTerm and open internally
- fix(OnyxSelect): fix bug where OnyxSelect closed when clicking the clear button

- implemented `useManagedState` composable, which allows to optionally manage State internally.


packages/sit-onyx/src/components/OnyxSelect/OnyxSelect.stories.ts
```diff
@@ -38,20 +38,22 @@ const meta: Meta<typeof OnyxSelect> = {
       /**
        * Decorator that simulates the load more functionality so we can show it in the stories.
        */
-      (story, ctx) => ({
-        components: { story },
-        setup: () => {
-          const { isLazyLoading, handleLoadMore, options } = useLazyLoading(ctx.args.options);
+      (story, ctx) => {
```
larsrickert, line 41: The search example does no longer work

packages/sit-onyx/src/components/OnyxSelect/OnyxSelect.vue
```diff
@@ -29,13 +32,26 @@ const emit = defineEmits<{
...
    * Emitted if lazy loading is triggered / the users scrolls to the end of the options.
    * See property `lazyLoading` for enabling the lazy loading.
    */
   lazyLoad: [];
 }>();
 
+const searchTerm = useManagedState(
+  toRef(() => props.searchTerm),
+  "",
+  (v) => emit("update:searchTerm", v ?? ""),
```
larsrickert, line 46: The default value seems a little redudant here, maybe we can to this in the `useManagedState` internally?

Because it is passed as second parameter here and also for the `?? ""`

packages/sit-onyx/src/components/OnyxSelect/OnyxSelect.vue
```diff
@@ -29,13 +32,26 @@ const emit = defineEmits<{
...
+  "update:open": [open: boolean];
   /**
    * Emitted if lazy loading is triggered / the users scrolls to the end of the options.
    * See property `lazyLoading` for enabling the lazy loading.
    */
   lazyLoad: [];
 }>();
 
+const searchTerm = useManagedState(
+  toRef(() => props.searchTerm),
```
larsrickert, line 44: Is there a reason why toRef instead of computed?

packages/sit-onyx/src/components/OnyxMiniSearch/OnyxMiniSearch.vue
```diff
@@ -61,7 +61,8 @@ defineExpose({
       class="onyx-mini-search__clear"
       :aria-label="t('select.clearSearch')"
       tabindex="-1"
-      @mousedown.prevent="emit('clear')"
+      @mousedown.prevent
+      @click="emit('clear')"
```
larsrickert, line 65: Isn't this the same as before?

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxSelect/OnyxSelect.vue
```diff
@@ -29,13 +32,26 @@ const emit = defineEmits<{
...
   "update:searchTerm": [searchTerm: string];
+  "update:open": [open: boolean];
   /**
    * Emitted if lazy loading is triggered / the users scrolls to the end of the options.
    * See property `lazyLoading` for enabling the lazy loading.
    */
   lazyLoad: [];
 }>();
 
+const searchTerm = useManagedState(
```
larsrickert, line 43: Managing the search term internally would only make sense if we also implement the filter logic, right?

packages/sit-onyx/src/components/OnyxMiniSearch/OnyxMiniSearch.vue
JoCa96: This is the bug fix for `fix(OnyxSelect): fix bug where OnyxSelect closed when clicking the clear button`
I added a comment explaining the issue.

packages/sit-onyx/src/components/OnyxSelect/OnyxSelect.stories.ts
JoCa96: Works for me, can you extend on this? What is the issue?

packages/sit-onyx/src/components/OnyxSelect/OnyxSelect.vue
JoCa96: I wouldn't think so. It should not have any additional side-effects in my opinion

packages/sit-onyx/src/components/OnyxSelect/OnyxSelect.stories.ts
larsrickert: https://github.com/SchwarzIT/onyx/assets/67898185/9a6a943f-2e5c-4e8e-af3b-0d53d5a4de59

Also when re-opening the select, the focus is not set on the search inut

packages/sit-onyx/src/components/OnyxSelect/OnyxSelect.vue
larsrickert: Sure it has no side effects but imho if its not used anywhere we don't need to add it :D
Otherwise its a little useless / unnecessary complexity

packages/sit-onyx/src/components/OnyxSelect/OnyxSelect.vue
JoCa96: for this use-case, there doesn't seem to be a funtional difference between the two.
Looking at the respective implementations `toRef` seems to have less overhead.
Also it seems to be the recommended way:
`toRef() is useful when you want to pass the ref of a prop to a composable function` (https://vuejs.org/api/reactivity-utilities.html#toref)

packages/sit-onyx/src/components/OnyxSelect/OnyxSelect.vue
JoCa96: The fallback value here is redundant, I removed it.

packages/sit-onyx/src/components/OnyxSelect/OnyxSelect.vue
JoCa96: Well, I would say it still has the benefit of of the user not needing to care about caching the searchterm:
```vue
<OnyxSelect @update:searchTerm="val => options = all.filter(...)"
```
This way the component still works as expected.

packages/sit-onyx/src/components/OnyxSelect/OnyxSelect.stories.ts
JoCa96: Cannot reproduce this 🤔 Maybe it's because I forgot to push?

JoCa96: The other issue is an existing issue

larsrickert: I still have the issue but only when using it on the "Docs" page. The single Story works fine

packages/sit-onyx/src/components/OnyxSelect/OnyxSelect.vue
```diff
@@ -452,3 +467,4 @@ const selectInputProps = computed(() => {
   }
 }
 </style>
+../../composables/useManagedState
```
larsrickert, line 470: Unused import :)

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxSelect/OnyxSelect.vue
JoCa96:
```suggestion
```

larsrickert (approved):

PR #1425, opened 2024-06-26
fix(OnyxPageLayout): stabilize layout structure

### LINKED ISSUE #484: Define stabilize slot switching in the alpha-test-app layout demo
Origin: https://github.com/SchwarzIT/onyx/pull/470/files#r1512515713

In `apps/alpha-test-app/src/views/LayoutDemoView.vue`, find a way to stabilize the breaking layout on frequent config switching without using key:
Image

### Steps to reproduce the issue:
1. remove the `:key` setting displayed above.
2. start the alpha-test-app
3. open http://localhost:5173/layout-demo
4. select "Sidebar Options" > "Overlay Sidebar"
5. close the overlay sidebar via close button
6. select "Footer Options" > "Full Footer"
7. select "Sidebar Options" > "Fixed Sidebar"

Now the whole layout is broken. 
Investigation showed that the computed `pageModifier` in `packages/sit-onyx/src/components/OnyxPageLayout/OnyxPageLayout.vue` did not fire when the sidebar slot was removed and re-added so the selected class is still `"onyx-page--main-footer""` even though there is a side bar.

### Implementation Hints

Idea: instead of hiding the slot parents with v-if, just let them be empty? maybe the grid itself can handle it already (the column/row use max-content anyways).

### Review

- __playwright screenshot__
- [ ] approved by designer @jannick-ux
- [ ] approved by dev __todo__

BoppLi:
Closes #484

Stabilizes OnyxPageLayout by using v-show instead of v-if.

### DIFF
--- a/packages/sit-onyx/src/components/OnyxPageLayout/OnyxPageLayout.vue
+++ b/packages/sit-onyx/src/components/OnyxPageLayout/OnyxPageLayout.vue
@@ -4,7 +4,7 @@ import type { OnyxPageLayoutProps } from "./types";
 
 const props = defineProps<OnyxPageLayoutProps>();
 
-const slots = defineSlots<{
+defineSlots<{
   /** Main content area of the page */
   default(): unknown;
   /** Optional sidebar of the page */
@@ -17,14 +17,10 @@ const slots = defineSlots<{
 
 const pageModifier = computed(() => {
   let mode = "";
-  if (!slots.footer && slots.sidebar) mode = "onyx-page--side-main";
-  if (slots.footer && (!slots.sidebar || props.hideSidebar)) mode = "onyx-page--main-footer";
-  if (slots.footer && slots.sidebar) {
-    if (props.footerAsideSidebar) {
-      mode = "onyx-page--side-main-footer-partial";
-    } else {
-      mode = "onyx-page--side-main-footer-full";
-    }
+  if (props.footerAsideSidebar) {
+    mode = "onyx-page--footer-partial";
+  } else {
+    mode = "onyx-page--footer-full";
   }
   return mode;
 });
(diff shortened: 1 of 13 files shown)


packages/sit-onyx/src/components/OnyxPageLayout/OnyxPageLayout.vue
```diff
@@ -17,31 +17,27 @@ const slots = defineSlots<{
...
     </aside>
     <main class="onyx-page__main">
       <slot></slot>
     </main>
-    <footer v-if="slots.footer" class="onyx-page__footer">
+    <footer v-show="slots.footer" class="onyx-page__footer">
       <slot name="footer"></slot>
     </footer>
-    <div v-if="slots.toasts" class="onyx-page__toasts">
+    <div v-show="slots.toasts" class="onyx-page__toasts">
```
larsrickert, line 40: Toast will be displayed fixed using the popover API, right so I think this slot is obsolete :)

packages/sit-onyx/src/components/OnyxPageLayout/OnyxPageLayout.vue
```diff
@@ -17,31 +17,27 @@ const slots = defineSlots<{
...
+    mode = "onyx-page--footer-full";
   }
   return mode;
 });
 </script>
 
 <template>
   <div class="onyx-page" :class="pageModifier">
-    <aside v-if="slots.sidebar && !props.hideSidebar" class="onyx-page__sidebar">
+    <aside v-show="slots.sidebar && !props.hideSidebar" class="onyx-page__sidebar">
```
larsrickert, line 31: I would prefer to not render the sidebar/footer if its not set, otherwise it might be weird if it shows in the dev tools although it is not needed.

Can you extend why the slot switching is not stable in the first place? Slots and properties are reactive so I assume the issue is related to the application, not onyx itself

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxPageLayout/OnyxPageLayout.vue
BoppLi: I thought about that but as soon as we get the bottom bar we might need to rethink the approach?

packages/sit-onyx/src/components/OnyxPageLayout/OnyxPageLayout.vue
BoppLi: The computed in line 18 `pageModifier` did not get updated with the described steps to reproduce.

But actually, with the change that it now only depends on 1 prop, it's stable already - we can keep v-if 🎉

larsrickert (approved):

PR #1430, opened 2024-06-27
feat: implement `OnyxToastProvider` and `useToast`

larsrickert:
Relates to #1286

- Implement `OnyxToastProvider` component and `useToast` composable for showing toasts in the application
- also fix toast width after clarification with Jannick. Toasts are always 40rem wide, independent of how large the content is. They only shrink when the screen size gets smaller than 40rem

## Checklist


packages/sit-onyx/src/components/OnyxToastProvider/OnyxToastProvider.stories.ts
```diff
@@ -0,0 +1,86 @@
...
+
+const codeTemplate = `
+  <div>
+    <OnyxButton label="Show toast" @click="showExampleToast" />
+    <OnyxToastProvider />
+  </div>
+`;
+
+/**
+ * Toast provider that is used to display toasts inside the application.
```
BoppLi, line 15: A thought from when I looked into approving the Toast ticket:


Could you also add such a top description in 
https://storybook.onyx.schwarz/?path=/docs/support-toast--docs
to point out that it is recommended to be used in the ToastProvider?

JoCa96 (changes requested):

JoCa96 (approved):

Issue #1415, opened 2024-06-26
Implement OnyxMenuItem


MajaZarkova:
### Why?

A follow up ticket on #833

### Acceptance criteria

- [ ] keyboard navigation is implemeneted

### Definition of Done

- The following component requirements are implemented:
- [ ] covered by component tests
- [ ] screenshot tests are updated
- [ ] follow-up tickets were created if necessary
- [ ] Storybook can show the feature
- [ ] Storybook code snippet of new/changed examples are checked that they are generated correctly
- [ ] Set task to `In Approval`

**Approval**

- Storybook 
- [ ] approved by designer @jannick-ux
- [ ] approved by dev 

### Implementation details
<OnyxMenuItem>
    <OnyxListItem
        :selected="props.active"
        v-bind="menuButton?.listItem"
        class="onyx-future-nav-item"
        @click="emit('click')"
      >
        <a
          v-if="props.href"
          class="onyx-future-nav-item__anchor"
          v-bind="menuButton?.menuItem({ active: props.active })"
          :href="props.href ?? 'javascript:void(0)'"
        >
          <slot></slot>
        </a>
       <button
          v-else
          v-bind="menuButton?.menuItem({ active: props.active })"
       >
       </button
  </OnyxListItem>
</OnyxMenuItem>

PR #1434, opened 2024-06-27
Implement OnyxMenuItem

MajaZarkova:
Relates to #1415 

This is a new component that is intended to be used inside OnyxUserMenu

## Checklist


packages/sit-onyx/playwright/snapshots/components/OnyxUserMenu/future/User-menu-chromium-linux.png
BoppLi, line 1: do we really need hover + focus-visible in the screenshot even though they are identical?

BoppLi (changes requested):
maybe add a changeset?

packages/sit-onyx/src/components/OnyxMenuItem/OnyxMenuItem.vue
```diff
@@ -0,0 +1,70 @@
+<script setup lang="ts">
```
JoCa96, line 1: Per my understanding, the `OnyxMenuItem` can do everything that the `OnyxNavItem` can, do, correct?

Can we use the `OnyxNavItem` as a thin wrapper around `OnyxNavItem`, that just restricts the props a bit?
As far as I remember, the only difference was that the `danger` color should not be allowed in the OnyxNavItem?

packages/sit-onyx/playwright/snapshots/components/OnyxUserMenu/future/User-menu-chromium-linux.png
larsrickert: I'd say yes because like this we ensure that the flyout is open both on hover and via keyboard :) If we one of them out we are not sure that they both work correctly

packages/sit-onyx/src/components/OnyxMenuItem/OnyxMenuItem.vue
```diff
@@ -0,0 +1,70 @@
...
+        'onyx-list-item--active': props.active,
+        [`onyx-list-item--${props.color}`]: props.color,
+        'onyx-list-item--disabled': props.disabled,
+      }"
+      v-bind="menuButton?.menuItem({ active: props.active })"
+      @click="emit('click')"
+    >
+      <slot></slot>
+    </button>
+  </OnyxListItem>
```
JoCa96, line 47: We can group this logic together:
```suggestion
    <component
      is="props.href ? 'a' : 'button'"
      :class="{
        'onyx-menu-item': true,
         'onyx-menu-item__anchor': props.href,
         'onyx-menu-item__button': !props.href,
        'onyx-list-item--active': props.active,
        [`onyx-list-item--${props.color}`]: props.color,
        'onyx-list-item--disabled': props.disabled,
      }"
      v-bind="menuButton?.menuItem({ active: props.active })"
      :href="props.href"
      @click="emit('click')"
    >
      <slot></slot>
    </component>
  </OnyxListItem>
```

packages/sit-onyx/src/components/OnyxUserMenu/future/OnyxUserMenu.ct.tsx
```diff
@@ -0,0 +1,71 @@
+import type { Locator } from "@playwright/test";
```
JoCa96, line 1: We need some tests for the keyboard interactions

packages/sit-onyx/playwright/snapshots/components/OnyxUserMenu/future/User-menu-chromium-linux.png
BoppLi: maybe use new roww rather to make one of the MenuItems active / hovered?

packages/sit-onyx/src/components/OnyxMenuItem/types.ts
```diff
@@ -0,0 +1,20 @@
+import type { OnyxColor } from "../../types";
+
+export type OnyxMenuItemProps = {
+  /**
+   * URL that the menu item points to.
```
JoCa96, line 5: We should also explain, that this property will switch between an anchor and a button element.

packages/sit-onyx/playwright/snapshots/components/OnyxUserMenu/future/User-menu-chromium-linux.png
BoppLi: ok 👍

packages/sit-onyx/src/components/OnyxMenuItem/types.ts
```diff
@@ -0,0 +1,20 @@
...
+   */
+  href?: string;
+  /**
+   * Whether the menu item is (visually) active.
+   */
```
JoCa96, line 10:
```suggestion
   * Whether the menu item is currently active.
   * This is the case if the given link is currently opened.
   */
```

packages/sit-onyx/src/components/OnyxMenuItem/OnyxMenuItem.vue
```diff
@@ -0,0 +1,70 @@
...
+      class="onyx-future-nav-item__anchor"
+      v-bind="menuButton?.menuItem({ active: props.active })"
+      :href="props.href"
+      @click="emit('click')"
+    >
+      <slot></slot>
+    </a>
+    <button
+      v-else
+      class="onyx-menu-item__button onyx-list-item"
```
JoCa96, line 36: For the `disabled` state we should set disabled="true" if it's a button and the aria-disabled="true" on the menuitem: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menuitem_role#description

JoCa96 (changes requested):

packages/sit-onyx/src/components/OnyxMenuItem/OnyxMenuItem.vue
MajaZarkova: Yes you are right. The only difference is the color. Did you mean to use the OnyxMenuItem as a thin wrapper around OnuxNavItem?

JoCa96: The `OnyxMenuItem` is more powerful. When we wrap it inside the `OnyxNavItem`, we can limit the props to what we want :)

packages/sit-onyx/src/components/OnyxMenuItem/OnyxMenuItem.vue
```diff
@@ -0,0 +1,67 @@
...
+    :disabled="props.disabled"
+    class="onyx-menu-item"
+  >
+    <component
+      :is="props.href ? 'a' : 'button'"
+      :class="{
+        'onyx-list-item': !props.href,
+        'onyx-menu-item__anchor': props.href,
+        'onyx-menu-item__button': !props.href,
+        [`onyx-list-item--${props.color}`]: !props.href && props.color,
```
JoCa96, line 32: same as above
```suggestion
```

packages/sit-onyx/src/components/OnyxMenuItem/OnyxMenuItem.vue
```diff
@@ -0,0 +1,67 @@
...
+    :active="props.active"
+    v-bind="menuButton?.listItem"
+    :color="props.color"
+    :disabled="props.disabled"
+    class="onyx-menu-item"
+  >
+    <component
+      :is="props.href ? 'a' : 'button'"
+      :class="{
+        'onyx-list-item': !props.href,
```
JoCa96, line 29: We must use bem classes of the component itself, the bem class of another component must not be used. Otherwise we will have unexpected side-effects in this component when editing the `OnyxListItem`.
```suggestion
```

JoCa96: As a trick you could apply `display: contents` to the `anchor` and `button` element.

JoCa96 (changes requested):

packages/sit-onyx/src/components/OnyxMenuItem/OnyxMenuItem.vue
MajaZarkova: display: contents messes up our keyboard navigation

JoCa96: Then we need to override the style manually

JoCa96 (approved):

Issue #1240, opened 2024-06-04
Implement Table empty + focus outline


BoppLi:
### Depends on 

- #1225

### Why?

Provide missing features for OnyxTable.

### Design

[Figma table definitions](https://www.figma.com/design/YfEUBOHk4J4nYrk04geswG/Onyx-Component-Library?node-id=5129-26998&t=X6MwF6EN8gsHzc0G-4)

![image](https://github.com/SchwarzIT/onyx/assets/151019977/31e31b22-0009-4a00-a2ea-a54236ef8ee4)

### Out of scope

- #1421

### Acceptance criteria

- [x] offer empty state
- [x] implement focus style when accessing the table with tab

### Definition of Done

- [x] covered by component tests
- [x] screenshot tests are updated
- [x] follow-up tickets were created if necessary
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature
- [x] Storybook code snippet of new/changed examples are checked that they are generated correctly
- [x] Set task to `In Approval`

**Approval**

- [Storybook Empty Table](https://storybook.onyx.schwarz/?path=/story/data-table--empty)
- [Storybook focused table](https://storybook.onyx.schwarz/?path=/story/data-table--default) -> access table with tab.
- [x] approved by designer @jannick-ux
- [x] approved by dev @larsrickert 

### Implementation details

- Use the exisiting empty component
- prefer CSS implementation, e.g. with `&:has(tbody tr)` selector

mj-hof:
@jannick-ux Open question
- Reconsider skeleton for simple table to be replace by a lading indicator style?

jannick-ux:
loading indicator and skeleton state have two different purposes. Skeleton is for building up the component. Loading indicator is for processing an input. Thats the reason why we commited to skip the loading indicator for the simple table.

PR #1437, opened 2024-06-27
feat(OnyxTable): implement empty state + define focus outline

BoppLi:
Relates to #1240

- refactor(OnyxTable): split default slot to distinguish thead and tbody
- feat(OnyxTable): implement empty state + define focus outline

## Checklist


packages/sit-onyx/src/components/OnyxTable/OnyxTable.stories.ts
```diff
@@ -112,3 +111,21 @@ export const LimitedHeightAndWidth = {
...
+export const Empty = {
+  args: {
+    default: () => [getTableHeader()],
+  },
+} satisfies Story;
+
+/**
+ * This example shows a table with an empty `tbody` and without a `thead`.
+ */
+export const EmptyWithoutHeader = {
```
larsrickert, line 127: Not sure if we need to create an example for this since usually the most tables will have headers :)

packages/sit-onyx/src/components/OnyxTable/OnyxTable.vue
```diff
@@ -1,20 +1,44 @@
...
+const { t } = injectI18n();
+
 const { densityClass } = useDensity(props);
+
+/**
+ * Returns an empty message if no table body was found
+ * or no rows were found inside a table body.
+ */
+const isEmptyMessage = computed(() => {
+  const tableBody = slots.default?.().find((data) => data.type === "tbody");
```
larsrickert, line 38: I am not sure if we should do this since it can be flaky / have edge cases.
Jonathan mentioned that he doesn't like to iterate over slots.

We could implement it via plain CSS and just set the empty to `display: none`

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxTable/OnyxTable.vue
BoppLi: > if we should do this

what do you mean -> 
1) look into our slot to find out if we should show an empty text
2) use v-if on the onyx-table__empty tbody?

I don't see the connection of your first paragraph and the last sentence ^^

BoppLi: if we aren't allowed to look into the slots.default, this means either

1) we are forced to split the default slot into a "thead" and a "tbody" slot
or
2) we need to add a prop to determine if we should show "empty", and ignore what's in the slot

larsrickert: Plain CSS solution could look something like this:

```css
.onyx-table {
    &__empty {
      display: none;
    }

    &:not(:has(tbody tr)) {
      .onyx-table__empty {
        display: block;
      }
    }
}
```

larsrickert: > > if we should do this
> 
> what do you mean ->
> 
> 1. look into our slot to find out if we should show an empty text
> 2. use v-if on the onyx-table__empty tbody?
> 
> I don't see the connection of your first paragraph and the last sentence ^^

I meant I am not sure if we should look into the slot content and search for a tbody element

larsrickert: Example why this can be flaky:

- we can not force the project to pass in a `tbody`. You can also just pass `<tr>` which is valid
- HTML will insert a tbody automatically when its rendered
- so if we filter the slot content, there will be no tbody element, although there is table data available
- when checking for `:has(tbody tr)` with CSS to check if the table is empty, this will be always correct because since HTML will add a `tbody` if you don't provide one

BoppLi: ok, got it 👍

BoppLi: I'll have to refactor that onyx-table__empty itself isn't a tbody then. ^^ else the condition never is fulfilled

larsrickert: > I'll have to refactor that onyx-table__empty itself isn't a tbody then. ^^ else the condition never is fulfilled

Either so or we could change the selector to:

```css
&:not(:has(tbody tr:not(:has(.onyx-table__empty))))
```

But thats a lot of :not and :has :D

packages/sit-onyx/src/components/OnyxTable/OnyxTable.stories.ts
BoppLi: ok :) it's still covered by playwright

packages/sit-onyx/src/components/OnyxTable/OnyxTable.vue
```diff
@@ -29,6 +45,25 @@ const { densityClass } = useDensity(props);
         ]"
       >
         <slot></slot>
+
+        <!-- info: we can't use tbody here because we determine via CSS
+         whether to show "empty" when no tbody is present -->
+        <div class="onyx-table__empty">
```
larsrickert, line 51: I think a <div> is not valid HTML for a table since a table
image

packages/sit-onyx/src/components/OnyxTable/OnyxTable.vue
```diff
@@ -159,6 +198,24 @@ $border: var(--onyx-1px-in-rem) solid var(--onyx-color-base-neutral-300);
...
+
+      &-content {
+        display: flex;
+        justify-content: center;
+      }
+      .onyx-icon {
+        color: var(--onyx-color-text-icons-neutral-medium);
+      }
+    }
+    &:not(:has(tbody tr)) {
```
larsrickert, line 212: This does not work if the project only provides <tr> without <tbody>

Maybe we should check for `<td>` instead?

image

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxTable/OnyxTable.vue
```diff
@@ -159,6 +205,16 @@ $border: var(--onyx-1px-in-rem) solid var(--onyx-color-base-neutral-300);
...
     width: 100%;
 
+    &__empty {
+      &-content {
+        display: flex;
+        justify-content: center;
+      }
+      .onyx-icon {
+        color: var(--onyx-color-text-icons-neutral-medium);
+      }
```
larsrickert, line 215: Shouldn't we stick to the default `color="neutral"` property of the icon to be uniform here? Also has better contrast in my opinion :)

larsrickert (changes requested):

apps/alpha-test-app/src/views/HomeView.vue
```diff
@@ -162,7 +162,7 @@ const toast = useToast();
 
     <div class="page" :class="[`onyx-density-${activeDensityOption}`]">
       <section class="page__intro">
-        <OnyxHeadline is="h1">Component usages</OnyxHeadline>
+        <OnyxHeadline is="h1">Component usages 2</OnyxHeadline>
```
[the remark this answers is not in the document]
BoppLi: yep... wasn't trusting the hot reload, used this to see it works ^^'

packages/sit-onyx/src/components/OnyxTable/OnyxTable.vue
BoppLi: needs to be answered by @jannick-ux , I was sticking to his design.

BoppLi: I asked for feedback in [our teams channel](https://teams.microsoft.com/l/message/19:8500e48ce8d14968a18d1bd2d0dd293e@thread.tacv2/1719592845329?tenantId=d04f4717-5a6e-4b98-b3f9-6918e0385f4c&groupId=969ca3b3-9c78-411f-8629-eb770bbfdc7b&parentMessageId=1719592845329&teamName=onyx&channelName=Design&createdTime=1719592845329)

BoppLi: martin says to use the default color. I'll update the screenshots.

larsrickert (approved):

PR #1447, opened 2024-06-28
style: sort imports with prettier

### LINKED ISSUE #519: Add prettier config to sort imports
We should enable prettier to sort imports so we have less git changes / merge conflicts

larsrickert:
closes #519

- Add prettier plugin to sort imports (in the same way VSCode does it).


JoCa96 (commented):
Is there a benefit over using the default eslint rule https://eslint.org/docs/latest/rules/sort-imports ?
The eslint rule also seems to work with `.vue` files

larsrickert:
> Is there a benefit over using the default eslint rule https://eslint.org/docs/latest/rules/sort-imports ? The eslint rule also seems to work with `.vue` files

Yes because eslint use custom rules to sort the imports and you can not configure it to sort exactly like VSCode / Typescript language server which causes conflicts if you have enabled VS Code to also sort imports.

`prettier-plugin-organize-imports` uses the`organizeImports` feature of the TypeScript language service API which is 1:1 what VS Code uses under the hood

JoCa96 (approved):

PR #1455, opened 2024-06-28
fix(OnyxSelect): do not submit form when pressing Enter

### LINKED ISSUE #1431: Fix premature form submission while checking options in OnyxSelect
### Steps to reproduce

1. open [this playground example](https://playground.onyx.schwarz/#eNp9lG1v2jAQx7+KZU0qlSBMWl8xqLZOvNg0rdWY9gaj1SQHuHNszw8pFeK77+w8ENqqQkLx3e8ud/8750A/G5NVAeiETl1uhfFEcrWdMeodo8SBD+aaKVEabT05WNgcycbqklxg0MXHk4fcqqf9AiTknjQIo074kUY7o0gylWvlPNlxVUhYhHUpPJmRwSWZXRMuwfrBfWN1Ic/BuU2QE/Lu4FLW31wGyKr4v3y/yiRfgzzeX/YS19yt8QKPmHnJFMHCUsgEq+HGSGB0iB1ibLRg89FCjsNn6Jor/J2xN43pJey85Y9rsPbpLGDRM6eg1YtaU09YKco6WK5iL9NxPQXUHA8eSiO5BzwRMi1ERXLJncPpRFVHWyuKEebzXCiwjCYMwY22ZUcWUGrS4VjKJ5dEzoyFCpRHpD8RBJo0mOg009ZEXqnAGa5GV4yemGpU6gIkUr0++0ASCd3zPccOoZHjjBDOj1rsOx7qmD5SBulFHGBnmeh6+N2Lm2XoR6GgOey0LMAidnc6ZVnW5yz8C8JC0VrGrbrjKG89kTGOBJ+m425QdIgXB0eyEdvswWmF9+oQUUZzXRqBW96VNCHJE31cSv34Ldm8DZAWLMXsIP/7iv3B4ZWKO3ZnwYGt4la3Ps/tFlDK6J4vfsAenzsnjiXEjX/D+ROcliHWWGM3IS4HblfHpWq/pnsv1PaXm+89KNc2FQuNZH1NGMXvxJc3Wj+V+yG7SnFMHVHFPxXYmBMFjMI6T4//AWvJipw=)
2. select a fruit via keyboard (pressing `Enter`)

=> now the Bug appears: The form is submitted, even though the user might want to select more than only one option with the keyboard. And the value is not even displayed yet in the OnyxSelectInput while the form is already sent.

Image

larsrickert:
closes #1431

- do not submit form when pressing Enter

## Checklist

### DIFF
--- a/packages/headless/src/composables/comboBox/createComboBox.ts
+++ b/packages/headless/src/composables/comboBox/createComboBox.ts
@@ -173,6 +173,11 @@ export const createComboBox = createBuilder(
     };
 
     const handleKeydown = (event: KeyboardEvent) => {
+      if (event.key === "Enter") {
+        // prevent submitting the form when pressing enter
+        event.preventDefault();
+      }
+
       if (!isExpanded.value && isKeyOfGroup(event, OPENING_KEYS)) {
         onToggle?.();
         if (event.key === " ") {
--- a/packages/headless/src/composables/comboBox/createComboBox.ts
+++ b/packages/headless/src/composables/comboBox/createComboBox.ts
@@ -174,7 +174,7 @@ export const createComboBox = createBuilder(
 
     const handleKeydown = (event: KeyboardEvent) => {
       if (event.key === "Enter") {
-        // prevent submitting the form when pressing enter
+        // prevent submitting on pressing enter when the combo box is used inside a <form>
         event.preventDefault();
       }
 
--- a/.changeset/moody-kiwis-swim.md
+++ b/.changeset/moody-kiwis-swim.md
@@ -1,0 +1,5 @@
+---
+"sit-onyx": patch
+---
+
+fix(OnyxSelect): do not submit form when pressing Enter
--- a/.changeset/rotten-apricots-battle.md
+++ b/.changeset/rotten-apricots-battle.md
@@ -1,0 +1,5 @@
+---
+"@sit-onyx/headless": patch
+---
+
+fix(createComboBox): prevent submitting form when pressing enter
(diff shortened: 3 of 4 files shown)


JoCa96 (approved):

packages/headless/src/composables/comboBox/createComboBox.ts
```diff
@@ -173,6 +173,11 @@ export const createComboBox = createBuilder(
     };
 
     const handleKeydown = (event: KeyboardEvent) => {
+      if (event.key === "Enter") {
+        // prevent submitting the form when pressing enter
+        event.preventDefault();
+      }
```
BoppLi, line 179: maybe we should only do this while the flyout is open?

because if you hit enter when it's closed, it might be fine to actually send a form.

also
```suggestion
      if (event.key === "Enter") {
        // prevent submitting on pressing enter when the combo box is inside a form
        event.preventDefault();
      }
```

larsrickert: The select has to be focussed in order for the enter key event to fire.
But if the select is focussed, enter should always either open the flyout or select a value if I understood it correctly, right @JoCa96 ?

BoppLi: ah, if enter also opens the flyout then yes, it should always be prevented. thx!

BoppLi: yep 👍
image

BoppLi (approved):
thank you!

PR #1462, opened 2024-07-01
fix(nuxt): require locale mapping for i18n integration

markbrockhoff:
This PR fixes an issue with the registration order warning of the nuxt module.
It also requires the user to specify a mapping for onyx locales to the locales from nuxt-i18n specific to their project.
This fixes a bug where all projects would display all locales of onyx as supported languages even if they were not specified within the project.
An issue and PR were already created for nuxt-i18n to support loading locales from npm packages. (https://github.com/nuxt-modules/i18n/issues/2999)
This way it would be possible to decouple onyx and nuxt-i18n while preserving easy support and customizability. This PR should make the onyx module useable until the feature is available inside nuxt-i18n.

## Checklist

### DIFF
--- a/.changeset/old-dogs-sin.md
+++ b/.changeset/old-dogs-sin.md
@@ -1,0 +1,8 @@
+---
+"@sit-onyx/nuxt": major
+"docs": patch
+---
+
+Require mapping of onyx locales to the project ones.
+
+This change was necessary because registering all languages supported by onyx would force the project to also support them due to the way locales are merged by nuxt-i18n.
--- a/apps/docs/src/development/packages/nuxt.md
+++ b/apps/docs/src/development/packages/nuxt.md
@@ -45,11 +45,31 @@ If your Nuxt project uses both modules, onyx will automatically detect it and us
 
 ### Setup
 
-The setup is automated, @sit-onyx/nuxt will check if @nuxtjs/i18n is registered as well and register all onyx translations and add a plugin to handle the translations for onyx using vue-i18n.
+To merge the locales provided by onyx with your own, you need to define a mapping.
+
+```ts [nuxt.config.ts]
+export default defineNuxtConfig({
+  modules: ["@sit-onyx/nuxt", "@nuxtjs/i18n"],
+  i18n: {
+    defaultLocale: "en_US",
+    langDir: "./i18n",
+    locales: [
+      { code: "en_US", file: "en-US.json", name: "English (US)" },
+      { code: "de", file: "de-DE.json", name: "Deutsch" },
+    ],
+  },
+  onyx: {
+    i18n: {
+      registerLocales: {
+        "de-DE": "de-DE",
+        en_US: "en-US",
+      },
+    },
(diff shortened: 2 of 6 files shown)


apps/docs/src/development/packages/nuxt.md
```diff
@@ -45,12 +45,32 @@ If your Nuxt project uses both modules, onyx will automatically detect it and us
...
+    langDir: "./i18n",
+    locales: [
+      { code: "en_US", file: "en-US.json", name: "English (US)" },
+      { code: "de", file: "de-DE.json", name: "Deutsch" },
+    ],
+  },
+  onyx: {
+    i18n: {
+      registerLocales: {
+        en_US: "en-US",
```
larsrickert, line 64: Is the mapping also needed when the key and value is the same? If yes, we should maybe also add an example for this here so people might not think they only need it when they are different

packages/nuxt/src/module.ts
```diff
@@ -68,13 +82,12 @@ export default defineNuxtModule<ModuleOptions>({
...
-          locales: [
-            // we need to use .js files instead of .ts because the .ts files would be compiled to .js in the build step, so
-            // when projects use this nuxt module, .ts files will throw a "can not find file" error
-            { code: "de-DE", file: "de-DE.js" },
-            { code: "en-US", file: "en-US.js" },
-            { code: "ko-KR", file: "ko-KR.js" },
-          ],
+          // we need to use .js files instead of .ts because the .ts files would be compiled to .js in the build step, so
+          // when projects use this nuxt module, .ts files will throw a "can not find file" error
+          locales: Object.entries(options.i18n?.registerLocales ?? {}).map(([code, language]) => ({
```
larsrickert, line 87: Can't we just check which locales the user has defined in nuxt i18n and only register onyx locales then?

.changeset/old-dogs-sin.md
```diff
@@ -0,0 +1,8 @@
+---
+"@sit-onyx/nuxt": patch
```
larsrickert, line 2:
```suggestion
"@sit-onyx/nuxt": major
```

This would be a major / breaking change because the user has to change something in his config

larsrickert (changes requested):

packages/nuxt/src/module.ts
markbrockhoff: Yes but actually no. :D
We could do that but as the locales can be defined in a lot of different ways in different places this would be really fragile and error prone. It would also not work if the user uses different locale codes than BCP-47.
And most important: This is only intended as a temporary fix to get the module working in the first place. Later nuxt-i18n should handle all the merging and mapping for the locales.

.changeset/old-dogs-sin.md
markbrockhoff: If you really want this, sure, but as long as we're in alpha mode, it should not create a "real" major release anyway right?

larsrickert (approved):

PR #1470, opened 2024-07-01
feat(OnyxTable): create prop to handle table- vs page-scrolling

### LINKED ISSUE #1242: Fix table header sticky issue in page context
In the following ticket it was defined:
- #840 
> The header sticks to the top of the scroll-area **when the page is scrolled, with the table on it**

During the initial implementation it once worked in the alpha test app.  
Now it does not work anymore:
Image

Reproduce bei starting the alpha test app and opening http://localhost:5173/. resize the height of the browser window a lot, then scroll down. the table header does not stick below the nav-bar as expected.

BoppLi:
Closes #1242

feat(OnyxTable): create prop to switch from table-scrolling to page-scrolling
- make alpha table example more vue-ish
- feat(demo-app): remember component selection in localStorage

## Checklist

### DIFF
--- a/.changeset/six-parents-appear.md
+++ b/.changeset/six-parents-appear.md
@@ -2,4 +2,4 @@
 "sit-onyx": minor
 ---
 
-feat(OnyxTable): create prop to switch from table-scrolling to page-scrolling
+feat(OnyxTable): add `withPageScrolling` prop to switch from table-scrolling to page-scrolling
--- a/.changeset/six-parents-appear.md
+++ b/.changeset/six-parents-appear.md
@@ -1,0 +1,5 @@
+---
+"sit-onyx": minor
+---
+
+feat(OnyxTable): create prop to switch from table-scrolling to page-scrolling
--- a/packages/sit-onyx/src/components/OnyxTable/types.ts
+++ b/packages/sit-onyx/src/components/OnyxTable/types.ts
@@ -13,7 +13,7 @@ export type OnyxTableProps = DensityProp & {
    * Whether the table will scroll internally when a CSS max height/width is exceeded.
    * - with scrolling internally, the table header will stick to the top of the **table**
    * - without scrolling internally, the table header will stick to the top of the **page**
-   * Never set a max-height/width on the table without setting withInternalScrolling to true.
+   * - Never set a max-height/width on the table without setting `withInternalScrolling` to true.
    */
   withInternalScrolling?: boolean;
 };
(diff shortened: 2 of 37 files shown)


packages/sit-onyx/src/components/OnyxTable/types.ts
```diff
@@ -9,4 +9,11 @@ export type OnyxTableProps = DensityProp & {
    * If `true`, vertical borders will be displayed in addition to the default horizontal borders.
    */
   withVerticalBorders?: boolean;
+  /**
+   * Whether the table will scroll internally when a CSS max height/width is exceeded.
+   * - with scrolling internally, the table header will stick to the top of the **table**
+   * - without scrolling internally, the table header will stick to the top of the **page**
+   * - Never set a max-height/width on the table without setting `withInternalScrolling` to true.
+   */
+  withInternalScrolling?: boolean;
```
larsrickert, line 18: I don't get the purpose of this prop to be honest^^ If the table height is limited via CSS, it should always be scrollable

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxTable/types.ts
BoppLi: we could flip the default - let's discuss it tomorrow

BoppLi: image
 done :D

.changeset/six-parents-appear.md
```diff
@@ -0,0 +1,5 @@
+---
+"sit-onyx": minor
+---
+
+feat(OnyxTable): create prop to switch from table-scrolling to page-scrolling
```
larsrickert, line 5:
```suggestion
feat(OnyxTable): add `withPageScrolling` prop to switch from table-scrolling to page-scrolling
```

packages/sit-onyx/src/components/OnyxTable/types.ts
```diff
@@ -9,4 +9,11 @@ export type OnyxTableProps = DensityProp & {
    * If `true`, vertical borders will be displayed in addition to the default horizontal borders.
    */
   withVerticalBorders?: boolean;
+  /**
+   * Whether the table will be displayed at it's full size and is scrollable only in the page context.
```
larsrickert, line 13:
```suggestion
   * Whether the table will be displayed at it's full size and is scrollable only in the page context.
```

I am not sure about the wording "display at its full size" that could be misleading because people the property only affects scrolling and not size

larsrickert (changes requested):

larsrickert (approved):

Issue #764, opened 2024-03-26
Storybook: Implement theme selection


BoppLi:
Make the themes selectable in storybook

PR #1517, opened 2024-07-05
fix(docs): use correct theme for VitePress

larsrickert:
Relates to #764

Currently our VitePress docs uses the twogo theme which is not correct. It should use the default onyx theme.
This is fixed in this PR. Also the Storybook theme switch now uses dynamic imports for the theme CSS styles

### DIFF
--- a/apps/docs/src/development/theming.data.ts
+++ b/apps/docs/src/development/theming.data.ts
@@ -1,0 +1,27 @@
+import { defineLoader } from "vitepress";
+
+export type Data = {
+  themes: string[];
+};
+
+declare const data: Data;
+export { data };
+
+/**
+ * Build-Time data loader to get a list of available languages
+ * @see https://vitepress.dev/guide/data-loading
+ */
+export default defineLoader({
+  watch: ["../../../../packages/sit-onyx/src/styles/themes/*.css"],
+  load(watchedFiles): Data {
+    return {
+      themes: watchedFiles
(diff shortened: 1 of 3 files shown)


apps/docs/src/development/theming.data.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export type Data = {
+  themes: string[];
+};
+
+declare const data: Data;
+export { data };
+
+/**
+ * Build-Time data loader to get a list of available languages
```
BoppLi, line 11:
```suggestion
 * Build-Time data loader to get a list of available themes and prioritizes the onyx theme
```

packages/sit-onyx/.storybook/theme-switch.ts
```diff
@@ -1,15 +1,24 @@
...
-const themes = import.meta.glob("../src/styles/themes//*.css", { eager: true });
-export const ONYX_THEMES = Object.keys(themes)
-  .map((filePath) => filePath.split("/").at(-1)!.replace(".css", ""))
-  .sort((a, b) => {
-    if (a === "onyx") return -1;
-    if (b === "onyx") return 1;
+const themes = import.meta.glob("../src/styles/themes/*.css");
+
+/**
+ * Map of all available onyx themes.
```
BoppLi, line 8:
```suggestion
 * Map of all available onyx themes that puts the onyx theme up front.
```

BoppLi (approved):

BoppLi (approved):

Issue #1301, opened 2024-06-13
Implement OnyxToastProvider and useToast composable


BoppLi:
### Depends on

- #1286

### Why?

A composable to control toasts to appear, no matter where it needs to be created. 

### Design

- [Figma](https://www.figma.com/design/YfEUBOHk4J4nYrk04geswG/Onyx-Component-Library?node-id=4922-23696&t=bsuwI8wAhjcT7zdW-4)

### Acceptance criteria

- [x] useToast works together with the OnyxToastProvider component
- [x] useToast can show different kinds of toasts (see implementation details)
- [x] uses [provide-inject](https://vuejs.org/guide/components/provide-inject)

#### Toast provider features
- [x] width of the toasts: define min and max width. respect "padding" on small screens.

### Definition of Done

- [x] example usage added to `apps/alpha-test-app/src/views/HomeView.vue`
- [x] LayoutDemo is updated `apps/alpha-test-app/src/views/LayoutDemoView.vue`, the Toast works as expected in combination with other overlays
- [x] covered by component tests
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature (?)
- [x] Storybook code snippet of new/changed examples are checked that they are generated correctly

**Approval**

- [Storybook](https://storybook.onyx.schwarz/?path=/docs/feedback-toast--docs)
- [Icon docs](https://onyx.schwarz/icons.html), click an icon to see the "Copied" success toast
- [x] approved by designer @jannick-ux
- [x] approved by dev @BoppLi 

### Implementation Details

- aria live -> do the polite interruption of the screen reader
- usage examples for how the api might look:
```ts
toast.show({ title: "Hello World", status: "warning", duration: 6000, onclick: () => {/* your callback */} });
toast.success({ title: "Successfully saved data", icon: emojiHappy2 });
toast.error({ title: "Something went wrong", description: "Please contact us.", closable: true });
```

### Reference implementations 

- https://vue-blueprint.schwarz/packages/vue-core.html#usetoast

BoppLi:
@larsrickert can you please change the useToast example in the `apps/alpha-test-app/src/views/HomeView.vue` to be triggered by a button that is listed as all other components, so it does not appear onMounted but only when "summoned"? :)
currently it's not visible in the component selection dropdown and "attention grabbing" when reloading ^^

BoppLi:
@larsrickert  change request: 
- if possible, add a link to docs about the mentioned `Vue plugin`
[...]

PR #1520, opened 2024-07-05
refactor: rename toast components

larsrickert:
Relates to #1301

As discussed with Jannick, we aligned the naming between UX and dev

## Checklist


packages/sit-onyx/playwright/snapshots/components/OnyxToastMessage/Toast-message-truncation--webkit-linux.png
BoppLi, line 1: was the intention to have the description be truncated after 3 lines? it does not fill the full 3 lines, or is there more above? (does not start with T but "est")
annoying that we can't capture scroll bars ^^'

larsrickert: It is limited at 3 lines and there are more than 3 in this screenshot example but it will scroll down to test that it works :) Thats the reason why it starts with "est" instead of "Test" because its already scrolled down

BoppLi (approved):

Issue #733, opened 2024-03-22
Implement truncation for OnyxSelect options


larsrickert:
### Depends on

- #564
- #883 (refactoring of select)

### Why?

The OnyxSelect should allow to set the label truncation for each option individually.

### Design

[Figma link](https://www.figma.com/file/YfEUBOHk4J4nYrk04geswG/onyx---Design-System-Figma-Library?type=design&node-id=1263-9259&mode=dev)

### Acceptance criteria

- the options can be truncated with
  - [x] ellipsis (default)
  - [x] multiline
- [x] if multiline truncation is used, the max height to only show X options still works

### Definition of Done

- [x] covered by component tests
- [x] screenshot tests are updated
- [x] follow-up tickets were created if necessary
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature
- [x] Set task to `In Approval`

**Approval**

- [Storybook](https://storybook.onyx.schwarz/?path=/story/form-select--multiline-options&args=open:!true)
- [x] approved by designer @jannick-ux
- [x] approved by dev @BoppLi 

### Implementation details

- Use existing `onyx-truncation-ellipsis` and `onyx-truncation-multiline` classes
- Use existing type `TruncationType` for the property

PR #1524, opened 2024-07-08
feat(OnyxSelect): Implement truncation for OnyxSelect options

MajaZarkova:
Relates to #733 

Implemented new truncation property using the existing 'TruncationType', which will apply the existing 'onyx-truncation-ellipsis' and 'onyx-truncation-multiline' classes

## Checklist


packages/sit-onyx/src/components/OnyxSelect/OnyxSelect.vue
```diff
@@ -388,6 +390,7 @@ const selectInputProps = computed(() => {
 
 .onyx-select-wrapper {
   @include layers.component() {
+    --max-height: 20rem;
```
BoppLi, line 393: the var would probably be overridden directly on <OnyxSelect in the project right?
then maybe enhance the name so it's clear that the max height refers to the flyout part?

packages/sit-onyx/src/components/OnyxSelect/OnyxSelect.vue
```diff
@@ -360,6 +361,7 @@ const selectInputProps = computed(() => {
                 :active="option.value === activeValue"
                 :icon="option.icon"
                 :density="props.density"
+                :truncation="props.truncation"
```
BoppLi, line 364: could you please add a storybook example? did you test it in combination with the custom-options?

BoppLi: info: it works properly with custom-options 👍
image

but having one storybook example with multiline flyout would be helpful :)

.changeset/chatty-ducks-play.md
```diff
@@ -0,0 +1,5 @@
+---
+"sit-onyx": minor
+---
+
+Implement truncation for OnyxSelect options
```
BoppLi, line 5: not sure if it _needs_ to be done like that, but I guess the majority of our changelog messages are in the syntax

```suggestion
feat(OnyxSelect): Implement truncation for OnyxSelectOptions
```

packages/sit-onyx/src/components/OnyxSelect/OnyxSelect.ct.tsx
```diff
@@ -93,6 +98,29 @@ test.describe("Empty screenshots", () => {
   });
 });
 
+test.describe("Truncated options screenshots", () => {
```
BoppLi, line 101: could you hover on one of the options so we can verify that the hover background covers the whole multiline option?

BoppLi (changes requested):

packages/sit-onyx/src/components/OnyxSelect/OnyxSelect.ct.tsx
```diff
@@ -93,6 +98,29 @@ test.describe("Empty screenshots", () => {
...
+    component: (column, row) => (
+      <div>
+        <OnyxSelect
+          label="Label"
+          listLabel="List label"
+          options={MOCK_LONG_LABELED_OPTIONS}
+          density={column}
+          truncation={row}
+        />
+      </div>
```
BoppLi, line 116: the div should be redundant

```suggestion
      <OnyxSelect
        label="Label"
        listLabel="List label"
        options={MOCK_LONG_LABELED_OPTIONS}
        density={column}
        truncation={row}
      />
```

packages/sit-onyx/src/components/OnyxSelect/OnyxSelect.vue
BoppLi, line 364: shouldn't multiline also work with the same truncation?
image

packages/sit-onyx/src/components/OnyxSelect/types.ts
```diff
@@ -84,6 +89,10 @@ export type OnyxSelectProps<TValue extends SelectOptionValue = SelectOptionValue
      * If you want to use a button instead, use the `optionsEnd` slot.
      */
     lazyLoading?: SelectLazyLoading;
+    /**
+     * How to truncate the option label / text content if it exceeds the max width.
+     */
+    truncation?: TruncationType;
```
BoppLi, line 95: question: shouldn't we have the same pattern as in the checkboxGroup and let the truncation be declared on the options-level? BaseSelectOption is shared between checkboxGroup and select and already offers `truncation` 
```
export type SelectOption<TValue extends SelectOptionValue = SelectOptionValue> = Pick<
  BaseSelectOption<TValue>,
  "value" | "label" | "disabled" | "truncation"
> 
```

packages/sit-onyx/src/components/OnyxSelectOption/types.ts
```diff
@@ -14,4 +15,8 @@ export type OnyxSelectOptionProps = Pick<OnyxListItemProps, "active" | "density"
    * Optional icon.
    */
   icon?: string;
+  /**
+   * How to truncate the option label / text content if it exceeds the max width.
+   */
+  truncation?: TruncationType;
```
BoppLi, line 21: see other comment, OnyxCheckbox e.g. inherits `truncation` from `BaseSelectOption`

BoppLi (changes requested):

packages/sit-onyx/src/components/OnyxSelect/types.ts
MajaZarkova: In this case, what happens with the checkAll truncation?

BoppLi: I think the check all will always be ellipsis but it's not expected to have such a long text that truncation gets relevant

packages/sit-onyx/src/components/OnyxSelect/OnyxSelect.ct.tsx
```diff
@@ -34,6 +34,12 @@ const MOCK_LONG_LABELED_OPTIONS = Array.from({ length: 10 }, (_, index) => ({
   label: `Long labeled option ${index + 1} `.repeat(4),
 })) satisfies SelectOption[];
 
+const MOCK_MULTILINE_LONG_LABELED_OPTIONS = Array.from({ length: 10 }, (_, index) => ({
+  value: index,
+  truncation: "multiline",
+  label: `Long labeled option ${index + 1} `.repeat(4),
+})) satisfies SelectOption[];
+
```
BoppLi, line 42:
```suggestion
MOCK_LONG_LABELED_OPTIONS.map(option => ({ ...option, truncation: "multiline" }) satisfies SelectOption[];
```

BoppLi (approved):

packages/sit-onyx/playwright/snapshots/components/OnyxSelect/Select-truncated--firefox-linux.png
BoppLi, line 1: the compact ellipsis is missing the highlighting :O

BoppLi: maybe try to only select by part of the text, like only "Long labeled option 1", you don't need to have the full string as long as there is a unique match

packages/sit-onyx/src/components/OnyxSelect/OnyxSelect.ct.tsx
```diff
@@ -93,6 +103,30 @@ test.describe("Empty screenshots", () => {
...
+        listLabel="List label"
+        options={
+          row === "ellipsis" ? MOCK_LONG_LABELED_OPTIONS : MOCK_MULTILINE_LONG_LABELED_OPTIONS
+        }
+        density={column}
+      />
+    ),
+    beforeScreenshot: async (component) => {
+      await openFlyout(component);
+      const option = component.getByText(`Long labeled option 1 `.repeat(4));
```
BoppLi, line 124: I'm referring to here :)

```suggestion
      const option = component.getByText(`Long labeled option 1 `);
```

BoppLi (dismissed):
everything except the one screenie is perfect now ^^

packages/sit-onyx/playwright/snapshots/components/OnyxSelect/Select-truncated--firefox-linux.png
BoppLi: see comment below

MajaZarkova: I updated the screenshots and it's fixed now :)

BoppLi (approved):

packages/sit-onyx/playwright/snapshots/components/OnyxSelect/Select-truncated--firefox-linux.png
MajaZarkova: I was missing 'await' in from of .hover()

BoppLi (approved):

Issue #1490, opened 2024-07-03
Implement theme distribution


BoppLi:
### Acceptance criteria

- [x] Pipeline is extended to extranct the Figma theme tokens
- [x] Explain the concept in the docs + add an example

### Out of scope

- #764 
- #1493

### Definition of Done

- [x] Docs are updated: how to import a theme
- [x] updated version + documentation is deployed
- [x] Set task to `In Approval`

**Approval**

- [Deployed Docs](https://onyx.schwarz/development/theming.html)
- [Storybook Theme Selection on Topbar](https://storybook.onyx.schwarz/?path=/docs/buttons-button--docs)
- [x] approved by designer @jannick-ux
- [x] approved by dev @BoppLi

jannick-ux:
We have to check the color palettes with the implementet color-values. The colors for Lidl and Kaufland seems to be incorrect in the storybook component. 
An extra meeting for this would make sense here. Who is the dedicated developer here?

larsrickert:
> The colors for Lidl and Kaufland seems to be incorrect in the storybook component.

We imported them from Figma :) I sent you a meeting to check this

jannick-ux:
The reason was my private monitor, that displayed the colors incorrectly.

PR #1528, opened 2024-07-08
feat(nuxt): support theming

markbrockhoff:
Relates to #1490

This PR adds the option to easily theme a nuxt project using the onyx module. Specifying the "theme" config option will automatically inject the correct stylesheet.

## Checklist


packages/nuxt/src/module.ts
```diff
@@ -4,6 +4,11 @@ import type { ModuleHooks as NuxtI18nModuleHooks } from "@nuxtjs/i18n";
 import * as onyx from "sit-onyx";
 
 export interface ModuleOptions {
+  /**
+   * The onyx theme to use. See: [https://onyx.schwarz/development/theming.html](https://onyx.schwarz/development/theming.html)
+   * If no theme is selected the default "onyx" theme will be used.
+   */
+  theme?: "kaufland" | "lidl" | "twogo";
```
larsrickert, line 11: Should we maybe maintain this type in sit-onyx itself and export it there?

apps/docs/src/development/packages/nuxt.md
```diff
@@ -38,6 +38,21 @@ If you are also using the [@nuxtjs/i18n](https://i18n.nuxtjs.org/) module, make
...
+```ts [nuxt.config.ts]
+export default defineNuxtConfig({
+  modules: ["@sit-onyx/nuxt"],
+  onyx: {
+    theme: "lidl",
+  },
+});
+```
+
+For all supported themes see: [Theming](/development/theming#themes)
```
larsrickert, line 54: Could you also add a short note in the theming docs that says something like (in the sense of)
"If you are using Nuxt, check here"

larsrickert (changes requested):
Nice addition to the Nuxt module! Thanks for contributing :)

packages/nuxt/src/module.ts
markbrockhoff: Sure, where would you like it to be?
Also: I would love to auto generate those but that would most likely require the use of macros which I'm unsure of if this is something we want inside the codebase...

markbrockhoff: I just put it in styles/themes/index.ts for now. Let me know if you see a more fitting place ;)

larsrickert: I'd suggest to move it to the types folder so we also don't need the additional entry in the package.json :)
Also maybe rename it to `OynxTheme`, what do you think?

markbrockhoff: Sure, no problem. I just placed it close to the themes so it won't be as easy to miss when adding new ones. But I guess this shouldn't happen frequently.

larsrickert (approved):

Issue #1518, opened 2024-07-05
Fix OnyxUserMenu captures click events before they reach the actual target


markbrockhoff:
### Version

1.0.0-beta.0

### Description

Hi, while trying to use the OnyxUserMenu I encountered the following bug:

Whenever I try to click something inside the menu it just closes, blocking any interaction with the elements inside of it.
Using the keyboard to interact with the element works as expected but click events seem to be blocked at some point.

### Reproduction steps

Here's a playground reproducing the issue. The example is basically the same as in the [docs](https://onyx.schwarz/development/theming.html#dark-mode), just without vueuse.

[Playground](https://playground.onyx.schwarz/#eNp9Uk1PAjEQ/SuTekATPg56QjAB5ACJYERvTUyzDLDQbZu2u2II/91pF9g1IfTUznsz895MD2xgTLvIkXVZzyU2NR4c+tyAFGrd58w7zl64SjOjrYcDzNXvfiaKobDNeP9yaN9Q5eVrpKW2i2SDGYbgxGMGR1hZnQFnLvUtTSTOnmsFLa4qCgmJKFeJVs5DUtWDfqDeN5bC7hoPxOl1Sr2kjh7UyUjhkV4AvUolCGNaSmRIXsZ7QSQMoWgK6FwS4Y5aetz7gUVxAk+Vzh4hp8up1lRvFLxqknvhntjXZlC0Mr1ESXk1R5xBp2rUqXc6i+v8txU5pS8K1FDWpEWR/lW6bm+dVrTNQ8gI/TKTSrRz41MaKWddiEjAhJT6Zxpj3ubYPMdJXbK7Et862l2XLu8WSWhBDi6YF3aNvoTHixnNsQaS+VwS+wb4gU7LPGgsacNcLUl2jRfVTuK3SdX60433HpU7mwpCA/MY+fEnjW5Yr+Q+tp9iHldHmuJ3gTbUpAGGwTrPjn95Lg0B)

### Confirmations

- [X] I have checked that there is not already a bug report for my issue.
- [X] I confirm that are no sensitive information included in this bug report. e.g. credentials, internal project details (for employees of the Schwarz Group) etc.

JoCa96:
Thanks for reporting! Will be fixed soon.

markbrockhoff:
Just as a note: It looks like not only the user menu is affected also a normal OnyxNavButton seems to not trigger any event when clicking it. Maybe the issue arrises from the Flyout which wrapps all of them?

PR #1460, opened 2024-07-01
feat(headless): implement headless Menu

JoCa96:
Relates to #1415, #883
Closes #1518 

- implemented headless feature: `createNavigationMenu`
- headless MenuButton:
  -  now takes an `isExpandedRef` and `onToggle` via it's options
  - `flyout` element is removed as it is not needed
  - removed hover and focus toggle features and moved them to the onyx component directly as these are non spec features

PR #1533, opened 2024-07-09
feat(@sit-onyx/headless): implement global event listener

JoCa96:
Some files/functions were added in preparation for #1460
Relates to #1518

* Rename `*.ct.ts` ts helper files to fix playwright import issue 
* Use `ref` instead of callback
* Update builder type to allow for no parameters and usage of reservered prop `ref`


packages/headless/src/composables/helpers/useGlobalListener.ts
```diff
@@ -0,0 +1,71 @@
...
+  event: DocumentEventMap[K],
+) => void;
+
+export type UseGlobalEventListenerOptions<K extends DocumentEventType> = {
+  type: K;
+  listener: GlobalListener<K>;
+  disabled?: Ref<boolean>;
+};
+
+const GLOBAL_LISTENERS = reactive(new Map<DocumentEventType, Set<GlobalListener>>());
```
larsrickert, line 14: This is global state, right? I am not sure if this behaves correctly when used with SSR (e.g. Nuxt), so it might be shared across all clients

packages/headless/src/composables/helpers/useGlobalListener.ts
```diff
@@ -0,0 +1,71 @@
...
+export const useGlobalEventListener = <K extends DocumentEventType>({
+  type,
+  listener,
+  disabled,
+}: UseGlobalEventListenerOptions<K>) => {
+  onBeforeMount(() => {
+    if (!disabled) {
+      addGlobalListener(type, listener);
+      return;
+    }
```
larsrickert, line 61: Isn't this already covered by the watch below?

packages/headless/src/composables/tooltip/createTooltip.ts
```diff
@@ -106,6 +107,9 @@ export const createTooltip = createBuilder((options: CreateTooltipOptions) => {
 
   return {
     elements: {
+      root: {
```
larsrickert, line 110: Technically this is a breaking change right? Because if you used this composable, you need to change your code to bind the root element which was not needed before

packages/headless/src/utils/vitest.ts
```diff
@@ -0,0 +1,36 @@
...
+export const mockVueLifecycle = () => {
+  const { callbacks } = vi.hoisted(() => ({
+    callbacks: {
+      onBeforeUnmountedCb: null as Callback | null,
+      onUnmountedCb: null as Callback | null,
+    },
+  }));
+
+  vi.mock("vue", async (original) => ({
+    ...((await original()) as object),
```
larsrickert, line 25:
```suggestion
    ...((await original()) as typeof import("vue")),
```

packages/headless/src/utils/vitest.ts
```diff
@@ -0,0 +1,36 @@
...
+    onBeforeMount: (cb: Callback) => cb(),
```
larsrickert, line 26:
```suggestion
    onBeforeMount: vi.fn().mockImplementation((cb: Callback) => cb())
```

Should we use vi.fn() here so they can be spyed on?

packages/sit-onyx/playwright.config.ts
```diff
@@ -40,9 +40,5 @@ export default defineConfig({
     },
   },
   /* Configure projects for major browsers */
-  projects: [
-    { name: "chromium", use: { ...devices["Desktop Chrome"] } },
-    { name: "firefox", use: { ...devices["Desktop Firefox"] } },
-    { name: "webkit", use: { ...devices["Desktop Safari"] } },
-  ],
+  projects: [{ name: "chromium", use: { ...devices["Desktop Chrome"] } }],
```
larsrickert, line 43: Why did you remove firefox and webkit? Is this unintentionally?

larsrickert (changes requested):

packages/sit-onyx/playwright.config.ts
JoCa96: Yes, whoops 🙈

packages/headless/src/composables/helpers/useGlobalListener.ts
JoCa96: Actually, `watchEffect` seems like a good choice here:

packages/headless/src/composables/helpers/useGlobalListener.ts
JoCa96: Yes, it's global state. But listeners are only added on the client onBeforeMount, so everything should be fine in that regard.

packages/headless/src/composables/tooltip/createTooltip.ts
JoCa96: Updated the Changeset

larsrickert (approved):

Issue #1539, opened 2024-07-10
Implement truncation prop for OnyxCheckboxGroup, OnyxRadioGroup and OnyxSelect


MajaZarkova:
Follow up on #733 

### Why?

Property truncation should be inherited from OnyxBaseSelectOption and be used as a fallback value, meaning if a truncation is not given on option level this value should be used for all options.

### Acceptance criteria

- [x] Default value og truncation should be 'ellipsis'

### Definition of Done

- [x] covered by component tests
- [x] screenshot tests are updated
- [x] follow-up tickets were created if necessary
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature
- [x] Storybook code snippet of new/changed examples are checked that they are generated correctly
- [x] Namings are aligned with Figma

**Approval**

- Storybook 
- [x] approved by dev @BoppLi

PR #1541, opened 2024-07-10
feat(OnyxSelect, OnyxRadioGroup, OnyxCheckboxGroup): Implement truncation on higher level

MajaZarkova:
Relates to #1539 

Up until now the truncation property was on option-level only for OnyxSelect, OnyxRadioGroup and OnyxCheckboxGroup. The user had to set the property for each option separately. In this PR, the truncation property is added to the grouped component too, so that the user can set the truncation only once for every option. This property is used only as a fallback if the truncation is not set on option-level.

## Checklist

### DIFF
--- a/packages/sit-onyx/src/components/OnyxCheckbox/OnyxCheckbox.vue
+++ b/packages/sit-onyx/src/components/OnyxCheckbox/OnyxCheckbox.vue
@@ -14,7 +14,6 @@ const props = withDefaults(defineProps<OnyxCheckboxProps<TValue>>(), {
   disabled: false,
   loading: false,
   required: false,
-  truncation: "ellipsis",
   skeleton: false,
 });
 
--- a/packages/sit-onyx/src/components/OnyxRadioButton/OnyxRadioButton.vue
+++ b/packages/sit-onyx/src/components/OnyxRadioButton/OnyxRadioButton.vue
@@ -11,7 +11,6 @@ const props = withDefaults(defineProps<OnyxRadioButtonProps<TValue>>(), {
   required: false,
   checked: false,
   loading: false,
-  truncation: "ellipsis",
 });
 
 const emit = defineEmits<{
--- a/packages/sit-onyx/src/components/OnyxSelectOption/OnyxSelectOption.vue
+++ b/packages/sit-onyx/src/components/OnyxSelectOption/OnyxSelectOption.vue
@@ -6,7 +6,6 @@ import type { OnyxSelectOptionProps } from "./types";
 const props = withDefaults(defineProps<OnyxSelectOptionProps>(), {
   active: false,
   multiple: false,
-  truncation: "ellipsis",
 });
 
 defineSlots<{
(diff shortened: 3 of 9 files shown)


packages/sit-onyx/playwright/snapshots/components/OnyxCheckbox/Checkbox-other--chromium-linux.png
BoppLi, line 1: oh no, the required feature got broken :O

BoppLi (changes requested):
<ignore this, it was an obsolete comment :D

packages/sit-onyx/src/components/OnyxRadioButton/OnyxRadioButton.vue
```diff
@@ -11,7 +11,6 @@ const props = withDefaults(defineProps<OnyxRadioButtonProps<TValue>>(), {
   required: false,
   checked: false,
   loading: false,
-  truncation: "ellipsis",
```
BoppLi, line 14: actually,  I think for the stories/playwright tests, we might even keep it. the group will always overwrite it regardless :)

on top, if the support component is used by a project team directly, they will not have to worry about it 😊

packages/sit-onyx/src/components/OnyxRadioButton/OnyxRadioButton.vue
```diff
@@ -11,7 +11,6 @@ const props = withDefaults(defineProps<OnyxRadioButtonProps<TValue>>(), {
   required: false,
   checked: false,
   loading: false,
```
BoppLi, line 13:
```suggestion
  loading: false,
  truncation: "ellipsis",
```

packages/sit-onyx/src/components/OnyxSelectOption/OnyxSelectOption.vue
```diff
@@ -6,7 +6,6 @@ import type { OnyxSelectOptionProps } from "./types";
 const props = withDefaults(defineProps<OnyxSelectOptionProps>(), {
   active: false,
   multiple: false,
```
BoppLi, line 8:
```suggestion
  multiple: false,
  truncation: "ellipsis",
```

packages/sit-onyx/src/components/OnyxCheckbox/OnyxCheckbox.vue
```diff
@@ -14,7 +14,6 @@ const props = withDefaults(defineProps<OnyxCheckboxProps<TValue>>(), {
   disabled: false,
   loading: false,
   required: false,
```
BoppLi, line 16:
```suggestion
  required: false,
  truncation: "ellipsis",
```

BoppLi (changes requested):

packages/sit-onyx/playwright/snapshots/components/OnyxCheckbox/Checkbox-other--chromium-linux.png
BoppLi: see suggestions for solution below. :)

packages/sit-onyx/src/components/OnyxRadioButton/OnyxRadioButton.vue
MajaZarkova: I'm checking in the parent component if the truncation prop of the child is undefined and in this case it will always be set and the parent property will be ignored?

MajaZarkova: For now I just added truncation prop for the screenshot tests :)

BoppLi (approved):

Issue #1397, opened 2024-06-24
Implement OnyxSelect API refactoring


BoppLi:
### Why?

We want to improve the DX for the majority of devs that use OnyxSelect.

### Acceptance criteria

- [x] Prio 1 of the ticket: finalize the concept. it's just a draft that is written below.

Changes inside of OnyxSelect
- [x] change modelValue property to only `TValue`/`TValue[]`
- [x] per default, resolve initial modelValues to find the labels in the given `options`
- [x] add new prop to overwrite the displayed labels: (to be defined whether  `string`/`string[]`)
- [x] add new prop to overwrite filter function, which defaults to our utility func `normalizedIncludes`. 
  - [x] defaults means, when the `useManagedState` says so.
  - [x] call filter function inside of OnyxSelect
  - [x] can be deactivated

### Definition of Done

- [x] example usage added to `apps/alpha-test-app/src/views/HomeView.vue`
- [x] covered by component tests
- [x] follow-up tickets were created if necessary
  - #1649
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature
- [x] Storybook code snippet of new/changed examples are checked that they are generated correctly
- [x] Set task to `In Approval`

**Approval**

- [Select with modelValue "mango" (= just the value)](https://storybook.onyx.schwarz/?path=/story/form-select--default&args=open:!false;modelValue:mango)
- [Select with integrated search](https://storybook.onyx.schwarz/?path=/story/form-select--with-search)
- [x] approved by designer @mj-hof 
- [x] approved by dev @larsrickert 

#### Sync example

```vue
<script>
const modelValue = ref<string>("1");
const options = ref(Array(10).fill(null).map(n => ({value: `${n}`, label: `${n}`})));
</script>

<template>
<OnyxSelect v-model="modelValue" :options="options" />
</template>
```

#### Async example

```vue
<script>
const modelValue = ref<string>("2"); // has initial selection
const options = ref<SelectOptions<string>[]>([])
const load = async () => {
   await setTimeout(2000);
   options.value = Array(10).fill(null).map(n => ({value: `${n}`, label: `label ${n}`}))
}

onMounted(() => load());
</script>

<template>
<OnyxSelect v-model="modelValue" :options="options" :selectionLabel="`label ${modelValue}`"  /> <!-- ideally would show "label 2" from the beginning -->
</template>
```

#### Filtering/Pagination:
```vue
<script>
const modelValue = ref<string>("2"); // has initial selection
[...]

PR #1564, opened 2024-07-12
refactor(OnyxSelect)!: restrict modelValue to only contain values 

BoppLi:
Relates to #1397

- OnyxSelects `modelValue` now only needs TValue, not SelectOption<TValue>
- OnyxSelect determines the displayed labels by comparing `modelValue` with `options`. This can be overridden by setting the new prop `valueLabel`.
- OnyxSelect now filters the options internally when `searchTerm` is set. This can be overridden by setting the new prop `manualSearch`.

## Checklist

- [x] If a new component is added, at least one playwright test is added


packages/sit-onyx/src/components/OnyxSelect/OnyxSelect.vue
```diff
@@ -347,14 +375,16 @@ const selectInputProps = computed(() => {
...
                 :key="option.value.toString()"
                 v-bind="
                   headlessOption({
                     value: option.value,
                     label: option.label,
                     disabled: option.disabled,
-                    selected: arrayValue.some(({ value }) => value === option.value),
+                    selected:
+                      arrayValue.some((value) => value === option.value) ||
+                      props.modelValue === option.value,
```
BoppLi, line 387: TODO: need help

this change ewas necessary, else the `test("should pass headless accessibility tests (select only)"` failed on the step `expectToBeSelected` - arrayValue was not yet updated at that time, but modelValue was... 
is there a way to stabilize the playwright test without this redundant workaround here?

BoppLi: solution note: in arrayValue, we did not check model value to be undefined but to be falsy. in the test, the ID was 0 so it failed.

packages/sit-onyx/src/components/OnyxSelect/types.ts
```diff
@@ -17,10 +17,17 @@ export type SelectSearchProps =
        * Property is managed internally, when undefined.
        */
       searchTerm?: string;
+      /**
+       * A filtered subset of the `options`.
+       * As default, onyx will handle the search by comparing
+       * the option labels with the `searchTerm`.
+       */
+      filteredOptions?: SelectOption<TValue>[];
```
JoCa96, line 25: Let's sync about the API here. When we are on the same page here, we can ensure a more stable API :) @MajaZarkova @larsrickert @BoppLi Do you have time today?
```
      withSearch: true | {
        /**
         * A filter function for synchronous filtering of the provided `options`. 
         * The default filter function will compare
         * the option labels with the `searchTerm`.
         */
        filterFunc: (option: SelectOption<TValue>, index: number, array: SelectOption<TValue>[]) => boolean;
      } | {
        loading?: boolean;
      }
```

BoppLi: note: as discussed, for this PR this would be too much. implemented another approach removing the filteredOptions for now

JoCa96 (dismissed):

JoCa96 (approved):

Issue #1348, opened 2024-06-18
Implement OnyxStepper


larsrickert:
### Open question

- @jannick-ux  Do we plan to create an additional component for entering very large numbers or numbers with many decimal places? If so, we don't need to implement the BigInt and custom decimal type for this ticket
- @jannick-ux Idea: Should we use monospace font?

### Why?

We want to support a stepper component that is optimised for entering numbers.

### Design

[Figma](https://www.figma.com/design/YfEUBOHk4J4nYrk04geswG/Onyx-Component-Library?node-id=5041-2580&t=zdXvmCyp6tpZ7T07-4)

### Acceptance criteria

- [x] design is implemented (and numbers are always centered)
- [x] supports positive, negative and decimal numbers
- [x] supports even and decimal steps
- [x] supports both dot and comma for entering decimal values
- [ ] can optionally put into "strip step" mode so that when manually entering values that do not fit the step size, a stepMismatch error is shown
- [ ] can optionally define a "precision" for the displayed value to always display the given number of decimals places, e.g. "3.00"
- [x] decrease/increase button is disabled when min/max value is reached
- [x] can be set to loading state
- the following native HTML properties are supported:
  - [x] autofocus
  - [x] autocomplete
- [x] error handling / error messages are supported for min, max and step property (see [this comment](https://github.com/SchwarzIT/onyx/issues/1273#issuecomment-2160150647))

### Out of scope

- [ ] can optionally put into "strip step" mode so that when manually entering values that do not fit the step size, a stepMismatch error is shown
- [ ] can optionally define a "precision" for the displayed value to always display the given number of decimals places, e.g. "3.00"
- [ ] example usage added to `apps/alpha-test-app/src/views/HomeView.vue`
- [ ] example usage added to `apps/alpha-test-app/src/views/FormDemo.vue`

### Definition of Done

- The following component requirements are implemented:
  - [x] `skeleton`
  - [x] `density`
- [x] covered by component tests
- [x] screenshot tests are updated
- [x] follow-up tickets were created if necessary -> #1587 
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature
- [x] Storybook code snippet of new/changed examples are checked that they are generated correctly
- [x] Set task to `In Approval`

**Approval**

- [Storybook](https://storybook.onyx.schwarz/?path=/docs/form-stepper--docs)
- [x] approved by designer @jannick-ux
- [x] approved by dev @BoppLi 
### Implementation details

[...]

PR #1568, opened 2024-07-16
feat(OnyxStepper): Implement basic OnyxStepper

MajaZarkova:
Relates to #1348 

A component that is optimized for entering numbers.

1.  supports positive, negative and decimal numbers
2.  supports even and decimal steps
3.  supports both dot and comma for entering decimal values
4.  decrease/increase button is disabled when min/max value is reached
5.  can be set to loading state
6.  autofocus is supported
7.  autocomplete is supported
8.  skeleton and density styles are implemented

## Checklist

### DIFF
--- a/packages/sit-onyx/src/components/OnyxStepper/OnyxStepper.vue
+++ b/packages/sit-onyx/src/components/OnyxStepper/OnyxStepper.vue
@@ -1,0 +1,261 @@
+<script lang="ts" setup>
+import minus from "@sit-onyx/icons/minus.svg?raw";
+import plus from "@sit-onyx/icons/plus.svg?raw";
+import { computed } from "vue";
+import { useDensity } from "../../composables/density";
+import { useCustomValidity } from "../../composables/useCustomValidity";
+import { injectI18n } from "../../i18n";
+import OnyxFormElement from "../OnyxFormElement/OnyxFormElement.vue";
+import OnyxIcon from "../OnyxIcon/OnyxIcon.vue";
+import OnyxLoadingIndicator from "../OnyxLoadingIndicator/OnyxLoadingIndicator.vue";
+import OnyxSkeleton from "../OnyxSkeleton/OnyxSkeleton.vue";
+import type { OnyxStepperProps } from "./types";
+
+const props = withDefaults(defineProps<OnyxStepperProps>(), {
+  modelValue: 0,
+  step: 1,
+  stripStep: false,
+  placeholder: "0",
+  disabled: false,
+  readonly: false,
+  loading: false,
+  skeleton: false,
+});
+
+const { t } = injectI18n();
+
+const emit = defineEmits<{
+  /** Emitted when the input value changes. */
+  "update:modelValue": [value: number];
+  /**
+   * Emitted when the current value changes but only when the input is blurred.
+   */
+  change: [value: number];
+  /**
+   * Emitted when the input is focussed.
+   */
+  focus: [];
+  /**
+   * Emitted when the input is blurred.
+   */
+  blur: [];
+  /**
+   * Emitted when the validity state of the input changes.
+   */
+  validityChange: [validity: ValidityState];
+}>();
+
+const { vCustomValidity, errorMessages } = useCustomValidity({ props, emit });
+
+const { densityClass } = useDensity(props);
+
+/**
+ * Current value (with getter and setter) that can be used as "v-model" for the native input.
+ */
+const value = computed({
+  get: () => props.modelValue,
+  set: (value) => emit("update:modelValue", value),
+});
+
+const handleChange = (event: Event) => {
+  const inputValue = (event.target as HTMLInputElement).valueAsNumber;
+  emit("update:modelValue", inputValue);
+};
+
+const handleCounterButton = (event: Event, operation: "add" | "substract") => {
+  event.preventDefault();
+  if (props.disabled || props.readonly || props.loading) return;
+
+  if (operation === "add") {
+    const newValue = value.value + props.step;
+
+    if (props.max && newValue > props.max) return;
+    value.value = value.value + props.step;
+  } else {
+    const newValue = value.value - props.step;
+
+    if (props.min && newValue < props.min) return;
+    value.value = value.value - props.step;
+  }
+};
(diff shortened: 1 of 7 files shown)


packages/sit-onyx/src/components/OnyxStepper/OnyxStepper.vue
```diff
@@ -0,0 +1,261 @@
...
+          :placeholder="props.placeholder"
+          :readonly="props.readonly"
+          :required="props.required"
+          :step="props.step"
+          :title="props.hideLabel ? props.label : undefined"
+          @change="handleChange"
+          @focus="emit('focus')"
+          @blur="emit('blur')"
+        />
+        <span
```
JoCa96, line 124: Please prefer to use a native button

packages/sit-onyx/src/components/OnyxStepper/OnyxStepper.vue
```diff
@@ -0,0 +1,261 @@
...
+          tabindex="0"
+          role="button"
+          class="onyx-stepper__counter"
+          :class="{ 'onyx-stepper__counter--disabled': props.max && props.max === value }"
+          :aria-label="t('add')"
+          @click="handleCounterButton($event, 'add')"
```
JoCa96, line 130: We can use the native [`stepUp`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/stepUp) and [`stepDown`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/stepDown) function here:
```suggestion
          @click="inputRef?.stepUp()"
```

packages/sit-onyx/src/components/OnyxStepper/OnyxStepper.vue
```diff
@@ -0,0 +1,261 @@
...
+
+const { t } = injectI18n();
+
+const emit = defineEmits<{
+  /** Emitted when the input value changes. */
+  "update:modelValue": [value: number];
+  /**
+   * Emitted when the current value changes but only when the input is blurred.
+   */
+  change: [value: number];
```
JoCa96, line 33: The change event is not used anywhere, as far as I see
```suggestion
```

packages/sit-onyx/src/components/OnyxStepper/OnyxStepper.vue
```diff
@@ -0,0 +1,261 @@
...
+          :aria-label="t('substract')"
+          @click="handleCounterButton($event, 'substract')"
+          @keydown.enter="handleCounterButton($event, 'substract')"
+        >
+          <OnyxIcon :icon="minus" />
+        </span>
+        <OnyxLoadingIndicator v-if="props.loading" class="onyx-stepper__loading" type="circle" />
+        <input
+          v-else
+          v-model="value"
```
JoCa96, line 105: `v-model` can produce a different result here, than our `@change` event handler, as it doesn't use `valueAsNumber`.

JoCa96 (changes requested):

packages/sit-onyx/src/components/OnyxStepper/OnyxStepper.vue
MajaZarkova: Should I use valueAsNumber instead?

packages/sit-onyx/src/components/OnyxStepper/OnyxStepper.vue
```diff
@@ -0,0 +1,261 @@
...
+      <div class="onyx-stepper__wrapper">
+        <span
+          tabindex="0"
+          role="button"
+          class="onyx-stepper__counter"
+          :class="{ 'onyx-stepper__counter--disabled': props.min && props.min === value }"
+          :aria-label="t('substract')"
+          @click="handleCounterButton($event, 'substract')"
+          @keydown.enter="handleCounterButton($event, 'substract')"
+        >
```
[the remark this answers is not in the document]
MajaZarkova: Initially I used buttons, but then issues with styling appeared. Because both buttons have the same class, when I hover or click on the plus button the styles are applied on the other button too. Do you know why is this happening?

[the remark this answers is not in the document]
MajaZarkova: I pushed the changes with native buttons you can check them out
@JoCa96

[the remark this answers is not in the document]
JoCa96: @MajaZarkova This is an issue with our `OnyxFormElement`. I created a PR to address this issue: #1576

JoCa96 (changes requested):

packages/sit-onyx/src/components/OnyxStepper/OnyxStepper.vue
```diff
@@ -0,0 +1,238 @@
...
+          :disabled="props.disabled || props.loading"
+          :min="props.min"
+          :max="props.max"
+          :name="props.name"
+          :placeholder="props.placeholder"
+          :readonly="props.readonly"
+          :required="props.required"
+          :step="props.step"
+          :title="props.hideLabel ? props.label : undefined"
+          @change="handleChange"
```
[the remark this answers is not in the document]
MajaZarkova: I thought I removed it, sorry

packages/sit-onyx/src/components/OnyxStepper/OnyxStepper.vue
```diff
@@ -0,0 +1,253 @@
...
+  switch (direction) {
+    case "down":
+      inputRef.value.stepDown();
+      break;
+    case "up":
+      inputRef.value.stepUp();
+  }
+
+  emit("update:modelValue", inputRef.value.valueAsNumber);
+};
```
JoCa96, line 69: This could be shortened like this:
```suggestion
const handleClick = (direction: "stepDown" | "stepUp") => {
  if (!inputRef.value) return;
  inputRef.value[direction]();
  emit("update:modelValue", inputRef.value.valueAsNumber);
};
```

JoCa96 (changes requested):

JoCa96 (approved):

Issue #1444, opened 2024-06-28
Adjust toast densities


larsrickert:
depends on #1210 

### Design

[Figma](https://www.figma.com/design/YfEUBOHk4J4nYrk04geswG/Onyx-Component-Library?node-id=4922-23696&t=aZGUUl9ztQowv9Yj-4)

![Image](https://github.com/SchwarzIT/onyx/assets/67898185/a4fa75a1-68f1-413e-bb4d-ff3b6ed1014c)

### Acceptance criteria

- [x] the design for the toast densities is checked and updated if needed
- [x] the updated design is implemented
- [x] the naming of the toast / toast message / toast provider is aligned between DEV and UX

### Definition of Done

- The following component requirements are implemented:
  - [x] `density`
- [x] covered by component tests
- [x] screenshot tests are updated
- [x] updated version + documentation is deployed

**Approval**

- [Storybook](https://storybook.onyx.schwarz/?path=/story/support-toastmessage--default)
- [x] approved by designer @jannick-ux
- [x] approved by dev @MajaZarkova

Issue #1210, opened 2024-05-29
Implement (refactor) density to be based on figma tokens


BoppLi:
### Depends on

- Follow-up for #1194

### Why?

The density concept got tokenized in figma. we need to align in dev.
-> Import the new density tokens and update all places where we handle paddings depending on the density currently in code.

### Acceptance criteria

- [x] The imported css files from figma get the correct selector on top (using our existing density classes)
- all density related CSS style got adapted to the new concept
  - [x] fix: Gap between checkbox+label in select (flex container) is fixed currently
  - [x] fix: Distance between switch and label is fixed currently
  - [x] fix: Checkbox padding on compact is 10px, it should be 8px. adjust the label padding accordingly to match the total height of 32px (for a 1-lined label)
- [x] Update the docs where needed, e.g. like the [other token docs](https://onyx.schwarz/tokens/spacings.html)
- [x] verify whether we still need to support our denstity mixins. Note: Yes it is still needed because some components have exceptions that can not be covered with the density tokens

### Definition of Done

- [x] covered by component tests
- [x] screenshot tests are updated
- [x] follow-up tickets were created if necessary
    - #1640
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature

**Approval**

- [Storybook](https://storybook.onyx.schwarz)
- [Deployed Docs](https://onyx.schwarz/tokens/spacings.html)
- [x] approved by designer @jannick-ux
- [x] approved by dev @BoppLi

PR #1601, opened 2024-07-23
refactor: implement new density tokens for all components

larsrickert:
Relates to #1210, #1444

Implement new density CSS variables for all components.

## Checklist


packages/sit-onyx/playwright/snapshots/components/OnyxToast/breakpoint-2xs-chromium-linux.png
BoppLi, line 1: the distance between headline and "x" also needs a density token
image
(density-medium, see figma)

BoppLi (changes requested):
reviewed all screenshots with @jannick-ux

packages/sit-onyx/playwright/snapshots/components/OnyxIconButton/Icon-button-loading--chromium-linux.png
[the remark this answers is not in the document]
larsrickert: Because this is actually a bug fix. If you take a look at the [current icon button](https://storybook.onyx.schwarz/?path=/story/buttons-iconbutton--primary), for compact or cozy density it has a different size when loading than when not-loading. With the new tokens this is fixed so the size stays the same

packages/sit-onyx/playwright/snapshots/components/OnyxToast/breakpoint-2xs-chromium-linux.png
BoppLi: could you modify e.g. the toast density screenshot test to show that?

larsrickert: I discussed this with Jannick and we dropped this feature since it leads to a less pretty gap when the toast is not hovered because by default its only shown on hover

packages/sit-onyx/src/components/OnyxButton/OnyxButton.vue
```diff
@@ -49,24 +49,10 @@ const emit = defineEmits<{
...
-  @include density.default {
-    --onyx-button-height: 2.5rem;
-    --onyx-button-padding-vertical: var(--onyx-spacing-2xs);
-  }
-
-  @include density.cozy {
-    --onyx-button-height: 3rem;
-    --onyx-button-padding-vertical: var(--onyx-spacing-sm);
-  }
+  --onyx-button-padding-vertical: var(--onyx-density-xs);
```
BoppLi, line 55: why do you keep `--onyx-button-padding-vertical` unlike in other components?
is there a pattern when you keep it / when you drop it?

packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxNavItem/OnyxNavItem.vue
```diff
@@ -42,7 +42,7 @@ defineSlots<{
 .onyx-nav-item {
   @include layers.component() {
     .onyx-external-link-icon {
-      margin-left: calc(-1 * var(--onyx-spacing-sm));
+      margin-left: calc(-1 * var(--onyx-menu-item-gap));
```
BoppLi, line 45: is it certain that OnyxNavItem with external lik icon will never be used without a parent OnyxMenuItem? else, the var would be unresolved right?

packages/sit-onyx/playwright/snapshots/components/OnyxToast/breakpoint-2xs-chromium-linux.png
BoppLi: ah, ok!

.github/workflows/import-figma.yml
```diff
@@ -25,24 +25,39 @@ jobs:
...
+
+      - name: 🛠️ Import default density variables
+        run: |
+          pnpm run @sit-onyx/figma-utils import-variables -k "${{ vars.FIGMA_FILE_KEY }}" -t "${{ secrets.FIGMA_TOKEN }}" -d "../sit-onyx/src/styles/variables" -m default -s ":where(:root), :where(.onyx-density-{mode})" -n density-
+        working-directory: packages/figma-utils
+
+      - name: 🛠️ Import remaining densities
+        run: |
+          pnpm run @sit-onyx/figma-utils import-variables -k "${{ vars.FIGMA_FILE_KEY }}" -t "${{ secrets.FIGMA_TOKEN }}" -d "../sit-onyx/src/styles/variables" -m compact cozy -s ":where(.onyx-density-{mode})" -n density-
         working-directory: packages/figma-utils
```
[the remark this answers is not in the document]
larsrickert: Yes because the :root selector must only be applied to the default density :)

packages/sit-onyx/src/components/OnyxButton/OnyxButton.vue
larsrickert: Because it is used multiple times (for the button and the skeleton).
If a variable is only used once, I removed it, otherwise I kept it

packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxNavItem/OnyxNavItem.vue
BoppLi: aah right ^^ thank you :)

BoppLi (approved):

Issue #363, opened 2024-02-20
Implement OnyxButton "active" animation


MajaZarkova:
**Having the hover state colors as a basis, the active state provides an animation, where the default-state-color expands circular from the center to the edges. Outline-button and plain-button  expand in transparent background.**

**Design**
https://www.figma.com/file/YfEUBOHk4J4nYrk04geswG/onyx---Design-System-Figma-Library?node-id=633%3A20349&mode=dev

**Acceptance criteria**
- [x] When the button is clicked there should be bubbled animation like in Figma
- [x] Check if the animation works on touch device

**Definition of Done**
 - [x] covered by component tests
 - [x] updated version + documentation is deployed
 - [x] Storybook can show the feature

**Review**
- [Storybook link](https://storybook.onyx.schwarz/?path=/story/buttons-button--primary)
- [x] approved by designer @jannick-ux 
- [x] approved by dev @BoppLi 

**Implementation details**

![Image](https://github.com/SchwarzIT/onyx/assets/60856270/bf22f1b9-5d30-4435-9cf6-55f5f77f0377)

**Reference implementations**
https://stackoverflow.com/questions/23934749/fill-element-from-center-on-hover

PR #1609, opened 2024-07-24
feat: implement OnyxButton active animation

oemueller:
Relates to #363 

Add ripple active animation to `OnyxButton`.

## Checklist

- [x] Playwright test is added


packages/sit-onyx/src/components/OnyxButton/OnyxButton.vue
```diff
@@ -185,6 +197,24 @@ const emit = defineEmits<{
...
+      #{$p}__icon {
+        position: relative;
+        z-index: 1;
+        pointer-events: none;
+      }
+      #{$p}__loading {
+        pointer-events: none;
+        z-index: 1;
+      }
+    }
```
larsrickert, line 216: Could you explain why this is needed? :)

packages/sit-onyx/src/components/OnyxButton/types.ts
```diff
@@ -35,6 +35,10 @@ export type OnyxButtonProps = DensityProp &
      * Whether to show a skeleton button.
      */
     skeleton?: boolean;
+    /**
+     * Whether to enable ripples on interaction
+     */
+    rippleAnimation?: boolean;
```
larsrickert, line 41: The button should always have the ripple animation, so we don't need to provide a property to disable it

packages/sit-onyx/src/components/OnyxRipple/OnyxRipple.stories.ts
```diff
@@ -0,0 +1,24 @@
+import { defineStorybookActionsAndVModels } from "@sit-onyx/storybook-utils";
+import type { Meta, StoryObj } from "@storybook/vue3";
+import OnyxRipple from "./OnyxRipple.vue";
+
+/**
+ * OnyxRipple provides visual feedback when a user interacts with the component. Supporting both touch and mouse events.
+ */
+const meta: Meta<typeof OnyxRipple> = {
+  title: "Support/OnyxRipple",
```
larsrickert, line 9:
```suggestion
  title: "Support/Ripple",
```

Also the example preview looks quiet empty, maybe we can add a button/text or similar to showcase it?

packages/sit-onyx/src/components/OnyxRipple/types.ts
```diff
@@ -0,0 +1,6 @@
+export type OnyxRippleProps = {
+  enterDuration?: string;
+  leaveDuration?: string;
```
larsrickert, line 3: I think for now its fine to define a fixed duration so its uniform across all components :)

We have some pre-defined durations from UX, maybe we can use `var(--onyx-duration-sm)`?
image

packages/sit-onyx/src/components/OnyxRipple/types.ts
```diff
@@ -0,0 +1,6 @@
...
+  color?: string;
```
larsrickert, line 4: Would prefer to set this via a CSS variable instead of a prop

packages/sit-onyx/src/components/OnyxRipple/OnyxRipple.vue
```diff
@@ -0,0 +1,90 @@
...
+  terminateOnPointerUp: false,
+});
+
+const config = computed<RippleConfig>(() => ({
+  color: props.color,
+  container: rippleTrigger,
+  terminateOnPointerUp: props.terminateOnPointerUp,
+}));
+
+const rippleTrigger = ref<HTMLElement>();
```
larsrickert, line 19: This is accesed in the computed before its defined :) Can you move it before the computed?

packages/sit-onyx/src/components/OnyxRipple/OnyxRipple.vue
```diff
@@ -0,0 +1,90 @@
...
+
+const { ripples, hideRipple, events } = useRipple(config);
+</script>
+
+<template>
+  <span ref="rippleTrigger" class="onyx-ripple" v-on="events">
+    <transition-group name="onyx-ripple" @after-enter="hideRipple($event as HTMLElement)">
+      <span
+        v-for="[key, r] in ripples"
```
larsrickert, line 28: Why are there multiple riples? One should be sufficient :)

packages/sit-onyx/src/components/OnyxRipple/OnyxRipple.vue
```diff
@@ -0,0 +1,90 @@
...
+
+const config = computed<RippleConfig>(() => ({
+  color: props.color,
+  container: rippleTrigger,
+  terminateOnPointerUp: props.terminateOnPointerUp,
+}));
+
+const rippleTrigger = ref<HTMLElement>();
+
+const { ripples, hideRipple, events } = useRipple(config);
```
larsrickert, line 21: In combination with the other comments in this PR, would it be possible to make the animation CSS only?

larsrickert (changes requested):
Thanks for your contribution!

packages/sit-onyx/src/components/OnyxButton/OnyxButton.vue
oemueller: pointer-events: none is needed so the Ripple can recieve MouseEvents. I managed to remove the z-index.

packages/sit-onyx/src/components/OnyxRipple/OnyxRipple.stories.ts
oemueller: added an generic example via template

packages/sit-onyx/src/components/OnyxRipple/types.ts
oemueller: changed it, duration-sm is fine for enter, for leave it's way too long, I kept 100ms

packages/sit-onyx/src/components/OnyxRipple/OnyxRipple.vue
oemueller: it's inspired by Material designs ripples, it feels good that you get a fresh ripple on every interaction 🙂. Most of the time there won't be more than 1 visible at a time.

packages/sit-onyx/src/components/OnyxButton/types.ts
oemueller: should I also remove the ripple-animation css class?

packages/sit-onyx/src/components/OnyxRipple/types.ts
```diff
@@ -0,0 +1,6 @@
...
+  terminateOnPointerUp?: boolean;
```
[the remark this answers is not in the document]
oemueller: I removed it, normally we would not set it. So no props left here 🙂

packages/sit-onyx/src/components/OnyxRipple/OnyxRipple.vue
oemueller: the animation is CSS only, only the creation of the ripples and some small calculations are Javascript.

packages/sit-onyx/src/components/OnyxButton/types.ts
BoppLi: Hint: should be disabled when the button is disabled

larsrickert: > should I also remove the ripple-animation css class?

If its no longer needed yes :)

packages/sit-onyx/src/components/OnyxButton/OnyxButton.stories.ts
```diff
@@ -73,6 +73,7 @@ export const WithIcon = {
   args: {
     label: "Button",
     icon: checkSmall,
+    rippleAnimation: true,
```
larsrickert, line 76: This is a leftover, right?

packages/sit-onyx/src/components/OnyxButton/OnyxButton.vue
```diff
@@ -185,6 +195,22 @@ const emit = defineEmits<{
...
+    $p: &;
+
+    &--ripple-animation {
+      &:hover:enabled {
+        --onyx-button-background-color: var(--onyx-button-background-hover-color);
+      }
+      #{$p}__label,
+      #{$p}__icon {
+        position: relative;
+        pointer-events: none;
```
larsrickert, line 207: When using `pointer-events: none` on the label, it is no longer markable/copyable via mouse...do you know if there is a way to get rid of this? Maybe by using the click handler on the parent?

packages/sit-onyx/src/components/OnyxRipple/types.ts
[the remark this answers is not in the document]
larsrickert: Nice thanks, can you delete the file/props then?

packages/sit-onyx/src/components/OnyxRipple/OnyxRipple.ct.tsx
```diff
@@ -0,0 +1,15 @@
+import { expect, test } from "../../playwright/a11y";
+import OnyxRipple from "./OnyxRipple.vue";
+
+test("should trigger some ripples", async ({ mount }) => {
+  // ARRANGE
+  const component = await mount(
+    <OnyxRipple enter-duration="2000ms" style="width: 500px; height: 500px;" />,
```
larsrickert, line 7:
```suggestion
    <OnyxRipple style="width: 500px; height: 500px;" />,
```

The prop no longer exists, right?

packages/sit-onyx/src/composables/useRipple.ts
```diff
@@ -0,0 +1,86 @@
...
+    const offsetX = x - containerRect.value.left;
+    const offsetY = y - containerRect.value.top;
+
+    const obj = {
+      left: offsetX - radius + "px",
+      top: offsetY - radius + "px",
+      radius: Math.round(radius * 2) + "px",
+      time: Date.now(),
+      backgroundColor: config.value.color,
+      rippleId: "d" + Date.now(),
```
larsrickert, line 47: We could also use 
`import { createId } from "@sit-onyx/headless"` for this

packages/sit-onyx/src/utils/math.spec.ts
```diff
@@ -0,0 +1,23 @@
...
+    bottom: 40,
+    width: 40,
+    height: 40,
+    toJSON: vi.fn(),
+  };
+
+  test("should calculate the distance to the furthest corner", () => {
+    const x = 20;
+    const y = 17;
+    const distance = distanceToFurthestCorner(x, y, rect as DOMRect);
```
larsrickert, line 20:
```suggestion
  const rect: DOMRect = {
    x: 0,
    y: 0,
    left: 0,
    top: 0,
    right: 40,
    bottom: 40,
    width: 40,
    height: 40,
    toJSON: vi.fn(),
  };

  test("should calculate the distance to the furthest corner", () => {
    const x = 20;
    const y = 17;
    const distance = distanceToFurthestCorner(x, y, rect);
```

packages/sit-onyx/src/composables/useRipple.ts
```diff
@@ -0,0 +1,86 @@
+import { computed, reactive, ref, type Ref } from "vue";
+import { distanceToFurthestCorner } from "../utils/math";
+
+export type RippleConfig = {
+  color: string;
+  terminateOnPointerUp: boolean;
```
larsrickert, line 6: I think this prop is unused so we can remove it :)

packages/sit-onyx/src/components/OnyxButton/OnyxButton.vue
```diff
@@ -41,7 +42,8 @@ const emit = defineEmits<{
     :autofocus="props.autofocus"
     @click="emit('click')"
   >
-    <OnyxIcon v-if="props.icon && !props.loading" :icon="props.icon" />
+    <OnyxRipple />
```
larsrickert, line 45: Can you disable the ripple when the button is disabled or loading?

packages/sit-onyx/src/composables/useRipple.ts
```diff
@@ -0,0 +1,86 @@
...
+  left: string;
+  top: string;
+  radius: string;
+  time: number;
+  backgroundColor: string;
+  rippleId: string;
+  fadeIn: boolean;
+};
+
+export const useRipple = (config: Ref<RippleConfig>) => {
```
larsrickert, line 20: Is it possible to also implement a reverse animation on mouse up? So the ripple "shrinks" back instead of hiding instantly

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxButton/OnyxButton.vue
oemueller: ok now I'm exposing the event handlers to the parent - so the Button in this case. And the parent registers them.
Not sure if there's a nicer way ...

packages/sit-onyx/src/composables/useRipple.ts
oemueller: I changed it like that .. I personally like the other version more, maybe you can decide with Jannick 🙂

packages/sit-onyx/src/components/OnyxButton/OnyxButton.vue
```diff
@@ -22,14 +24,18 @@ const emit = defineEmits<{
    */
   click: [];
 }>();
+
+const rippleRef = ref<ComponentInstance<typeof OnyxRipple> | null>(null);
+
+const rippleEvents = computed(() => (rippleRef.value ? rippleRef.value.events : undefined));
```
larsrickert, line 30:
```suggestion
const rippleRef = ref<ComponentInstance<typeof OnyxRipple>>();
const rippleEvents = computed(() => rippleRef.value?.events);
```

packages/sit-onyx/src/components/OnyxButton/OnyxButton.vue
```diff
@@ -22,14 +24,18 @@ const emit = defineEmits<{
...
 </script>
 
 <template>
   <OnyxSkeleton v-if="props.skeleton" :class="['onyx-button-skeleton', densityClass]" />
   <button
     v-else
     :class="[
-      'onyx-button',
+      'onyx-button onyx-button--ripple-animation',
```
larsrickert, line 38:
```suggestion
      'onyx-button',
```

packages/sit-onyx/src/components/OnyxRipple/OnyxRipple.stories.ts
```diff
@@ -0,0 +1,37 @@
...
+  }),
+};
+
+export default meta;
+type Story = StoryObj<typeof OnyxRipple>;
+
+/**
+ * Default properties like color and durations can be overridden. Click anywhere to create a ripple.
+ */
+export const Default = {
```
larsrickert, line 23: The example does not work for me

packages/sit-onyx/src/components/OnyxRipple/OnyxRipple.vue
```diff
@@ -0,0 +1,86 @@
...
+
+defineExpose({
+  events,
+});
+</script>
+
+<template>
+  <span ref="rippleTrigger" class="onyx-ripple">
```
larsrickert, line 20:
```suggestion
  <span ref="rippleTrigger" class="onyx-ripple" aria-hidden="true">
```

Should we hide the ripple for screen readers? Since its only an animation

packages/sit-onyx/src/components/OnyxRipple/OnyxRipple.vue
```diff
@@ -0,0 +1,86 @@
...
+        }"
+        :data-rippleid="key"
+      ></span>
+    </transition-group>
+  </span>
+</template>
+
+<style lang="scss">
+@use "../../styles/mixins/layers.scss";
+@use "../../styles/mixins/density.scss";
```
larsrickert, line 41: The density import is unused :)

Also can you wrap the styles in `@include layers.component() {}` as described [here](https://onyx.schwarz/principles/contributing.html#developing-components)? :)

packages/sit-onyx/src/components/OnyxRipple/types.ts
larsrickert, line 1: Can this file be deleted?

packages/sit-onyx/src/composables/useRipple.ts
larsrickert: Yes I agree that it looks a little weird...I will check this with Jannick

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxRipple/OnyxRipple.vue
oemueller: yes why not, although the **span** should not be shown/interpreted by any screen reader

packages/sit-onyx/src/components/OnyxRipple/OnyxRipple.stories.ts
oemueller: I see, thats because the events are now triggered on the parent. 

shall we link to the OnyxButton story ?
I guess we won't be able to trigger anything exposed by the Ripple component ?

packages/sit-onyx/src/composables/useRipple.ts
larsrickert: Sorry, it was a misunderstanding on my side. As discussed with Jannick, your previous behaviour is correct

packages/sit-onyx/src/components/OnyxRipple/OnyxRipple.stories.ts
larsrickert: Yes thats fine for me

larsrickert (approved):

PR #1644, opened 2024-07-26
feat: improve headless typings

### LINKED ISSUE #1643: Implement improved headless typings

JoCa96:
Closes #1643

- Replaced incorrect usage of `HtmlHTMLAttributes` with `HTMLAttributes`: `HTMLAttributes` is the correct supertype, while `HtmlHTMLAttributes` actually refers to attributes of an `html` element.
- implemented `createElRef`, which creates a special writeable computed that references a DOM Element and unwraps components. 
- `createElRef` returns a `HeadlessElRef`, which is a special subtype of `WritableComputedRef`. It uses a type differentiator, so no other `Ref` types can be assigned to it.
- updated `createBuilder` `HeadlessElement` typings to only accept `HeadlessElRef` for `ref` keys. So `createElRef` must be used to create these.


larsrickert (changes requested):
Can you add a breaking change changeset for your changes?

JoCa96:
> Can you add a breaking change changeset for your changes?

Done

larsrickert (approved):

PR #1681, opened 2024-08-01
fix playground

### LINKED ISSUE #1489: Fix playground doesn't seem to use the selected onyx version
### Repro steps

1. Open this [Playground example](https://playground.onyx.schwarz/#eNp9Uk1v1DAQ/Ssjn3fXQnBaQsWHKgEHWlEuSJaQiWcTF8e27EnYarX/nbGTbQNCvXnmvZl57yUn8S7G3TSi2Ismt8lGAqd990YJykpARhrjlfJ2iCERnODGPxw/ojbOetzU6o4wRkxwhkMKAyiRLW0DA0q8Vr6R81bewQXhEJ0m5AqgMXaC1umc+VoZ2HbJmm0bPGnenpSoNCauj4It9P4Fo7dOP3QpjN40ck25jMWr72GEVntAYwmoR2iDQQi+vh0eCLQ3bBLnhp0QEubRUYYeE+4aWcyvVCxmlwsAC1TPr7FGsjl+NXJlWXmxEXOS20HH3X0OnnM/lQG1ABz6Hmqn9J6i3HPVE8W8l7I1nmcNst6080jSx0FeqG9/Imm+nklab/DITCXKvrPyZ75PmQM+2O6f620YonWYbiLZ4P9WoZ0Lvz/XHqURN5d+22P76z/9+7wIvuUwMU2oxCNGOnVIM3x99wWP/H4Eh2BGx+xnwK+YgxuLxpn2nr8+y17xqtpPNUzru2/5+kjo88VUEVrTqHwl+Nf/8Iz1J7kvd69WKf6YMJWdHGD5uJnE+Q/40Rr9)
2. The example was created with the `alpha` version of onyx, but instead the newest version is selected
3. When the version is manually changed to an `alpha` version, the `OnyxStepper` shouldn't be rendered, as it didn't exist in that version. Its HTML is rendered anyway, even tough without the correct styling.

### Expectation

- The selected version should sync to the URL and loaded when a playground is opened with this link
- The selected version of `sit-onyx` should be actually loaded (e.g., `OnyxStepper` shouldn't be rendered and give an error when an alpha version is selected).

larsrickert:
fixes #1489

- fix `Module '"sit-onyx"' has no exported member 'OnyxHeadline'` error
- version selects: show "beta" or "latest" version initially instead of concrete version number
- support search for version selects
- persist selected onyx and typescript version (vue version is already persisted)

I created a bug report for the `@vue/repl` package: https://github.com/vuejs/repl/issues/269


apps/playground/src/composables/useStore.ts
```diff
@@ -50,16 +52,32 @@ export const useStore = () => {
...
         return { "sit-onyx": version };
       }),
     },
     // initialize repl with previously serialized state
     location.hash.slice(1),
   );
 
+  watch(onyxVersion, (newVersion) => {
+    updateQueryParam("onyxVersion", newVersion);
+  });
```
BoppLi, line 67: question: shouldn't we use immediate here, so even when you currently make a playground demo with the latest version and use htat for a bug report, that currently latest version will be included in the URL?

BoppLi: or would this cause a loop when you open a link and it tries to write it, which triggers the writing again and again ..?

apps/playground/src/composables/useStore.ts
```diff
@@ -50,16 +52,32 @@ export const useStore = () => {
...
+  watch(
+    () => store.typescriptVersion,
+    (newVersion) => {
+      updateQueryParam("typescriptVersion", newVersion);
+    },
+  );
+
+  const updateQueryParam = (key: string, value: string) => {
+    const url = new URL(location.href);
```
BoppLi, line 76: should there be an early return if the same version is set again, the URL/history does not need to update?
e.g. our select emits update:modelValue even when you click the current value again, or maybe it's even to prevent the possible loop mentioned in the other comment.
image

BoppLi (approved):

apps/playground/src/composables/useStore.ts
larsrickert: Vue will already take care of this so if you have a ref with value "beta" and you change it to "beta", no watcher or reactive updates will be triggered

apps/playground/src/composables/useStore.ts
larsrickert: Sure we could do this

BoppLi (approved):
Thank you! 😊

Issue #1697, opened 2024-08-06
Fix missing matrix screenshots for OnyxRadioGroup


BoppLi:
https://teams.microsoft.com/l/message/19:3a7010e95b8749edac6271ea346a7bc4@thread.tacv2/1722925096614?tenantId=d04f4717-5a6e-4b98-b3f9-6918e0385f4c&groupId=969ca3b3-9c78-411f-8629-eb770bbfdc7b&parentMessageId=1722925096614&teamName=onyx&channelName=Developers&createdTime=1722925096614

![Image](https://github.com/user-attachments/assets/095f2aae-7544-4cfe-b551-7d982a3a83fe)

larsrickert:
@jannick-ux I renamed the `headline` property to `label` for the radio and checkbox group so its uniform with our other form components. Can you also rename it in Figma?

PR #1707, opened 2024-08-06
fix: add required aria label for `OnyxRadioGroup` and `OnyxCheckboxGroup`

larsrickert:
Relates to #1697

- add matrix screenshot tests for radio group
- according to the [checkbox](https://www.w3.org/WAI/ARIA/apg/patterns/checkbox/) and [radio group](https://www.w3.org/WAI/ARIA/apg/patterns/radio/), the group should have an aria label. I removed the current optional `headline` property and replaced it with a required `label` prop (+ optional `hideLabel`) like we already have for all other form components

## Checklist


packages/sit-onyx/playwright/snapshots/components/OnyxRadioGroup/Radio-group-vertical--chromium-linux.png
BoppLi, line 1: could you use "Dummy 1" so it looks more like a real-world-ish example? ^^

packages/sit-onyx/src/components/OnyxCheckboxGroup/OnyxCheckboxGroup.vue
```diff
@@ -52,9 +52,13 @@ const checkAllLabel = computed(() => {
...
-  <fieldset :class="['onyx-checkbox-group', densityClass]" :disabled="props.disabled">
-    <legend v-if="props.headline" class="onyx-checkbox-group__label">
-      <OnyxHeadline is="h3">{{ props.headline }}</OnyxHeadline>
+  <fieldset
+    :class="['onyx-checkbox-group', densityClass]"
+    :disabled="props.disabled"
+    :aria-label="props.label"
+  >
+    <legend v-if="!props.hideLabel" class="onyx-checkbox-group__label">
+      <OnyxHeadline is="h3">{{ props.label }}</OnyxHeadline>
```
BoppLi, line 61: in case this is visible, shouldn't aria-label be udnefined and instead e.g. aria-described-by can be used pointing to the legend?
I have the feeling that else, a screen reader might read out the label twice.

packages/sit-onyx/src/components/OnyxRadioGroup/OnyxRadioGroup.vue
```diff
@@ -38,10 +37,11 @@ const handleChange = (selected: boolean, value: TValue) => {
     :class="['onyx-radio-button-group', densityClass, requiredTypeClass]"
     :disabled="props.disabled"
     role="radiogroup"
+    :aria-label="props.label"
```
BoppLi, line 40: see other comment on checkbox group

packages/sit-onyx/src/components/OnyxRadioGroup/types.ts
```diff
@@ -22,11 +24,6 @@ export type OnyxRadioGroupProps<TValue extends SelectOptionValue = SelectOptionV
        * The selected radio button option.
        */
       modelValue?: TValue;
```
BoppLi, line 26: the previous comment made a good point, could you maybe extend the comment on `RequiredMarkerProp`, like `No marker will be visible if the label of the element is hidden.`

context:
> ```
>  /**
>   * Required mode: `optional` will show an `(optional)` text after the label for optional inputs.
>   * `required` will show an `*` indicator for required inputs after the label instead.
>   * No marker will be visible if the label of the element is hidden.
>   * @default undefined By default the parents setting is used, if none is defined on any `required` is the default.
>   */
>  requiredMarker?: RequiredMarkerType;
> ```

BoppLi (changes requested):

packages/sit-onyx/src/components/OnyxCheckboxGroup/OnyxCheckboxGroup.vue
larsrickert: I checked this with Jonathan, he said thats fine :)

BoppLi (approved):

Issue #1734, opened 2024-08-12
Fix select bugs


larsrickert:
OnyxSelect bugs I noticed when working on #1722:

1. Clicking the gap between label and icon or clicking on the padding does not open the select. Only when clicking **exactly** on the label or icon

![Image](https://github.com/user-attachments/assets/6a1d0ff1-3bfd-4604-8a62-3f0ec105302d)

2. When changing the value programatically, the page jumps (because the select tries to scroll into view) although its not opened

https://github.com/user-attachments/assets/322faac2-ad46-4f4e-a91b-649aa914f14d

larsrickert:
@MajaZarkova When opening the select, it does scroll but not "far enough" so the select option is not visible


![Image](https://github.com/user-attachments/assets/0f9ae163-36d3-436d-91ff-7b4000739777)

MajaZarkova:
> [@MajaZarkova](https://github.com/MajaZarkova) When opening the select, it does scroll but not "far enough" so the select option is not visible
> 
> ![Image](https://github.com/user-attachments/assets/0f9ae163-36d3-436d-91ff-7b4000739777)

Follow up bug ticket #1781

PR #1756, opened 2024-08-15
fix(OnyxSelect): Fix OnyxSelect opening behavior

MajaZarkova:
Relates to #1734

Clicking the gap between label and icon or clicking on the padding did not open the select.

## Checklist


packages/sit-onyx/src/components/OnyxSelectInput/OnyxSelectInput.vue
```diff
@@ -122,7 +122,7 @@ const blockTyping = (event: KeyboardEvent) => {
     ]"
     v-bind="rootAttrs"
   >
-    <OnyxFormElement v-bind="props" :error-messages="errorMessages">
+    <OnyxFormElement v-bind="props" :error-messages="errorMessages" @click="emit('click')">
```
larsrickert, line 125: This would also emit the click event when clicking on the tooltip, message, label etc. which I am not sure if this should be the case.

Maybe we need to limit it to the input area only?

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxSelectInput/OnyxSelectInput.vue
MajaZarkova: But then it won't open when you click on the icon

larsrickert: The icon is inside the `onyx-select-input__wrapper` so setting the click handler there should also work when clicking the icon :)

.changeset/khaki-roses-confess.md
```diff
@@ -0,0 +1,5 @@
+---
+"sit-onyx": patch
+---
+
+Fix OnyxSelect opening behavior
```
larsrickert, line 5:
```suggestion
fix(OnyxSelect): open flyout when clicking the input gap or padding
```

larsrickert (changes requested):

larsrickert (approved):

PR #1759, opened 2024-08-16
refactor: PR suggestions for #1745

larsrickert:
PR suggestions for #1745 :)

Adressed points:
- add `useOpenDirection` composable so we can re-use it in the future for other components
- remove flyout ref from calculations since it is not needed to check in which direction more space is available


packages/sit-onyx/src/composables/useOpenDirection.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  /**
+   * Recursively finds the first parent element with hidden overflow.
+   */
+  const findParentWithHiddenOverflow = (element?: Element): Element | undefined => {
+    if (!element) return undefined;
+
+    const style = getComputedStyle(element);
+    if (style.overflow === "hidden") {
+      // if the element has hidden overflow, the flyout would we cut off at this element so we need to use
```
MajaZarkova, line 33:
```suggestion
      // if the element has hidden overflow, the flyout would be cut off by this element so we need to use
```

packages/sit-onyx/src/composables/useOpenDirection.ts
```diff
@@ -0,0 +1,52 @@
...
+  };
+
+  return {
+    /**
+     * Direction in which the flyout etc. should open to.
+     */
+    openDirection,
+    /**
+     * Detects in which direction a flyout etc. should be opened, depending on the available space in each direction.
+     * Should only be called onBeforeMount oder later to support server side rendering.
```
MajaZarkova, line 48:
```suggestion
     * Should only be called onBeforeMount or later to support server side rendering.
```

Did you mean this?

MajaZarkova (changes requested):

MajaZarkova (changes requested):
Looks like the composable doesn't work correctly
image

MajaZarkova (approved):

MajaZarkova (approved):

PR #1767, opened 2024-08-19
fix(listbox): Fix auto scrolling behavior when it's closed

MajaZarkova:
Relates to #1734

A window auto scroll behavior was detected when programmatically changing the active option inside a listbox without opening it.

## Checklist

### DIFF
--- a/packages/headless/src/composables/comboBox/createComboBox.ts
+++ b/packages/headless/src/composables/comboBox/createComboBox.ts
@@ -227,7 +227,7 @@ export const createComboBox = createBuilder(
       multiple,
       controlled: true,
       activeOption,
-      isExpanded: isExpanded.value,
+      isExpanded,
       onSelect: handleSelect,
     });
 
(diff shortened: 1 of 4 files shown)


packages/headless/src/composables/comboBox/createComboBox.ts
```diff
@@ -227,6 +227,7 @@ export const createComboBox = createBuilder(
       multiple,
       controlled: true,
       activeOption,
+      isExpanded: isExpanded.value,
```
JoCa96, line 230: We need to pass the ref, otherwise isExpanded is not updated for the listbox

JoCa96 (changes requested):

packages/headless/src/composables/comboBox/createComboBox.ts
JoCa96:
```suggestion
      isExpanded,
```

JoCa96 (approved):

Issue #1198, opened 2024-05-28
Implement OnyxSelect flyout opening direction and alignment


BoppLi:
### Open question

- Should we also support to manually set a fixed open direction?

### Why?

The flyout of the `OnyxSelect` should detect in which direction it should open (top or bottom), depending on in which direction more space is available.
It should also support to align left/right if not using the full parent width

### Design

[Figma](https://www.figma.com/design/YfEUBOHk4J4nYrk04geswG/Onyx-Component-Library?node-id=4462-57174&m=dev)

### Acceptance criteria

- [x] if more space is available above the select, its opened to the top
- [x] if more space is available below the select, its opened to the bottom
- [x] by default the flyout uses the full width of the select
- [x] the width can optionally be limited and aligned left/right

### Definition of Done

- [x] covered by component tests
- [x] screenshot tests are updated
- [x] follow-up tickets were created if necessary -> #1764 
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature
- [x] Storybook code snippet of new/changed examples are checked that they are generated correctly
- [x] Set task to `In Approval`

**Approval**

- [Storybook top open direction](https://storybook.onyx.schwarz/?path=/story/form-select--with-top-open-direction)
- [Storybook Right alignment](https://storybook.onyx.schwarz/?path=/story/form-select--with-right-alignment)
- [x] approved by designer @jannick-ux
- [x] approved by dev @larsrickert 

### Implementation Details

- [ViusalViewport](https://developer.mozilla.org/en-US/docs/Web/API/VisualViewport)
- don't change the direction on scroll -> try out how it feels.
- prefer to solve the width things with CSS only

larsrickert:
@MajaZarkova When using the select with search, the flyout width increases to more than the content width, is this intended?


![Image](https://github.com/user-attachments/assets/66c7f4c8-6b8b-4c98-ace3-4cb710f53756)

PR #1777, opened 2024-08-22
fix(OnyxMiniSearch): Adjust the size to the search input

MajaZarkova:
Relates to #1198 

The native input component has a default property size with value 20, which is approximately 100px. This was preventing us to make the input smaller. I adjusted the size property to 1 for the purposes of OnyxMiniSearch component.

## Checklist


packages/sit-onyx/src/components/OnyxMiniSearch/OnyxMiniSearch.vue
```diff
@@ -45,13 +45,18 @@ defineExpose({
 </script>
 
 <template>
-  <div :class="['onyx-mini-search', densityClass]" v-bind="rootAttrs">
+  <div
+    :class="['onyx-mini-search', densityClass]"
+    v-bind="rootAttrs"
+    :style="{ '--onyx-placeholder-character-count': input?.placeholder.length }"
```
larsrickert, line 51:
```suggestion
    :style="{ '--onyx-placeholder-character-count': input?.placeholder.length }"
```

Using the placeholder like this is not reactive so it will not update if the language changes. Would prefer to create a computed for the placeholder text and use it here + as input placeholder

packages/sit-onyx/src/components/OnyxMiniSearch/OnyxMiniSearch.vue
```diff
@@ -97,6 +103,7 @@ defineExpose({
       font-style: normal;
       flex-grow: 1;
       min-width: 0;
+      width: calc(var(--onyx-placeholder-character-count) * 1ch);
```
larsrickert, line 106: Shouldn't it be min width so that the input width can be bigger than the placeholder?

Do we still need the `size` property with this approach?

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxMiniSearch/OnyxMiniSearch.vue
MajaZarkova: We need the size property because the native input component has a default property size with value 20, which is approximately 100px.

packages/sit-onyx/src/components/OnyxMiniSearch/OnyxMiniSearch.vue
```diff
@@ -45,13 +47,18 @@ defineExpose({
 </script>
 
 <template>
-  <div :class="['onyx-mini-search', densityClass]" v-bind="rootAttrs">
+  <div
+    :class="['onyx-mini-search', densityClass]"
+    v-bind="rootAttrs"
+    :style="{ '--onyx-placeholder-character-count': placeholder?.length }"
```
larsrickert, line 53:
```suggestion
    :style="{ '--onyx-placeholder-character-count': placeholder.length }"
```

The placeholder is always defined :)

larsrickert (dismissed):

larsrickert (approved):

Issue #939, opened 2024-04-16
Implement icon from figma downloading cli tool


mj-hof:
### Discussed in https://github.com/SchwarzIT/onyx/discussions/184

<div type='discussions-op-text'>

<sup>Originally posted by **JoCa96** January 23, 2024</sup>
Currently the designers need to maintain the Icons in two places: Figma & and our internal Schwarz Icon Repo. 
We should evaluate if its possible to use Figma as a Single-Source-of-Truth.</div>

### Why

Currently the designers need to maintain the Icons in two places: Figma & and our internal Schwarz Icon Repo. 
We should evaluate if its possible to use Figma as a Single-Source-of-Truth. The [Icon Library - UX Schwarz IT](https://www.figma.com/design/ZtlUPkO9ATUCV42Wkg7P0o/Icon-Library---UX-Schwarz-IT?node-id=3270-179&m=dev) Figma file should we used for the icon source and for the `@sit-onyx/icons` npm package, the icons should be imported automatically (on a schedule, e.g. daily) from this Figma file and released.

### Acceptance criteria
- [x] Download icon SVGs from Figma
- [x] Get meta data (category, icon name, alias names)
- [x] Optimise downloaded icons
- [x] If changes exists: Publish new version of `@sit-onyx/icons` with changeset + changelog
- [x] check when to run the pipeline regularly (e.g. daily, weekly etc.)
- [x] Implement generically (only download + meta data), so it can be used also for other icon libraries (e.g. pass Figma file as parameter etc.). Maybe similar or as enhancement of our Figma CLI utility?
- [x] is an onyx independent solution
- [x] [Icon docs](https://onyx.schwarz/icons.html) supports search by alias names 

#### Follow up tickets

- #1802
- #1803

**Approval**

- [Icon docs](https://onyx.schwarz/icons.html)
- [Figma CLI docs](https://onyx.schwarz/development/packages/figma-utils.html#import-icons)
- [x] approved by designer @jannick-ux
- [x] approved by dev @MajaZarkova 

### Implementation Details
[Figma REST API](https://www.figma.com/developers/api#get-file-nodes-endpoint)

larsrickert:
@MajaZarkova For review: #1813

PR #1794, opened 2024-08-26
feat(icons): implement changeset generation

larsrickert:
Relates to #939

Generate changeset with list of new, deleted and modified icons whenever we import icons from Figma so they are released as npm package.

I also removed UX from codeowners for the icons so they are not added as required reviewer. The UX review process will be done in Figma so we devs should be responsible to approve the technical import via CI to make sure everything is imported correctly.


packages/icons/scripts/changeset.ts
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  console.log("Wrote changeset");
+}
+
+/**
+ * Gets a list of tracked git filenames.
+ */
+async function gitLsFiles(flags: string) {
+  const { stdout } = await exec(`git ls-files --exclude-standard ${flags} src/assets`);
```
JoCa96, line 68: I would prefer to use [`git status --porcelain`](https://git-scm.com/docs/git-status#_output) 

> Give the output in an easy-to-parse format for scripts. This is similar to the short output, but will remain stable across Git versions and regardless of user configuration. See below for details.

Using that, we would also be able to handle renamed files.

JoCa96 (dismissed):

packages/icons/scripts/changeset.ts
JoCa96: But this is optional and can be done in a follow-up ticket or PR.

larsrickert: Good idea :) Done

MajaZarkova (approved):

Issue #1603, opened 2024-07-23
Implement removal and linting of native event name usage


JoCa96:
### Why?

Currently, we are re-emitting events like `click` or `change`.
This creates complications and confusion for our users:
- the emit names are already reserved by native HTML events
- the emit payload is not compatible
- the custom emits don't work with Vue [event](https://vuejs.org/guide/essentials/event-handling.html#event-modifiers) and [key](https://vuejs.org/guide/essentials/event-handling.html#key-modifiers) modifiers
- there is a performance overhead of listening for the native events and re-emitting them
- native events don't behave like Vue events (bubbling, etc.)

### Acceptance criteria

- [x] all onyx emits with native event names have been removed or renamed
- [x] a lint rule enforcing non-native event/emit name collisions is added. (can be a follow-up ticket)
- [x] a storybook helper/util is provided that allows relevant events to be logged

### Definition of Done

- [x] follow-up tickets were created if necessary
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature
- [x] Storybook code snippet of new/changed examples are checked that they are generated correctly

**Approval**

- Docs: New Storybook util is documented here: https://onyx.schwarz/development/packages/storybook-utils.html#withnativeeventlogging 
- Storybook: Most components now have an `Relevant HTML events` section, where native HTML events are documented: e.g.: 

1. Go To https://storybook.onyx.schwarz/?path=/story/form-input--default 
2. Open the "add-on sidebar" if necessary
3. Scroll to the `Relevant HTML events` section and you will find the documented elements

- [x] approved by dev @larsrickert 

### Implementation details

Make use of `eslint-plugin-vue/lib/utils`:

```js
const utils = require("eslint-plugin-vue/lib/utils");
```

#### Reference implementations

See the following as examples for similar lint rules:
- [vue-require-explicit-emits](https://eslint.vuejs.org/rules/require-explicit-emits.html#vue-require-explicit-emits)

MajaZarkova:
- [PR open](https://github.com/SchwarzIT/onyx/pull/1799)
- The PR doesn't have a changeset yet
- A follow up ticket should be created regarding this [comment](https://github.com/SchwarzIT/onyx/pull/1799#discussion_r1745203585)
- There is a replay on the other two comments with the issues I have encountered.
- The last point of this ticket (the part in Implementation details) is not implemented

PR #1799, opened 2024-08-26
feat(components): Remove/rename click event on every component

MajaZarkova:
Relates to #1603 

Remove/rename native event names on every component

## Checklist

### DIFF
--- a/packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxNavButton/OnyxNavButton.vue
+++ b/packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxNavButton/OnyxNavButton.vue
@@ -14,9 +14,9 @@ const props = withDefaults(defineProps<OnyxNavButtonProps>(), {
 
 const emit = defineEmits<{
   /**
-   * Emitted when the nav item is clicked (via click or keyboard).
+   * Emitted when the nav button is clicked (via click or keyboard).
    */
-  click: [href: string];
+  navButtonClick: [href: string];
 }>();
 
 const slots = defineSlots<{
@@ -38,7 +38,7 @@ const handleParentClick = () => {
   if (isMobile?.value && hasChildren.value && !isMobileChildrenOpen.value) {
     isMobileChildrenOpen.value = true;
   } else if (props.href) {
-    emit("click", props.href);
+    emit("navButtonClick", props.href);
   }
 };
 </script>
--- a/packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxNavItem/OnyxNavItem.vue
+++ b/packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxNavItem/OnyxNavItem.vue
@@ -11,7 +11,7 @@ const emit = defineEmits<{
   /**
    * Emitted when the nav item is clicked.
    */
-  click: [href: string];
+  navItemClick: [href: string];
 }>();
 
 defineSlots<{
@@ -27,7 +27,7 @@ defineSlots<{
     class="onyx-nav-item"
     :active="props.active"
     :href="props.href ?? 'javascript:void(0)'"
-    @click="props.href && emit('click', props.href)"
+    @click="props.href && emit('navItemClick', props.href)"
   >
     <slot>
       <span>{{ props.label }}</span>
--- a/packages/sit-onyx/src/components/OnyxSelectInput/OnyxSelectInput.vue
+++ b/packages/sit-onyx/src/components/OnyxSelectInput/OnyxSelectInput.vue
@@ -29,7 +29,7 @@ const emit = defineEmits<{
   /**
    * Emitted when the select input is clicked (and is not disabled).
    */
-  click: [];
+  selectInputClick: [];
   /**
    * Emitted when the validity state of the input changes.
    */
@@ -125,7 +125,7 @@ const blockTyping = (event: KeyboardEvent) => {
     <OnyxFormElement v-bind="props" :error-messages="errorMessages">
       <template #default="{ id }">
         <!-- eslint-disable-next-line vuejs-accessibility/click-events-have-key-events, vuejs-accessibility/no-static-element-interactions -- Disabled rules are needed here, because of the click event. Otherwise clicking on the padding or gap won't open the select -->
-        <div class="onyx-select-input__wrapper" @click="emit('click')">
+        <div class="onyx-select-input__wrapper" @click="emit('selectInputClick')">
           <OnyxLoadingIndicator
             v-if="props.loading"
             class="onyx-select-input__loading"
--- a/packages/sit-onyx/src/components/OnyxToastMessage/OnyxToastMessage.vue
+++ b/packages/sit-onyx/src/components/OnyxToastMessage/OnyxToastMessage.vue
@@ -21,7 +21,7 @@ const emit = defineEmits<{
   /**
    * Emitted when the toast is clicked. Requires `clickable` property to be enabled.
    */
-  click: [];
+  toastMessageClick: [];
   /**
    * Emitted when the toast should be closed.
    */
@@ -53,7 +53,7 @@ const icon = computed(() => {
     :class="[`onyx-toast-message--${props.color}`, densityClass]"
     :role="props.color === 'danger' || props.color === 'warning' ? 'alert' : 'status'"
     :aria-label="props.clickable ? props.headline : undefined"
-    @click="props.clickable && emit('click')"
+    @click="props.clickable && emit('toastMessageClick')"
   >
     <div class="onyx-toast-message__wrapper">
       <OnyxIcon v-if="icon" :icon="icon" />
--- a/packages/sit-onyx/src/components/OnyxButton/OnyxButton.vue
+++ b/packages/sit-onyx/src/components/OnyxButton/OnyxButton.vue
@@ -18,13 +18,6 @@ const props = withDefaults(defineProps<OnyxButtonProps>(), {
 
 const { densityClass } = useDensity(props);
 
-const emit = defineEmits<{
-  /**
-   * Emitted when the button is clicked (and is not disabled).
-   */
-  click: [];
-}>();
-
 const rippleRef = ref<ComponentInstance<typeof OnyxRipple>>();
 const rippleEvents = computed(() => rippleRef.value?.events ?? {});
 </script>
@@ -44,7 +37,6 @@ const rippleEvents = computed(() => rippleRef.value?.events ?? {});
     :type="props.type"
     :aria-label="props.loading ? props.label : undefined"
     :autofocus="props.autofocus"
-    @click="emit('click')"
     v-on="rippleEvents"
   >
     <OnyxRipple v-if="!props.disabled && !props.loading" ref="rippleRef" />
--- a/packages/sit-onyx/src/components/OnyxIconButton/OnyxIconButton.vue
+++ b/packages/sit-onyx/src/components/OnyxIconButton/OnyxIconButton.vue
@@ -18,11 +18,6 @@ defineSlots<{
   /** Slot for an custom icon. Will have no effect if property `icon` is passed. */
   default(): unknown;
 }>();
-
-const emit = defineEmits<{
-  /** Emitted when the button is clicked (and is not disabled). */
-  click: [];
-}>();
 </script>
 
 <template>
@@ -41,7 +36,6 @@ const emit = defineEmits<{
     ]"
     :disabled="props.disabled || props.loading"
     :autofocus="props.autofocus"
-    @click="emit('click')"
   >
     <OnyxLoadingIndicator v-if="props.loading" type="circle" />
     <OnyxIcon v-else-if="props.icon" :icon="props.icon" />
--- a/packages/sit-onyx/src/components/OnyxLink/OnyxLink.vue
+++ b/packages/sit-onyx/src/components/OnyxLink/OnyxLink.vue
@@ -9,13 +9,6 @@ const props = withDefaults(defineProps<OnyxLinkProps>(), {
   withExternalIcon: "auto",
 });
 
-const emit = defineEmits<{
-  /**
-   * Emitted when the link is opened (via click or keyboard).
-   */
-  click: [];
-}>();
-
 defineSlots<{
   /**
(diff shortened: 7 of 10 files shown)


packages/sit-onyx/src/components/OnyxIconButton/OnyxIconButton.ct.tsx
```diff
@@ -10,25 +10,19 @@ import OnyxIconButton from "./OnyxIconButton.vue";
...
 
   // ARRANGE
   const component = await mount(OnyxIconButton, setup);
 
   await test.step("clickable by default", async () => {
-    // ACT
-    await component.click();
     // ASSERT
-    expect(clicks).toBe(1);
+    await expect(component).toBeEnabled();
```
larsrickert, line 25: Why did you remove the test for the click events? Would prefer to still keep them so we can test that the click events are actually emitted

packages/sit-onyx/src/components/OnyxInput/OnyxInput.vue
```diff
@@ -94,9 +85,8 @@ const patternSource = computed(() => {
             :maxlength="props.maxlength"
             :aria-label="props.hideLabel ? props.label : undefined"
             :title="props.hideLabel ? props.label : undefined"
-            @change="handleChange"
-            @focus="emit('focus')"
-            @blur="emit('blur')"
+            @focus="emit('inputFocus')"
+            @blur="emit('inputBlur')"
```
larsrickert, line 89: Would prefer to still call them focus and blur and pass the native event with them

larsrickert: But thats my personal opinion, maybe @JoCa96 disagrees on that

packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxNavItem/OnyxNavItem.vue
```diff
@@ -27,7 +27,7 @@ defineSlots<{
     class="onyx-nav-item"
     :active="props.active"
     :href="props.href ?? 'javascript:void(0)'"
-    @click="props.href && emit('click', props.href)"
+    @click="props.href && emit('navItemClick', props.href)"
```
larsrickert, line 30: Should we remove the click event here so it bubbles the native click event?

packages/sit-onyx/src/components/OnyxSelectInput/OnyxSelectInput.vue
```diff
@@ -29,7 +29,7 @@ const emit = defineEmits<{
   /**
    * Emitted when the select input is clicked (and is not disabled).
    */
-  click: [];
+  selectInputClick: [];
```
larsrickert, line 32: What do you think about just `inputClick`? Since it should be clear that this component is the select :)

packages/sit-onyx/src/components/OnyxStepper/OnyxStepper.vue
```diff
@@ -110,8 +110,8 @@ const decrementLabel = computed(() => t.value("stepper.decrement", { stepSize: p
           :step="props.step"
           :title="props.hideLabel ? props.label : undefined"
           @change="wasTouched = true"
-          @focus="emit('focus')"
-          @blur="emit('blur')"
+          @focus="emit('inputFocus')"
+          @blur="emit('inputBlur')"
```
larsrickert, line 114: Same comment as for input

packages/sit-onyx/src/components/OnyxTextarea/OnyxTextarea.vue
```diff
@@ -112,9 +103,8 @@ const handleInput = (event: Event) => {
             :aria-label="props.hideLabel ? props.label : undefined"
             :title="props.hideLabel ? props.label : undefined"
             @input="handleInput"
-            @change="handleChange"
-            @focus="emit('focus')"
-            @blur="emit('blur')"
+            @focus="emit('textareaFocus')"
+            @blur="emit('textareaBlur')"
```
larsrickert, line 107: Same comment as for input

packages/sit-onyx/src/components/OnyxToastMessage/OnyxToastMessage.vue
```diff
@@ -21,7 +21,7 @@ const emit = defineEmits<{
   /**
    * Emitted when the toast is clicked. Requires `clickable` property to be enabled.
    */
-  click: [];
+  toastMessageClick: [];
```
larsrickert, line 24: Would prefer to still call this click but I guess thats not the intention of the ticket...

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxIconButton/OnyxIconButton.ct.tsx
MajaZarkova: I removed it because the test is failing. Event though the event is captured by Storybook

packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxNavItem/OnyxNavItem.vue
MajaZarkova: I remember that we wanted to emit the href on click. This is the reason why I renamed the event

packages/sit-onyx/src/components/OnyxIconButton/OnyxIconButton.ct.tsx
JoCa96: Please don't remove failing tests, we need to fix the issue instead

packages/sit-onyx/src/components/OnyxInput/OnyxInput.vue
JoCa96: The issue is that `focus` and `blur` do not bubble, so currently the user cannot listen for them on the `OnyxInput` component.
1. We don't provide any emits for `focus` and `blur`
**Reasoning:** While these don't bubble, they still have a capture phase, so `@focus.capture=...` and `@blur.capture=...` will still work. Also, `focusin` and `focusout` do bubble and could be used instead.
**Caveat:** Only works well for `OnyxInput`, where we have a single focusable element in our component.
2. We use `const { rootAttrs, restAttrs } = useRootAttrs();` and bind `restAttrs` to the input element.
**Reasoning:** This way all event-listeners and attributes of the user are bound to the input element. Also works for all other event-listeners.
**Caveat:** Might infer with our own aria-bindings 
3. Rename the emits with a common name for all our components: I would propose `activate` and `deactivate`. These event names are not colliding with any native event 
namings, which is the main purpose of the renaming.

I would personally prefer to do option 1 for now, create a follow-up ticket for 3 and wait for user-feedback.

packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxNavItem/OnyxNavItem.vue
```diff
@@ -11,7 +11,7 @@ const emit = defineEmits<{
   /**
    * Emitted when the nav item is clicked.
    */
-  click: [href: string];
+  navItemClick: [href: string];
```
JoCa96, line 14: Preferably we use a emit name that describes the intent of the user.
So I would prefer to rename the emit to `navigate` (also for the navbutton).
```suggestion
  navigate: [href: string];
```

packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxNavButton/OnyxNavButton.vue
```diff
@@ -14,9 +14,9 @@ const props = withDefaults(defineProps<OnyxNavButtonProps>(), {
 
 const emit = defineEmits<{
   /**
-   * Emitted when the nav item is clicked (via click or keyboard).
+   * Emitted when the nav button is clicked (via click or keyboard).
    */
-  click: [href: string];
+  navButtonClick: [href: string];
```
JoCa96, line 19: Preferably we use a emit name that describes the intent of the user.
So I would prefer to rename the emit to `navigate` (also for the navitem).
```suggestion
  navigate: [href: string];
```

packages/sit-onyx/src/components/OnyxSelectInput/OnyxSelectInput.vue
```diff
@@ -125,7 +125,7 @@ const blockTyping = (event: KeyboardEvent) => {
     <OnyxFormElement v-bind="props" :error-messages="errorMessages">
       <template #default="{ id }">
         <!-- eslint-disable-next-line vuejs-accessibility/click-events-have-key-events, vuejs-accessibility/no-static-element-interactions -- Disabled rules are needed here, because of the click event. Otherwise clicking on the padding or gap won't open the select -->
-        <div class="onyx-select-input__wrapper" @click="emit('click')">
+        <div class="onyx-select-input__wrapper" @click="emit('inputClick')">
```
JoCa96, line 128: We should be able to remove this click handler here without issues, this will also allow us to remove the "eslint" exception above

```suggestion
        <div class="onyx-select-input__wrapper">
```

packages/sit-onyx/src/components/OnyxToastMessage/OnyxToastMessage.vue
JoCa96: We should remove the event completely. When it is not clickable and the developer listens for a click event anyways, we won't be able to circumvent that anyways.
Alternatively I would propose a renaming to a more generic verb, maybe `trigger`?

JoCa96 (changes requested):

packages/sit-onyx/src/components/OnyxIconButton/OnyxIconButton.ct.tsx
MajaZarkova: check the native button for enabled instead

MajaZarkova: Our component represents a native button. I tried with 'component.getByLabel(...)', the test couldn't find the element.

packages/sit-onyx/src/components/OnyxSelectInput/OnyxSelectInput.vue
MajaZarkova: This is creating side effects with opening the select component.

packages/sit-onyx/src/components/OnyxInput/OnyxInput.vue
MajaZarkova: I removed the focus and blur events, but I haven't created a follow up ticket for the third point.

larsrickert (changes requested):
Could you please also include a changeset with all breaking changes so the user knows which events  he needs to rename when updating onyx

packages/sit-onyx/src/components/OnyxToastMessage/OnyxToastMessage.vue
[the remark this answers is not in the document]
JoCa96: Listening for a click event on a non-interactive element is a user error, and unfortunately something we cannot easily protect the user from. There is the [no-static-element-interactions lint rule](https://vue-a11y.github.io/eslint-plugin-vuejs-accessibility/rules/no-static-element-interactions.html), but I doubt that it is advanced enough to infer it here. We can consider offering an onyx-version of the lint rule, though.

Shadowing the native event is not a solution here, for the reasons already layed out in my [other comment](https://github.com/SchwarzIT/onyx/pull/1799/files/078d2e2ebd16de6ab5a28c8e803a55fd6d87031b#r1745203585).

Alternatively, we could prove a custom event name. But I don't think it's necessary until we get conflicting feedback.

packages/sit-onyx/src/components/OnyxNavBar/OnyxNavBar.vue
```diff
@@ -21,11 +21,11 @@ const emit = defineEmits<{
    * Emitted when the app area (where logo and app name are placed) is clicked.
    * Usually the user should be redirected to the home page then.
    */
-  appAreaClick: [];
+  navigateToStart: [event: MouseEvent];
   /**
    * Emitted when the back button is clicked.
    */
-  backButtonClick: [];
+  navigateBack: [event: MouseEvent];
```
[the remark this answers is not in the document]
JoCa96: As discussed: @jannick-ux confirmed that these two interactions are explicitly defined by the design, therefore the naming should mirror that :)

packages/sit-onyx/src/components/OnyxIconButton/OnyxIconButton.ct.tsx
JoCa96: done

JoCa96 (dismissed):

.changeset/hip-kids-decide.md
```diff
@@ -0,0 +1,5 @@
+---
+"sit-onyx": major
+---
+
+feat: Remove/Rename emits that collide with native event names
```
larsrickert, line 5:
```suggestion
feat: Remove/Rename emits that collide with native event names

- OnyxInput, OnyxTextarea and OnyxStepper: remove `focus` and `blur` event
- OnyxInput and OnyxTextarea: `change` event is now the native HTML event so to get the input value, change `$event` to `$event.target.value`
- OnyxNavBar: rename `appAreaClick` to `navigateToStart` and `backButtonClick` to `navigateBack`
- OnyxNavButton: rename `click` to `navigate`, also the native MouseEvent is now passed as second parameter
- OnyxNavItem: rename `click` to `navigate`
- OnyxRadioButton: `change` event is now the native HTML event so to get the input value, change `$event` to `$event.target.checked`
- OnyxSelectInput: rename `click` to `inputClick`
- OnyxToastMessage: `click` event is now the native HTML event so it will always be emitted, even when `clickable` prop is not set.
```

We should be as detailed as possible for the changelog, otherwise the user will not know how to migrate to the new version

larsrickert (changes requested):

.changeset/hip-kids-decide.md
JoCa96: Updated the changeset

larsrickert (approved):

Issue #1366, opened 2024-06-20
Adjust selected text contrast in dark mode


BoppLi:
Currently it is hard to detect which text is selected in input controls like OnyxInput, OnyxTextarea, OnyxSelect when darkmode is active:
![Image](https://github.com/SchwarzIT/onyx/assets/151019977/bcba9ac8-80f4-4f17-a18e-4cba7fa34aeb)

The currently used token is `--onyx-color-base-neutral-200`

Info: **text color** as well as background color can be adjusted however you like.
https://developer.mozilla.org/en-US/docs/Web/CSS/::selection
![Image](https://github.com/SchwarzIT/onyx/assets/151019977/599a505e-f7ce-41ab-b0fa-7b8df7b24d90)

jannick-ux:
In the figma library the default color 500 is used with an opacity of .30 (in darkmode) and .20 (in light mode).
Do we have to alignhere? Or what works best for dev?

jannick-ux:
Update: After a discussion in the onyx daily we detected, that this problem is only in read-only and disabled state. We decided to use the same marking color like with the default state (primary). Screenshot below.


![Image](https://github.com/user-attachments/assets/a78906fa-7bcb-41eb-8ff5-87e0a75df1f1)
![Image](https://github.com/user-attachments/assets/6b31f1ee-1b78-4234-b167-ffe7920b11b7)

PR #1827, opened 2024-09-04
fix(OnyxInput, OnyxTextarea): Adjust selected text color

MajaZarkova:
Relates to #1366 

The selected text background color for readonly and disabled state was not visible. We are keeping the same color as it is in the default state.

## Checklist


.changeset/tender-walls-walk.md
```diff
@@ -0,0 +1,5 @@
+---
+"sit-onyx": patch
+---
+
+fix(OnyxInput, OnyxTextarea): Adjust selected text color
```
JoCa96, line 5: Please use a descriptive text that lets people understand the changes without additional context:
```suggestion
fix(OnyxInput, OnyxTextarea): Adjust selection highlighting for better contrast
```

I tend to reuse the text from the changeset in the PR description, that way you're don't need to write it twice :)

packages/sit-onyx/src/styles/mixins/input.scss
```diff
@@ -49,7 +49,6 @@
 
     // :read-only is valid for readonly and disabled state so we put shared styles for both states here
     &:has(#{$base-selector}__native:read-only) {
-      --selection-color: var(--onyx-color-base-neutral-200);
```
JoCa96, line 52: Please add a screenshot test, so that we can avoid regressions

JoCa96 (changes requested):

JoCa96 (approved):

Issue #1783, opened 2024-08-22
Implement updated components details for onyx.schwarz


JoCa96:
### Why?

The component implementation status details on https://onyx.schwarz/ are out of date.
Show 6 rows of components instead of 3 - the rest can be triggered by "show more" function.

Update the mention components below following dates and status:

#### 07/2024 - Implemented
- [x] Nav bar
- [x] Select
- [x] Tooltip
- [x] Stepper

#### 10/2024 - In progress
- [x] Pagniation
- [x] Headlines

#### 11/2024 - In progress
- [x] Search
- [x] Filters
 - [x] Table

#### 12/2024 - In progress
- [x] Data grid
- [x] Sidebar

#### 01/2025 - Planned
- [x] Datepicker
- [x] Timepicker
- [x] Dialog

### Acceptance criteria

- [x] The component implementation status are updated

### Definition of Done

- [x] updated version + documentation is deployed

**Approval**

- Deployed Docs at https://onyx.schwarz/ (Scroll down to components)
- [x] approved by ux @jannick-ux

JoCa96:
Issue with `eslint-plugin-vue`: https://github.com/vuejs/eslint-plugin-vue/issues/2557

PR #1856, opened 2024-09-11
chore: update component card data

JoCa96:
Relates to #1783 

- Update the implementation status of the component overview on the main page according to #1783.
- Require `href` of `ComponentCardProps` to be defined when the component state is `implemented`.


apps/docs/src/.vitepress/components/ComponentCard.vue
```diff
@@ -18,7 +14,21 @@ export type ComponentCardProps = {
...
-};
+} & (
+  | {
+      /** Component status. */
+      status: Extract<ComponentStatus, "in-progress" | "planned">;
+      /** Link to the components storybook. */
+      href?: string;
+    }
+  | {
+      status: Exclude<ComponentStatus, "in-progress" | "planned">;
```
larsrickert, line 25:
```suggestion
      status: Exclude<ComponentStatus, "implemented">;
      /** Link to the components storybook. */
      href?: string;
    }
  | {
      status: Extract<ComponentStatus, "implemented">;
```

Isn't it easier to just pick the implemented status instead of excluding all others?

larsrickert (approved):

PR #1880, opened 2024-09-17
chore: add relevant events logging

JoCa96:
A new Storybook Util `withNativeEventLoggingFor` is introduced.
It generates Storybook `ArgTypes` that allow us to document and log relevant native events in Storybook.
These can be merged with other `ArgTypes`.
The native events will be documented in a new section "Relevant HTML Events".

Relates to #1603


.changeset/tender-zebras-flow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sit-onyx/storybook-utils": minor
+---
+
+feat: added withNativeEventLogging to log and document native events
```
larsrickert, line 5: Should we also document this utility in our docs?

packages/storybook-utils/src/actions.ts
```diff
@@ -23,6 +23,40 @@ export const enhanceEventArgTypes: ArgTypesEnhancer = ({ argTypes }) => {
...
+ * ```
+ *
+ * @param relevantEvents a list of event names that should be logged
+ * @returns Storybook ArgTypes object
+ */
+export const withNativeEventLogging = (relevantEvents: (keyof Events)[]) =>
+  relevantEvents.reduce((argTypes, eventName) => {
+    const action = eventName.replace(/^on/, "").toLowerCase();
+    argTypes[eventName] = {
+      name: eventName,
```
larsrickert, line 49:
```suggestion
      name: action,
```

Would prefer to show the event with the on prefix

packages/storybook-utils/src/actions.ts
```diff
@@ -23,6 +23,40 @@ export const enhanceEventArgTypes: ArgTypesEnhancer = ({ argTypes }) => {
...
+      control: false,
+      description: `[${action}_event](https://developer.mozilla.org/en-US/search?q=element+${action}+event)`, // unfortunately there is no static path for all events
```
larsrickert, line 51:
```suggestion
      description: `Native HTML [${action}](https://developer.mozilla.org/en-US/search?q=element+${action}+event) event`, // unfortunately there is no static path for all events
```

larsrickert, line 51: (optional) we could maintain a map of "real" links for very common events like click, change, focus etc. and use the generic search link for all other events, what do you think?

If its too much manual maintenance we don't have to do it :)

larsrickert (changes requested):

packages/storybook-utils/src/actions.ts
JoCa96: Done, I had some fun with typescript :D

packages/storybook-utils/src/events.ts
```diff
@@ -0,0 +1,777 @@
...
+ * Removes a prefix from a string type.
+ * @example ```ts
+ * type Trimmed = TrimStart<"onclick", "on">;
+ * const trimmed: Trimmed = "click";
+ * ```
+ */
+type TrimStart<
+  TString extends string,
+  Prefix extends string,
+> = TString extends `${Prefix}${infer Base}` ? Base : TString;
```
larsrickert, line 28:
```suggestion
  TPrefix extends string,
> = TString extends `${TPrefix}${infer TBase}` ? TBase : TString;
```

Isn't it a convention to prefix generic types with T?

packages/storybook-utils/src/events.ts
```diff
@@ -0,0 +1,777 @@
...
+
+/**
+ * The doc type ensures that a name has a valid URL that includes the name.
+ */
+type Docs<Name extends string> = {
+  name: Name;
+  url: `${string}${Name}${string}`;
+};
+
+export const EVENT_DOC_MAP: EventDocMap = {
```
larsrickert, line 45: Nice TypeScript magic :D
Couldn't we map the event.name from the key? If I see correctly, the EVENT_DOC_MAP is always the event.name with the `on` prefix

larsrickert (changes requested):
👏

packages/storybook-utils/src/events.ts
larsrickert: I meant the keys from our own `EVENT_DOC_MAP` :) So when we use it in the Storybook util, we could extract the event name from the key so we don't need to maintain them manually

larsrickert (approved):

PR #1898, opened 2024-09-25
chore: improve docs page performance by only showing default example

JoCa96:
There is a noticeable lag on when typing into the input of the [`OnyxInput`](https://storybook.onyx.schwarz/?path=/docs/form-input--docs).

### Cause
I am concluding that the issue lie with Storybook:
- I performed a cursory performance check using the Vue Dev Tools and there are no apparent performance issues with our Vue components.
- The Storybook input elements are also affected

### Mitigation
I removed all, but the Default Story example from the docs page.
This should mitigate bigger lagging issues for now.


MajaZarkova (approved):

larsrickert:
@JoCa96 The issue with this "solution" is that the code snippets for most examples can now not be seen/copied which lowers their value quiet a lot :/

Issue #1614, opened 2024-07-24
Implement visual feedback in clickable vs non-clickable group-navItem


BoppLi:
Change the pointer cursor in the nav bar for a NavItem and ContextArea into arrow cursor 

[Example](https://storybook.onyx.schwarz/?path=/story/navigation-navbar--default&globals=viewport:md) - Item 2 is clickable (has a link itself), but Item 3 only has children but is not itself clickable.
Right now, both look exactly the same on desktop - both have cursor-pointer etc.
On mobile, we have a distinction but this would not work on desktop.
![Image](https://github.com/user-attachments/assets/ff4fb1f4-4ec0-4ee1-8342-4f007e7cf7f3)

mj-hof:
Is a indicator for nested items needed?

PR #1901, opened 2024-09-30
fix(OnyxNavBar): Replace cursor pointer with default

MajaZarkova:
Relates to #1614 

Replace cursor pointer with default for OnyxNavButton and it's nested items

## Checklist


JoCa96 (changes requested):
Now the NavButtons never show a `pointer` cursor.
I should still be shown for navbuttons, that have a direct action (aka no children)

JoCa96 (approved):

Issue #757, opened 2024-03-25
Implement OnyxForm component


JoCa96:
### Definition of Done

- The following component requirements are implemented:
  - [ ] `skeleton`
  - [ ] `density`
  - [ ] example usage added to `apps/alpha-test-app/src/views/HomeView.vue`
- [ ] covered by component tests
- [ ] screenshot tests are updated
- [ ] follow-up tickets were created if necessary
- [ ] updated version + documentation is deployed
- [ ] Storybook can show the feature
- [ ] Storybook code snippet of new/changed examples are checked that they are generated correctly
- [ ] Set task to `In Approval`

**Approval**

- Storybook 
- Deployed Docs 
- [ ] approved by designer @jannick-ux
- [ ] approved by dev

BoppLi:
From Retro: 

![Image](https://github.com/SchwarzIT/onyx/assets/151019977/a7cb6606-4233-4b54-b54e-92a6238c9e2b)

BoppLi:
brainstorming: modes for showing validation errors / validation on user side
- initial
- onInput/onChange
- onBlur
- onSubmit

PR #1902, opened 2024-09-30
feat: Implement OnyxForm with disabled state injection

JoCa96:
Relates to #757 

- Implement `OnyxForm` component: Wrapper around Form, which allows setting the `disabled` state for all child form elements
- Add E2E and unit tests
- Add Storybook Documentation
- Update all form elements to use injected `disabled` state


packages/sit-onyx/src/components/OnyxForm/OnyxForm.core.ts
```diff
@@ -0,0 +1,101 @@
...
+  computed(() => {
+    const prop = props[key] as FormInjected<FormInjectedProps[TKey]> | undefined;
+    if (prop != undefined && prop !== FORM_INJECTED_SYMBOL) {
+      return prop;
+    }
+    const formProp = formProps?.[key];
+    if (formProp !== undefined) {
+      return formProp;
+    }
+    return defaultValue;
```
larsrickert, line 60:
```suggestion
    return formProps?.[key] ?? defaultValue;
```

packages/sit-onyx/src/components/OnyxForm/OnyxForm.core.ts
```diff
@@ -0,0 +1,101 @@
...
+  });
+
+const createFormInjectionContext =
+  (formProps?: FormInjectedProps) =>
+  (
+    props: Reactive<LocalProps>,
+  ): { [TKey in keyof FormInjectedProps]: Ref<FormInjectedProps[TKey]> } => ({
+    get disabled() {
+      return createCompute(formProps, props, "disabled", false);
```
larsrickert, line 69: Like this, a new computed is created every time the value is accessed, right?

packages/sit-onyx/src/components/OnyxForm/OnyxForm.stories.ts
```diff
@@ -0,0 +1,46 @@
...
+import type { Meta, StoryObj } from "@storybook/vue3";
+import { h } from "vue";
+import { OnyxInput } from "../..";
+import OnyxButton from "../OnyxButton/OnyxButton.vue";
+import OnyxStepper from "../OnyxStepper/OnyxStepper.vue";
+import OnyxForm from "./OnyxForm.vue";
+
+/**
+ * This component can be used in to wrap form components like OnyxInput/OnyxTextarea/OnyxStepper.
+ * Currently it can be used to set all form elements to disabled.
```
larsrickert, line 11: We will most likely not maintain this comment in the future
```suggestion
 * It can be used to e.g. set all form components to disabled.
```

packages/sit-onyx/src/components/OnyxForm/OnyxForm.stories.ts
```diff
@@ -0,0 +1,46 @@
...
+      h(OnyxInput, { label: "Favorite password", type: "password", modelValue: "incorrect" }),
+      h(OnyxStepper, { label: "Number of hairs", min: 0, modelValue: 23 }),
+      h(
+        "div",
+        {
+          style: { display: "flex", gap: "0.5rem" },
+        },
+        [
+          h(OnyxButton, { label: "Reset", type: "reset", mode: "outline" }),
+          h(OnyxButton, { label: "Submit", type: "submit", formmethod: "dialog" }), // we use formmethod `dialog` to avoid a page load on submit
```
larsrickert, line 38: Shouldn't we set `method="dialog"` on the `<form>` itself instead?
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#method

packages/sit-onyx/src/components/OnyxForm/OnyxForm.vue
```diff
@@ -0,0 +1,28 @@
...
+   * Any form content
+   */
+  default?(): unknown;
+}>();
+
+provideFormContext(props);
+
+const { densityClass } = useDensity(props);
+</script>
+<template>
```
larsrickert, line 19:
```suggestion
</script>

<template>
```

New line for my inner monk :D

packages/sit-onyx/src/components/OnyxForm/OnyxForm.vue
```diff
@@ -0,0 +1,28 @@
+<script setup lang="ts">
+import { useDensity } from "../../composables/density";
+import { provideFormContext } from "./OnyxForm.core";
+import type { OnyxFormProps } from "./types";
+
+const props = defineProps<OnyxFormProps>();
```
larsrickert, line 6: Should we set some mostly used defaults? Like method "dialog"?

packages/sit-onyx/src/components/OnyxForm/OnyxForm.core.ts
```diff
@@ -0,0 +1,101 @@
+import { computed, inject, provide, type InjectionKey, type Reactive, type Ref } from "vue";
+
+const FORM_INJECTION_KEY = Symbol() as InjectionKey<ReturnType<typeof createFormInjectionContext>>;
+
+/**
+ * Props on the `OnyxForm` component.
+ * These are injected, so that they can be used in the form child components.
+ */
+export type FormInjectedProps = {
+  disabled: boolean;
```
larsrickert, line 10:
```suggestion
  disabled?: boolean;
```

Disabled is optional, right?

packages/sit-onyx/src/components/OnyxSelect/types.ts
```diff
@@ -70,9 +71,17 @@ export type SelectModelValueProps<TValue extends SelectOptionValue> =
...
   AutofocusProp &
   Pick<BaseSelectOption, "truncation"> & {
+    /**
+     * Whether the select should be disabled.
+     */
+    disabled?: FormInjected<boolean>;
+    /**
+     * Whether the select should be readonly.
+     */
+    readonly?: boolean;
```
larsrickert, line 84: Why don't we just not `Omit<>` readonly here? Seems to be the same

packages/sit-onyx/src/components/OnyxSelectInput/OnyxSelectInput.vue
```diff
@@ -179,7 +182,7 @@ const blockTyping = (event: KeyboardEvent) => {
             type="button"
             :aria-label="t('select.toggleDropDown')"
             tabindex="-1"
-            :disabled="props.readonly || props.disabled || props.loading"
+            :disabled="disabled || props.loading"
```
larsrickert, line 185: Did you remove readonly intentionally? Shouldn't the select be non-openable if its readonly?

packages/sit-onyx/.storybook/formInjected.ts
larsrickert, line 1: This looks nearly identical to what we have for the managed symbol....
What do you think about creating a generic `createSymbolArgTypeEnhancer` where you can pass the symbol so we don't need redundant code?

packages/sit-onyx/src/components/OnyxForm/OnyxForm.stories.ts
larsrickert, line 1: Imo it feels weird to have both the OnyxForm component and the [form examples](https://storybook.onyx.schwarz/?path=/docs/examples-form--docs).

Should we combine them?

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxForm/OnyxForm.stories.ts
JoCa96: Sounds good, will do a s a follow-up

packages/sit-onyx/src/components/OnyxForm/OnyxForm.core.ts
JoCa96: Not for this type, otherwise it will mess up the derived types

packages/sit-onyx/src/components/OnyxSelect/types.ts
JoCa96: done

packages/sit-onyx/.storybook/formInjected.ts
JoCa96: done

packages/sit-onyx/src/components/OnyxForm/OnyxForm.core.ts
JoCa96: done

packages/sit-onyx/src/components/OnyxForm/OnyxForm.vue
JoCa96: out of scope of this PR

packages/sit-onyx/src/components/OnyxForm/OnyxForm.stories.ts
larsrickert: Yes, technically both work but shouldn't we recommend the `<form method="dialog">` instead?
Since our button does not have a formmethod properrty

larsrickert (dismissed):

packages/sit-onyx/src/components/OnyxForm/OnyxForm.stories.ts
JoCa96: But the form doesn't have it either 😅

larsrickert (approved):

Issue #641, opened 2024-03-15
Storybook: Implement A11y addon


JoCa96:
### Why?

We want to emphasize our focus on accessibility.
By adding the [addon-a11y](https://storybook.js.org/addons/@storybook/addon-a11y) add-on to our Storybook, we are able to show our commitment to accessibility and the quality of our components. 

### Acceptance criteria

- [x] The Accessibility tab of the add-on is available in our Storybook Stories
- [x] If possible, share the axe configuration (not the ignore rules)

### Definition of Done

- [x] follow-up tickets were created if necessary
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature
- [x] Set task to `In Approval`

**Approval**

- [Storybook](https://storybook.onyx.schwarz/?path=/story/buttons-iconbutton--primary), the ticket is about the accessibility tab right next to 'Controls' and 'Actions'.
- [x] approved by designer @jannick-ux
- [x] approved by dev @JoCa96 

### Implementation details

- Follow the instructions on https://storybook.js.org/addons/@storybook/addon-a11y

#### Reference implementations

- [Trinity](https://ods-lidl-web-vuejs-storybook.apps.01.cf.eu01.stackit.cloud/?path=/story/ods-addonbadge--default)

larsrickert:
@MajaZarkova Do you know which rules the addon applies/checks by default? Since we decided to aim for AA, we should ensure that all rules for AA are also included in the Storybook addon :)

MajaZarkova:
I don't, but I can investigate

MajaZarkova:
@larsrickert , I found the list with the rules -> https://github.com/dequelabs/axe-core/blob/develop/doc/rule-descriptions.md

larsrickert:
> [@larsrickert](https://github.com/larsrickert) , I found the list with the rules -> https://github.com/dequelabs/axe-core/blob/develop/doc/rule-descriptions.md

These are just all available rules, right? We should check which of them are actually checked by the Storybook addon. E.g. it might only check a few of them by default and we might need to configure it to check more. But I am not sure 🤔

PR #1921, opened 2024-10-03
chore: Impelement a11y addon

MajaZarkova:
Relates to #641 

Impelement a11y addon in Storybook

### DIFF
--- a/packages/sit-onyx/package.json
+++ b/packages/sit-onyx/package.json
@@ -50,13 +50,11 @@
   "devDependencies": {
     "@axe-core/playwright": "^4.10.0",
     "@sit-onyx/headless": "workspace:^",
+    "@storybook/addon-a11y": "^8.3.4",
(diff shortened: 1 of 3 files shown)


packages/sit-onyx/package.json
```diff
@@ -55,5 +55,8 @@
     "sass": "catalog:",
     "vue": "catalog:",
     "vue-component-type-helpers": "^2.1.6"
+  },
+  "dependencies": {
+    "@storybook/addon-a11y": "^8.3.4"
```
JoCa96, line 60: Please install the storybook add-on as *devDependencies* (using `pnpm i -D @storybook/addon-a11y`).
If a package is listed in the *dependencies*, it will be installed when `sit-onyx` itself is installed by someone. We don't want that as this addon is only used for development and build time.

JoCa96 (changes requested):

PR #1926, opened 2024-10-07
chore: switch to sass-embedded and remedy sass API deprecation warnings

JoCa96:
- add `sass-embedded` to catalog and use it instead of `sass`
- add workspace package `@sit-onyx/shared` for shared configs
  - add `vite.config.base.ts` file for shared Vite base config
  - right now we can't use a module import because of an [open Vite issue](https://github.com/vitejs/vite/issues/5370).
  We have to use a relative import instead.


apps/demo-app/vite.config.ts
```diff
@@ -1,7 +1,7 @@
-import vue from "@vitejs/plugin-vue";
 import { defineConfig } from "vite";
+import { VITE_BASE_CONFIG } from "./node_modules/@sit-onyx/shared/vite.config.base";
 
 // https://vitejs.dev/config/
 export default defineConfig({
-  plugins: [vue()],
+  ...VITE_BASE_CONFIG,
```
larsrickert, line 6: Would prefer to use Vite's mergeConfig so it is deep merged correctly:
```ts
import { mergeConfig, UserConfig } from "vite";
import { VITE_BASE_CONFIG } from "./node_modules/@sit-onyx/shared/vite.config.base";

// https://vitejs.dev/config/
export default mergeConfig(VITE_BASE_CONFIG, {} satisfies UserConfig);
```

apps/docs/vite.config.ts
```diff
@@ -1,10 +1,11 @@
-import vue from "@vitejs/plugin-vue";
 import { fileURLToPath, URL } from "node:url";
 import { defineConfig } from "vite";
+import { VITE_BASE_CONFIG } from "./node_modules/@sit-onyx/shared/vite.config.base";
 
 // https://vitejs.dev/config
 export default defineConfig({
-  plugins: [vue()],
+  ...VITE_BASE_CONFIG,
+  plugins: [],
```
larsrickert, line 8:
```suggestion
```

Isn't this obsolete?

packages/shared/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "@sit-onyx/shared",
+  "version": "1.0.0",
+  "description": "",
+  "private": true,
+  "type": "module",
+  "devDependencies": {
```
larsrickert, line 7:
```suggestion
  "peerDependencies": {
```

They are more likely peerDependencies, right?

pnpm-workspace.yaml
```diff
@@ -6,4 +6,5 @@ catalog:
   "@vue/compiler-dom": "3.5.11"
   "chart.js": "^4.4.4"
   "vue": "3.5.11"
-  "sass": "^1.79.3"
+  "sass": "1.79.4"
```
larsrickert, line 9: Can't we remove sass then?

larsrickert (changes requested):

apps/docs/vite.config.ts
JoCa96: Hi, not completely for the vitepress config we have to remove the vue plugin for it to work.
But Storybook doesn't work without the vue plugin, so I moved the overwrite to the vitepress config.

apps/demo-app/vite.config.ts
JoCa96: oh, interesting didn't know about that, thanks!

JoCa96: @larsrickert I dropped the `mergeConfig` approach again, as it causes issues (it doesn't merge exisiting plugins) and currently solves none. The spread operator behaves exactly as expected :)

packages/shared/package.json
JoCa96: done

apps/demo-app/vite.config.ts
larsrickert: I fear that this approach will cause unexpected issues in the future if we add/override custom options which might override ones from VITE_BASE_CONFIG because they are not merged correctly :/

What if we implement a `createViteConfig({})` utility method where overrides can be passed which uses `mergeConfig` from vite under the hood plus takes care of also merging the plugins?

JoCa96: We can do that when needed :) Would prefer to keep it simple as it is right now

larsrickert (approved):

Issue #1375, opened 2024-06-21
Implement override option to show error on :invalid instead of :user-invalid


BoppLi:
### Why?

 Form validation: Custom error should always be displayed, not only on when the user interacted with the component. Maybe also let the developer decide whether the errors show instantly or only after interaction.

### Acceptance criteria

- [x] offer an option to switch between invalid/user-invalid mode (like we do with the required/optional class / props)

### Definition of Done

- [x] covered by component tests
- [x] follow-up tickets were created if necessary
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature
- [x] Storybook code snippet of new/changed examples are checked that they are generated correctly
- [x] Set task to `In Approval`

**Approval**

- [Storybook](https://storybook.onyx.schwarz/?path=/story/form-elements-form--show-error)
- [x] approved by dev @MajaZarkova 

### Implementation details

- try to find an implementation where we don't need to copy styles for :invalid and :user-invalid
- test this feature very well with Playwright / Vitest
- make sure that both build in, and custom errors work as expected

PR #1943, opened 2024-10-09
feat(OnyxForm): implement showError property

JoCa96:
Relates to #1375 

- feat(OnyxForm): implement `showError` mode
- updated all relevant FormElements to have and use `showError` prop
- created dedicated `FormProps` and `FormInjectedProps` type for easier re-use and shared description
- fix(OnyxForm): update of injected prop not working when initially using default
- enabled and configured type test from vitest for ensuring vue workaround doesn't cause issues
- fix linting issue in `useDensity` computed type


.changeset/beige-cherries-explain.md
MajaZarkova, line 1: Why do we have the ripple animation changeset in this PR?

packages/sit-onyx/src/components/OnyxForm/OnyxForm.stories.ts
MajaZarkova, line 1: I noticed that on the input example we have a boolean
image

But this is not the case for the form component. Is this intentional?

MajaZarkova (changes requested):

.changeset/beige-cherries-explain.md
JoCa96: I merged incorrectly, fixed the issue now :)

packages/sit-onyx/src/components/OnyxForm/OnyxForm.stories.ts
JoCa96: Good catch :) But that is intentional. Forms will only provide the form injected props, but not inject them themselves.

PR #1951, opened 2024-10-11
chore: fix cyclic dependencies

ChristianBusshoff:
- moved @sit-onyx/storybook-utils and @sit-onyx/eslint-plugin from root to the repos where they are directly used
- chore(storybook-utils): Renamed internal `index.css`, which is already exported as `styles.css` to have the same name internally
- chore(storybook-utils): Removed `sit-onyx` dependency:
  - Load colors from DOM using `getComputedStyle(el).getPropertyValue(property)`
  - Removed `brand` defaults for `createTheme` to keep users from using our brand details
  - Moved `ONYX_BREAKPOINTS` constant and `OnyxBreakpoint` type to `@sit-onyx/shared`
    - `@sit-onyx/shared` is now published as we have direct build and runtime dependencies to it
    -  `turbo` configuration has been updated, so that it builds shared beforehand
 - updated `turbo` to determine dependencies itself
   - duplicated `peerDependencies` from the *workspace* into `devDependencies` as `turbo` will not consider them otherwise (See also: https://github.com/vercel/turborepo/commit/1eb3333f048ddc246caafbde920dd122e66b7e87)
- updated and added override for `sass-embedded` to fix incompatible vite plugin types (different version usages of `sass-embedded` will result in dedicated `vite` dependencies and therefore inequal types)


package.json
```diff
@@ -16,7 +16,7 @@
     "publint:all": "pnpm -r --parallel --aggregate-output exec publint",
     "prepare": "simple-git-hooks",
     "gh:show-report": "./scripts/show-last-test-report.sh",
-    "preinstall": "npx only-allow pnpm"
+    "preinstall": "npx only-allow pnpm && pnpm list --depth=-1 --long | grep cycle && echo 'Cyclic dependency detected!' && exit 1 || echo 'No cyclic dependencies'"
```
JoCa96, line 19: `pnpm list --depth=-1 --long` doesn't output anything for me, can you check the script please?

JoCa96 (changes requested):

packages/shared/tsconfig.json
larsrickert, line 1: Would prefer to use the same tsconfig that we use for the other packages too:

```json
{
  "extends": "@tsconfig/node22/tsconfig.json",
  "include": ["src"],
  "exclude": ["src/**/*.ct.*", "src/**/*.spec.*"],
  "compilerOptions": {
    "composite": true,
    "rootDir": "./src",
    "baseUrl": ".",
    "outDir": "dist",
    "declaration": true,
    "types": ["node"]
  }
}
```

larsrickert: Or at least remove all the uncommented code

packages/sit-onyx/package.json
```diff
@@ -34,6 +34,7 @@
     "url": "https://github.com/SchwarzIT/onyx/issues"
   },
   "scripts": {
+    "td": "turbo watch --filter=sit-onyx dev",
```
larsrickert, line 37: What is this script and when should it be used?

packages/shared/src/index.ts
larsrickert, line 1: Is this empty file intended?

packages/icons/package.json
```diff
@@ -22,6 +22,7 @@
     "./*": "./src/assets/*"
   },
   "scripts": {
+    "build": "tsc",
```
larsrickert, line 25: Whats the purpose of this? The dist folder is not published in the npm package

packages/shared/src/breakpoints.ts
larsrickert, line 1: Should we re-export the breakpoints in sit-onyx so the project can use them for individual logic without needing to install @sit-onyx/shared?

packages/sit-onyx/src/styles/grid.ct.tsx
```diff
@@ -1,7 +1,7 @@
 /* eslint-disable playwright/expect-expect */
 import type { Locator, Page } from "@playwright/test";
+import { ONYX_BREAKPOINTS, type OnyxBreakpoint } from "@sit-onyx/shared/breakpoints";
```
larsrickert, line 3: Why is there a dedicated `/breakpoints` path?
Shouldn't we just export all relevant exports directly from `@sit-onyx/shared`?

packages/storybook-utils/src/preview.ts
```diff
@@ -40,7 +35,15 @@ const themes = {
  * export default preview;
  * ```
  */
-export const createPreview = <T extends Preview = Preview>(overrides?: T) => {
+export const createPreview = <T extends Preview = Preview>(
+  brandDetails: BrandDetails,
+  overrides?: T,
```
larsrickert, line 40: Would prefer to make the brandDetails optional as second parameter since most projects probably don't care about this

packages/storybook-utils/src/style.css
larsrickert, line 1: Why did you remove the export? onyx CSS variables are used in this file so we need to import them :)
The storybook utils are independent from the onyx component library so you could use the onyx theme without setting up sit-onyx itself so we can not rely on that the variables are imported in other places

packages/storybook-utils/src/theme.ts
larsrickert, line 1: If the `sit-onyx/themes/onyx.json` is unused now you can delete the file and also adjust `.github/workflows/import-figma.yml` to not import it any longer since it was only needed/used here

packages/storybook-utils/package.json
```diff
@@ -28,16 +28,18 @@
...
     "storybook-dark-mode": ">= 4"
   },
   "dependencies": {
     "deepmerge-ts": "^7.1.3"
   },
   "devDependencies": {
-    "vue": "catalog:"
+    "vue": "catalog:",
+    "@sit-onyx/shared": "workspace:^",
+    "@sit-onyx/icons": "workspace:^"
```
larsrickert, line 43:
```suggestion
    "vue": "catalog:"
```

These two are already peerDependencies so they can be removed here

packages/storybook-utils/src/theme.ts
```diff
@@ -1,33 +1,40 @@
...
  * Creates a custom theme for Storybook that uses onyx colors.
  *
  * @see https://storybook.js.org/docs/react/configure/theming#create-a-theme-quickstart
  */
-export const createTheme = (
-  options?: Pick<ThemeVarsPartial, "base" | "brandTitle" | "brandImage" | "brandUrl">,
-) => {
-  const base = options?.base ?? "light";
-  const primaryColor = onyxVariables["onyx-color-themed-primary-500"];
+export const createTheme = (brandDetails: BrandDetails, base: "light" | "dark" = "light") => {
```
larsrickert, line 21: Why is base not included in the brandDetails?

packages/vitepress-theme/package.json
```diff
@@ -34,5 +34,8 @@
   "dependencies": {
     "@fontsource-variable/source-code-pro": "^5.1.0",
     "@fontsource-variable/source-sans-3": "^5.1.0"
+  },
+  "devDependencies": {
+    "sit-onyx": "workspace:^"
```
larsrickert, line 39: onyx is already a peerDependency, why is it added as devDependency as well?

turbo.json
```diff
@@ -20,10 +16,18 @@
     "test:components": {
       "outputs": ["playwright-report", "test-results"]
     },
+    "sit-onyx#test:components": {
+      "dependsOn": ["sit-onyx#build"]
+    },
```
larsrickert, line 21: Why are the tests dependent on the build? Playwright builds the components separately

larsrickert (changes requested):

packages/sit-onyx/package.json
JoCa96: This script watches all dependencies of `sit-onyx#dev` and runs them

turbo.json
JoCa96: You are right, what I wanted to achieve was "tests depend on the build of dependencies". Should be fixed thorugh the last commit.

packages/icons/package.json
JoCa96: Done

packages/shared/src/breakpoints.ts
JoCa96: good idea, done

packages/vitepress-theme/package.json
JoCa96: - updated turbo to determine dependencies itself

    -  duplicated peerDependencies from the workspace into devDependencies as turbo will not consider them otherwise (See also: vercel/turborepo@1eb3333)

packages/storybook-utils/package.json
JoCa96: updated turbo to determine dependencies itself

    duplicated peerDependencies from the workspace into devDependencies as turbo will not consider them otherwise (See also: vercel/turborepo@1eb3333)

packages/sit-onyx/src/styles/grid.ct.tsx
JoCa96: The two exports to be bundled together, as they are address different run targets

packages/storybook-utils/src/preview.ts
JoCa96: No, because if anybody uses it should not default to our details.
This will create the impression that it is an official `onyx` product.

packages/shared/tsconfig.json
JoCa96: done

packages/shared/src/index.ts
JoCa96: yes, otherwise the bundler complains

packages/storybook-utils/src/style.css
JoCa96: Okay, then I need to move this to the `sit-onyx` package to be able to remove the dependency

packages/storybook-utils/src/theme.ts
JoCa96: Because the base is not related to the brand.

packages/storybook-utils/src/preview.ts
larsrickert: I meant to not provide any defaults but make the parameter optional.
If not provided, it will fall back to the default Storybook logo. We shouldn't force projects to pass any brand details, there might not be any

packages/storybook-utils/src/theme.ts
JoCa96: Done

packages/storybook-utils/src/style.css
JoCa96: Okay, I was able to do use a relative import

packages/sit-onyx/src/index.ts
```diff
@@ -1,5 +1,7 @@
 import "./styles/index.scss";
 
+export { ONYX_BREAKPOINTS } from "@sit-onyx/shared/breakpoints";
```
larsrickert, line 3:
```suggestion
export { ONYX_BREAKPOINTS, type OnyxBreakpoint } from "@sit-onyx/shared/breakpoints";
```

packages/sit-onyx/src/styles/grid.ct.tsx
larsrickert: Should we then instead use a `@sit-onyx/shared/node` or `@sit-onyx/shared/browser` path so we don't have to create a new path in the future for every new feature? With the current name its limited to only include breakpoints

packages/storybook-utils/src/style.css
```diff
@@ -1,4 +1,4 @@
-@import url("sit-onyx/themes/onyx.css");
+@import url("../../sit-onyx/src/styles/variables/themes/onyx.css");
```
larsrickert, line 1: This works in our local monorepo but will break when used in a real project since the relative path is not resolvable then...we need to import from `sit-onyx` instead

larsrickert (changes requested):

packages/sit-onyx/src/styles/grid.ct.tsx
JoCa96: Ideally, yes. But looking at the PR, its age and scope this is too much effort for now. Fo-Up was created: [#2056](https://github.com/SchwarzIT/onyx/issues/2056)

packages/storybook-utils/src/style.css
```diff
@@ -0,0 +1,161 @@
...
+.sb-previewBlock,
+.sb-argstableBlock {
+  display: none;
+}
+
+/* 
+  Copy of ../sit-onyx/src/styles/variables/themes/onyx.css
+  TODO: Find a way to automate this
+ */
+:where(:root),
```
larsrickert, line 41: Hmm I don't really like this copy...what do you think about just replacing / hard-coding the two variables in line 8 and 23 since we only need those two values?

larsrickert (changes requested):

packages/storybook-utils/src/preview.ts
JoCa96: done

packages/storybook-utils/src/style.css
larsrickert: Oh, you‘re right

Issue #83, opened 2024-01-09
Define setup tooling for web-api-support checks (CanIUse)


larsrickert:
### Why?

Check if we can e.g. setup eslint (or a similar tool) to check/warn for Web API support / browser compatibility in our code.
So that we are notified when we try to use a Web API that is not supported by X percent of browsers.

Also look into the necessity of polyfills etc for supporting older browsers in the company

### Acceptance criteria

- [x] Pipeline should check browser support of the web APIs we use
- [x] Define a list of relevant browser
- [x] If there are issues that get out of scope, create tickets
- [x] Update the Docs to reflect which browsers + versions are supported (ideally it is connected to a browserlists file)

### Definition of Done

- [x] follow-up tickets were created if necessary
    - #1970 
    - #1997 
- [x] updated version + documentation is deployed

**Approval**

- [Documentation link](https://onyx.schwarz/development/#browser-support)
- [x] approved by ux @mj-hof 
- [x] approved by dev @JoCa96 

### Implementation details

- Info: ask A.S. from the department "Workplace Platform Design" about which relevant browsers & versions we have to support

oemueller:
Proposed browserlist: https://browsersl.ist/#q=%3E+0.5%25%2C+last+2+versions%2C+Firefox+ESR%2C+not+dead

oemueller:
with `npx browserslist` we can get a list of supported browsers -> based on that we can render a table of all supported browsers in the Docs

oemueller:
Desktop browser requirements: Edge, Safari (last 2 Versions), Firefox ESR (Source: Workplace Platform Design)

PR #1956, opened 2024-10-14
feat: add browsers list, linting plugin and Docs section with supported browsers

oemueller:
Relates to #83 

## Checklist


oemueller:
Preview of browser section on Onyx Docs:
Bildschirmfoto 2024-10-15 um 19 10 21

.changeset/brave-days-happen.md
```diff
@@ -0,0 +1,6 @@
+---
+"docs": minor
+"sit-onyx": minor
```
larsrickert, line 3:
```suggestion
```

There are no relevant changes for the users, right? So we don't need to publish a new onyx version

.eslintrc.cjs
```diff
@@ -18,6 +22,7 @@ module.exports = {
     // see: https://github.com/vuejs/eslint-config-typescript#vueeslint-config-typescriptrecommended
     "@vue/eslint-config-typescript/recommended",
     "@vue/eslint-config-prettier/skip-formatting",
+    "plugin:compat/recommended",
```
larsrickert, line 25: Is this the line that will check all our code to match the browserlistrc?

apps/docs/package.json
```diff
@@ -26,5 +27,11 @@
     "vitepress": "^1.4.0",
     "vue": "catalog:",
     "vue-chartjs": "^5.3.1"
+  },
+  "imports": {
+    "#src/.vitepress/browser-loader.data": {
+      "storybook": "./src/.vitepress/browser-loader.data.mock.ts",
+      "default": "./src/.vitepress/browser-loader.data.ts"
+    }
```
larsrickert, line 35: Why is this needed?

apps/docs/src/.vitepress/browser-loader.data.ts
```diff
@@ -0,0 +1,72 @@
...
+import path, { dirname } from "path";
+import { fileURLToPath } from "url";
+import { defineLoader } from "vitepress";
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = dirname(__filename);
+const browserslistRcPath = path.resolve(__dirname, "../../../../.browserslistrc");
+
+const load = async () => {};
+
```
larsrickert, line 12:
```suggestion
```

This is obsolete, right?

apps/docs/src/.vitepress/browser-loader.data.ts
```diff
@@ -0,0 +1,72 @@
+import fs from "fs";
+import https from "https";
+import path, { dirname } from "path";
+import { fileURLToPath } from "url";
+import { defineLoader } from "vitepress";
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = dirname(__filename);
+const browserslistRcPath = path.resolve(__dirname, "../../../../.browserslistrc");
```
larsrickert, line 9:
```suggestion
const browserslistRcPath = fileURLToPath(new URL("../../../../.browserslistrc", import.meta.url));
```

Usually we do this like this so we save 2 lines :)

apps/docs/src/.vitepress/browser-loader.data.ts
```diff
@@ -0,0 +1,72 @@
...
+  coverage: number;
+  id: string;
+  name: string;
+  versions: Record<string, number | undefined>;
+};
+
+export interface Data {
+  browsers: Browser[];
+  coverage: number;
+  config: string;
```
larsrickert, line 23:
```suggestion
```

Is config unused?

apps/docs/src/.vitepress/browser-loader.data.ts
```diff
@@ -0,0 +1,72 @@
...
+ * More information can be found on https://browsersl.ist
+ */
+export default defineLoader({
+  async load(): Promise<Data> {
+    return new Promise((resolve, reject) => {
+      let browserRules = "";
+
+      try {
+        const data = fs.readFileSync(browserslistRcPath, "utf8");
+        const [_firstLine, ...lines] = data.split("\n").filter((l) => !!l);
```
larsrickert, line 42: Omitting the first line might not work as expected if comments are removed/added to the file

Should we instead remove all lines that are comments (start with a a `#`) to be safer?

apps/docs/src/.vitepress/browser-loader.data.ts
```diff
@@ -0,0 +1,72 @@
...
+        browserRules = lines.join("").trim();
+      } catch (e) {
+        console.error("could not read .browserslistrc");
+      }
+
+      const url = `https://browsersl.ist/api/browsers?q=${browserRules}`;
+
+      https
```
larsrickert, line 50: Is there a reason you are using https instead if the default fetch?

apps/docs/vite.config.ts
```diff
@@ -5,6 +5,7 @@ import { VITE_BASE_CONFIG } from "./node_modules/@sit-onyx/shared/vite.config.ba
 // https://vitejs.dev/config
 export default defineConfig({
   ...VITE_BASE_CONFIG,
+  optimizeDeps: { exclude: ["fsevents"] },
```
larsrickert, line 8: Why is this needed? Removing it also works fine for me :)

apps/docs/src/.vitepress/components/OnyxHomePage.vue
larsrickert, line 1: I appreciate your detailed approach here 👍 
But at least for me, all the coverage and browser version numbers are a bit overwhelming. In fact, what does the browser coverage actually mean?^^ That we only support 44.33% or that 44.33% of users are using Chrome on Android?

tl;rd: I have 2 sugesstions:
1. Maybe you simplify the browser support to something like this + and additional note that onyx might work with older versions but it is not tested/guaranteed: image
2. We move the component to the [getting started page](https://onyx.schwarz/development) since it seems more fitting to me there

What do you think?

larsrickert (changes requested):
Thanks for your contribution! I left a few comments, let me know what you think!

larsrickert:
With the current approach, only JavaScript features are linted, right? So if we use CSS features that are e.g. not supported in firefox, we will not know

.eslintrc.cjs
oemueller: that's the plugin to check our code, runs as part of `lint:ci:all`

apps/docs/package.json
oemueller: because the file normally runs in Node.js environment and this would throw an error in Storybook context. 
This is the Storybook recommended way to mock the import.

https://vitepress.dev/guide/data-loading

apps/docs/src/.vitepress/browser-loader.data.ts
oemueller: will check :-)

apps/docs/src/.vitepress/components/OnyxHomePage.vue
oemueller: makes sense, shall we include mobile browsers as well ?

oemueller:
we could try out https://stylelint.io/ to lint our CSS, will create a follow up ticket for that

apps/docs/src/.vitepress/components/OnyxHomePage.vue
larsrickert: Sure why not :)

apps/docs/src/.vitepress/components/BrowsersList.vue
```diff
@@ -0,0 +1,63 @@
...
+const filteredBrowsers = popularBrowsers
+  .map((pb) => {
+    const b = browsers.find((b) => b.id === pb);
+    if (b)
+      return {
+        ...b,
+        minVersion: b?.versions ? Object.keys(b.versions).sort()[0] : "",
+      };
+  })
+  .filter((b) => !!b);
```
larsrickert, line 17:
```suggestion
const filteredBrowsers = browsers
  .filter((browser) => popularBrowsers.includes(browser.id))
  .map((browser) => {
    return {
      ...browser,
      minVersion: Object.keys(browser.versions).sort()[0],
    };
  });
```

apps/docs/src/.vitepress/components/BrowsersList.vue
```diff
@@ -0,0 +1,63 @@
...
+      <p>Version ≥ {{ minVersion }}</p>
+    </li>
+  </ul>
+</template>
+
+<style lang="scss" scoped>
+.browsersList {
+  display: grid;
+  grid-template-columns: repeat(auto-fill, minmax(160px, 1fr));
+  row-gap: var(--onyx-spacing-lg);
```
larsrickert, line 40:
```suggestion
  display: flex;
  flex-wrap: wrap;
  gap: var(--onyx-spacing-xl);
```

We could use flex + flex-wrap instead so we don't need to maintain the 160px width manually and instead let the components take whatever width they need :) What do you think?

apps/docs/src/.vitepress/components/BrowsersList.vue
```diff
@@ -0,0 +1,63 @@
...
+</script>
+
+<template>
+  <ul class="browsersList">
+    <li v-for="{ name, minVersion, id } in filteredBrowsers" :key="name" class="browser">
+      <img
+        :src="`/assets/browsers/${id}.svg`"
+        :alt="`${name} icon`"
+        width="40px"
+        height="40px"
```
larsrickert, line 27: Should we (also) set the size via CSS? Otherwise the images are not exactly 40x40px:
image

apps/docs/src/.vitepress/components/BrowsersList.vue
```diff
@@ -0,0 +1,63 @@
...
+    line-height: 1.3;
+  }
+  &__name {
+    padding-top: var(--onyx-spacing-xs);
+    font-weight: 600;
+  }
+  &__image {
+    position: relative;
+    transform: translateX(-4px);
+  }
```
larsrickert, line 61: Might be cleaner to remove the whitespace from the images itself, what do you think?
I can provide you the edited images

apps/docs/src/assets/browsers/chrome.svg
larsrickert, line 1: Where did you get the browser logos from? Are they copyright free / free to use?

apps/docs/src/development/index.md
```diff
@@ -154,8 +154,24 @@ Be careful when overriding styles globally since it will affect **EVERY** compon
 
 :::
 
+### Browser Support
```
larsrickert, line 157: Would prefer to move this to the bottom of the page and change the headline to h2 :)

apps/docs/src/.vitepress/components/BrowsersList.vue
larsrickert, line 1: I like the new design, good job!

apps/docs/src/development/index.md
```diff
@@ -154,8 +154,24 @@ Be careful when overriding styles globally since it will affect **EVERY** compon
...
+import BrowsersList from "../.vitepress/components/BrowsersList.vue"
+</script>
+
+This list of supported browsers is based on our `Browserslist` config and data from [caniuse](https://caniuse.com)
+
+Onyx might also work in older versions, but we cannot guarantee full compatibility.
+
+<BrowsersList />
+
+Find our .browserslistrc [here](https://github.com/SchwarzIT/onyx/blob/main/.browserslistrc)
```
larsrickert, line 169:
```suggestion
## Browser Support

<script lang="ts" setup>
import BrowsersList from "../.vitepress/components/BrowsersList.vue"
</script>

Onyx works best with the following browser versions. Older versions might also work but we can't guarantee full compatibility.

<BrowsersList />

::: details More information (.browserslistrc file)

<<< @include ../../../../.browserslistrc
:::
```

What do you think about rephrasing it like this and include the browserlistrc content directly?

apps/docs/src/development/index.md
larsrickert, line 1: @JoCa96 If we list supported browsers in our docs, we should also make sure that they are actually working, so e.g. use those browsers and versions in our Playwright tests. What do you think?

apps/docs/src/assets/browsers/chrome.svg
oemueller: they are from https://github.com/browserslist/browsersl.ist which is under MIT license so should be fine

apps/docs/src/development/index.md
oemueller: like it!

apps/docs/src/.vitepress/components/BrowsersList.vue
oemueller: sure, fill free to change them :-)

.eslintrc.cjs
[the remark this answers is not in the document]
oemueller: fixed + rebased :-)

apps/docs/src/development/index.md
```diff
@@ -154,8 +154,25 @@ Be careful when overriding styles globally since it will affect **EVERY** compon
...
+<<< @include ../../../../.browserslistrc{txt}
+:::
+
+## Next steps
 
 - [Theming](/development/theming), e.g. for setting up the dark mode
 - [i18n](/development/i18n) if you are supporting multiple languages in your application
 - [Grid](/development/grid) for easily implementing custom layouts
+
+---
```
larsrickert, line 178:
```suggestion
```

Is this a leftover or intended? :)

larsrickert (approved):

apps/docs/src/development/index.md
oemueller: leftover 😄

apps/docs/src/development/index.md
oemueller: shall we create follow up ticket ?

larsrickert: #1997

I will refine it next week

larsrickert (approved):

Issue #1602, opened 2024-07-23
Implement skeleton prop injection


JoCa96:
### Why?

From @jannick-ux:
> Skeleton modes are used for the initial loading of the entire page. So when I am directed to a page, the skeleton mode of the components is shown until the components themselves or the entire page is loaded.
Loading mode, on the other hand, is used when the components themselves no longer need to be loaded, but the data is processed or validated via a button click, for example (classic example “processing”).

To make this easier for the user and reflect this global/page state, we should enable the user to trigger the skeleton mode globally via an injection.

### Acceptance criteria

- `skeleton` boolean ref can be provided on an app level
- all components with a skeleton prop also inject the `skeleton` boolean
- feature is added to the DemoApp

### Definition of Done

- [ ] covered by component tests
- [ ] screenshot tests are updated
- [ ] follow-up tickets were created if necessary
- [ ] updated version + documentation is deployed
- [ ] Storybook can show the feature
- [ ] Storybook code snippet of new/changed examples are checked that they are generated correctly

**Approval**

- Storybook 
- Deployed Docs 
- [ ] approved by dev 

### Implementation details

Demo app:

```ts
import { SkeletonModeInjectionKey } from "sit-onyx";
import { provide } from "vue";

…

const skeleton = ref(true);
provide(SkeletonInjectionKey, skeleton);
```

JoCa96:
created follow-up: [#1967](https://github.com/SchwarzIT/onyx/issues/1967)

PR #1962, opened 2024-10-15
feat(demoApp): Implement skeleton injected prop in DemoApp

MajaZarkova:
Relates to #1602 

Implement skeleton injected prop in DemoApp

## Checklist

### DIFF
--- a/apps/demo-app/src/views/HomeView.vue
+++ b/apps/demo-app/src/views/HomeView.vue
@@ -79,6 +79,7 @@ const show = computed(() => {
 const densityOptions = DENSITIES.map((value: string) => ({
   value,
   label: capitalize(value),
+  skeleton: false,
 })) satisfies SelectOption[];
 
 const activeDensityOption = ref(densityOptions[1].value);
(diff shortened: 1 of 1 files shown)


JoCa96:
The demo app doesn't start for me:
Screenshot 2024-10-15 at 12 59 46

MajaZarkova:
> The demo app doesn't start for me: Screenshot 2024-10-15 at 12 59 46

Did you build it? I'm building it first and then preview

JoCa96:
@MajaZarkova Interesting, `preview` works. So it's only in the dev mode. I will investigate...

apps/demo-app/src/views/HomeView.vue
```diff
@@ -79,6 +79,7 @@ const show = computed(() => {
 const densityOptions = DENSITIES.map((value: string) => ({
   value,
   label: capitalize(value),
+  skeleton: false,
```
JoCa96, line 82: I don't think this does anything..?
```suggestion
```

JoCa96 (approved):

apps/demo-app/src/views/HomeView.vue
MajaZarkova: I've added this line because when we switch to skeleton mode, the density radio group is  picking up on the provided property and sets the density options to skeleton too. I don't think this is out intention.

Issue #1967, opened 2024-10-16
Fix skeleton prop injection for OnyxRadioGroup


JoCa96:
- [x] Fix skeleton prop injection for `OnyxRadioGroup`
 - when there are no (yet) options given and skeleton is injected `true`, the radiogroup
- [x] Add OnyxPageLayout example

larsrickert:
Does it also make sense to let the user provide the skeleton in any component to turn all child components in skeleton?
Example:

- you have a page with 3 sections
- the user wants to make the 2nd section skeleton but not the others
- currently this would not work because its only possible on page or form level, but the 2nd section might not be a form

JoCa96:
We can add instructions on how to use the inject

MajaZarkova:
follow up tickets: #2015 #2016

Issue #1955, opened 2024-10-14
Fix skeleton prop injection through OnyxPageLayout and OnyxForm


MajaZarkova:
### Why?

From @jannick-ux:
> Skeleton modes are used for the initial loading of the entire page. So when I am directed to a page, the skeleton mode of the components is shown until the components themselves or the entire page is loaded.
Loading mode, on the other hand, is used when the components themselves no longer need to be loaded, but the data is processed or validated via a button click, for example (classic example “processing”).

To make this easier for the user and reflect this global/page state, we should enable the user to trigger the skeleton mode via an injection. the injection is already implemented. OnyxPageLyout and OnyxForm are providing this property. The issue arises when we have OnyxForm nested inside an OnyxPageLayout. For example if we set the skeleton prop to true on OnyxPageLayout layer, it doesn't reach the components inside the OnyxForm.

### Acceptance criteria

- [ ] components inside an OnyxForm can inject (have access to) the skeleton prop on OnyxPageLayout layer, only if it's not set on OnyxForm layer.
- [ ] skeleton prop on OnyxForm layer tакеs advantage before the skeleton prop on OnyxPageLayout layer.
- [ ] A description for the prop is added and shown in stroybook

### Implementation details

see 
- https://github.com/SchwarzIT/onyx/blob/ebd8a1bda95b63920eabfd34250c00f0dc93a217/packages/sit-onyx/src/components/OnyxForm/OnyxForm.core.ts
- [createCompute](https://github.com/SchwarzIT/onyx/blob/8445b570ee04de3ea240b0705d5e213cd5b1396d/packages/sit-onyx/src/components/OnyxForm/OnyxForm.core.ts#L42)

### Definition of Done

- [ ] covered by component tests
- [ ] screenshot tests are updated
- [ ] follow-up tickets were created if necessary
- [ ] updated version + documentation is deployed
- [ ] Storybook can show the feature
- [ ] Storybook code snippet of new/changed examples are checked that they are generated correctly

**Approval**

- Storybook 
- Deployed Docs 
- [ ] approved by dev

PR #1974, opened 2024-10-17
chore(OnyxPageLayout): Add skeleton OnyxPageLayout example

MajaZarkova:
Relates to #1967 

Add skeleton OnyxPageLayout example

## Checklist

### DIFF
--- a/packages/sit-onyx/src/components/OnyxPageLayout/OnyxPageLayout.stories.ts
+++ b/packages/sit-onyx/src/components/OnyxPageLayout/OnyxPageLayout.stories.ts
@@ -1,5 +1,7 @@
 import type { Meta, StoryObj } from "@storybook/vue3";
 import { h } from "vue";
+import OnyxButton from "../OnyxButton/OnyxButton.vue";
+import OnyxRadioGroup from "../OnyxRadioGroup/OnyxRadioGroup.vue";
 import OnyxPageLayout from "./OnyxPageLayout.vue";
 
 /**
@@ -41,6 +43,13 @@ const meta: Meta<typeof OnyxPageLayout> = {
 
 export default meta;
 type Story = StoryObj<typeof OnyxPageLayout>;
+
+const EXAMPLE_OPTIONS = [
+  { label: "Green", value: 1 },
+  { label: "Red", value: 2 },
+  { label: "Pink", value: 3 },
+  { label: "Yellow", value: 4 },
+];
 
 /** A standard page with some content. */
 export const Default = {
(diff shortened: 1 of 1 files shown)


packages/sit-onyx/src/components/OnyxPageLayout/OnyxPageLayout.stories.ts
```diff
@@ -93,3 +102,24 @@ export const WithPartialFooter = {
...
+        label: "Color choices",
+        options: EXAMPLE_OPTIONS,
+        modelValue: EXAMPLE_OPTIONS[0].value,
+      }),
+      h(
+        "div",
+        {
+          style: { display: "flex", gap: "0.5rem", paddingLeft: "0.5rem" },
+        },
+        [h(OnyxButton, { label: "Reset", mode: "outline" }), h(OnyxButton, { label: "Submit" })],
```
larsrickert, line 121: If the buttons are labeled like this, I'd prefer to wrap it in a form/OnyxForm so its a "correct" form example :) And maybe add a little spacing between the radio group and the buttons

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxPageLayout/OnyxPageLayout.stories.ts
MajaZarkova: We have an issue there. The provided skeleton prop doesn't work when we have an OnyxForm wrapped in OnyxPageLayout. There is a ticket for that -> https://github.com/SchwarzIT/onyx/issues/1955

larsrickert: Ah okay, then maybe use a regular `<form>` instead :) Imo we should not promote creating forms without any kind of form component since validation etc. does not work then

packages/sit-onyx/src/components/OnyxPageLayout/OnyxPageLayout.stories.ts
```diff
@@ -93,3 +104,38 @@ export const WithPartialFooter = {
     footerAsideSidebar: true,
   },
 } satisfies Story;
+
+/** A standard page in skeleton mode. */
+export const WithSkeletonComponents = {
+  args: {
+    skeleton: true,
+    default: () => [
+      h("form", {}, [
```
larsrickert, line 113:
```suggestion
      h("form", { style: { display: "flex", flexDirection: "column", gap: "1rem" } }, [
```

larsrickert (changes requested):

larsrickert (approved):

PR #1976, opened 2024-10-18
feat(OnyxStepper): implement strip step and precision

ChristianBusshoff:
- added strip step Mode (values that do not fit the step size, a stepMismatch error is shown)
- added precision (always display the given number of decimals places)
- applied the spin button aria pattern


packages/sit-onyx/src/components/OnyxStepper/types.ts
```diff
@@ -36,6 +36,15 @@ export type OnyxStepperProps = FormInjectedProps &
...
     step?: number;
+    /**
+     * Ensures that the value is a multiple of the step size.
+     */
+    stripStep?: boolean;
+    /**
+     * Defines the number of decimal places to display.
+     */
+    precision?: number;
+
```
JoCa96, line 47: As discussed:
```ts
    /**
     * Incremental step.
     * @deprecated
     */
    step?: number; // step-mismatch + step-increment


    /**
     * The smallest allowed value and rounded precision
     */
    precision?: number; // step-mismatch => uses :step="props.precision" for the validation
    /**
     * The increment number
     * @default precision is the default stepSize
     */
    stepSize?: number; //  step-increment => number which is used for increment/decrement
```

packages/sit-onyx/src/components/OnyxStepper/OnyxStepper.vue
```diff
@@ -105,15 +135,22 @@ const decrementLabel = computed(() => t.value("stepper.decrement", { stepSize: p
           :placeholder="props.placeholder"
           :readonly="props.readonly"
           :required="props.required"
-          :step="props.step"
+          :step="props.stripStep ? props.step : 'any'"
           :title="props.hideLabel ? props.label : undefined"
-          @change="wasTouched = true"
+          @change="handleChange"
         />
```
JoCa96, line 141:
```vue
@keydown.up.prevent="handleClick('stepUp')"
@keydown.down.prevent="handleClick('stepDown')"
```

packages/sit-onyx/src/components/OnyxStepper/OnyxStepper.vue
```diff
@@ -91,7 +121,7 @@ const decrementLabel = computed(() => t.value("stepper.decrement", { stepSize: p
         <input
           v-else
           ref="inputRef"
-          v-model.number="value"
+          v-model.number="displayValue"
```
JoCa96, line 124:
```suggestion
         :value="displayValue"
```

JoCa96 (changes requested):

packages/sit-onyx/src/components/OnyxStepper/OnyxStepper.vue
```diff
@@ -57,19 +53,54 @@ const value = computed({
...
 
-const handleClick = (direction: "stepUp" | "stepDown") => {
-  if (!inputRef.value) return;
+function countDecimalPlaces(num: number) {
+  const numStr = num.toString();
+  if (numStr.includes(".")) {
+    return numStr.split(".")[1].length;
+  }
+  return 0;
+}
```
JoCa96, line 62: We can use the standard math functions, this does then works for tens and decimals
```suggestion
function countDecimalPlaces(num: number) {
  return Math.floor(Math.log10(num));
}
```

packages/sit-onyx/src/components/OnyxStepper/OnyxStepper.vue
```diff
@@ -57,19 +53,54 @@ const value = computed({
...
+    value.value = isNaN(newValue) ? undefined : parseFloat(newValue.toFixed(precision));
+    inputRef.value.value = newValue.toFixed(precision);
+    wasTouched.value = true;
+  }
 };
 
-const incrementLabel = computed(() => t.value("stepper.increment", { stepSize: props.step }));
-const decrementLabel = computed(() => t.value("stepper.decrement", { stepSize: props.step }));
+const incrementLabel = computed(() =>
+  t.value("stepper.increment", { stepSize: props?.stepSize ? props.stepSize : props.precision }),
```
JoCa96, line 98: Please use a `computed` to determine the stepSize, so the logic and default value have a single source of truth:
```ts
const determinedStepSize = computed(() =>  props.stepSize ?? props.precision);
```

packages/sit-onyx/src/components/OnyxStepper/OnyxStepper.vue
```diff
@@ -57,19 +53,54 @@ const value = computed({
...
+  } else {
+    value.value = roundedValue;
+  }
+};
+const handleChange = () => {
+  if (!inputRef.value) return;
+  const newValue = parseFloat(inputRef.value.value);
+  const precision = countDecimalPlaces(props.precision);
+  if (props.stripStep && newValue % props.precision !== 0) {
+    inputRef.value.value = value.value !== undefined ? value.value.toString() : "";
```
JoCa96, line 89: We should only use a single method to modify the value of the input. Preferably we use `v-model` or `:value`

packages/sit-onyx/src/components/OnyxStepper/OnyxStepper.vue
ChristianBusshoff: Math.floor(Math.log10(num)) isn't working for every Number. e.g. 1.5 would be 0

.changeset/olive-jobs-attend.md
```diff
@@ -0,0 +1,5 @@
+---
+"sit-onyx": minor
+---
+
+feat(OnyxStepper): implemnt `stripStep` and `precision`
```
JoCa96, line 5: typo
```suggestion
feat(OnyxStepper): implement `stripStep` and `precision`
```

JoCa96: Please mention the deprecation

PR #1981, opened 2024-10-21
fix: Fix uncaught TypeError for components with symbol props

### LINKED ISSUE #1980: [Bug]: Recent addition of skeleton symbols broke input components
### Version

sit-onyx@1.0.0-beta.55

### Description

It seems that #1938 broke components that handle input, like `OnyxInput`, `OnyxButton`, `OnyxStepper` and probably some more, throwing a very unhelpful error message from Vue:

```
Uncaught TypeError: Cannot convert a Symbol value to a string
    at styleValue (runtime-core.esm-bundler.js:4403:5)
    at getInvalidTypeMessage (runtime-core.esm-bundler.js:4386:25)
    at validateProp (runtime-core.esm-bundler.js:4345:14)
    at validateProps (runtime-core.esm-bundler.js:4317:5)
    at initProps (runtime-core.esm-bundler.js:4018:5)
    at setupComponent (runtime-core.esm-bundler.js:7829:3)
    at mountComponent (runtime-core.esm-bundler.js:5179:7)
    at processComponent (runtime-core.esm-bundler.js:5145:9)
    at patch (runtime-core.esm-bundler.js:4663:11)
    at mountChildren (runtime-core.esm-bundler.js:4895:7)
```

### Reproduction steps

1. install `sit-onyx@1.0.0-beta.55`
2. add a `<OnyxInput />` to your template

### Confirmations

- [X] I have checked that there is not already a bug report for my issue.
- [X] I confirm that are no sensitive information included in this bug report. e.g. credentials, internal project details (for employees of the Schwarz Group) etc.

JoCa96:
Closes #1980

**Background:**
1. Vue Prop Runtime Validation Vue validates prop types at runtime (e.g., `typeof prop === "boolean"`). If a prop doesn't pass the validation, an error is logged (only in DEV mode): `Invalid prop: type check failed for prop "${name}". Expected ${expectedTypes}.`

1. Vue Prop Type inference Vue supports different ways to define the types of props, which are used for the runtime validation. We define the props using [type-based declaration](https://vuejs.org/guide/typescript/composition-api.html#typing-component-props). This means that the runtime types are inferred based on the typescript types.

**Cause:**
In our case, Vue's compiler inferred the prop type used for runtime validation as `Boolean`, even though the typescript type is `typeof unique symbol | boolean`. This causes the runtime validation to fail and trying to log an error. Because Vue tries to create an error message using implicit string conversions and symbols cannot be implicitly [converted to a string,](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/toString#description) the JS runtime throws: `Uncaught (in promise) TypeError: can't convert symbol to string`

**Fix:**
Replaced all prop uses of `typeof <unique symbol>` with just `symbol`, this fixes the type inference and the runtime validation. Follow-up issues for Vue core will be created.

### DIFF
--- a/.changeset/tender-foxes-burn.md
+++ b/.changeset/tender-foxes-burn.md
@@ -1,0 +1,5 @@
+---
+"sit-onyx": patch
+---
+
+fix: Fix uncaught TypeError for components with symbol props
--- a/packages/sit-onyx/src/components/OnyxForm/OnyxForm.core.ts
+++ b/packages/sit-onyx/src/components/OnyxForm/OnyxForm.core.ts
@@ -55,7 +55,7 @@ export type FormInjectedProps = {
  * Symbol for the injected form injected properties.
  */
 export const FORM_INJECTED_SYMBOL = Symbol("FORM_INJECTED_SYMBOL");
-export type FORM_INJECTED = typeof FORM_INJECTED_SYMBOL;
+export type FORM_INJECTED = symbol; // we can't use `typeof FORM_INJECTED_SYMBOL` as vue is unable to infer its type: https://github.com/SchwarzIT/onyx/issues/1980
 /**
  * Prop type used by form child elements, which indicates that the prop value is taken from the parent form by default.
  * The props **MUST** use `FORM_INJECTED_SYMBOL` as default value.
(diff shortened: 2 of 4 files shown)


larsrickert (approved):
Can you create a follow up ticket to keep an eye on Vue and remove this workaround when its fixed in Vue?

Issue #1042, opened 2024-05-07
Migrate to eslint 9


larsrickert:
Follow up on #1041.
Wait for release of [Vue eslint config](https://github.com/vuejs/eslint-config-typescript/issues/76)

PR #1984, opened 2024-10-22
chore: migrate to eslint v9

JoCa96:
Relates to #1042


eslint.config.js
```diff
@@ -0,0 +1,190 @@
...
+
+/**
+ * General Vue and TypeScript lint rules without typechecked lint rules.
+ * Linting with typechecking has been disabled by default to improve lint performance.
+ * Enabling the projectService and typechecked lint rules increases linting times drastically (from 20s to about 100s for the whole project)
+ */
+const generalVueTsConfig = {
+  name: "general-vue-ts",
+  files: ["**/*.{js,jsx,ts,tsx,vue}"],
+  extends: [
```
larsrickert, line 22: Extends seems to be a removed feature which only works here since we use `tsEslint.config()` where I am not sure of we should use it and if/how long it will still be supported in the future

JoCa96: `extends` is a quality-of-life feature of the `typescript-eslint` package and for neither there is any indication that they will be deprecated.
The `extends` here is not intended and not used as a backwards compatiblity feature.
Instead we use it to improve readability and reduce duplications.
For all the reasons I think we are good here.

Issue #1852, opened 2024-09-11
Implement basic OnyxDataGrid useTableFeature composable


JoCa96:
### Depends on
- #1660 

### Description

The composable provides the basic API for implementing table features.

This is done by taking the provided user table data and transform and extend it.
Every feature can perform some transformations and wrap rendered slots.

In the first step, we implement the basic `useTableFeature` composable.
It should be enough to implement a `Sort` feature.

### Implementation details

The basic `useTableFeature` composable API should look like this:

```ts
// we need to intersect the empty array, so the array entry types are inferred
export const useTableFeatures = <
  T extends TableFeature<any, any, any, any, any, any>[] | [],
  TFeatureName extends symbol = ExtractTFeatureName<T>,
>(
  features: T,
) => {
return {
    /** Takes the column definition and maps all, calls mutation func and maps at the end to RendererCell */
    enrichTableData,
    /** Takes the column definition and creates a RenderHeader for each, adds actions from features */
    enrichHeaders,
    // the combined `watch` for all features
    watcheSources,
  };
}

export type TableEntry = {
  id: string | number;
  [key: PropertyKey]: unknown;
};


export type TableFeature<
  TEntry extends TableEntry,
  TFeatureName extends symbol,
> = {
  /**
   * Unique name and identifier of the table feature
   */
  name: TFeatureName;

  /**
   * An array of reactive states that should trigger a table re-calcualtion
   */
  watch: WatchSource[];

  /**
   * Allows modifying the table state as a whole.
   */
  mutation?: {
    func: (state: EntryState<TEntry>[]) => EntryState<TEntry>[];
  };

  /**
   * Allows the modification of the header columns before render.
   */
  header?: {
    /** actions are shown after the header label, later `listComponent` can be added */
    actions?: { iconComponent: Component, onTrigger?: (event: MouseEvent) => void }[];
  };
};
```

### Reference implementation

- https://github.com/SchwarzIT/onyx/blob/757-onyx-form/packages/sit-onyx/src/components/poc/OnyxDataGrid/OnyxDataGrid.feature.ts
- https://github.com/SchwarzIT/onyx/blob/13b8bb182b1fe048b1a7573f50c954a41721c378/packages/sit-onyx/src/components/poc/OnyxDataGrid/features/sorting/sorting.ts

### Design

https://www.figma.com/design/YfEUBOHk4J4nYrk04geswG/Onyx-Component-Library?node-id=4435-44768&node-type=text&m=dev

[...]

PR #2004, opened 2024-10-28
chore(OnyxDataGrid): Implement basic OnyxDataGrid feature API

JoCa96:
Relates to #1852 

Implemented basic `useTableFeature` API based on a simple sorting feature.
The API and the feature set will be extended in future tickets.


packages/sit-onyx/src/components/OnyxDataGrid/OnyxDataGrid.vue
```diff
@@ -0,0 +1,44 @@
...
+watch(
+  [columns, data, ...watchSources],
+  ([newColumns, newData]) => {
+    renderCols.value = createRendererColumns(newColumns);
+    renderRows.value = createRendererRows(newData, newColumns);
+  },
+  { immediate: true },
+);
+</script>
+<template>
```
larsrickert, line 42:
```suggestion
</script>

<template>
```

Newline for my inner monk 🙈

packages/sit-onyx/src/components/OnyxDataGrid/OnyxDataGrid.vue
```diff
@@ -0,0 +1,44 @@
...
+   */
+  data: TEntry[];
+}>();
+
+const withSorting = useTableSorting<TEntry>();
+
+const { watchSources, createRendererRows, createRendererColumns } = useTableFeatures([withSorting]);
+
+const renderCols: Ref<DataGridRendererColumn<TEntry, object>[]> = ref([]);
+const renderRows: Ref<DataGridRendererRow<TEntry, DataGridMetadata>[]> = ref([]);
```
larsrickert, line 29:
```suggestion
const renderCols = ref<DataGridRendererColumn<TEntry, object>[]>([]);
const renderRows = ref<DataGridRendererRow<TEntry, DataGridMetadata>[]>([]);
```

Is there a reason why you don't define it like this?

packages/sit-onyx/src/components/OnyxDataGrid/OnyxDataGrid.vue
```diff
@@ -0,0 +1,44 @@
...
+  columns: (keyof TEntry)[];
+  /**
+   * The data that should be used to fill the table.
+   */
+  data: TEntry[];
+}>();
+
+const withSorting = useTableSorting<TEntry>();
+
+const { watchSources, createRendererRows, createRendererColumns } = useTableFeatures([withSorting]);
```
larsrickert, line 26: Aren't the features passed as props by the project?

packages/sit-onyx/src/components/OnyxDataGrid/OnyxDataGrid.vue
```diff
@@ -0,0 +1,44 @@
...
+
+const { columns, data } = toRefs(props);
+
+watch(
+  [columns, data, ...watchSources],
+  ([newColumns, newData]) => {
+    renderCols.value = createRendererColumns(newColumns);
+    renderRows.value = createRendererRows(newData, newColumns);
+  },
+  { immediate: true },
```
larsrickert, line 39:
```suggestion
  { immediate: true },
```

Do we need a deep watcher here?

packages/sit-onyx/src/components/OnyxDataGrid/OnyxDataGrid.vue
```diff
@@ -0,0 +1,44 @@
...
+   * The data that should be used to fill the table.
+   */
+  data: TEntry[];
+}>();
+
+const withSorting = useTableSorting<TEntry>();
+
+const { watchSources, createRendererRows, createRendererColumns } = useTableFeatures([withSorting]);
+
+const renderCols: Ref<DataGridRendererColumn<TEntry, object>[]> = ref([]);
```
larsrickert, line 28:
```suggestion
const renderColumns: Ref<DataGridRendererColumn<TEntry, object>[]> = ref([]);
```

Isn't it a bad practice to use abbreviations for variable names?

packages/sit-onyx/src/components/OnyxDataGrid/features/index.ts
```diff
@@ -0,0 +1,178 @@
...
+  /**
+   * Allows the modification of the headers.
+   */
+  header?: {
+    /**
+     * Adds header icon button(s).
+     * `iconComponent` of an action is shown after the header label.
+     * The components must be ARIA-conform buttons.
+     */
+    actions?: (column: keyof TEntry) => {
```
larsrickert, line 32: Imho the name `actions` can be confusing since there might also be actions before the header, not only after it.
What do you think about making it flexible here and provide a `before` as well as `after` property where any components can be passed? Or we add a property `position: "before" | "after"` to each array entry

packages/sit-onyx/src/components/OnyxDataGrid/features/index.ts
```diff
@@ -0,0 +1,178 @@
...
+ * The properties are then used to render the data grid.
+ *
+ * Make use of the `watchSources` to trigger re-rendering when state changes occur.
+ * @example
+ * ```vue
+ * <script setup lang="ts">
+ * // ...
+ * // imports, props, emits, etc.
+ * const withHeaderButton = useTableHeaderButton<TEntry>();
+ *
```
larsrickert, line 79:
```suggestion
```

This is unused in the example

packages/sit-onyx/src/components/OnyxDataGrid/features/index.ts
```diff
@@ -0,0 +1,178 @@
...
+ *   { immediate: true },
+ * );
+ * </script>
+ * <template>
+ *   <OnyxDataGridRenderer :columns="renderCols" :rows="renderRows" />
+ * </template>
+ * ```
+ */
+export const useTableFeatures = <
+  T extends TableFeature<TEntry, symbol>[] | [],
```
larsrickert, line 102:
```suggestion
  T extends TableFeature<TEntry, symbol>[] | [],
```

Why is the `| []` needed?

packages/sit-onyx/src/components/OnyxDataGrid/features/index.ts
```diff
@@ -0,0 +1,178 @@
...
+
+      return {
+        key: column,
+        component: () =>
+          h("div", { class: getBemClass("header-cell") }, [
+            String(column),
+            ...iconActions.map((a) => h(a)),
+          ]),
+        props: {},
+      };
```
larsrickert, line 128: Not sure if I like this inline component  + CSS in JS...what do you think about creating a .vue component for this?

packages/sit-onyx/src/components/OnyxDataGrid/features/index.ts
```diff
@@ -0,0 +1,178 @@
...
+    });
+  };
+
+  const createRendererRows = (
+    entries: TEntry[],
+    columns: (keyof TEntry)[],
+  ): DataGridRendererRow<TEntry, DataGridMetadata>[] => {
+    const mutations = features.map((f) => f.mutation).filter((m) => !!m);
+    const shallowCopy = [...entries];
```
larsrickert, line 137: Why do you make a shallow copy here? If needed, should we do a `structuredClone` instead?
Because otherwise, the user data might be manipulated because objects are passed as reference

packages/sit-onyx/src/i18n/locales/en-US.json
```diff
@@ -4,6 +4,13 @@
   "apply": "Apply",
   "back": "Back",
   "close": "Close",
+  "data-grid": {
+    "head": {
+      "sorting-action-asc": "Click to sort the table ascending by the {field} column.",
+      "sorting-action-desc": "Click to sort the table descending by the {field} column.",
+      "sorting-action-none": "Click to reset the sorting order."
```
larsrickert, line 11:
```suggestion
      "sorting-action-asc": "Sort the table ascending by the {field} column.",
      "sorting-action-desc": "Sort the table descending by the {field} column.",
      "sorting-action-none": "Reset the sorting order."
```

The action can also be triggered via keyboard, right? So technically "Click to" is not correct

packages/sit-onyx/src/components/OnyxDataGrid/features/sorting/SortAction.vue
```diff
@@ -0,0 +1,41 @@
...
+
+const buttonLabel = computed(() => {
+  const nextDirection = nextSortDirection(props.sortDirection);
+  return t.value(`data-grid.head.sorting-action-${nextDirection}`, {
+    field: props.columnLabel,
+  });
+});
+</script>
+<template>
+  <OnyxIconButton color="neutral" :icon="icon" size="12px" :label="buttonLabel" />
```
larsrickert, line 40:
```suggestion
  <OnyxIconButton color="neutral" :icon="icon" :label="buttonLabel" />
```

The icon button does not have a size property (yet).

Also, are you sure that this is the correct design? Looks quiet big + no gap on hover
image

image

packages/sit-onyx/src/components/OnyxDataGrid/features/sorting/sorting.ts
```diff
@@ -0,0 +1,68 @@
+import { computed, h, ref } from "vue";
+import { createFeature } from "..";
+import { injectI18n } from "../../../../i18n";
+import type { DataGridEntry } from "../../types";
+import SortAction from "./SortAction.vue";
+
+export type SortDirection = "asc" | "desc" | "none";
```
larsrickert, line 7:
```suggestion
export type SortDirection = "asc" | "desc";
```

What do you think about making undefined the "none" state?

packages/sit-onyx/src/i18n/locales/en-US.json
```diff
@@ -4,6 +4,13 @@
   "apply": "Apply",
   "back": "Back",
   "close": "Close",
+  "data-grid": {
```
larsrickert, line 7:
```suggestion
  "dataGrid": {
```

We usually use camelCase for translation keys :)

packages/sit-onyx/src/components/OnyxDataGrid/features/sorting/SortAction.vue
```diff
@@ -0,0 +1,41 @@
...
+);
+
+const buttonLabel = computed(() => {
+  const nextDirection = nextSortDirection(props.sortDirection);
+  return t.value(`data-grid.head.sorting-action-${nextDirection}`, {
+    field: props.columnLabel,
+  });
+});
+</script>
+<template>
```
larsrickert, line 39:
```suggestion
</script>

<template>
```

Inner monk :D

packages/sit-onyx/src/components/OnyxDataGrid/OnyxDataGrid.stories.ts
```diff
@@ -0,0 +1,21 @@
+import type { Meta, StoryObj } from "@storybook/vue3";
+import OnyxDataGrid from "./OnyxDataGrid.vue";
+
+const meta: Meta<typeof OnyxDataGrid> = {
+  title: "DataGrid/DataGrid",
```
larsrickert, line 5:
```suggestion
  title: "Data/DataGrid",
```

Wouldn't the data category be more fitting here? Would also be aligned with Figma then

packages/sit-onyx/src/components/OnyxDataGrid/features/sorting/SortAction.vue
larsrickert, line 40: If no sort order is set, the icon should only be shown on hover: https://www.figma.com/design/YfEUBOHk4J4nYrk04geswG/Onyx-Component-Library?node-id=4435-44766&m=dev

packages/sit-onyx/src/components/OnyxDataGrid/features/sorting/sorting.ts
```diff
@@ -0,0 +1,68 @@
...
+      return "none";
+    case "none":
+    default:
+      return "asc";
+  }
+};
+
+const SORTING_FEATURE = Symbol("Sorting");
+
+export const useTableSorting = createFeature(<TEntry extends DataGridEntry>() => {
```
larsrickert, line 23:
```suggestion
export const useDataGridSorting = createFeature(<TEntry extends DataGridEntry>() => {
```

Its about the data grid, not the table right?

packages/sit-onyx/src/components/OnyxDataGrid/features/index.ts
```diff
@@ -0,0 +1,178 @@
+import { h, type Component, type WatchSource } from "vue";
+import type { DataGridRendererColumn, DataGridRendererRow } from "../../..";
+import type { DataGridEntry, DataGridMetadata } from "../types";
+
+export type TableFeature<TEntry extends DataGridEntry, TFeatureName extends symbol> = {
```
larsrickert, line 5:
```suggestion
export type DataGridFeature<TEntry extends DataGridEntry, TFeatureName extends symbol> = {
```

Isn't this about the data grid, not the table?

larsrickert: Same for other docs/functions in this file, e.g. `useDataGridFeature` instead of `useTableFeature`

packages/sit-onyx/src/components/OnyxDataGrid/features/sorting/sorting.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  const handleClick = (column: keyof TEntry) => {
+    if (sortColumn.value === column) {
+      sortDirection.value = nextSortDirection(sortDirection.value);
+    } else {
+      sortColumn.value = column;
+      sortDirection.value = nextSortDirection();
+    }
+    if (sortDirection.value === "none") {
+      sortColumn.value = undefined;
```
larsrickert, line 37: Would this cause a second re-render / watcher trigger because `sortColumn.value` is already changed before

packages/sit-onyx/src/components/OnyxDataGrid/features/sorting/sorting.ts
```diff
@@ -0,0 +1,68 @@
...
+    }
+  };
+
+  const sortData = (data: TEntry[]) => {
+    const column = sortColumn.value;
+    if (!column || sortDirection.value === "none") {
+      return;
+    }
+    const multiplicand = sortDirection.value === "asc" ? 1 : -1;
+    data.sort((a, b) => intlCompare.value(String(a[column]), String(b[column])) * multiplicand);
```
larsrickert, line 47: Sort modifies the array in place but this is not critical because we pass cloned data, right?

packages/sit-onyx/src/components/OnyxDataGrid/OnyxDataGrid.stories.ts
```diff
@@ -0,0 +1,21 @@
...
+export default meta;
+type Story = StoryObj<typeof OnyxDataGrid>;
+
+export const Default = {
+  args: {
+    columns: ["name", "age", "birthday"],
+    data: [
+      { id: 1, name: "Alice", age: 30, birthday: new Date("1990-01-01") },
+      { id: 2, name: "Bob", age: 25, birthday: new Date("1995-06-15") },
+      { id: 3, name: "Charlie", age: 35, birthday: new Date("1998-02-11") },
```
larsrickert, line 18:
```suggestion
      { id: 1, name: "Alice", age: 30, birthday: new Date("1990-01-01") },
      { id: 2, name: "Charlie", age: 35, birthday: new Date("1998-02-11") },
      { id: 3, name: "Bob", age: 25, birthday: new Date("1995-06-15") },
```

We could make the default order unsorted so the user can notice a difference between no sorting and ascending

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxDataGrid/features/sorting/sorting.ts
JoCa96: No, as long as they happen in the same tick. Vue.js batches the changes afaik.

JoCa96: See https://vuejs.org/guide/essentials/watchers.html#callback-flush-timing

packages/sit-onyx/src/components/OnyxDataGrid/features/sorting/sorting.ts
JoCa96: Yes, for the mutations, it's even intentional :)

packages/sit-onyx/src/components/OnyxDataGrid/OnyxDataGrid.vue
JoCa96: Yes, that will come in later PRs

packages/sit-onyx/src/components/OnyxDataGrid/OnyxDataGrid.vue
JoCa96: Afaik, deep watchers are only used for function watchers: `() => watchedData`. We only use `Ref`s which are already tracking deep access.

packages/sit-onyx/src/components/OnyxDataGrid/features/index.ts
JoCa96: Yes, we can certainly do that when we have requirements for it.

packages/sit-onyx/src/components/OnyxDataGrid/OnyxDataGrid.vue
JoCa96: I added a comment, doing it this way to avoid `UnwrapRef` issues. There are mismatched types when binding it to the renderer, otherwise.

packages/sit-onyx/src/components/OnyxDataGrid/features/sorting/sorting.ts
JoCa96: done

packages/sit-onyx/src/components/OnyxDataGrid/OnyxDataGrid.vue
```diff
@@ -0,0 +1,44 @@
...
+  OnyxDataGridRenderer,
+  type DataGridEntry,
+  type DataGridMetadata,
+  type DataGridRendererColumn,
+  type DataGridRendererRow,
+} from "../..";
+import { useTableFeatures } from "./features";
+import { useTableSorting } from "./features/sorting/sorting";
+
+const props = defineProps<{
```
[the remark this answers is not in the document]
JoCa96: I extracted the props and added an experimental annotation. But we don't even export the the grid, so there should not be any issues.
Would like to extend the props and feature configuration in the follow-up PRs.

packages/sit-onyx/src/components/OnyxDataGrid/features/index.ts
JoCa96: done

packages/sit-onyx/src/components/OnyxDataGrid/features/index.ts
JoCa96: used now :)

packages/sit-onyx/src/components/OnyxDataGrid/features/index.ts
JoCa96: We had issues in the past with `structuredClone` and `ref`/`reactive` properties that might exist in the entries.
Also, conceptually, I want the mutation funcs to only modify the array itself and not the entries. Therefore, I wrapped the `TEntry` with `Readonly`, so that we write protect the entries via TypeScript.

JoCa96: We will probably freeze the entries for real when we implement the `mapping` part of the feature API.

JoCa96: `toRaw` does not do a deep unref

packages/sit-onyx/src/components/OnyxDataGrid/OnyxDataGridRenderer/OnyxDataGridRenderer.vue
```diff
@@ -42,6 +42,10 @@ const props = withDefaults(defineProps<OnyxDataGridRendererProps<TEntry, TMetada
 
 .onyx-data-grid {
   @include layers.component() {
+    &-header-cell {
```
larsrickert, line 45: This is obsolete now because of the HeaderCell, right?

packages/sit-onyx/src/components/OnyxDataGrid/features/sorting/SortAction.vue
JoCa96: out-of-scope for now

larsrickert (approved):

Issue #742, opened 2024-03-22
Implement additional data storybook examples for OnyxSelect options


larsrickert:
### Depends on

- #564
- #568
- #566
- #883

### Why?

The select option should be extended to support displaying additional data.

### Design

[Figma link](https://www.figma.com/design/YfEUBOHk4J4nYrk04geswG/Onyx-Component-Library?node-id=4418-27008&m=dev).

### Acceptance criteria

- [x] a badge can be shown
- [x] a tag can be shown
- [x] a suffix text can be displayed

#### Out of scope

- keyboard shortcut
- consider implementing a dedicated component for keyboard shortcuts (e.g. with a prop to switch between windows and mac style)

### Definition of Done

- [x] covered by component tests
- [x] screenshot tests are updated
- [x] follow-up tickets were created if necessary
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature
- [x] Set task to `In Approval`

**Approval**

- [Storybook](https://storybook.onyx.schwarz/?path=/story/form-elements-select--with-custom-options)
- [x] approved by designer @mj-hof 
- [x] approved by dev @larsrickert 

### Implementation details

_add details if necessary, e.g.,_

- suffix text should support string values (not only keyboard shortcuts)
- for flexibility, use slots

BoppLi:
Info: started in feat/742-additional-select-data, but aborted. there are plans to refactor flyout contents to be slot based. this impacts the way this feature is implemented greatly.

PR #2019, opened 2024-10-29
chore(OnyxSelect): Add new OnyxSelect storybook example

MajaZarkova:
Relates to #742 

Added new storybook example which includes custom options with additional data added via slot.


packages/sit-onyx/src/components/OnyxSelect/OnyxSelect.stories.ts
```diff
@@ -515,3 +528,25 @@ export const WithRightAlignment = {
     placeholder: "Right alignment...",
   },
 } satisfies Story;
+
+/**
+ * This example shows a single select with options containing additional data, passed as slot.
+ */
+export const WithAdditionalData = {
```
larsrickert, line 535: We already have an example with custom data, would prefer to replace it with your example

larsrickert (changes requested):

larsrickert (approved):

Issue #986, opened 2024-04-22
Implement "OnyxMoreList" component


MajaZarkova:
### Depends on

- #874 
- #867

### Why?

This component should be used inside the navigationBar. While resizing the navItems should be grouped inside one flyout.
Therefore, we want to provide a generic "OnyxMoreList" component, where any list of components can be used to show any "more indicator" component.

### Acceptance criteria

- [x] should show all elements, that fully fit into the available width
- [x] remaining items are hidden and a more indicator is shown (e.g. +3 more)
- [x] when resizing, the elements visibility is re-checked and updated if needed

### Definition of Done

- [x] covered by component tests
- [x] follow-up tickets were created if necessary
    - #2171
    - #2228
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature

**Approval**

- [Storybook](https://storybook.onyx.schwarz/?path=/docs/support-morelist--docs)
- [x] approved by designer @jannick-ux
- [x] approved by dev @MajaZarkova

larsrickert:
Reference implementation for a chip list that we can re-use:
```ts
<script lang="ts" setup>
import { ScuChip } from '@scu/vue';
import { useIntersectionObserver } from '@vueuse/core';
import { computed, onBeforeUnmount, ref, watchEffect } from 'vue';

export interface Chip {
  label: string;
  icon?: string;
}

const props = defineProps<{
  chips: Chip[];
  color?: 'dark';
  /**
   * If `true`, all chips are shown and will be wrapped to the next lines if they exceed the containers max width.
   * Otherwise only chips that fit into the container width are shown and a remaining items indicator is shown.
   */
  wrap?: boolean;
}>();

const wrapper = ref<HTMLDivElement | null>(null);

const visibleChips = ref(0);
const remainingChips = computed(() => props.chips.length - visibleChips.value);
const observer = ref<IntersectionObserver>();
onBeforeUnmount(() => observer.value?.disconnect());

watchEffect(() => {
  if (observer.value && props.chips.length) {
    observer.value.disconnect();
    const chips = Array.from(wrapper.value?.querySelectorAll('scu-chip') ?? []);
    chips.forEach((chip) => observer.value?.observe(chip));
  }
});

watchEffect(() => {
  if (!wrapper.value || props.wrap) return;
  observer.value?.disconnect();

  observer.value = new IntersectionObserver(
    (res) => {
      // res contains all changed chips (not all available chips)
[...]

PR #2022, opened 2024-10-30
feat: implement `OnyxMoreList` component

larsrickert:
Relates to #986

Implement a generic `OnyxMoreList` component that can be used to render any list of components with a "+ more" indicator.
I will integrate it in following PRs into the nav bar as well as implementing the positioning of the more indicator which is currently not visible at all.

## Checklist


packages/sit-onyx/src/components/OnyxMoreList/TestWrapper.ct.vue
```diff
@@ -0,0 +1,47 @@
...
+import OnyxNavButton from "../OnyxNavBar/modules/OnyxNavButton/OnyxNavButton.vue";
+import { NAV_BAR_MORE_LIST_INJECTION_KEY } from "../OnyxNavBar/types";
+import OnyxMoreList from "./OnyxMoreList.vue";
+import type { MoreListSlotBindings } from "./types";
+
+const props = defineProps<{
+  /**
+   * Number of components to show. Can also be decimal.
+   */
+  count?: number;
```
MajaZarkova, line 11: Why do we need decimal?

MajaZarkova (approved):

packages/sit-onyx/src/components/OnyxMoreList/TestWrapper.ct.vue
larsrickert: This is just for the Playwright test so you can e.g. pass 2.5 to test that then only 2 are fully shown and the third is hidden :)

packages/sit-onyx/src/composables/useMoreList.ts
```diff
@@ -0,0 +1,216 @@
...
+        (changeEntries) => {
+          // changeEntries contains all changed component visibilities (not all available components)
+          // if component is shown, intersectionRatio is 1, otherwise its completely or partially hidden
+          const shownIds: string[] = [];
+          const hiddenIds: string[] = [];
+
+          changeEntries.forEach((entry) => {
+            const elementId = Array.from(options.componentRefs).find(([_, element]) => {
+              return getTemplateRefElement(unref(element))?.isSameNode(entry.target);
+            })?.[0];
```
JoCa96, line 121: `isSameNode` is a [legacy function](https://developer.mozilla.org/en-US/docs/Web/API/Node/isSameNode):
```suggestion
            const elementId = Array.from(options.componentRefs).find(([_, element]) => 
              getTemplateRefElement(unref(element))? === entry.target
            )?.[0];
```

MajaZarkova (approved):

PR #2030, opened 2024-10-30
chore(docs): add attribution for contributors

JoCa96:
- load contributor and team data from GitHub API
- add attribution for bots and contributors
- make use of `VPTeamPage` components


JoCa96:
@larsrickert Was there a specific reason why we didn't use all the team components?
```
import {
  VPTeamPage,
  VPTeamPageTitle,
  VPTeamMembers,
  VPTeamPageSection
} from 'vitepress/theme' 
```

apps/docs/src/.vitepress/components/OnyxHomePage.stories.ts
```diff
@@ -18,6 +18,7 @@ type Story = StoryObj<typeof OnyxHomePage>;
 export const Default = {
   args: {
     data: {
+      contributors: [],
```
larsrickert, line 21: Is this a leftover?

apps/docs/src/about/team.md
```diff
@@ -1,72 +1,128 @@
...
-import { VPTeamMembers } from 'vitepress/theme'
+import { data } from "../index.data";
+import {
+  VPTeamPage,
+  VPTeamPageTitle,
+  VPTeamMembers,
+  VPTeamPageSection
+} from 'vitepress/theme'
+
+const shuffleArray = (array) =>
```
larsrickert, line 14: I get your intention with shuffeling but imho this is confusing. What do you think about ordering the core team by title and the contributors by number/size of contributions

apps/docs/src/about/team.md
```diff
@@ -1,72 +1,128 @@
...
     name: 'Jannick Keller',
+    core: true,
     title: 'Lead Designer',
-    links: [
-      { icon: 'github', link: 'https://github.com/jannick-ux' },
-    ]
   },
   {
-    avatar: 'https://www.github.com/JoCa96.png',
+    login: "flubnau",
```
larsrickert, line 37: Flo is not shown currently, propably because he hasn't contributed to GitHub yet.
But we still want to show him, right?

apps/docs/src/about/team.md
```diff
@@ -1,72 +1,128 @@
...
     title: 'Product Owner',
-    links: [
-      { icon: 'github', link: 'https://github.com/mj-hof' },
-    ]
   },
   {
-    avatar: 'https://www.github.com/jannick-ux.png',
+    login: "jannick-ux",
     name: 'Jannick Keller',
+    core: true,
```
larsrickert, line 33: Since you are the "Engineering Lead", would it be more unified if Jannick is the "Design Lead" instead of "Lead Designer"?

larsrickert (changes requested):

larsrickert:
> @larsrickert Was there a specific reason why we didn't use all the team components?
> 
> ```
> import {
>   VPTeamPage,
>   VPTeamPageTitle,
>   VPTeamMembers,
>   VPTeamPageSection
> } from 'vitepress/theme' 
> ```

In my opinion, the page looks a little weird with all those components. Would prefer personally to use regular h1 headline and description. I guess these components are mainly intended to be used on the home page to fit its look. We could also think about moving the team + contributors to the home page
image

apps/docs/src/about/team.md
```diff
@@ -1,72 +1,128 @@
...
+    </template>
 
-<VPTeamMembers size="small" :members="members" />
+  </VPTeamPageTitle>
+  
+  <VPTeamMembers size="medium" :members="shuffleArray(coreMembers)" />
+  <VPTeamPageSection>
+    <template #title>Thank you to all contributors 🙏</template>
+    <template #members>
+      <VPTeamMembers size="small" :members="shuffleArray(contributors)" />
```
larsrickert, line 119: I am not sure how this will scale in the future if we might have > 50 contributors since the page will get very large then...

apps/docs/src/about/team.md
JoCa96: He should show up now, as he create an issue

JoCa96:
@larsrickert I created a follow-up ticket: [#2116](https://github.com/SchwarzIT/onyx/issues/2116)

.gitignore
```diff
@@ -15,7 +15,7 @@ node_modules/
 .npmrc
 **/.cache
 **/.vitepress/cache
-**/.vitepress/config.ts.timestamp-*.mjs
+**/.vitepress/*.ts.timestamp*.mjs
```
larsrickert, line 18:
```suggestion
*.ts.timestamp*.mjs
```

When starting the docs, I also get another timestamp file which is not inside the .vitepress folder, should we just ignore them all?

larsrickert (approved):

Issue #573, opened 2024-03-11
Implement success state for OnyxFormElement


larsrickert:
### Depends on

- #189

### Why?

There are cases where it is necessary to communicate an explicit `success` state for an FormElement:
- To highlight the validity of important inputs (E.g., a password input on sign-up)
- To show validity after an asynchronous validation (E.g. Chosen username is available).
- To show the successful execution of an action related to the input (E.g., Value copied to the clipboard)

The (validation) `valid` state independent of the `success` state:
- the `success` state is set explicitly and **not** the implicit result of validation
- the validation `error` state has higher priority than the `success`
- the `success` state does not change or affect the validation
- for now only support for inputs (text, select) and not for textarea

### Design

https://www.figma.com/design/EBi98T51T9fGfjrYYfsulg/onyx---Archive?node-id=3427-73120&t=WFV3kII28sKTTZft-4

### Acceptance criteria
- [x] A new type `FormMessage` is created
- [x] `message`, `successMessage`, `errorMessage` all use the new type
    -  * Message precedence from highest to lowest:
      1. `errorMessage`
      2. html native validation error message
      3. `successMessage`
      4. `message`
- [x] when `successMessage` is shown, a green checkmark symbol is shown on the right -> Follow up #2143 
- [x] All existing components that use the `OnyxFormElement` are extended to support the new API
- [x] Follow-up from #1731: Refactor to only show `onyx-form-element__footer` when any message is shown: Right now the gap is from the parent is applied even when no message is visible

### Definition of Done

- [x] covered by component tests
- [x] screenshot tests are updated
- [x] follow-up tickets were created if necessary
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature
- [x] Set task to `In Approval`

**Review**

- [Input](https://storybook.onyx.schwarz/?path=/story/form-elements-input--success-input)
- [Textarea](https://storybook.onyx.schwarz/?path=/story/form-elements-textarea--success-message)
- [x] approved by designer @jannick-ux
- [x] approved by dev @JoCa96 

### Implementation details

_Consider that success might be shown without message, but error should never be shown without message. Also, do we still need a message for a11y reasons?_ 

```ts
/**
  * If it's just a string it acts as `FormErrorMessage.shortMessage`
  */
export type FormErrorMessage = string | {
  /**
   * A short error message preview that is shown permanently beneath the input field
   */
  shortMessage: string;
  /**
   * An extended message that is shown temporarily after hovering over the message tooltip.
[...]

PR #2031, opened 2024-10-31
feat(OnyxForm): Add successMessage to useCustomValidity composable

MajaZarkova:
Relates to #573 

Add successMessage prop to useCustomValidity composable. This property will be provided in OnyxForm component and later injected by it's child input components to represent the success state.

### DIFF
--- a/packages/sit-onyx/src/composables/useCustomValidity.ts
+++ b/packages/sit-onyx/src/composables/useCustomValidity.ts
@@ -6,13 +6,17 @@ import type { BaseSelectOption } from "../types";
 import { areObjectsFlatEqual } from "../utils/objects";
 import { getFirstInvalidType, transformValidityStateToObject } from "../utils/validity";
 
-export type CustomErrorType = string | FormErrorMessages;
+export type CustomMessageType = string | FormMessages;
 
 export type CustomValidityProp = {
   /**
    * Custom error message to show. Will only show up after the user has interacted with the input.
    */
-  customError?: CustomErrorType;
+  customError?: CustomMessageType;
+  /**
+   * Custom success message to show. Will only show up after the user has interacted with the input.
+   */
+  customSuccess?: CustomMessageType;
 };
 
 export type UseCustomValidityOptions = {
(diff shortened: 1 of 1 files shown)


packages/sit-onyx/src/composables/useCustomValidity.ts
```diff
@@ -6,13 +6,17 @@ import type { BaseSelectOption } from "../types";
...
 export type CustomValidityProp = {
   /**
    * Custom error message to show. Will only show up after the user has interacted with the input.
    */
-  customError?: CustomErrorType;
+  customError?: CustomMessageType;
+  /**
+   * Custom success message to show. Will only show up after the user has interacted with the input.
+   */
+  customSuccess?: CustomMessageType;
```
larsrickert, line 19: For errors there are native HTML error messages so we call this property `customError`.
For success tho, there are no default success messages, so I'd prefer to e.g. call this `successMessage`.

We could also think about refactoring our existing `message` and `messageTooltip` props to merge them to use a single `CustomMessageType` property

packages/sit-onyx/src/composables/useCustomValidity.ts
```diff
@@ -45,39 +49,38 @@ export const TRANSLATED_INPUT_TYPES = Object.keys(
...
+  return customMessage;
 };
 
 /**
- * Returns a string combining short + long message or just the customError if it was provided as single string.
+ * Returns a string combining short + long message or just the customMessage if it was provided as single string.
  * Will be used e.g. for customInvalidity and showing a tooltip e.g. in RadioButtons
  */
-export const getCustomErrorText = (customError?: CustomErrorType): string | undefined => {
+export const getCustomText = (customError?: CustomMessageType): string | undefined => {
```
larsrickert, line 83:
```suggestion
export const getCustomText = (customError?: CustomMessageType): string | undefined => {
```

What do you think about renaming this to e.g. `getFormMessageText`? So its more clear what it does
Similar, `getCustomMessages` could be renamed to `getFormMessages`

packages/sit-onyx/src/composables/useCustomValidity.ts
```diff
@@ -207,6 +210,22 @@ export const useCustomValidity = (options: UseCustomValidityOptions) => {
...
+  const successMessages = computed<FormMessages | undefined>(() => {
+    if (validityState.value === undefined || !validityState.value.valid) return;
+
+    const errorType = getFirstInvalidType(validityState.value);
+    const customErrors = getCustomMessages(options.props.customError);
+
+    // a custom error message always is considered first
+    if (customErrors || errorType === "customError") {
+      if (customErrors) return;
+    }
```
larsrickert, line 222:
```suggestion
```

Why is this needed? If custom errors would exist `validatyState.value-valid` would be false which is already covered in the if above :)

larsrickert (changes requested):

packages/sit-onyx/src/composables/useCustomValidity.ts
MajaZarkova: yes, good point here :)

packages/sit-onyx/src/composables/useCustomValidity.ts
```diff
@@ -207,6 +210,14 @@ export const useCustomValidity = (options: UseCustomValidityOptions) => {
     };
   });
 
+  const successMessages = computed<FormMessages | undefined>(() => {
+    if (validityState.value === undefined || !validityState.value.valid) return;
+
+    const customMessage = getFormMessages(options.props.customMessage);
+
+    return customMessage;
+  });
```
larsrickert, line 219:
```suggestion
  const successMessages = computed<FormMessages | undefined>(() => {
    if (validityState.value === undefined || !validityState.value.valid) return;
    return getFormMessages(options.props.customMessage);
  });
```

We can make this more compact :)

packages/sit-onyx/src/composables/useCustomValidity.ts
```diff
@@ -6,13 +6,17 @@ import type { BaseSelectOption } from "../types";
...
 export type CustomValidityProp = {
   /**
    * Custom error message to show. Will only show up after the user has interacted with the input.
    */
-  customError?: CustomErrorType;
+  customError?: CustomMessageType;
+  /**
+   * Custom success message to show. Will only show up after the user has interacted with the input.
+   */
+  customMessage?: CustomMessageType;
```
larsrickert, line 19:
```suggestion
  /**
   * Success message to show. Will only show up after the user has interacted with the input.
   */
  successMessage?: CustomMessageType;
```

Is this property only intended to be the green success state? i yes, what do you think about renaming it like this so its clearer?

larsrickert (changes requested):

larsrickert (approved):

Issue #736, opened 2024-03-22
Implement design token update for docs


larsrickert:
### Why?

Since we also have tokens with a very long value, we want to update the design for all design tokens as shown below.

### Design

[Figma link](https://www.figma.com/file/EBi98T51T9fGfjrYYfsulg/onyx---Archive?type=design&node-id=3337-4232&mode=dev)

![Image](https://github.com/SchwarzIT/onyx/assets/67898185/dea68fb0-17d4-4d4a-92a0-fd5d2d3946c7)

### Acceptance criteria

- the design is updated for:
  - [x] spacings
  - [x] border radius
  - [x] shadows

#### Out of scope:

- colors

### Definition of Done

- covered by tests
  - [x] visual tests (Playwright screenshots)
- [x] updated version + documentation is deployed

**Approval**

- [Deployed Docs](https://onyx.schwarz/variables/shadows.html)
- [x] approved by designer @jannick-ux
- [x] approved by dev @JoCa96

mj-hof:
@jannick-ux will deliver the mobile behaviour

PR #2032, opened 2024-10-31
docs: update variable design

ChristianBusshoff:
relates to #736 

- Update variable design for docs
- fix monospace font not displayed on Typography page
- fix docs Storybook not working correctly because of invalid tsconfig setting


apps/docs/src/.vitepress/components/DesignVariable.vue
```diff
@@ -97,10 +102,17 @@ const emit = defineEmits<{
     &--no-value {
       padding: var(--onyx-spacing-3xs) var(--onyx-spacing-2xs);
     }
+    &--no-padding {
+      padding: 0;
+    }
   }
 
   &__value {
+    color: var(--onyx-color-base-neutral-500);
```
larsrickert, line 111:
```suggestion
    color: var(--onyx-color-text-icons-neutral-soft);
```

Variable as used in Figma :) Has better dark mode contrast

apps/docs/src/.vitepress/components/DesignVariable.vue
```diff
@@ -97,10 +102,17 @@ const emit = defineEmits<{
...
     font-weight: 600;
+    font-size: 0.8125rem;
+    border-top: 1px solid var(--onyx-color-base-neutral-300);
```
larsrickert, line 114:
```suggestion
    border-top: var(--onyx-1px-in-rem) solid var(--onyx-color-base-neutral-300);
```

We want to always use rem

apps/docs/src/.vitepress/components/DesignVariable.vue
```diff
@@ -97,10 +102,17 @@ const emit = defineEmits<{
...
     }
+    &--no-padding {
+      padding: 0;
+    }
   }
 
   &__value {
+    color: var(--onyx-color-base-neutral-500);
     font-weight: 600;
+    font-size: 0.8125rem;
```
larsrickert, line 113: (optional) we also have a `onyx-text--small` class for this so we don't need to remember this weird rem value all the time :D

apps/docs/src/.vitepress/components/OnyxShadowVariables.vue
```diff
@@ -12,13 +12,7 @@ const shadows = [
...
 
 <template>
-  <DesignVariableCard
-    v-for="shadow in shadows"
-    :key="shadow"
-    :name="shadow"
-    class="variable"
-    hide-value
-  >
+  <DesignVariableCard v-for="shadow in shadows" :key="shadow" :name="shadow" class="variable">
```
larsrickert, line 15: I guess the `hide-value` property is now unused so we can remove it from the `DesignVariableCard`

apps/docs/src/.vitepress/components/DesignVariableCard.vue
larsrickert, line 1: Could you also please check the responsive design? :)
image

apps/docs/src/.vitepress/components/DesignVariable.vue
larsrickert, line 1: On the colors page when copying a value, the token padding is off (should not have left padding bot more top padding)

image

apps/docs/src/.vitepress/components/DesignVariable.vue
```diff
@@ -54,6 +55,9 @@ const emit = defineEmits<{
 
     <OnyxIcon v-else class="variable__copy" :icon="copyIcon" color="primary" />
   </button>
+  <div v-if="props.value && props.type === 'value'" class="variable__value">
```
larsrickert, line 58: In my opinion when looking at the design, the value should be part of the card component instead of the variable to keep things separated :) E.g. on the colors page its not needed

larsrickert (changes requested):

JoCa96 (approved):

Issue #2018, opened 2024-10-29
Implement OnyxTabs


JoCa96:
### Why?

We want to offer a tabs component where content can be grouped in tabs for structuring page or sidebar content.

### Design

[Figma](https://www.figma.com/design/YfEUBOHk4J4nYrk04geswG/Onyx-Component-Library?node-id=7162-15700&t=w1vkWbCstQZGcKGn-4)

### Acceptance criteria

- [x] the design is implemented
- [x] only one tab can be active at the same time
- [x] a tab can be disabled
- [x] arbitrary custom content can be placed in the tab (e.g. icon, badge etc.)
- [x] the [accessibility pattern](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/) is implemented

#### Follow up tickets

- #2079

### Implementation details

- inactive tabs should not be rendered (v-if) to increase performance and unwanted side-effects (e.g. onMounted triggered although the content is not visible)
- we can use provide/inject to inject which tab is active
- tab value/ID should allow any `PropertyKey` (string, number, symbol)

##### Option 1 (recommendation by @larsrickert):

| Advantages                                                    | Disadvantages                                   |
| ------------------------------------------------------------- | ----------------------------------------------- |
| Tab and panel content are grouped together (less error prone) | Unable to have same content for multiple panels |

```vue
<script lang="ts" setup>
const activeTab = ref("tab-1");
</script>

<template>
  <OnyxTabs v-model="activeTab">
    <OnyxTab value="tab-1" label="Tab 1"> Content for tab 1... </OnyxTab>

    <OnyxTab value="tab-2" label="Tab 2">
      <template #tab="{ label, active }">
        {{ label }} is active: {{ active }}
        <OnyxBadge dot />
      </template>

      Content for tab 1...
    </OnyxTab>
  </OnyxTabs>
</template>
```

##### Option 2:

| Advantages                                      | Disadvantages                                                                                              |
| ----------------------------------------------- | ---------------------------------------------------------------------------------------------------------- |
| Panel allows shared content for multiple panels | v-if has to be done on every panel by the user                                                             |
|                                                 | tab and panel for the same tab are separated in the code                                                   |
|                                                 | user must make sure that tab and panel IDs are matching (can lead to unexpected behavior if changing them) |

```vue
<script lang="ts" setup>
const activeTab = ref("tab-1");
</script>

<template>
  <OnyxTabs v-model="activeTab">
[...]

PR #2040, opened 2024-11-05
feat: implement basic `OnyxTabs` component

larsrickert:
Relates to #2018

Implement basic `OnyxTabs` and `OnyxTab` component.

## Checklist


packages/sit-onyx/src/components/OnyxTab/OnyxTab.vue
```diff
@@ -0,0 +1,117 @@
...
+  /**
+   * Optional slot to override the tab content. By default, the `label` property will be displayed.
+   */
+  tab?(): unknown;
+}>();
+
+const { densityClass } = useDensity(props);
+const tabsContext = inject(TABS_INJECTION_KEY);
+
+const isActive = computed(() => tabsContext?.headless.state.isVisible.value(props.value) ?? false);
```
JoCa96, line 23: We should use the same name for the state (*active* vs. *selected*) to reduce cognitive overhead:
```suggestion
const isSelected = computed(() => tabsContext?.headless.state.isSelected.value(props.value) ?? false);
```

JoCa96, line 23: IMHO, the template would be cleaner, if we use computed to extract the elements.
Also, it seems like the `isVisible` function is redundant, when we use ´tab.["aria-selected"]´
```suggestion
const tab = computed(() => tabsContext?.headless.elements.tab.value({ value: props.value }));
const tabpanel = computed(() =>
  tabsContext?.headless.elements.tabpanel.value({ value: props.value }),
);
const isActive = computed(() => tab.value?.["aria-selected"] ?? false);
```

packages/sit-onyx/src/components/OnyxTab/types.ts
```diff
@@ -0,0 +1,9 @@
+import type { DensityProp } from "../../composables/density";
+
+export type OnyxTabProps = DensityProp & {
+  value: PropertyKey;
```
JoCa96, line 4: Please add a jsdoc

packages/sit-onyx/src/components/OnyxTabs/types.ts
```diff
@@ -0,0 +1,36 @@
...
+ * Injection key to provide relevant context data from the tabs parent to the
+ * individual child tab components.
+ */
+export const TABS_INJECTION_KEY = Symbol() as InjectionKey<{
+  /**
+   * Headless composable provided for child tabs for correct accessibility and behavior.
+   */
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any -- needed to fix type error when using generics inside the OnyxTabs
+  headless: ReturnType<typeof createTabs<any>>;
```
JoCa96, line 30:
```suggestion
  headless: ReturnType<typeof createTabs<PropertyKey>>
```

JoCa96 (changes requested):

packages/headless/src/composables/tabs/createTabs.ts
```diff
@@ -25,16 +25,26 @@ export const createTabs = createBuilder(<T extends PropertyKey>(options: CreateT
...
     if (!idMap.has(value)) {
       idMap.set(value, { tabId: useId(), panelId: useId() });
     }
     return idMap.get(value)!;
   };
 
+  const isVisible = computed(() => {
+    return (value: T) => getId(value).tabId === getId(unref(options.selectedTab)).tabId;
+  });
+
```
JoCa96, line 40: Isn't that the same as `tab(value).['aria-selected']`.?

packages/sit-onyx/src/components/OnyxTabs/types.ts
larsrickert: I added any because everything else causes a type issue due to the generics in OnyxTabs.vue. Do you have an idea how to solve this?

JoCa96: Yes, you need to pass through the generic "TValue" type from the OnyxTab component.
To achieve that you can wrap the provided value of `TABS_INJECTION_KEY` into a function, which can inject the generic.

JoCa96 (approved):

PR #2058, opened 2024-11-07
chore: fix dev script of all packages to use turborepo

larsrickert:
Relates to #1951 

Currently our dev scripts (e.g. of `sit-onyx`) do not work in a clean repository. The issue is that `sit-onyx` depends on other packages (like icons) to be build first.

I aligned all `dev` scripts for all packages to be run with turborepo so all dependencies are managed/build correctly.

### DIFF
--- a/apps/demo-app/package.json
+++ b/apps/demo-app/package.json
@@ -4,7 +4,7 @@
   "version": "1.0.0-beta.63",
   "type": "module",
   "scripts": {
-    "dev": "turbo run turbo:dev --filter demo-app",
+    "dev": "turbo run turbo:dev --filter=.",
     "turbo:dev": "vite",
     "build": "vue-tsc && vite build",
     "preview": "vite preview"
--- a/apps/demo-app/package.json
+++ b/apps/demo-app/package.json
@@ -4,8 +4,8 @@
   "version": "1.0.0-beta.63",
   "type": "module",
   "scripts": {
-    "dev": "turbo run vite --filter demo-app",
-    "vite": "vite",
+    "dev": "turbo run turbo:dev --filter demo-app",
+    "turbo:dev": "vite",
     "build": "vue-tsc && vite build",
     "preview": "vite preview"
   },
--- a/apps/docs/package.json
+++ b/apps/docs/package.json
@@ -4,7 +4,7 @@
   "type": "module",
   "private": true,
   "scripts": {
-    "dev": "turbo run turbo:dev --filter docs",
+    "dev": "turbo run turbo:dev --filter=.",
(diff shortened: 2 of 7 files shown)


JoCa96:
What about putting a command in the workspace root instead of having to maintain it in all `package.json` files:
`turbo watch xx:xx --filter=.`
This would then be called using `pnpm -w dev` in the respective workspace.

apps/demo-app/package.json
```diff
@@ -4,8 +4,8 @@
   "version": "1.0.0-beta.63",
   "type": "module",
   "scripts": {
-    "dev": "turbo run vite --filter demo-app",
-    "vite": "vite",
+    "dev": "turbo run turbo:dev --filter demo-app",
```
JoCa96, line 7: IMHO, we should use the `watch` command, which will rebuild dependencies if necessary:
```suggestion
    "dev": "turbo watch turbo:dev --filter demo-app",
```
I got it to work now, maybe we can sync if the issue only affected me.

apps/demo-app/package.json
```diff
@@ -4,8 +4,8 @@
...
+    "turbo:dev": "vite",
```
JoCa96, line 8: I would prefer a non-turbo specific command name, e.g.:
- "dev:run"
- "build:watch"
- "watch"

apps/demo-app/package.json
larsrickert: Watch does not work as expected for me. E.g. if using `turbo watch` in the demo app and then e.g. make changes to the OnyxButton, turbo will re-build `sit-onyx` like it should but the demo app will be shut down and not work any longer...

larsrickert:
> What about putting a command in the workspace root instead of having to maintain it in all `package.json` files: `turbo watch xx:xx --filter=.` This would then be called using `pnpm -w dev` in the respective workspace.

I like the idea. Did you get it to work? If I do this, turbo tells me that no workspaces/packages have been found and nothing is run

apps/demo-app/package.json
larsrickert: I wanted to indicate that the `turbo:dev` script should not be run manually. It is only needed for turborepo.
Would you still rename it?

JoCa96: I would prefer it, but it's not a requirement for me

apps/demo-app/package.json
JoCa96: -_- `turbo watch` broke again, so I can't even test it

JoCa96 (approved):

JoCa96 (changes requested):
Wait, can't we use `pnpm`? https://pnpm.io/filtering#--filter-package_name-4

JoCa96 (approved):

Issue #2053, opened 2024-11-06
Define OnyxSupportButton


JoCa96:
### Related to

- #1008

### Acceptance criteria

- [ ] A new created refined `Implement OnyxSupportButton` ticket
- [ ] Replaced `OnyxIconButton` in `packages/sit-onyx/src/components/OnyxDataGrid/features/sorting/SortAction.vue`
- [ ] Ticket was created

larsrickert:
See #2139

PR #2074, opened 2024-11-11
chore(OnyxDataGrid): Add tests and demo

JoCa96:
Relates to #1852 

- Added ct and unit tests
- Added page to Demo app
- Extended feature prop jsdoc


apps/demo-app/src/views/DataGridView.vue
```diff
@@ -0,0 +1,39 @@
...
+    <div class="onyx-grid-container">
+      <OnyxHeadline is="h1">Data-Grid example</OnyxHeadline>
+      <article class="data-grid-settings">
+        <OnyxSwitch v-model="sortingEnabled" label="Enable sorting" />
+      </article>
+      <OnyxDataGrid :features :data :columns="['name', 'age']" />
+    </div>
+  </OnyxPageLayout>
+</template>
+<style lang="scss" scoped>
```
larsrickert, line 34:
```suggestion
</script>

<template>
  <OnyxPageLayout>
    <div class="onyx-grid-container">
      <OnyxHeadline is="h1">Data-Grid example</OnyxHeadline>
      <article class="data-grid-settings">
        <OnyxSwitch v-model="sortingEnabled" label="Enable sorting" />
      </article>
      <OnyxDataGrid :features :data :columns="['name', 'age']" />
    </div>
  </OnyxPageLayout>
</template>

<style lang="scss" scoped>
```

I though #2063 added a linting rule for this? :D

apps/demo-app/src/views/DataGridView.vue
```diff
@@ -0,0 +1,39 @@
...
+    enabled.push(DataGridFeatures.useSorting());
+  }
+  return enabled;
+});
+</script>
+<template>
+  <OnyxPageLayout>
+    <div class="onyx-grid-container">
+      <OnyxHeadline is="h1">Data-Grid example</OnyxHeadline>
+      <article class="data-grid-settings">
```
larsrickert, line 27: Is `<article>` the correct element here? Isn't it used for things like block posts? Maybe use `<section>` instead?

packages/sit-onyx/playwright/snapshots/components/OnyxDataGrid/features/sorting/data-grid-sorting-asc-chromium-linux.png
larsrickert, line 1: Shouldn't there be a little gap between the header label and the sort icon? imo it looks weird in hover state like this

packages/sit-onyx/src/components/OnyxDataGrid/features/sorting/SortAction.ct.tsx
```diff
@@ -0,0 +1,12 @@
+import { test } from "../../../../playwright/a11y";
+import { executeMatrixScreenshotTest } from "../../../../playwright/screenshots";
+import SortAction from "./SortAction.vue";
+
+test.describe("Screenshot tests", () => {
+  executeMatrixScreenshotTest({
+    rows: ["none", "asc", "desc"] as const,
```
larsrickert, line 7:
```suggestion
    rows: ["none", "asc", "desc"],
```

packages/sit-onyx/src/components/OnyxDataGrid/features/sorting/SortAction.ct.tsx
```diff
@@ -0,0 +1,12 @@
...
+    columns: ["default"],
```
larsrickert, line 8: Would prefer to also include hover, active and focus-visible here

packages/sit-onyx/src/components/OnyxDataGrid/features/sorting/sorting.ct.tsx
```diff
@@ -0,0 +1,73 @@
+import type { Locator } from "@playwright/test";
+import { expect, test } from "../../../../playwright/a11y";
+import SortingDataGrid from "../../../examples/DataGrid/SortingDataGrid.vue";
+
+const getTestArray = () => [
```
larsrickert, line 5:
```suggestion
const getTestData = () => [
```

packages/sit-onyx/src/components/OnyxDataGrid/features/sorting/sorting.ct.tsx
```diff
@@ -0,0 +1,73 @@
...
+  expect(rows).toHaveLength(data.length);
+
+  await expectOrderedText(
+    rows,
+    data.map((item) => item.a),
+  );
+  await expect(component).toHaveScreenshot("data-grid-sorting-initial.png");
+
+  // ACT
+  await component.getByLabel("Sort the table ascending by the a column.").click();
```
larsrickert, line 43:
```suggestion
  await component.getByLabel("Sort the data grid ascending by the a column.").click();
```

Accessibility question: For the end user / screen reader, should we call it data grid or table?

packages/sit-onyx/src/components/OnyxDataGrid/features/sorting/sorting.ct.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+  // ASSERT
+  rows = await getFirstColumn();
```
larsrickert, line 46: You don't have to re-assign this every time because the locator stays the same, only the resolved content changes but the content is evaluated lazily when calling `expect`.

So calling this once is enough:
```ts
  const rows = await component.locator("tbody tr td:first-of-type").all();
```

packages/sit-onyx/src/components/OnyxDataGrid/features/sorting/sorting.ct.tsx
larsrickert: Same for the other files :) reduced cognitive overhead since the return value is the `data` property of the data grid :)

packages/sit-onyx/src/components/OnyxDataGrid/features/sorting/sorting.ts
```diff
@@ -53,7 +53,7 @@ export const useDataGridSorting = createFeature(
       return !config || config?.[col]?.enabled === true;
     });
 
-    const locale = injectI18n()?.locale;
+    const locale = computed(() => injectI18n()?.locale.value ?? "en-us");
```
larsrickert, line 56: The default / fallback value is handeled by i18n so the locale is always defined:
image

You can just use it with:
```ts
const { locale } = injectI18n();
```

packages/sit-onyx/src/components/OnyxDataGrid/features/sorting/types.ts
```diff
@@ -31,19 +31,21 @@ export type SortState<TEntry extends DataGridEntry> = {
...
+         */
+        enabled: boolean;
+        /**
+         * A custom sorting function for this column.
+         * By default the `Intl.Collator` with the current locale is used.
+         */
+        sortFunc?: Compare<TEntry[TKey]>;
+      };
+    }
+  | undefined;
```
larsrickert, line 48: Why did you add undefined here? Would prefer to specify it where its used, e.g.

`{
myProp: SortColumnOptions | undefined
}`

packages/sit-onyx/src/components/OnyxDataGrid/types.ts
```diff
@@ -11,6 +11,24 @@ export type OnyxDataGridProps<
...
+   * <script setup lang="ts">
+   * import { ref, watch } from "vue";
+   * import type { DataGridEntry, OnyxDataGridProps } from "../../..";
+   * import { DataGridFeatures, OnyxDataGrid } from "../../..";
+   *
+   * const withSorting = DataGridFeatures.useSorting(sortOptions);
+   * </script>
+   *
+   * <template>
+   *   <OnyxDataGrid :columns :data :features=[withSorting] />
```
larsrickert, line 28:
```suggestion
   * import { DataGridFeatures, OnyxDataGrid, type DataGridEntry, type OnyxDataGridProps } from "sit-onyx";
   *
   * const withSorting = DataGridFeatures.useSorting(sortOptions);
   * </script>
   *
   * <template>
   *   <OnyxDataGrid :columns :data :features="[withSorting]" />
```

The user needs to import from `sit-onyx`, not from relative paths :)

larsrickert (changes requested):

larsrickert:
Do you want to add a changeset to release the JSDocs changes?

packages/sit-onyx/src/components/OnyxDataGrid/features/sorting/sorting.ct.tsx
JoCa96: The screen reader will announce it as "table", so we should use the term for the enduser.

packages/sit-onyx/playwright/snapshots/components/OnyxDataGrid/features/sorting/data-grid-sorting-asc-chromium-linux.png
JoCa96: The current sort button is a placeholder, until we have the SupportButton. See #2053 or the comment in the `SortAction.vue`

packages/sit-onyx/src/components/OnyxDataGrid/features/sorting/SortAction.ct.tsx
JoCa96: Seems redundant, as this is just the same as the `OnyxIconButton`

larsrickert (approved):

Issue #2079, opened 2024-11-13
Implement OnyxTabs font styles and overflow


larsrickert:
### Depends on

- #2018

### Why?

We want to offer multiple font styles for the `OnyxTabs` component and also make them horizontally scrollable if they exceed the max. available width (e.g. if many tabs are used).

### Design

[Figma](https://www.figma.com/design/YfEUBOHk4J4nYrk04geswG/Onyx-Component-Library?node-id=7162-15700&t=w1vkWbCstQZGcKGn-4)

### Acceptance criteria

- [x] Tabs should be horizontally scrollable if they exceed the max available width.
- [x] `h3` and `h4` font styles are implemented
- [x] The Storybook examples are updated to be more realistic
- [x] the bug is fixed that the tab content is not vertically centered if the tab is not selected
![image](https://github.com/user-attachments/assets/93777c6a-54c2-4788-bed9-a807f43dd10b)

#### Applicable ARIA Pattern

[Tabs pattern](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/)

### Definition of Done

- covered by tests
  - [x] visual tests (Playwright screenshots)
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature
- [x] Storybook code snippet of new/changed examples are checked that they are generated correctly
- [x] Namings are aligned with Figma

**Approval**

- [Storybook](https://storybook.onyx.schwarz/?path=/story/navigation-tabs--many-tabs)
- [x] approved by designer @jannick-ux
- [x] approved by dev @MajaZarkova

jannick-ux:
scroll behavior and font size is fine. Thanks.
Just one finding here regarding the colors: The deselected tabs should have the "color/text+icons/neutral/medium" variable, but it seems like the have the "intense" one. The intense-value is exclusively for the selected tab.

larsrickert:
> scroll behavior and font size is fine. Thanks. Just one finding here regarding the colors: The deselected tabs should have the "color/text+icons/neutral/medium" variable, but it seems like the have the "intense" one. The intense-value is exclusively for the selected tab.

The deselected already have the medium color, only the selected one has the intense one, see screenshot below. The difference is better visible in dark mode since medium and intense seem to be pretty similar in light mode

![Image](https://github.com/user-attachments/assets/e323d374-f7bb-4cc3-a327-fb19b5e76d1e)
[...]

PR #2089, opened 2024-11-13
feat(OnyxTabs): support horizontal scrolling when overflowing

larsrickert:
Relates to #2079

- support horizontal scrolling when not all tabs fit into the available width
- add property `size` to change the font style
- add CSS variable `--onyx-outline-width` and use it in all components for consistency
- make Storybook examples more realistic
- fix vertical alignment of tab text if its not selected

## Checklist


packages/sit-onyx/src/components/OnyxTab/OnyxTab.vue
MajaZarkova, line 1: I couldn't find the density styles in Figma. I noticed that the underline is different for the different densities, is that intentional?

larsrickert: You can open the component in the playground and then select the density there
image

MajaZarkova (approved):

packages/sit-onyx/src/components/OnyxTab/OnyxTab.vue
```diff
@@ -25,6 +25,7 @@ defineSlots<{
 const { densityClass } = useDensity(props);
 const tabsContext = inject(TABS_INJECTION_KEY);
 const skeleton = useSkeletonContext(props);
+const sizeClass = computed(() => `onyx-tab--${tabsContext?.size.value}`);
```
JoCa96, line 28: Wouldn't it be enough to set the size class on `.onyx-tabs__tablist` element?

larsrickert: In theory yes but I wanted to separate this since the font styles are responsibility of the tab and not of the parent (in my opinion) :)

Issue #1997, opened 2024-10-25
Implement Playwright testing for all browser that we support in browserlistrc


larsrickert:
https://github.com/SchwarzIT/onyx/pull/1956#discussion_r1808187248

oemueller:
Playwright does only support certain browsers, also each version is bound to the Playwright version. (Samsung internet is not supported at all)
Therefore it's hard to use the browserslist to automatically configure Playwright without a manual step. 
I guess we don't want the browserslist to be a dependency for the Playwright version we use. 

For now I would configure the Playwright "projects" / Browsers manually. 

And we should consider testing the Edge browser additionally or replace chromium with it. (Since it's the main browser to be supported)

`{
  name: 'Microsoft Edge',
  use: { ...devices['Desktop Edge'], channel: 'msedge' }, // or 'msedge-dev'
}`

JoCa96:
Sounds good to me :)

JoCa96:
With #4043 we updated the browserlist to match our actual target browsers, which we are already testing (except mobile versions)

PR #2130, opened 2024-11-20
chore: use Edge browser for Onyx Playwright tests

oemueller:
Relates to #1997

Use MS Edge in Playwright tests for Onyx, since this is the most important browser.
(Also Chromium based but the version is a few weeks older and has possible other minor differences) 
https://playwright.dev/docs/browsers#configure-browsers


packages/sit-onyx/playwright.config.ts
```diff
@@ -41,7 +41,7 @@ export default defineConfig({
   },
   /* Configure projects for major browsers */
   projects: [
-    { name: "chromium", use: { ...devices["Desktop Chrome"] } },
+    { name: "Microsoft Edge", use: { ...devices["Desktop Edge"], channel: "msedge" } },
```
larsrickert, line 44:
```suggestion
    { name: "edge", use: { ...devices["Desktop Edge"], channel: "msedge" } },
```

For the other browsers is also just the browser name and lowercase :) This name will be added to the screenshot names so I would keep it short here

larsrickert (changes requested):

.changeset/empty-drinks-sniff.md
larsrickert, line 1: We can omit the changeset since this is an internal only change that does not require a new npm package version to be released to our users :)

larsrickert (dismissed):

larsrickert:
> Shall we use Edge in other packages as well ?

@oemueller Sure why not so its the same for all packages :)

oemueller:
@JoCa96 can you approve ? all issues resolved :-)

apps/docs/playwright/snapshots/about-team-chromium-linux.png
JoCa96, line 1: Where are the new screenshots that replace these?

oemueller: there was a problem with the test file pattern in docs app so no tests where found. I changed it to: ` `**/*.ct.{ts,tsx}``

JoCa96 (approved):

JoCa96:
Approved :)

Issue #1818, opened 2024-08-30
Implement basic OnyxDatePicker


mj-hof:
### Depends on

- #1817

### Why?

We want to offer a date picker component where users can select a date or datetime, e.g. birthdays or meeting dates.
Because implementing a whole date picker from scratch is time consuming, we want to offer a "quick win" by already
offering a `OnyxDatePicker` component which uses the existing `OnyxInput` and the native HTML `<input type="date" />` for the flyout.
Like this, the component mostly looks and feels like onyx and we can provide the component quicker.

Later, we can refactor it and build our own flyout with more advanced features.

### Design

[Figma](https://www.figma.com/design/YfEUBOHk4J4nYrk04geswG/Onyx-Component-Library?node-id=7420-36065&t=pvuWNaMRTggogXnV-4)

### Acceptance criteria

- a date can be selected
  - [x] manually via keyboard
  - [x] via flyout
- a datetime can be selected
  - [x] manually via keyboard
  - [x] via flyout
- [x] the component considers the current i18n locale and the users browser preference for i18n
- [x] all features of the `OnyxInput` are supported, except properties that do not make sense for dates (e.g. type email)
- [x] if possible, the native HTML flyout aligns with the light/dark mode
- [x] a min and max date can be set
- [x] should work in all timezones

### Implementation details

- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#input_types
- values should be ISO timestamps
- for formatting, we should use the `Intl` API

```ts
export type OnyxDatePickerProps = OnyxInputProps & {
  type?: "date" | "datetime";
};
```

### Definition of Done

- The following component requirements are implemented:
  - [x] skeleton
  - [x] density
  - [x] example usage added to `apps/demo-app/src/views/HomeView.vue`
- covered by tests
  - [x] functional tests (Playwright or unit test)
  - [x] visual tests (Playwright screenshots)
- [x] follow-up tickets were created if necessary
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature
- [x] Storybook code snippet of new/changed examples are checked that they are generated correctly
- [x] Namings are aligned with Figma

**Approval**

- [Storybook](https://storybook.onyx.schwarz/?path=/docs/form-elements-datepicker--docs)
- [x] approved by designer @jannick-ux
- [x] approved by dev @MajaZarkova

larsrickert:
Idea for design: Show current timezone / be able to change timezone

larsrickert:
[...]

PR #2145, opened 2024-11-25
feat: implement basic `OnyxDatePicker` component

larsrickert:
Relates to #1818

Implement a basic date picker comonent.

## Checklist


packages/sit-onyx/src/components/OnyxDatePicker/OnyxDatePicker.vue
```diff
@@ -0,0 +1,150 @@
...
+          <OnyxLoadingIndicator
+            v-if="props.loading"
+            class="onyx-datepicker__loading"
+            type="circle"
+          />
+          <!-- key is needed to keep current value when switching between date and datetime type -->
+          <input
+            :id="inputId"
+            :key="props.type"
+            v-model="value"
```
JoCa96, line 102: Is there a particular reason why we don't use [`valueAsDate`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/valueAsDate)? Then we don't need to do the conversion magic ourselves 😅

larsrickert: Because valueAsDate does not work for `type="datetime-local"`:
image

larsrickert: So we can not use it for setting the value, but I updated the code to call `valueAsDate` for the `@input` event

PR #2196, opened 2024-12-02
feat(OnyxDatePicker): implement `min` and `max` property

larsrickert:
Relates to #1818

Also fix the bug where the value is always `undefined`, when using `type="datetime-local"` by removing the usage of `valueAsDate` which does not work with date+time.

## Checklist

### DIFF
--- a/packages/sit-onyx/src/composables/useCustomValidity.ts
+++ b/packages/sit-onyx/src/composables/useCustomValidity.ts
@@ -116,7 +116,7 @@ export const getFormMessageText = (customError?: CustomMessageType): string | un
  * ```
  */
 export const useCustomValidity = (options: UseCustomValidityOptions) => {
-  const { t } = injectI18n();
+  const { t, locale } = injectI18n();
 
   const validityState = ref<Record<keyof ValidityState, boolean>>();
   const isDirty = ref(false);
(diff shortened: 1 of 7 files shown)


JoCa96 (approved):

packages/sit-onyx/src/composables/useCustomValidity.ts
```diff
@@ -224,3 +225,16 @@ export const useCustomValidity = (options: UseCustomValidityOptions) => {
...
+  type: UseCustomValidityOptions["props"]["type"],
+  value?: DateValue,
+): string | undefined => {
+  if (!type || !["date", "datetime-local"].includes(type)) return value?.toString();
+
+  const date = value != undefined ? new Date(value) : undefined;
+  if (!isValidDate(date)) return value?.toString();
+
+  if (type === "date") return date.toLocaleDateString();
+  else if (type === "datetime-local") return date.toLocaleString();
```
JoCa96, line 239: Are we able to pass the local to `toLocaleDateString` and `toLocaleString`?

larsrickert: Done

JoCa96 (approved):

PR #2204, opened 2024-12-02
feat(OnyxSelect): update the hover and focus colors of the chevron icon

larsrickert:
Relates to #1818

The colors of the select chevron icon were updated in Figma. I updated them in this PR and also moved the button styles to the shared scss file so we can easily add such buttons other form elements (date picker, inputs etc.) in the future

## Checklist


JoCa96 (changes requested):
To be honest, there is a lot of complexity and edge-cases added just to show a custom icon. Can we skip this until we do the full-scale datepicker implementation? Then the implementation will change anyway.

larsrickert:
> To be honest, there is a lot of complexity and edge-cases added just to show a custom icon. Can we skip this until we do the full-scale datepicker implementation? Then the implementation will change anyway.

Where do you seethe complexity in this PR? Imo, the main changes are the screenshots which are only changed because the select chevron color changed in Figma which we need to implement anyways, independent from the date picker.

The date picker just uses the same icon styles as the select

JoCa96 (approved):

PR #2213, opened 2024-12-03
docs: correctly skip remote fetches in CI

larsrickert:
The `VITEPRESS_SKIP_REMOTE_FETCH` env variable for our docs was implemented incorrectly. When set, not all remote fetches where skipped so some CI pipelines still failed if triggered too often.


JoCa96:
Can we try https://github.com/SchwarzIT/onyx/pull/2207 first? :)

PR #2256, opened 2024-12-05
fix(sit-onyx): Fix missing "locale" JSONs in dist which are referenced by some type definitions

JoCa96:
- "locale" JSONs were missing in the dist, but referenced but some type definitions. These types then didn't work correctly.
  Fixed by adding a dedicated tsconfig, which takes care of "transpiling" the JSON files.
- fix "prop is an recognized symbol" error message occurring in cases where it's actually fine
- updated some validation messages to be clearer and follow the recommendations of https://webaim.org/


.changeset/warm-seals-approve.md
```diff
@@ -0,0 +1,5 @@
+---
+"sit-onyx": patch
+---
+
+Fix missing "locale" json, which are referenced by some type definitions
```
larsrickert, line 5:
```suggestion
Fix missing "locale" json, which are referenced by some type definitions.
Also update default error messages to be more descriptive
```

larsrickert (approved):

PR #2269, opened 2024-12-06
feat: implement `@sit-onyx/playwright-utils` package

larsrickert:
Relates to #1991

Implement `useMatrixScreenshotTest` and `adjustSizeToAbsolutePosition` playwright utilities.

## Checklist


packages/playwright-utils/src/screenshots/utils.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * Escapes the given string so its save to be used as CSS grid-area, e.g. by replacing whitespaces.
+ */
+export const escapeGridAreaName = (name: string) => {
+  return name.replace(/ /g, "-");
```
JoCa96, line 5: To be a bit pedantic here 😅: 
We should probably replace all unusual characters:
```suggestion
  return name.replace(/\W/g, "-");
```

packages/playwright-utils/src/screenshots/types.ts
```diff
@@ -0,0 +1,84 @@
...
+   * **IMPORTANT**: Should be avoided! If used, please include a comment why it is needed.
+   *
+   * @see https://playwright.dev/docs/accessibility-testing#disabling-individual-scan-rules
+   */
+  disabledAccessibilityRules?: string[];
+  /**
+   * If `true`, no padding will be added to the screenshots.
+   * By default a `1rem` padding is added around the component/screenshot.
+   */
+  disablePadding?: boolean;
```
JoCa96, line 74:
```suggestion
  removePadding?: boolean;
```

packages/playwright-utils/src/screenshots/types.ts
```diff
@@ -0,0 +1,84 @@
...
+  /**
+   * Function that returns the component for the given column and row.
+   */
+  component: (column: TColumn, row: TRow) => JSX.Element;
+  /**
+   * Optional callback to be executed before capturing the screenshot.
+   * Useful for performing `expect()` or e.g. hover, focus-visible state etc.
+   * Focus and mouse will be reset after each screenshot.
+   */
+  beforeScreenshot?: ScreenshotTestHook<TColumn, TRow>;
```
JoCa96, line 56: Would prefer to rename or clarify in the description that this is called before an **individual/single** screenshot is taken
```suggestion
  beforeScreenshot?: ScreenshotTestHook<TColumn, TRow>;
```

packages/playwright-utils/src/screenshots/index.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+        if (disabledAccessibilityRules.length) {
+          axeBuilder.disableRules(disabledAccessibilityRules);
+        }
+
+        const accessibilityScanResults = await axeBuilder.analyze();
+        expect(
+          accessibilityScanResults.violations,
+          `should pass accessibility checks for ${column} ${row}`,
+        ).toEqual([]);
```
JoCa96, line 57: Imho we should remove axe-core as a default dependency/include as it addresses a different concern and could be called as well in the `afterScreenshot` hook.

packages/playwright-utils/src/screenshots/index.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+        const id = `${row}-${column}`;
+
+        const image = (
+          <img
+            width={box?.width}
+            height={box?.height}
+            style={{ gridArea: escapeGridAreaName(id) }}
+            src={`data:image/png;base64,${Buffer.from(screenshot).toString("base64")}`}
```
JoCa96, line 66: Please avoid base64 conversion as it unnecessarily increases memory and process usages. I thought we already, addressed this? 🤔

JoCa96 (changes requested):
Please move axe-core checks out

packages/playwright-utils/src/screenshots/index.tsx
larsrickert: I refactored the utility with a new `hooks` pattern, what do you think?

JoCa96: Looks great! Amazing work :)

packages/playwright-utils/src/screenshots/index.tsx
larsrickert: Done. However, after local testing it seems to me that base64 conversion is actually faster than mocking the Playwright routes and doing actual requests.

I will refine a follow up ticket for this to check it: #2291

JoCa96 (approved):

PR #2276, opened 2024-12-09
feat: implement normalization using base class

ChristianBusshoff:
- add onyx-component as base-class to every component
- put normalize.apply in onyx-componet class
- add linting rool


packages/eslint-plugin/src/rules/require-root-class.cjs
```diff
@@ -0,0 +1,176 @@
...
+              }
+
+              return false;
+            });
+
+            if (!has) {
+              context.report({
+                node: tag,
+                loc: tag.loc,
+                message: "missing class",
```
larsrickert, line 152:
```suggestion
                message: "missing class",
```

Would be nice to include which class si missing

larsrickert (changes requested):
Nice :D The file size of our bundled `style.css` goes down from 335kB to 152kB 💪 

Could you also please update our contributing guide template for a new component to include the new class?

packages/eslint-plugin/src/rules/require-root-class.cjs
larsrickert, line 1: Puh thats a complex eslint rule 😅 
Is it possible to add tests for this?

larsrickert (changes requested):

packages/eslint-plugin/src/rules/require-root-class.cjs
ChristianBusshoff: I talked to Jonathan about testing on Friday and since there are no tests for linting yet (eslint-rule-tester is missing), I should create a follow-up ticket for testing instead of implementing it in this one.

packages/eslint-plugin/src/rules/require-root-class.cjs
ChristianBusshoff: changed it

larsrickert (approved):

packages/eslint-plugin/src/rules/require-root-class.cjs
```diff
@@ -0,0 +1,181 @@
...
+      description: "disallow adding root nodes to the template",
+    },
+    fixable: null,
+    schema: [],
+    messages: {
+      multipleRoot: "The template root requires exactly one element.",
+      textRoot: "The template root requires an element rather than texts.",
+      disallowedDirective: "The template root disallows 'v-for' directives.",
+      missingClass: "The root element is missing the 'onyx-component' class.",
+    },
```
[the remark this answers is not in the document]
ChristianBusshoff: which message should I change and how?

Issue #2139, opened 2024-11-25
Implement OnyxSystemButton component


larsrickert:
### Why?

We want to offer a system button component that is similar to the OnyxIconButton but which a different / more compact design.
The system button should be used in parent contexts like sorting in the table / data grid, accordion chevron or "three dot more flyout".

### Design

[Figma](https://www.figma.com/design/YfEUBOHk4J4nYrk04geswG/Onyx-Component-Library?node-id=7860-25394&t=OSbOnKQwlsvbe49h-4)

### Acceptance criteria

- [x] an icon can be shown
- [x] optionally a text can be shown
- the following colors are implemented:
    - [x] intense (default)
    - [x] medium
    - [x] soft
- [x] the button can be disabled
- [x] an example is created where the system button opens a flyout menu
- [x] an example is created where the system button shows a tooltip on hoverd
- the new system button is replaced/used in the following existing components:
    - [x] sort icon in data grid
    - ~~accordion (if implemented at this point)~~
- [x] When hovering the button after a few seconds (should be the same as normal tooltip) the label is show in a tooltip

### Implementation details

- no ripple active animation is needed
- aria pattern is not needed, we use the native button element

### Definition of Done

- The following component requirements are implemented:
  - [x] skeleton
- covered by tests
  - [x] visual tests (Playwright screenshots)
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature
- [x] Storybook code snippet of new/changed examples are checked that they are generated correctly
- [x] Namings are aligned with Figma

**Approval**

- [Storybook](https://storybook.onyx.schwarz/?path=/docs/buttons-systembutton--docs)
- [x] approved by designer @jannick-ux
- [x] approved by dev @MajaZarkova

PR #2284, opened 2024-12-09
feat(OnyxSystemButton): implement soft and medium color

larsrickert:
Relates to #2139

Implement soft and medium color to system button.

## Checklist


packages/sit-onyx/playwright/snapshots/components/OnyxSystemButton/System-button-icon-intense--edge-linux.png
MajaZarkova, line 1: I don't see a difference between the different densities. Maybe we don't need 3 columns for the icon

MajaZarkova (approved):

packages/sit-onyx/playwright/snapshots/components/OnyxSystemButton/System-button-icon-intense--edge-linux.png
larsrickert: Good catch! I talked with Jannick and this was a bug in Figma. The system button does not support density. I removed and Jannick changed it in Figma too

MajaZarkova (approved):

Issue #812, opened 2024-04-04
Implement normalization using base class


JoCa96:
### Depends on

- [x] #338 

### Why?

The `normalize.scss` is right now duplicated for every component.
This adds a lot of duplicated CSS rules into our global CSS.
We can optimize this by instead requiring a base class like `onyx-component` to be set on every component root.
Then the `onyx-component` will apply the normalization CSS.

### Acceptance criteria

- [x] Add global class `onyx-component` which applies the `normalize.scss`
- [x] Updated all components to use the new base class
- [x] Created an ESLint rule to enforce that the class is set on root

### Definition of Done

- [x] follow-up tickets were created if necessary
- [x] updated version + documentation is deployed
- [x] Set task to `In Approval`

**Approval**

- _add link to deployed docs_
- [x] approved by dev @larsrickert

### Implementation details
ESLint POC: 

```JS
/**
 * @fileoverview disallow adding multiple root nodes to the template
 * @author Przemyslaw Falowski (@przemkow)
 */
"use strict";

const { inspect } = require("util");

/**
 * Check whether the given start tag has specific directive.
 * @param {VElement} node The start tag node to check.
 * @param {string} name The directive name to check.
 * @param {string} [argument] The directive argument to check.
 * @returns {boolean} `true` if the start tag has the directive.
 */
function hasDirective(node, name, argument) {
  return Boolean(getDirective(node, name, argument));
}

/**
 * Get the directive which has the given name.
 * @param {VElement} node The start tag node to check.
 * @param {string} name The directive name to check.
 * @param {string} [argument] The directive argument to check.
 * @returns {VDirective | null} The found directive.
 */
function getDirective(node, name, argument) {
  return (
    node.startTag.attributes.find(
      /**
       * @param {VAttribute | VDirective} node
       * @returns {node is VDirective}
       */
      (node) =>
        node.directive &&
        node.key.name.name === name &&
        (argument === undefined ||
          (node.key.argument &&
            node.key.argument.type === "VIdentifier" &&
            node.key.argument.name) === argument),
    ) || null
  );
}

module.exports = {
  meta: {
    type: "problem",
    docs: {
      description: "disallow adding root nodes to the template",
    },
    fixable: null,
    schema: [],
    // TODO: Update messages
    messages: {
      multipleRoot: "The template root requires exactly one element.",
      textRoot: "The template root requires an element rather than texts.",
      disallowedElement: "The template root disallows '<{{name}}>' elements.",
[...]

PR #2310, opened 2024-12-11
docs: fix CSS layers order in deployed Storybook

larsrickert:
Relates to #812 

Currently many components styles are broken in our prod Storybook. Its propably caused by the normalization change but it works when running the Storybook locally and also for the Playwright screenshots.
 
Do you have any idea whats going on here? I remember that we had some weird issues with the CSS layers not being in the correct order in the deployed Storybook

Can be reproduced locally by running "pnpm build:storybook && pnpm preview"

### DIFF
--- a/packages/sit-onyx/.storybook/main.ts
+++ b/packages/sit-onyx/.storybook/main.ts
@@ -1,5 +1,12 @@
 import type { StorybookConfig } from "@storybook/vue3-vite";
+import { readFileSync } from "fs";
+import { fileURLToPath } from "node:url";
 import { mergeConfig } from "vite";
+
+const onyxLayers = readFileSync(
+  fileURLToPath(new URL("../src/styles/layers.scss", import.meta.url)),
+  "utf-8",
+);
 
 const config: StorybookConfig = {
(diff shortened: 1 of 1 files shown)


JoCa96:
Please add the TODO [#2311](https://github.com/SchwarzIT/onyx/issues/2311)

JoCa96 (approved):

packages/sit-onyx/.storybook/main.ts
```diff
@@ -27,6 +34,12 @@ const config: StorybookConfig = {
     <meta property="og:image:height" content="600" />
     <meta property="og:url" content="https://storybook.onyx.schwarz">
   `,
+  // somehow Storybook does not parse the CSS in the correct order so the CSS layers are messed up.
+  // to fix this, we add them manually before all other styles here
```
JoCa96, line 38:
```suggestion
  // somehow Storybook does not parse the CSS in the correct order so the CSS layers are messed up.
  // to fix this, we add them manually before all other styles here
  // TODO: Remove after [#2311](https://github.com/SchwarzIT/onyx/issues/2311)
```

PR #2315, opened 2024-12-11
fix(CI): fix CI caching, deploy and enable turbo remote cache

### LINKED ISSUE #2321: Fix CI caching and enable Turbo Remote Caching

JoCa96:
Closes #2321 
- Add deployment pipeline and CF manifest for [turborepo-remote-cache](https://ducktors.github.io/turborepo-remote-cache/)
- Remove GitHub pipeline caching of `.turbo` directories, only vitepress fetchs are cached now
- Configure all CI pipelines to make use of the remote turbo cache
- Add `varArgs` input to our GitHub `cf-push` custom action, which allows passing custom vars to be used in the manifest

### DIFF
--- a/apps/turbo-remote-cache/.cloud-foundry/manifest-prod.yml
+++ b/apps/turbo-remote-cache/.cloud-foundry/manifest-prod.yml
@@ -1,7 +1,7 @@
 applications:
   - name: turbo-remote-cache
     instances: 2
-    memory: 128MB
+    memory: 256MB
     disk_quota: 1GB
     docker:
       image: ducktors/turborepo-remote-cache
--- a/apps/turbo-remote-cache/.cloud-foundry/manifest-prod.yml
+++ b/apps/turbo-remote-cache/.cloud-foundry/manifest-prod.yml
@@ -1,7 +1,7 @@
 applications:
   - name: turbo-remote-cache
     instances: 2
-    memory: 265MB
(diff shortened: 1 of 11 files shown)


apps/turbo-remote-cache/.cloud-foundry/manifest-prod.yml
```diff
@@ -0,0 +1,14 @@
+applications:
+  - name: turbo-remote-cache
+    instances: 2
+    memory: 256MB
```
larsrickert, line 4: Are this the recommended settings for the docker image? Because Cloud Foundry charges per memory/RAM increasing it unnecessary might blow up the costs :)

larsrickert (approved):

apps/turbo-remote-cache/.cloud-foundry/manifest-prod.yml
JoCa96: Unfortunately, there are no recommend settings provided.
I played around with different memory settings, 128 MB also seemed to work fine.
Cloud Foundry then showed a max RAM usage of around 40-50%. 
So I would like to stay with 256 MB as we will have also way more workload after the merge.
Though, we can check the Cloud Foundry metrics after a week of real usage and still decide to downscale.

Anyway, the budget for the current setting was approved by @mj-hof.

Issue #1939, opened 2024-10-09
Implement OnyxDataGrid column action


JoCa96:
### Depends on
- #1852 

### Why?

This feature is needed for our collaboration project **PSF**.

In case of advanced filtering/sorting or multiple column actions being required, we need to show the actions in a special flyout.

This story is implemented on the basis of the advanced sorting mode.

### Design

[Figma > Data > Advanced Sorting](https://www.figma.com/design/YfEUBOHk4J4nYrk04geswG/Onyx-Component-Library?node-id=4435-44806&node-type=text&m=dev)

### Acceptance criteria
- [x] `TableFeature` type is extended to be able to add column actions
- [ ] the actions can be configured for an inline and a list view
- [x] The `OnyxDataGrid` shows the column actions in a flyout when
  - [x] there is more than one action defined
  - [x] an action only has a `listComponent` defined
- [x] The existing sorting feature is extended with a list component as shown in [Figma](https://www.figma.com/design/YfEUBOHk4J4nYrk04geswG/Onyx-Component-Library?node-id=4435-44806&node-type=text&m=dev)

### Implementation details

The basic `useTableFeature` composable API (defined in #1852) is extended like this:

```ts
export type TableFeature< /* ... */> = {
  // same as before ...

  /**
   * Allows the modification of the header columns before render.
   */
  header?: {
    /** actions are shown after the header label, later `listComponent` can be added */
    actions?: {
      label: string;
      /** optional icon component to be shown in the column header, if not defined list mode will be enforced */
      iconComponent?: Component;
      /** components to render in the list mode of the column actions, if multiple are defined, they will be grouped together */
      listComponents: Component | Component[];
      onTrigger: (event: MouseEvent) => void;
    }[];
  };
};
```

#### Applicable ARIA Pattern

- [MenuButton](https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/)
- [SortableTable](https://www.w3.org/WAI/ARIA/apg/patterns/table/examples/sortable-table/)

### Definition of Done

- covered by tests
  - [ ] functional tests (Playwright or unit test)
  - [ ] visual tests (Playwright screenshots)
- [x] follow-up tickets were created if necessary -> #2293 
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature
- [x] Storybook code snippet of new/changed examples are checked that they are generated correctly
- [x] Namings are aligned with Figma

**Approval**

[...]

PR #2322, opened 2024-12-12
feat(OnyxDataGrid): Render icons in flyout

MajaZarkova:
Relates to #1939 

When listItems are not set the icons will be rendered inside the flyout

## Checklist


JoCa96 (changes requested):
Using the iconComponents (or rather the `OnyxSystemButton`s) inside the list creates a whole suite of issues. 

Screenshot 2024-12-12 at 15 22 51

Either we
- add a new support component which renders to a `OnyxSystemButton` or listitem based on the context
- or we just make listItems a required property

packages/sit-onyx/src/components/OnyxDataGrid/OnyxDataGrid.stories.ts
```diff
@@ -36,27 +39,47 @@ export const HeaderInteractions = {
...
-            },
-            {
-              iconComponent: h(OnyxSystemButton, {
-                label: "Column options",
-                icon: sort,
-                color: "medium",
-              }),
-              listItems: [h(OnyxListItem, () => "Remove column")],
-            },
-          ],
```
JoCa96, line 59: Please keep the more generic approach with `iconComponent` and `listItems` render functions. We will need to support more complex use-cases like the search input.

JoCa96 (changes requested):

JoCa96 (approved):

PR #2330, opened 2024-12-13
fix(docs): fix css ordering in storybook

### LINKED ISSUE #2311: Define fix for layers not working correctly in Storybook
### Why?

https://github.com/SchwarzIT/onyx/pull/2310 is a temp fix, but we need to figure out whats going on:

> 
> Broken components in Storybook
> dev Currently many components styles are broken in our prod Storybook. Its propably caused by the normalization change but it works when running the Storybook locally and also for the Playwright screenshots.
>  
> Do you have any idea whats going on here? I remember that we had some weird issues with the CSS layers not being in the correct order in the deployed Storybook
>  
> Correct layers (when running pnpm dev):
>  
>  
> Somehow broken layers when deploying Storybook:
>  
> Can be reproduced locally by running "pnpm build:storybook && pnpm preview"
>  
> That probably means that the order in which the css is parsed, changed
>  
> The css layer order is determined by whatever happens first:
> Either a plain @layer declaration with the order of the declarations 
> or the order in which the layers are used
>  
>  
> When searching in the DOM, it seems like the layers definition is missing completely, in the local Storybook you can find it in the DOM <style> tags
>  

See also: https://github.com/SchwarzIT/onyx/issues/2311#issuecomment-2540820134

JoCa96:
Closes #2311

Vite doesn't consider the order of CSS includes/imports (See https://github.com/vitejs/vite/issues/3924).
This causes issues for us in production, where we rely on the layer definitions to be defined first.

As a fix, we disable CSS code splitting: There isn't really a downside to this, it's a single, slightly bigger CSS file instead of many smaller ones.

### DIFF
--- a/packages/sit-onyx/.storybook/main.ts
+++ b/packages/sit-onyx/.storybook/main.ts
@@ -1,12 +1,5 @@
 import type { StorybookConfig } from "@storybook/vue3-vite";
-import { readFileSync } from "fs";
-import { fileURLToPath } from "node:url";
 import { mergeConfig } from "vite";
-
-const onyxLayers = readFileSync(
-  fileURLToPath(new URL("../src/styles/layers.scss", import.meta.url)),
-  "utf-8",
-);
 
 const config: StorybookConfig = {
   stories: ["./pages/*.mdx", "../src/**/*.stories.ts"],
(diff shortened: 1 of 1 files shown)


larsrickert (changes requested):
This does not fix the issue for me...if you run "pnpm build:storybook && pnpm preview", the CSS layers are still broken

PR #2338, opened 2024-12-16
chore: enable build caching for playwright tests

JoCa96:
Relates to #2321

- Move build step as precondition for all checks and tests in the pipeline to create cache and avoid downstream build failures.
- Extracted shared playwright config into `@sit-onyx/shared` package as `PLAYWRIGHT_BASE_CONFIG` to reduce code duplications.
- Pass playwright shard configuration from `playwright.config.ts` file instead of using CLI arguments. See beneath for more details:

There was an issue where every shard executed a fresh build of all packages.
The cause was the `--shard` cli argument that differs for every shard.
Because [turbo considers](https://turbo.build/repo/docs/crafting-your-repository/caching#task-inputs) pass-through arguments in their hashing, there was always a hash mismatch.
Unfortunately, turbo doesn't support changing this behavior.

Luckily, playwright allows setting the sharding configuration via the `playwright.config.ts` file, where we can use environment variables. And because turbo can be configured to ignore specific environment variables, we can use them instead of CLI arguments.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,9 +5,9 @@
   "author": "Schwarz IT KG",
   "license": "Apache-2.0",
   "scripts": {
-    "build:all": "turbo build build:storybook",
-    "test:all": "turbo test:coverage",
-    "test:components:all": "turbo test:components test:integration --concurrency 1",
+    "build:all": "turbo run build build:storybook --summarize",
+    "test:all": "turbo run test:coverage --summarize",
+    "test:components:all": "turbo run test:components test:integration --summarize --concurrency 1",
     "format:all": "prettier --write .",
     "format:check:all": "prettier --check .",
     "lint:all": "eslint .",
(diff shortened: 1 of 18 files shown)


package.json
```diff
@@ -5,9 +5,9 @@
   "author": "Schwarz IT KG",
   "license": "Apache-2.0",
   "scripts": {
-    "build:all": "turbo build build:storybook",
-    "test:all": "turbo test:coverage",
-    "test:components:all": "turbo test:components test:integration --concurrency 1",
+    "build:all": "turbo run build build:storybook",
+    "test:all": "turbo run test:coverage",
+    "test:playwright:all": "turbo run test:playwright test:playwright --concurrency 1",
```
larsrickert, line 10:
```suggestion
    "test:playwright:all": "turbo run test:playwright --concurrency 1",
```

larsrickert (approved):

Issue #2308, opened 2024-12-11
Implement WaWi input replacement follow-up


JoCa96:
### Why?

We determined which discrepancies between WaWi and onyx exists: 

### Open questions / TODOs

- We will meet with the WaWi UI/UX team on Friday. Then we will present the current PoC state and afterward decide on the specifics.

### Depends on

- #2179 

### Acceptance criteria

The following topics are made possible via onyx changes and discussions

*significant:*

- [x] `noTruncation` property in onyx => todo
- [x] `isMaxLengthDisplayed` not in onyx => discuss
- [x] `rightIcon` not yet available in onyx => todo
- [x] `trailing`/`trailingLabel` not in onyx => todo
- [x] different error behavior => discuss
- [x] no `focus()`, `select()` => todo

*minor:*

- [x] Different `maxlength` visualization => discuss
- [x] Different `readonly` visualization => discuss

### Implementation details

The currrent PoC is pushed to the [jlc/poc-onyx-wux-input](https://dev.azure.com/schwarzit/lidl.wawi-core/_git/lidl-wawi?version=GBjlc%2Fpoc-onyx-wux-input) branch

### Definition of Done

- Should be covered by tests:
  - [ ] functional tests (Playwright or unit test)
  - [x] visual tests (Playwright screenshots)

- Make sure, that
  - [ ] follow-up tickets were created if necessary
  - [ ] updated version + documentation is deployed
  - [ ] Storybook can show the feature
  - [ ] Storybook code snippet of new/changed examples are checked that they are generated correctly
  - [ ] Namings are aligned with Figma

#### Approval

[Storybook]()
[Documentation]()

- Checked and approved by
  - [ ] designer 
  - [ ] dev

PR #2347, opened 2024-12-16
feat(i18n): Custom translation function can now be static

JoCa96:
Relates to #2308

### DIFF
--- a/packages/sit-onyx/src/i18n/index.ts
+++ b/packages/sit-onyx/src/i18n/index.ts
@@ -84,7 +84,7 @@ const createI18n = (options: ProvideI18nOptions = {}) => {
 
   // If the user provided a custom `t` function, it is used instead of the default.
   // Then we also skip the loading and applying of messages.
-  if (options.t) {
+  if (userT) {
     return { t: userT, locale };
   }
 
--- a/packages/sit-onyx/src/i18n/index.ts
+++ b/packages/sit-onyx/src/i18n/index.ts
@@ -1,4 +1,4 @@
-import { computed, inject, toRef, type App, type InjectionKey, type MaybeRef } from "vue";
+import { computed, inject, readonly, toRef, type App, type InjectionKey, type MaybeRef } from "vue";
 import type { FlattenedKeysOf, TranslationValue } from "../types/i18n";
 import type { DeepPartial } from "../types/utils";
 import enUS from "./locales/en-US.json";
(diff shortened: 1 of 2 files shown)


packages/sit-onyx/src/i18n/index.ts
```diff
@@ -87,11 +79,13 @@ const createI18n = (options: ProvideI18nOptions = {}) => {
    * Current locale.
    * @default "en-US"
    */
-  const locale = computed(() => unref(options?.locale) ?? "en-US");
+  const locale = toRef(options?.locale ?? "en-US");
```
larsrickert, line 82:
```suggestion
  const locale = toRef(options?.locale ?? "en-US");
```

Shouldn't we make the locale readonly? So we don't unintentionally change it internally

packages/sit-onyx/src/i18n/index.ts
```diff
@@ -87,11 +79,13 @@ const createI18n = (options: ProvideI18nOptions = {}) => {
...
+  const userT = options.t ? toRef(options.t) : undefined;
 
-  // If the user provided a custom `t` function it should be used instead of the default one
-  if (options.t) {
-    return { t: options.t, locale };
+  // If the user provided a custom `t` function, it is used instead of the default.
+  // Then we also skip the loading and applying of messages.
+  if (userT) {
+    return { t: userT, locale };
```
larsrickert, line 88:
```suggestion
  // If the user provided a custom `t` function, it is used instead of the default.
  // Then we also skip the loading and applying of messages.
  if (options.t) {
    return { t: toRef(options.t), locale };
```

Like this we could get rid of the `userT` variable

larsrickert (changes requested):

packages/sit-onyx/src/i18n/index.ts
```diff
@@ -87,11 +79,12 @@ const createI18n = (options: ProvideI18nOptions = {}) => {
...
    */
-  const locale = computed(() => unref(options?.locale) ?? "en-US");
+  const locale = computed(() => toValue(options?.locale) ?? "en-US");
 
-  // If the user provided a custom `t` function it should be used instead of the default one
+  // If the user provided a custom `t` function, it is used instead of the default.
+  // Then we also skip the loading and applying of messages.
   if (options.t) {
-    return { t: options.t, locale };
+    return { t: computed(() => toValue(options.t!)), locale } as const;
```
larsrickert, line 87: I am not sure if `!` is safe here since the if check only works initially but the computed function might be re-run later due to reactivity

larsrickert (changes requested):

larsrickert (approved):

PR #2401, opened 2024-12-23
feat: implement `OnyxAccordion` component

ChristianBusshoff:
- implement OnyxAccordion & AccordionItem


packages/sit-onyx/src/components/OnyxAccordionItem/OnyxAccordionItem.vue
```diff
@@ -0,0 +1,154 @@
...
+      <slot name="header"></slot>
+      <OnyxIcon :icon="chevronLeftSmall" class="onyx-accordion-item__header__icon" />
+    </summary>
+    <div class="onyx-accordion-item__panel">
+      <slot name="panel"></slot>
+    </div>
+  </details>
+</template>
+
+<style lang="scss">
```
larsrickert, line 77: Could you please check the styles with Figma? There are currently a few differences like text color, font weight etc.

Also, I'd prefer to use a flex layout instead of positioning the chevron icon absolute

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxAccordionItem/types.ts
```diff
@@ -0,0 +1,13 @@
+import type { DensityProp } from "../../composables/density";
+import type { SkeletonInjected } from "../../composables/useSkeletonState";
+
+export type OnyxAccordionItemProps = DensityProp & {
```
larsrickert, line 4: Would be nice if the open state can be controlled as prop so you could e.g. open it programatically or set it open initially on page load.

You don't have to do it in this PR

packages/sit-onyx/src/components/OnyxAccordionItem/OnyxAccordionItem.vue
```diff
@@ -0,0 +1,165 @@
...
+      accordionContext?.disabled.value || props.disabled ? 'onyx-accordion-item--disabled' : '',
+    ]"
+    :open="isOpen"
+    @toggle="isOpen = ($event.target as HTMLDetailsElement).open"
+  >
+    <summary
+      :id="'header-' + itemId"
+      class="onyx-accordion-item__header"
+      role="button"
+      :tabindex="accordionContext?.disabled.value || props.disabled ? -1 : 0"
```
larsrickert, line 58: We have this disabled logic with `||` multiple times in this file, maybe you could create a `isDisabled` computed that unifies this

packages/sit-onyx/src/components/OnyxAccordionItem/OnyxAccordionItem.vue
```diff
@@ -0,0 +1,165 @@
...
+      :aria-expanded="isOpen"
+      :aria-controls="'panel-' + itemId"
+      :aria-disabled="accordionContext?.disabled.value || props.disabled"
+    >
+      <div class="onyx-accordion-item__header__content">
```
larsrickert, line 63: In BEM naming only one `__` can be used so we would need to rename this, e.g. to `onyx-accordion-item__header-content`

larsrickert: Same for `__header__icon`

packages/sit-onyx/src/components/OnyxAccordionItem/OnyxAccordionItem.vue
```diff
@@ -0,0 +1,165 @@
...
+        <slot name="header"></slot>
+      </div>
+
+      <OnyxIcon
+        label="toggle-accordion-icon"
```
larsrickert, line 68:
```suggestion
```

The icon does not have a label prop

packages/sit-onyx/src/components/OnyxAccordionItem/OnyxAccordionItem.vue
```diff
@@ -0,0 +1,165 @@
...
+  </details>
+</template>
+
+<style lang="scss">
+@use "../../styles/mixins/layers";
+.onyx-accordion-item {
+  @include layers.component() {
+    border-bottom: var(--onyx-1px-in-rem) solid var(--onyx-color-component-border-neutral);
+    color: var(--onyx-color-text-icons-neutral-intense);
+    overflow: hidden;
```
larsrickert, line 90: Why is this needed? This would cut off things like tooltips or flyouts

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxAccordion/OnyxAccordion.ct.tsx
```diff
@@ -0,0 +1,99 @@
...
+        <template v-slot:header>Accordion Header 2</template>
+        Accordion Panel 2
+      </OnyxAccordionItem>
+    </OnyxAccordion>,
+  );
+
+  const firstHeader = component.locator(".onyx-accordion-item__header").first();
+  const secondHeader = component.locator(".onyx-accordion-item__header").nth(1);
+  const firstPanel = component.locator(".onyx-accordion-item__panel").first();
+  const secondPanel = component.locator(".onyx-accordion-item__panel").nth(1);
```
larsrickert, line 86:
```suggestion
  const firstHeader = component.getByRole('button', { name: 'Accordion Header 1' });
  const secondHeader = component.getByRole('button', { name: 'Accordion Header 2' });
  const firstPanel = component.getByLabel('Accordion Header 1');
  const secondPanel = component.getByLabel('Accordion Header 2');
```

We should try to avoid using `.locator()` and instead using Playwright web native selectors like `getByRole()` or `getByLabel()`.

You can use the VS Code Playwright extension to just click on elements and it will copy the selector :)

image

image

packages/sit-onyx/src/components/OnyxAccordion/OnyxAccordion.ct.tsx
larsrickert, line 1: You still need to run the Playwright screenshot pipeline to generate the screenshots :)

packages/sit-onyx/src/components/OnyxAccordion/OnyxAccordion.ct.tsx
larsrickert: Same for the OnyxAccordionItem tests

packages/sit-onyx/src/components/OnyxAccordionItem/OnyxAccordionItem.vue
```diff
@@ -0,0 +1,168 @@
...
+const isOpen = computed({
+  get: () => accordionContext?.openItems.value.has(itemId) || props.open || false,
+  set: (value: boolean) => {
+    accordionContext?.updateOpen(itemId, value);
+  },
+});
+
+const skeleton = useSkeletonContext(props);
+const { densityClass } = useDensity(props);
+const isDiabled = computed(() => accordionContext?.disabled.value || props.disabled);
```
larsrickert, line 38:
```suggestion
const isDisabled = computed(() => accordionContext?.disabled.value || props.disabled);
```

packages/sit-onyx/src/components/OnyxAccordionItem/OnyxAccordionItem.vue
```diff
@@ -0,0 +1,168 @@
...
+
+      cursor: pointer;
+      &:hover,
+      &:focus-visible {
+        background-color: var(--onyx-color-base-neutral-200);
+        outline: none;
+      }
+      &-content {
+        display: flex;
+        * {
```
larsrickert, line 111: Hmm not sure if we should apply `*` here...this was to change the font styles when e.g. `<h3>` is used, right? Maybe we should just remove the usage of h3 instead and use plain text for the header

larsrickert: @JoCa96 What do you think?

packages/sit-onyx/src/components/OnyxAccordionItem/OnyxAccordionItem.vue
```diff
@@ -0,0 +1,168 @@
...
+      width: 100%;
+      position: relative;
+      display: flex;
+
+      justify-content: space-between;
+      gap: var(--onyx-density-md);
+      align-items: center;
+      padding: var(--onyx-density-md);
+
+      cursor: pointer;
```
larsrickert, line 103: Cursor is not applied for me

packages/sit-onyx/src/components/OnyxAccordionItem/OnyxAccordionItem.vue
```diff
@@ -0,0 +1,168 @@
...
+      }
+
+      &-icon {
+        color: inherit;
+      }
+    }
+    &[open] &__header-icon {
+      transform: rotate(-90deg);
+    }
+    &:has(:focus-visible) {
```
larsrickert, line 124:
```suggestion
    &:has(&__header:focus-visible) {
```

Otherwise it would apply to any focus-visible elements, e.g. inputs inside the panel which is not intended I guess

packages/sit-onyx/src/components/OnyxAccordionItem/OnyxAccordionItem.vue
```diff
@@ -0,0 +1,168 @@
...
+      border-radius: var(--onyx-radius-md);
+      border-bottom: none;
+      z-index: 1;
```
larsrickert, line 127: Why is the z-index needed?

packages/sit-onyx/src/components/OnyxAccordionItem/OnyxAccordionItem.vue
```diff
@@ -0,0 +1,168 @@
...
+
+      &-icon {
+        color: inherit;
+      }
+    }
+    &[open] &__header-icon {
+      transform: rotate(-90deg);
+    }
+    &:has(:focus-visible) {
+      border-radius: var(--onyx-radius-md);
```
larsrickert, line 125: border radius does not seem to be applied
image

packages/sit-onyx/src/components/OnyxAccordionItem/OnyxAccordionItem.vue
```diff
@@ -0,0 +1,168 @@
...
+      }
+    }
+    &[open] &__header::after {
+      transform: rotate(90deg);
+    }
+    &__panel {
+      padding: var(--onyx-density-md);
+    }
+    &:has(&__header[aria-disabled="true"]) {
+      color: var(--onyx-color-text-icons-neutral-soft);
```
larsrickert, line 145: This would also change the text color of the panel, is this intended?

packages/sit-onyx/src/components/OnyxAccordionItem/OnyxAccordionItem.vue
```diff
@@ -0,0 +1,168 @@
...
+      .onyx-accordion-item__header {
+        color: var(--onyx-color-text-icons-neutral-soft);
+      }
+    }
+  }
+  &-skeleton {
+    padding: var(--onyx-density-md);
+    display: flex;
+    gap: var(--onyx-density-md);
+    border-bottom: 1px solid var(--onyx-color-component-border-neutral);
```
larsrickert, line 156:
```suggestion
    border-bottom: var(--onyx-1px-in-rem) solid var(--onyx-color-component-border-neutral);
```

packages/sit-onyx/src/components/OnyxAccordionItem/OnyxAccordionItem.vue
```diff
@@ -0,0 +1,168 @@
...
+
+    &__main {
+      width: calc(100% - 3 * var(--onyx-density-md));
```
larsrickert, line 159: Thats the same as `var(--onyx-density-2xl)`. Is there a reason why you are calculating here? Does it depend on some other styles, e.g. the padding value?

packages/sit-onyx/src/components/OnyxAccordionItem/OnyxAccordionItem.vue
```diff
@@ -0,0 +1,168 @@
...
+      height: var(--onyx-density-md);
+    }
+    &__icon {
+      height: var(--onyx-density-md);
```
larsrickert, line 163: If I see it correctly in Figma, the icon always has the same size for all densities. Only the padding changes

larsrickert: Same for the main element height

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxAccordionItem/OnyxAccordionItem.vue
ChristianBusshoff: yes, cause open panels can also be disabled

packages/sit-onyx/src/components/OnyxAccordionItem/OnyxAccordionItem.vue
ChristianBusshoff: it's 3 times the padding value

larsrickert: Then please add a new custom CSS variable like `--onyx-accordion-padding` and use this one so it behaves as expected when we change the padding in the future :)

larsrickert (changes requested):

larsrickert (approved):

PR #2416, opened 2024-12-26
update Stepper API

ChristianBusshoff:
- changed precision, stepSize and validStepSize


packages/sit-onyx/src/components/OnyxStepper/types.ts
```diff
@@ -33,18 +33,32 @@ export type OnyxStepperProps = FormInjectedProps &
     step?: number; // step-mismatch + step-increment
 
     /**
-     * The smallest allowed value and rounded precision
+     * Number of decimal places to show. Can also be negative. Value will be rounded if needed
+     * to match the specified precision.
+     *
+     * @example For `precision=2` with `modelValue=5`, "5.00" will be displayed.
+     * @example For `precision=-1` with `modelValue=60`, "100" will be displayed.
```
MajaZarkova, line 40: Just tested this, when I enter 60 -> 60,0 is displayed. Am I missing something or the description is wrong?

packages/sit-onyx/src/components/OnyxStepper/types.ts
```diff
@@ -33,18 +33,32 @@ export type OnyxStepperProps = FormInjectedProps &
...
+    stepSize?: number;
     /**
-     * Ensure no wrong number can be inputed
+     * Defines step size f0r valid/allowed values. Can be independent of the `stepSize` property.
+     * Uses the `min` property as base so if defining min=3 with validStepSize=2, only odd numbers will
+     *
+     * @example For `validStepSize` 0.01, only multiples of 0.01 are valid (useful for currencies)
+     * @example For `stepSize=4` `validStepSize=2`, only even numbers are valid and the user can adjust the value by 4 when clicking the +/- button.
+     * @example For `min=3` and `validStepSize=2`, only odd numbers will be valid
+     * @default undefined
```
MajaZarkova, line 58: This comment confuses me. Is this reffering to both validStepSize and stripStep?

MajaZarkova (changes requested):

packages/sit-onyx/src/components/OnyxStepper/types.ts
ChristianBusshoff: sorry, the description should be -2 instead of -1. I changed it.

packages/sit-onyx/src/components/OnyxStepper/types.ts
ChristianBusshoff: No, it's just the validStepSize. I copied the given description from the ticket, but I'm not sure if it's too long because it's also shown in Storybook.

packages/sit-onyx/src/components/OnyxStepper/types.ts
```diff
@@ -24,24 +24,39 @@ export type OnyxStepperProps = FormInjectedProps &
      */
     max?: number;
     /**
-     * Incremental step.
-     */
-    /**
+     * Ensure no wrong number can be inputed
```
MajaZarkova, line 27: This is not needed
```suggestion
     * Ensure no wrong number can be inputed
```

packages/sit-onyx/src/components/OnyxStepper/types.ts
MajaZarkova: If it was suggested in the ticket, then it's fine :)

MajaZarkova (changes requested):

larsrickert (approved):

PR #2418, opened 2024-12-27
fix(OnyxTooltip): fixed alignment of tooltip

ChristianBusshoff:
- added wedgePointsToCenter attribute to fix the issue


larsrickert (changes requested):
Imho the wedge should always be centered without needing to set a property.

larsrickert (changes requested):
I doubled checked with Jannick, when left/right aligned, the wedge should not be centered and should align like this:
image

We still need to ensure that the wedge looks correct in the info tooltip

PR #2450, opened 2025-01-08
fix(OnyxAvatar): update default initials

larsrickert:
relates to #2454, #2442

Previously the initials were taken from the first two words. Now they will be determined as described [here](https://github.com/SchwarzIT/onyx/issues/2454) by considering the locale.
If the username contains unsupported characters (e.g. for some Korean characters) a fallback icon will be displayed.

Example for "John Middlename Doe":

- Previously: "JM"
- Now: "JD"

The default slot has been removed in favor of the `initials` property to set custom initials.

## Checklist


packages/sit-onyx/src/components/OnyxAvatar/OnyxAvatar.vue
```diff
@@ -6,18 +6,13 @@ const props = withDefaults(defineProps<OnyxAvatarProps>(), {
   size: "48px",
 });
 
-const slots = defineSlots<{
```
JoCa96, line 9: Please leave the option in to use a slot, then we don't have a breaking change

larsrickert: Personally, I don't like having two ways for doing the same thing 🤔  Since slot also allows non-text content which we propably don't want to support, I'd prefer the propert.
What do you think?

packages/sit-onyx/src/components/OnyxAvatar/types.ts
```diff
@@ -2,13 +2,26 @@ import type { IconSize, IconSizeProp } from "../OnyxIcon/types";
 
 export type OnyxAvatarProps = IconSizeProp<Exclude<IconSize, "12px">> & {
   /**
-   * Label (e.g. the user name), required for accessibility.
-   * Will also determine the displayed initials.
+   * Username. Will determine the displayed initials.
    */
-  label: string;
+  username: DisplayName;
```
JoCa96, line 7: (Optional): When I read "username" I am always thinking "login name" e.g., `joca96` or `baracko`, so I would prefer something like:
```suggestion
  displayName: DisplayName;
```
```suggestion
  fullName: DisplayName;
```

packages/sit-onyx/src/utils/strings.ts
```diff
@@ -19,3 +19,35 @@ const removeDiacritics = (str: string) => str.normalize("NFD").replace(/[\u0300-
...
+    return;
+  }
+
+  const segmenter = new Intl.Segmenter(locale, { granularity: "word" });
+
+  const name = username.trim().toUpperCase();
+  const segments = Array.from(segmenter.segment(name));
+
+  if (segments.length === 1) return segments[0].segment.substring(0, 2);
+  return `${segments[0].segment.charAt(0)}${segments.at(-1)?.segment.charAt(0)}`;
```
JoCa96, line 39: For getting characters of a word, we should also use the segmenter API with granularity "grapheme", so that Unicode characters consisting of multiple code points are not split.

JoCa96 (changes requested):

packages/sit-onyx/src/utils/strings.ts
larsrickert: Done, can you please check if I implemented it correctly?

packages/sit-onyx/src/components/OnyxAvatar/types.ts
larsrickert: Currently this is aligned with the UserMenu. Would you also change it there too?

JoCa96: Sounds good :)

packages/sit-onyx/src/utils/strings.ts
JoCa96, line 1: Please add some unit tests for the new functions (maybe follow-up ticket/PR)

packages/sit-onyx/src/components/OnyxAvatar/types.ts
JoCa96: But you don't have to change it

JoCa96 (approved):

JoCa96 (approved):

JoCa96 (approved):

Issue #2293, opened 2024-12-10
Implement UX feedback for OnyxDataGrid moreActions


MajaZarkova:
### Open Questions/To-dos

- [x] Ask @JoCa96  and @jannick-ux about the linear order of the actions
       - no need for that

### Why?

This is a follow up ticket for #1939 

### Acceptance criteria

Following is implemented:
- [x] functional tests (Playwright or unit test)
- [x] If actions are added without listItems, the flyout should contain the icons
- [x] The flyout should open on click, not on hover

### Definition of Done

- The following component requirements are implemented:
  - [x] example usage added to `apps/demo-app/src/views/HomeView.vue`
- covered by tests
  - [x] functional tests (Playwright or unit test)
  - [x] visual tests (Playwright screenshots)
- [x] follow-up tickets were created if necessary
- [x] updated version + documentation is deployed
- [x] Storybook can show the feature
- [x] Storybook code snippet of new/changed examples are checked that they are generated correctly
- [x] Namings are aligned with Figma

**Approval**

- [Storybook](https://storybook.onyx.schwarz/?path=/story/data-datagrid--header-interactions)
- [x] approved by designer @jannick-ux
- [x] approved by dev @larsrickert

jannick-ux:
UX findings:

- [ ] when clicking on the "more button" again, please make it happen to close the flyout

PR #2487, opened 2025-01-13
feat(demo-app): Add data grid example in the demo-app

MajaZarkova:
Relates to #2293 

- Add example in the demo-app

### DIFF
--- a/apps/demo-app/src/views/DataGridView.vue
+++ b/apps/demo-app/src/views/DataGridView.vue
@@ -4,7 +4,7 @@ import {
   DataGridFeatures,
   OnyxDataGrid,
   OnyxHeadline,
-  OnyxListItem,
+  OnyxMenuItem,
   OnyxPageLayout,
   OnyxSwitch,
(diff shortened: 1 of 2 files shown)


apps/demo-app/src/views/DataGridView.vue
```diff
@@ -12,11 +23,40 @@ const data = [
...
+  watch: [],
+  header: {
+    actions: () => [
+      {
+        iconComponent: h(OnyxSystemButton, {
+          label: "Column options",
+          icon: sort,
+          color: "medium",
+        }),
+        menuItems: [h(OnyxListItem, () => "Pin column"), h(OnyxListItem, () => "Unpin column")],
```
larsrickert, line 37: The system button should use `OnyxMenuItem` since its used inside the flyout menu: https://storybook.onyx.schwarz/?path=/docs/buttons-systembutton-examples-flyoutmenu--docs

larsrickert (changes requested):

PR #2492, opened 2025-01-13
feat(OnyxFormElements): add smart autofocus

ChristianBusshoff:
-implemented smart autofocus on all formElements


packages/sit-onyx/src/composables/useAutoFocus.ts
```diff
@@ -0,0 +1,16 @@
+import { nextTick, onMounted, type Ref } from "vue";
+
+export const useAutofocus = (
+  inputRef: Ref<HTMLElement | null>,
+  props: { autofocus: boolean; loading?: boolean },
+) => {
+  onMounted(() => {
+    if (props.autofocus && !(props.loading ?? false)) {
+      nextTick(() => {
```
JoCa96, line 9: As far as I am aware, `nextTick` should not be necessary. When `onMounted` is called, everything is rendered, so no waiting should be needed.

JoCa96 (changes requested):
see comment

packages/sit-onyx/src/composables/useAutoFocus.ts
```diff
@@ -0,0 +1,12 @@
+import { onMounted, type Ref } from "vue";
+
+export const useAutofocus = (
+  inputRef: Ref<HTMLElement | null>,
+  props: { autofocus: boolean; loading?: boolean },
+) => {
+  onMounted(() => {
+    if (props.autofocus && !(props.loading ?? false) && document.activeElement !== inputRef.value) {
```
JoCa96, line 8: It is fine to directly use `!props.loading` here, as it will convert nullish values to false
```suggestion
    if (props.autofocus && !props.loading && document.activeElement !== inputRef.value) {
```

JoCa96, line 8: Very good idea to consider the loading state. But maybe we can improve this a bit by waiting for loading to change?

This would be my suggestion:
```ts
  // only register lifecycle hook, when autofocus is set
  if (!props.autofocus) {
    return;
  }

  onMounted(() => {
    // if not loading, we can ensure focus immediately
    if (!props.loading) {
      inputRef.value?.focus();
      return;
    }

    // otherwise we wait for loading to become `false`
    watch(
      () => !!props.loading,
      () => inputRef.value?.focus(),
      { once: true },
    );
  });
```

JoCa96 (approved):

PR #2498, opened 2025-01-14
fix(OnyxInput, OnyxTextarea): `maxlength` doesn't restrict the user from typing more than the allowed characters.

JoCa96:
Relates to #2308 

- `maxlength` doesn't restrict the user from typing more than the allowed characters.
- The previous behavior, which restricts the user from typing more than the allowed characters, can be achieved by setting `strictMaxlength`.

### DIFF
--- a/packages/sit-onyx/src/composables/useLenientMaxLengthValidation.ts
+++ b/packages/sit-onyx/src/composables/useLenientMaxLengthValidation.ts
@@ -85,8 +85,14 @@ export const useLenientMaxLengthValidation = (options: UseTextInputOptions) => {
     options.props.maxlength &&
     options.props.modelValue.length > options.props.maxlength
       ? {
-          longMessage: t.value("validations.rangeOverflow.fullError"),
-          shortMessage: t.value("validations.rangeOverflow.preview"),
+          longMessage: t.value("validations.tooLong.fullError", {
+            n: options.props.modelValue.length,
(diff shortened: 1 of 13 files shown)


packages/sit-onyx/src/composables/useLenientMaxLengthValidation.ts
```diff
@@ -0,0 +1,111 @@
...
+   * Maximum number of characters that are allowed to be entered.
+   * Warning: when the value is (pre)set programmatically,
+   * the input invalidity will not be detected by the browser, it will only turn invalid
+   * as soon as a user interacts with the input (types something).
+   */
+  maxlength?: number;
+  /**
+   * Restricts the user from typing more characters than allowed.
+   */
+  strictMaxlength?: boolean;
```
larsrickert, line 32: Since this prop only makes sense when `maxlength` is set, what do you think about combining them into one property like we are already doing for other props/components?

So the user can do:
```vue
<OnyxInput :maxlength="42" />
<OnyxInput :maxlength="{ max: 42, strict: true }" />
```

packages/sit-onyx/src/composables/useLenientMaxLengthValidation.ts
larsrickert, line 1: Can we also add Playwright tests for this?

larsrickert (approved):

PR #2529, opened 2025-01-16
chore: replace redundant useManagedState with Vue's defineModel

JoCa96:
The changes are mostly internal, but the typings were of `OnyxSelect` have changed:

- The `modelValue` now infers a specific subtype of `SelectOptionValue` and the `options` values must match.
- `withSearch`: Filtering of the options will not automatically disabled anymore when `searchTerm` is bound. Instead `noFilter` must be set.
- removed `Arrayable` type, imported from `vitest`, as it was an internal type. Replaced with our own type with the same name.
- removed redundant `MaybeReactiveSource` type, as it is now covered by Vue's `MaybeRefOrGetter` type


packages/sit-onyx/src/components/OnyxMiniSearch/OnyxMiniSearch.vue
```diff
@@ -13,29 +13,22 @@ defineOptions({ inheritAttrs: false });
...
   /**
    * Emitted when the clear button is clicked.
    */
   clear: [];
 }>();
 
+/**
+ * Current input/search value.
+ */
+const modelValue = defineModel<string>();
```
larsrickert, line 25: Previously, the emit value was required although the property was optional.
How is this type now? Can the emitted value also be undefined? We might need to define a default value in this case

packages/sit-onyx/src/components/examples/FormExample/FormExample.vue
```diff
@@ -59,7 +59,7 @@ const fruitOptions = ref(
     "Melon",
     "Raspberry",
     "Strawberry",
-  ].map<SelectOption>((option) => ({ value: option.toLowerCase(), label: option })),
+  ].map<SelectOption<string>>((option) => ({ value: option.toLowerCase(), label: option })),
```
larsrickert, line 62: We could also use `satisfies SelectOption` so the type is inferred automatically

packages/sit-onyx/src/components/OnyxSelect/OnyxSelect.vue
```diff
@@ -423,7 +424,7 @@ defineExpose({ input: computed(() => selectInput.value?.input) });
                   headlessOption({
                     value: option.value,
                     label: option.label,
-                    disabled: option.disabled,
+                    disabled: option.disabled as any,
```
larsrickert, line 427: Would prefer to get rid of the any here :)
Also the TODO in the next line seems to be fixed now

apps/demo-app/tsconfig.json
```diff
@@ -3,6 +3,7 @@
   "include": ["src/**/*"],
   "compilerOptions": {
     "rootDir": "./src",
+    "lib": ["DOM", "DOM.Iterable", "ES2022"],
```
larsrickert, line 6: Why is this change needed? Same question for the playground

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxSelect/OnyxSelect.vue
JoCa96: done

packages/sit-onyx/src/components/OnyxMiniSearch/OnyxMiniSearch.vue
JoCa96: done

apps/demo-app/tsconfig.json
JoCa96: Honestly, I don't know why it is needed *now*.
Without it, I am getting errors:
Screenshot 2025-01-17 at 08 58 30

The errors make sense, but why only after all these seemingly unrelated changes?
My assumption is, that the Vue compiler can handle the updated typings better and doesn't bail out (as early)?

larsrickert: Hmm shouldn't the headless code be inlined? Maybe this has something to do with the update to Vite 6 in #2530 ?

packages/sit-onyx/src/components/OnyxSelect/OnyxSelect.vue
```diff
@@ -424,7 +423,6 @@ defineExpose({ input: computed(() => selectInput.value?.input) });
                     value: option.value,
                     label: option.label,
                     disabled: option.disabled,
-                    // TODO: remove type cast once its fixed in Vue / vue-tsc version
                     selected: arrayValue.some((value: TValue) => value === option.value),
```
larsrickert, line 426:
```suggestion
                    selected: arrayValue.some((value) => value === option.value),
```

I guess the TODO was meant for this one

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxSelect/OnyxSelect.vue
JoCa96: the type hint here is still necessary

apps/demo-app/tsconfig.json
JoCa96: JavaScript is inlined, the typings are not inlined afaik

JoCa96: I don't think it is related, because there are no issues on main

larsrickert: I think we should find out why and fix this, otherwise propably all projects using onyx will have build errors...

JoCa96: I reverted the code that caused this

PR #2534, opened 2025-01-16
fix(OnyxFormElement): fix broken counter text when maxlength with counter is used

JoCa96:
- fixed broken counter text, which was caused by string conversion of an object
- remove import for `defineSlots` to get rid of vue warning
- counter text is now colored red, when the text is over the maxlength


packages/sit-onyx/src/components/OnyxFormElement/OnyxFormElement.ct.tsx
```diff
@@ -55,6 +55,28 @@ test("should render error message", async ({ mount }) => {
   await expect(message).toBeVisible();
 });
 
+test("should render counter text correctly", async ({ mount }) => {
```
larsrickert, line 58: Can you also add a screenshot test for this?
+ should we also test for strict max length?

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxFormElement/OnyxFormElement.ct.tsx
JoCa96: done; there are already tests for the strict max-length in place

PR #2568, opened 2025-01-21
fix(OnyxDataGrid): export sorting types

larsrickert:
Remove scoped namespace for exporting data grid feature types.

Feedback from @oemueller 

## Checklist

### DIFF
--- a/.changeset/twenty-swans-eat.md
+++ b/.changeset/twenty-swans-eat.md
@@ -3,3 +3,7 @@
 ---
 
 fix(OnyxDataGrid): export sorting types
+
+Previously, all data grid types were exported under the `DataGridFeatures` namespace so if you e.g. wanted to use the `SortState` type, you'd need to use `DataGridFeatures.SortState`.
+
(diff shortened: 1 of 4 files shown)


JoCa96 (commented):
I am not a fan tbh. We will be running into naming collisions at some point. So we either need to namespace types or put a prefix into their name (like DataGridSortState).
I prefer namespacing as it is better readable IMHO.

Which issue are we trying to solve with this PR specifically?

larsrickert:
> I am not a fan tbh. We will be running into naming collisions at some point. So we either need to namespace types or put a prefix into their name (like DataGridSortState). I prefer namespacing as it is better readable IMHO.
> 
> Which issue are we trying to solve with this PR specifically?

I think the main issue/confusing is that if you take a look at the docs or component types, the type will always be just `SortState`. So if you want to define your own variable with this type, you'd expect it to be imported by `import { SortState } from "sit-onyx";` but that does currently not work. Also the IDE will not auto-suggest the import so the user might think that the type is missing

oemueller:
I would be fine with _DataGridSortState_ :-) 
yes, the problem was with the missing auto-suggest in the IDE. 
I imagine that many users will complain about this.

PR #2570, opened 2025-01-21
feat(OnyxAccordion): implement review feedback

ChristianBusshoff:
- emit open state
- skeleton can be a number
- changing the exclusive mode is closing all open accordions


packages/sit-onyx/src/components/OnyxAccordionItem/OnyxAccordionItem.vue
```diff
@@ -27,10 +27,15 @@ defineSlots<{
 const itemId = useId();
 const accordionContext = inject(ACCORDION_INJECTION_KEY);
 
+const emit = defineEmits<{
+  (event: "update:open", accordionItem: string, value: boolean): void; // Event declaration
```
larsrickert, line 31:
```suggestion
  /** 
  * Emitted when the open state is changed.
  */
  "update:open": [isOpen: boolean]
```

To align with other component, we should use the new syntax :) Also I don't think we need to emit the itemId since its an internal ID used by the accordion

packages/sit-onyx/src/components/OnyxAccordion/OnyxAccordion.vue
```diff
@@ -31,10 +32,16 @@ const updateOpen = (id: string, isOpen: boolean) => {
...
     return;
   }
-  if (props.exclusive) openItems.value.clear();
+  if (exclusive.value) openItems.value.clear();
   openItems.value.add(id);
 };
 
+watch(exclusive, (newExclusive) => {
+  if (newExclusive) {
+    openItems.value.clear();
```
larsrickert, line 41: We only need to clear if more than 1 accordion is open. If exactly one is open thats fine.
Did you discuss with Jannick how this should behave? Maybe he decides that we don't want to handle this at all

packages/sit-onyx/src/components/OnyxAccordion/OnyxAccordion.vue
```diff
@@ -45,7 +52,10 @@ provide(ACCORDION_INJECTION_KEY as AccordionInjectionKey, {
 
 <template>
   <div :class="['onyx-component', 'onyx-accordion', densityClass]">
-    <slot></slot>
+    <template v-if="typeof skeleton === 'number'">
+      <OnyxAccordionItem v-for="i in skeleton" :key="i" skeleton />
```
larsrickert, line 56: Oh the skeleton count is defined by the items passed via the slot, I didn't think about this.
In this case I don't think we need additional logic here :)

larsrickert (changes requested):

larsrickert (approved):

Issue #2380, opened 2024-12-19
Fix OnyxLayoutGrid example


MajaZarkova:
RIght now default columns for the GridPlayground for a big screen is 16 columns. We should change it to 12.

- [x] Add CSS variable for DEV
- [x] Update GridPlayground Layout to transport correct information about behaviour and usage to our user
- [x] add CSS class `onyx-grid-span-full` that sets the column span to the max columns so it can be used independent on whether the max columns are 12, 16 or 20 columns + add docs for it

jannick-ux:
Design conception is finished
https://www.figma.com/design/YfEUBOHk4J4nYrk04geswG/Onyx-Component-Library?node-id=7409-19220&t=XHlm692KNqLFvSum-4

JoCa96:
> Design conception is finished https://www.figma.com/design/YfEUBOHk4J4nYrk04geswG/Onyx-Component-Library?node-id=7409-19220&t=XHlm692KNqLFvSum-4

Created a follow-up ticket: [#2691](https://github.com/SchwarzIT/onyx/issues/2691)

PR #2590, opened 2025-01-23
feat(grid): implement onyx-grid-span-full and fix issues regarding optional max columns

JoCa96:
Relates to #2380 

- feat(grid): new `onyx-grid-span-full` class which enables an element to always span the full row width.
- fix(grid)!: fix issues regarding optional max columns. 
  The default maximum number of columns is `12` (before it was `16`).
  The `onyx-grid-xl-20` class now also sets `16` columns for the `lg` breakpoint.
  The new `onyx-grid-lg-16` class sets `16` columns for the `lg` and `xl` breakpoint.

- chore(docs): update playground to support the changes


packages/sit-onyx/src/components/examples/GridPlayground/GridPlayground.vue
```diff
@@ -144,15 +145,16 @@ const currentBreakpoint = computed(() => {
...
           :options="alignmentOptions"
           direction="horizontal"
           :disabled="gridSettings.maxWidth === 'none'"
         />
 
         <OnyxRadioGroup
-          v-if="viewportSize.width.value >= ONYX_BREAKPOINTS.xl"
           v-model="gridSettings.maxColumns"
           label="Max columns"
+          label-tooltip="Defines the maximum number of columns the grid can have. This only applies to the XL breakpoint."
```
larsrickert, line 157: 16 also applies to lg breakpoint right? Not only XL

larsrickert (changes requested):

packages/sit-onyx/playwright/snapshots/components/examples/GridPlayground/EditGridElementDialog/default-edge-linux.png
larsrickert, line 1: (optional) We can increase the screenshot height so the buttons are not cut off

larsrickert (approved):

Issue #447, opened 2024-02-29
Implement vue-router integration


larsrickert:
### Why?

As a developer, I want to be able to easily integrate, the de facto standard, [Vue Router](https://router.vuejs.org/) with my in-application links and buttons. 

### Acceptance criteria

- [x] Clicking an "integrated" link must perform a navigation using the provided router, that means no actual browser navigation happens
- [x] All links must be native links (`a` tags) to support the basic browser features
- [x] We have a new `SharedLinkProps` type which is compatible to the Vue Routers [RouterLinkProps](https://router.vuejs.org/api/interfaces/RouterLinkProps.html)
  - [x] `OnyxLink` uses this new type
- [x] Via `createOnyx` a router can be passed in, the router API must match the [Vue Router API](https://router.vuejs.org/api/interfaces/Router.html#Interface-Router)
  - [x] This is ensured via type tests (like in packages/sit-onyx/src/components/OnyxForm/OnyxForm.core.spec-d.ts)
  - [x] Vue-Router must only be a dev dependency
- [x] Documentation has been updated

### Implementation details

- We provide an onyx specifc and internal composable `useLink`, similar to [useLink](https://router.vuejs.org/api/functions/useLink.html). This way we can reuse the composable and the type for all components which can behave like a link (see also #1427).
- Maybe the composable can/must be wrapped in a "RouterLink" component for handling SPA links vs. external links, etc.

```ts
/**
 * TODO: Provide appropriate descriptions
 */
export type ProvideRouterOptions = {
  // https://router.vuejs.org/api/interfaces/Router.html#push-
  push: (to: unknown) => void;
  // https://router.vuejs.org/api/interfaces/Router.html#replace-
  replace: (to: unknown) => void;
  // https://router.vuejs.org/api/interfaces/RouteLocationNormalizedLoadedGeneric.html#fullPath
  // for matching the currently active route
  readonly currentRoute: { fullPath: string };
};

export type OnyxPluginOptions = {
  i18n?: ProvideI18nOptions;
  router?: ProvideRouterOptions;
};

```

### Definition of Done

- The following component requirements are implemented:
  - [x] example usage added to `apps/demo-app/src/views/HomeView.vue`
- covered by tests
  - [x] functional tests (Playwright or unit test)
- [x] updated version + documentation is deployed

**Approval**

- [Storybook](https://storybook.onyx.schwarz/?path=/docs/support-routerlink--docs)
[...]

PR #2600, opened 2025-01-24
feat: implement Vue Router integration

larsrickert:
Relates to #447

Implement Vue Router integration by supporting and using OnyxRouterLink component and useLink composable.

## Checklist


packages/sit-onyx/src/composables/useLink.ts
```diff
@@ -0,0 +1,74 @@
...
+   *   </a>
+   * </template>
+   * ```
+   */
+  const navigate = (e: MouseEvent, href: string, target: LinkTarget) => {
+    if (!shouldHandleNavigation(e)) return;
+    const isExternal = isExternalLink(href);
+    if (isExternal || !router) window.open(href, target);
+    else router.push(href);
+  };
```
JoCa96, line 37: IMHO we should only handle the link when a router is given and it's internal.
Or rather `window.open` should be left to the default browser behavior.

External links can be handled by the browser, the the user has to set the target correctly.
But we could consider to set the target on the `a` accordingly for external links, when left undefined by the user.

If there is a reason that I am overseeing, we might need an explanatory comment in the code.

```suggestion
  const navigate = (e: MouseEvent, href: string) => {
    if (shouldHandleNavigation(e) && !isExternalLink(href) && router) {
      router.push(href);
    }
  };
```

packages/sit-onyx/src/components/OnyxRouterLink/OnyxRouterLink.ct.tsx
```diff
@@ -0,0 +1,133 @@
...
+    mount,
+    browserName,
+  }) => {
+    // eslint-disable-next-line playwright/no-conditional-in-test -- Alt does not work in safari
+    if (browserName === "webkit" && modifier === "Alt") {
+      // eslint-disable-next-line playwright/no-skipped-test -- Alt does not work in safari
+      return test.skip();
+    }
+
+    // ARRANGE
```
JoCa96, line 94: Conditions can be put directly into the skip command:
```suggestion
    test.skip(browserName === "webkit" && modifier === "Alt");

    // ARRANGE
```

JoCa96 (changes requested):

JoCa96 (approved):

Issue #2384, opened 2024-12-19
Implement OnyxColorSchemeDialog update to use internal onyx components


larsrickert:
### Why?

We currently already have the [OnyxColorSchemeDialog](https://storybook.onyx.schwarz/?path=/story/navigation-modules-colorschemedialog--default) component that can be used to switch between auto, light and dark mode.

We want to update its implementation internally to use the OnyxModalDialog and OnyxCard component to prevent custom/duplicated code.

### Depends on

- #2383 

### Design

[Storybook](https://storybook.onyx.schwarz/?path=/story/navigation-modules-colorschemedialog--default)

### Acceptance criteria

- [x] The `OnyxColorSchemeDialog` should appear like our default modal/dialog
- [x] the OnyxModalDialog is used where "Change appearance" is the headline and "Select the appearance of the application:" becomes the subtitle
- [x] the OnyxCard is used for the auto, light and dark selectable option

### Definition of Done

- Should be covered by tests:
  - [x] functional tests (Playwright or unit test)
  - [x] visual tests (Playwright screenshots)

- Make sure, that
  - [x] updated version + documentation is deployed
  - [x] Storybook can show the feature
  - [x] Storybook code snippet of new/changed examples are checked that they are generated correctly
  - [x] Namings are aligned with Figma

#### Approval

[Storybook](https://storybook.onyx.schwarz/?path=/docs/navigation-modules-colorschemedialog--docs)

- Checked and approved by
  - [x] designer @jannick-ux 
  - [x] dev @MajaZarkova

ChristianBusshoff:
If the opening state is set, the dialog doesn't close through the buttons. Is this the intended behavior?

PR #2623, opened 2025-01-27
docs(OnyxColorSchemeDialog): fix Storybook example when open property is set manually

larsrickert:
Relates to https://github.com/SchwarzIT/onyx/issues/2384#issuecomment-2611865738

Currently, when setting the `open` property manually, the close button did not close the dialog again.


packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxColorSchemeDialog/OnyxColorSchemeDialog.stories.ts
JoCa96, line 1: This is also an issue with the `OnyxModalDialog`

larsrickert: Good point, done

JoCa96 (approved):

Issue #456, opened 2024-03-01
Implement notification dot for icons


jannick-ux:
### Depends on

- #841 

#### Approval

[Storybook](https://storybook.onyx.schwarz/?path=/docs/support-notificationdot--docs)
- [x] approved by dev @JoCa96 
- [x] approved by ux @

PR #2635, opened 2025-01-29
feat: implement `OnyxNotificationDot` component

d-dimitrovv:
Relates to #2737, #456

Implement new OnyxNotificationDot component to support showing a dot in the top right corner of any component


packages/sit-onyx/src/components/OnyxNotificationDot/OnyxNotificationDot.vue
```diff
@@ -0,0 +1,49 @@
+<script lang="ts" setup>
+import OnyxBadge from "../OnyxBadge/OnyxBadge.vue";
+import type { OnyxNotificationDotProps } from "./types";
+
+const props = withDefaults(defineProps<OnyxNotificationDotProps>(), {
+  color: "warning",
+  hidden: false,
```
JoCa96, line 7: This property seems to be unused? Is there a use-case?

JoCa96 (changes requested):

packages/sit-onyx/src/components/OnyxNotificationDot/OnyxNotificationDot.vue
larsrickert: Good catch, I fixed it!
Yes the `hidden` property can be used to show the notification dot based on conditions, e.g. you can just show a notification bell icon inside the nav bar and only show the dot when there are unread messages, so e.g.:

```html
<OnyxNotificationDot :hidden="unreadNotifications.length > 0" />
```

So you don't need to use `v-if / v-else` and duplicate the bell icon

JoCa96 (approved):

PR #2661, opened 2025-02-03
feat: enhance Vue Router integration

larsrickert:
Relates to #447, https://github.com/SchwarzIT/onyx/issues/447#issuecomment-2624779076

- close mobile nav bar flyouts when route changes
- change nav app area to be a link (defaults to "/" but can be overridden)
- remove navigate event for nav button and nav item since they are no longer needed
- fix bug that mobile nav item with children does not open child view but instead directly opens link

## Checklist


packages/sit-onyx/src/components/OnyxNavBar/OnyxNavBar.ct.tsx
MajaZarkova, line 1: In Storybook, when I open the default example of the nav bar and click on the app area, a new storybook is opened inside of the example:

This is what I get:

image


Is it intentional?

MajaZarkova (changes requested):

packages/sit-onyx/src/components/OnyxNavBar/OnyxNavBar.ct.tsx
MajaZarkova: Same thing happanes if I click on the nav buttons

MajaZarkova (approved):

Issue #2496, opened 2025-01-14
Implement Header labels for OnyxDataGrid


JoCa96:
### Why?

Users want to see nice header texts

### Acceptance criteria

- [ ] The `OnyxDataGrid` is able to show custom labels in the header

### Definition of Done

- Should be covered by tests:
  - [x] functional tests (Playwright or unit test)
  - [x] visual tests (Playwright screenshots)

- Make sure, that
  - [x] follow-up tickets were created if necessary
  - [x] updated version + documentation is deployed
  - [x] Storybook can show the feature

#### Approval

[Storybook](https://storybook.onyx.schwarz/?path=/story/data-datagrid--default)

- Checked and approved by
  - [x] dev @larsrickert

PR #2677, opened 2025-02-04
feat(OnyxDataGrid): implement header labels

JoCa96:
Relates to #2496

- Adds optional `label` property to column config
- Added jsdocs for `ColumnConfig`


packages/sit-onyx/src/components/OnyxDataGrid/OnyxDataGrid.ct.tsx
```diff
@@ -1,17 +1,27 @@
...
+    expect(headers).toHaveLength(3);
+    await expect(component).toHaveScreenshot("data-grid-with-header-labels.png");
+  });
 
-  // ACT
-  await moreActionsBtn.click();
+  await test.step("should support header actions", async () => {
+    // ARRANGE
+    const moreActionsBtn = component.getByLabel("Toggle column actions").first();
+    const flyout = component.getByLabel('Choose an action for the column "name"');
```
larsrickert, line 18: Should we also test that if you pass a label, that this "Choose an action" label also uses the label instead of the key?

packages/sit-onyx/src/components/OnyxDataGrid/OnyxDataGrid.stories.ts
```diff
@@ -23,7 +23,11 @@ type Story = StoryObj<typeof OnyxDataGrid>;
 
 export const Default = {
   args: {
-    columns: ["name", "age", "birthday"],
+    columns: [
+      { key: "name", label: "Firstname" },
```
larsrickert, line 27:
```suggestion
      { key: "name", label: "First name" },
```

Is it "Firstname" or "First name"?

packages/sit-onyx/src/components/OnyxDataGrid/OnyxDataGrid.stories.ts
```diff
@@ -34,7 +38,11 @@ export const Default = {
 
 export const HeaderInteractions = {
   args: {
-    columns: ["name", "age", "birthday"],
+    columns: [
```
larsrickert, line 41: Instead of duplicating it, we could use `Default.args.columns` to re-use them from the default story

larsrickert (approved):

packages/sit-onyx/src/components/OnyxDataGrid/OnyxDataGrid.ct.tsx
JoCa96: done

larsrickert (approved):

Issue #2370, opened 2024-12-18
Implement Cell Renderer


JoCa96:
### Why?

As a user, I want to see different kind of data visualizations for different kind of columns.

### Design

[Figma](https://www.figma.com/design/YfEUBOHk4J4nYrk04geswG/Onyx-Component-Library?m=auto&node-id=4435-44202)

### Acceptance criteria

- [x] The `OnyxDataGrid` is able to display numbers differently from strings, when configured so via the columns
- [x] For type number
  - [x] monospace font is used
  - [x] is formatted using [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat)
- [ ] ~options/props for the type are generically made available to the feature user~ FoUp: #2783 

### Implementation details

- The `column` property of the [`OnyxDataGridProps`](https://github.com/SchwarzIT/onyx/blob/72add3ea86b22d52944e7634802c85f01601cedf/packages/sit-onyx/src/components/OnyxDataGrid/types.ts#L8) is extended to support a `type` property. For now, it will support the types `"string" | "number"`. Default is `"string"`.
- The [`createRendererRows`](https://github.com/SchwarzIT/onyx/blob/2b9f36cf1ddb11412850fabe5b9915af58f491af/packages/sit-onyx/src/components/OnyxDataGrid/features/index.ts#L167) function must be adapted. Based on the column `type` the cells component render function is switched. Make sure that the same render function is used.
- features should be able to register their own types with the respective render functions via the `DataGridFeature` API

`ColumnConfig` API types:
```ts
/**
 * Normalized config for internal usage
 */
export type NormalizedColumnConfig<TEntry extends DataGridEntry> = {
  key: keyof TEntry;
  label?: string;
  type?: PropertyKey;
};

/**
 * ColumnConfig for the enduser
 */
export type ColumnConfig<TEntry extends DataGridEntry> =
  | keyof TEntry
  | NormalizedColumnConfig<TEntry>;
```

Feature API types:
```ts
  /**
   * Defines a renderer for a type.
   */
  typeRenderer?: Record<PropertyKey, DataGridRendererCellComponent<TEntry>>;

  /**
   * Allows modifying of the column configuration. E.g. adding or modifying columns. 
   */
  modifyColumns?: {
    func: (columns: Readonly<NormalizedColumnConfig<TEntry>[]>) => NormalizedColumnConfig<TEntry>[];
  };
```

[...]

PR #2679, opened 2025-02-04
feat(OnyxDataGrid): Implement cell renderer

MajaZarkova:
Relates to #2370 

- New 'type' property is added to the ColumnConfig type. With this property a column can be of type 'string' or 'number'. Different styles will be applied depending on the type of the column.

## Checklist


packages/sit-onyx/src/components/OnyxDataGrid/features/index.ts
```diff
@@ -49,7 +49,7 @@ export type ColumnConfig<TEntry extends DataGridEntry, TTypes> =
  */
 export type NormalizedColumnConfig<TEntry extends DataGridEntry, TTypes = PropertyKey> = {
   key: keyof TEntry;
-  type?: TTypes;
+  type?: TTypes | "string" | "number";
```
larsrickert, line 52: We should rather change the generic definition, so:

```ts
<TTypes extends PropertyKey = "string" | "number">
```

packages/sit-onyx/src/components/OnyxDataGrid/OnyxDataGrid.stories.ts
```diff
@@ -74,6 +74,21 @@ export const HeaderInteractions = {
   },
 } satisfies Story;
 
+export const WithDifferentColTypes = {
+  args: {
+    columns: [
+      { key: "name", type: "string" },
```
larsrickert, line 80: Shouldn't we set string as default so you don't have to specify anything for just showing strings?

packages/sit-onyx/src/components/OnyxDataGrid/features/index.ts
```diff
@@ -219,15 +219,41 @@ export const useDataGridFeatures = <
     header: { component: HeaderCell },
     cell: { component: (props) => String(props.modelValue) },
   };
+
+  const numberFormatter = (value: TEntry[keyof TEntry] | undefined) => {
+    if (typeof value !== "number") {
```
larsrickert, line 224:
```suggestion
    if (typeof value !== "number" || typeof value === "bigint") {
```

The Intl. formatter also supports bigint

packages/sit-onyx/src/components/OnyxDataGrid/features/index.ts
```diff
@@ -219,15 +219,41 @@ export const useDataGridFeatures = <
...
+      return;
```
larsrickert, line 225: Should we really show nothing if its no number? Maybe we could use the fallbackRenderer instead

packages/sit-onyx/src/components/OnyxDataGrid/features/index.ts
```diff
@@ -219,15 +219,41 @@ export const useDataGridFeatures = <
...
+    }
+    const locale = injectI18n().locale;
+    const formatter = new Intl.NumberFormat(locale.value);
+
+    return formatter.format(value);
+  };
+
+  const setRenderer = (type: PropertyKey | undefined) => {
```
larsrickert, line 233:
```suggestion
  const setRenderer = (type?: PropertyKey) => {
```

larsrickert: Why is this function needed anyway? Can't we just always define those renderers?

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxDataGrid/features/index.ts
JoCa96: The `TTypes` are the inferred types from features. `"string" | "number"` are hardcoded default types. When defined as default for the generic, they would be overridden by the inferred types from the features. 
This would remove them from the selection.
So the type definition is correct as it currently is defined.

We could however either add a bit more docs or a dedicated type to make it clearer:
```suggestion
  type?: TTypes | DefaultTypes; // where type DefaultTypes = "string" | "number";
```

packages/sit-onyx/src/components/OnyxDataGrid/OnyxDataGrid.stories.ts
JoCa96: Yes, that already works :) `type` is optional.

packages/sit-onyx/src/components/OnyxDataGrid/features/index.ts
JoCa96: now tries to render any given value

packages/sit-onyx/src/components/OnyxDataGrid/features/index.ts
JoCa96: Moved it out into `DefaultSupportedTypes`

larsrickert (approved):

Issue #2699, opened 2025-02-05
Fix OnyxAccordionItem not closing when :open receives a false


ChristianBecker02:
### Version

1.0.0-beta.146

### Description

I am currently trying to add a open all, close all AccordionItems button, and stumbled over the fact that when the :open parameter is set to true, the AccordionItem is correctly opening, but the reverse part - setting it to false does not close the AccordionItem. 

Manually it works and in the showcase of the component is marked as it can be toggled between true and false.

But its not working with ref, computed or else and also not with a simple button in the playground.

### Reproduction steps

https://playground.onyx.schwarz/?onyxVersion=1.0.0-beta.146#eNp9U01PGzEQ/StT9wJSkj20pzRUQIVUeihVy9FS5XonG4PXtuzZbVCU/87Ym48NIKQc1vOeZ96852zEVQizvkMxF4ukowkEVrnmQgpKUkBC6sJX6UwbfCTYwJ17Wl9p7WNtvJucHm8J26F03RF5B1tYRt+CFMnQ1HNdii+jXhGXRwprKKh02rtEgOugXA0XmXW2VDbheUGXndPEw4B801g8O4eNdLCjz3plO+RLH8Zn6bb55qIaFuR1Fqw0WEXI3wCL2vSgrUqJ184yp0009ZR1kDIOoxSFxsSTdXfFl+XsAsx9QMfdBhmHBoW9nw0fV6hqjCMM4H5lEvCvfYJDR/heeKMW1Yn+ofRKxF70KTJeZRfTpbZGP+bIi6UcezX4UrEx2a3xODERQ4DTVoXZQ/KOn06JQO4AfjfzIZRcO0Y/59OKKKR5Vena8d0arenjzCFVLrTVnnr5D0nx7ESVcTWumSlF7scxbnk+JY5maZoX07Vvg7EY70J+H6cqlLX+/49So9jhZF/XK9SPb9Qf0k7wr4gJY8+mHDBSsUEa4Js/P3HN3wew9XWXLXwH/I3J2y5rHGjXnatZ9ohX1N4WM41r7tPNmtCl/VJZaHGj8Ms/59s7qx/lfpp9Hrn4t8eYe7KBOdpEYvsMR95bjQ==

### Confirmations

- [x] I have checked that there is not already a bug report for my issue.
- [x] I confirm that are no sensitive information included in this bug report. e.g. credentials, internal project details (for employees of the Schwarz Group) etc.

[...]

PR #2740, opened 2025-02-13
refactor(OnyxAccordion): move management of items open state to parent component

larsrickert:
Relates to #2699

The open state of all nested OnyxAccordionItem components is now fully moved to the OnyxAccordion to prevent recursive state logic. The OnyxAccordion now also supports a `v-model` for the currently opened items.
Also the open state of the accordion can now be forced by just setting the `modelValue` property without `v-model` / `@update:model-value` event

## Checklist


JoCa96 (approved):

packages/sit-onyx/src/components/OnyxAccordion/types.ts
```diff
@@ -2,7 +2,11 @@ import type { SkeletonInjected } from "src/composables/useSkeletonState";
 import type { InjectionKey, Ref } from "vue";
 import type { DensityProp } from "../../composables/density";
 
-export type OnyxAccordionProps = DensityProp & {
+export type OnyxAccordionProps<TValue extends PropertyKey> = DensityProp & {
+  /**
+   * Currently opened items. Will include the `value` property of the nested `OnyxAccordionItems`.
+   */
+  modelValue?: TValue[];
```
JoCa96, line 9: What about using an object based API? It is easier and more intuitive to use. Also it could be extended without breaking changes :) 

```suggestion
  modelValue?: { [key: TValue]?: Nullable<boolean> };
```

JoCa96 (approved):

packages/sit-onyx/src/components/OnyxAccordion/types.ts
larsrickert: I'd like to stick to the array API since the usage/typing for the project is easier imho and also all other libraries I found use an array for the open items (Element Plus, Nuxt UI, Prime Vue etc.) :)

PR #2744, opened 2025-02-13
fix: correctly detect links as external

larsrickert:
Relates to #447 

Fix detection of external links. Since there are way more variations for external than for internal links, we will just detect if the link is internal and treat all other links as external.

Previously only links starting with `http://` or `https://` we detected as external links, meaning all other links we treated as internal and navigated using the [provided Vue Router](https://onyx.schwarz/development/router.html).

Since this is incorrect for links like `mailto:`, `tel:` etc. this behavior has been fixed.
Now only links starting with `/` and `#` are treated as **internal**. All other links are treated as **external**.

## Checklist

### DIFF
--- a/packages/sit-onyx/src/utils/index.ts
+++ b/packages/sit-onyx/src/utils/index.ts
@@ -1,6 +1,6 @@
 /**
- * Checks whether the given link is external (starts with http:// or https://).
+ * Checks whether the given link is internal / a relative link of the current application.
  */
-export const isExternalLink = (href: string): href is `http${"s" | ""}://${string}` => {
-  return /^http(s?):\/\//.test(href);
+export const isInternalLink = (href: string) => {
+  return href.startsWith("/") || href.startsWith("#");
 };
--- a/packages/sit-onyx/src/utils/index.ts
+++ b/packages/sit-onyx/src/utils/index.ts
@@ -2,5 +2,5 @@
  * Checks whether the given link is internal / a relative link of the current application.
  */
 export const isInternalLink = (href: string) => {
-  return href.startsWith("/") || href.startsWith("#") || /^\.+\//.test(href);
+  return href === "" || href.startsWith("/") || href.startsWith("#") || /^\.+\//.test(href);
 };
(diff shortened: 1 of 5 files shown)


packages/sit-onyx/src/utils/index.ts
```diff
@@ -1,6 +1,6 @@
 /**
- * Checks whether the given link is external (starts with http:// or https://).
+ * Checks whether the given link is internal / a relative link of the current application.
  */
-export const isExternalLink = (href: string): href is `http${"s" | ""}://${string}` => {
-  return /^http(s?):\/\//.test(href);
+export const isInternalLink = (href: string) => {
+  return href.startsWith("/") || href.startsWith("#");
```
JoCa96, line 5: Do we also need to consider relative URLs? e.g. `href.startsWith("./")`

packages/sit-onyx/src/utils/index.ts
```diff
@@ -1,6 +1,6 @@
 /**
- * Checks whether the given link is external (starts with http:// or https://).
+ * Checks whether the given link is internal / a relative link of the current application.
  */
-export const isExternalLink = (href: string): href is `http${"s" | ""}://${string}` => {
-  return /^http(s?):\/\//.test(href);
+export const isInternalLink = (href: string) => {
```
JoCa96, line 4: What was the reason, again, why we don't check against the actual origin?
E.g.
```ts
const hrefUrl = new URL(href, window.location);
return location.origin === hrefUrl.origin;
```

JoCa96 (changes requested):

packages/sit-onyx/src/utils/index.ts
larsrickert: Because we would need to take care about SSR that we use this utility only onMounted

packages/sit-onyx/src/utils/index.ts
larsrickert: Done, I also included "../"

packages/sit-onyx/src/utils/index.ts
JoCa96: Per my understanding, we can just use:
```suggestion
export const isInternalLink = (href: string) => !URL.canParse(); // if it can be parsed the URL is a full URL with some protocol and domain
```

larsrickert: Nice! Seems to work fine, I updated it

JoCa96 (approved):

Issue #2691, opened 2025-02-04
Implement updated GridPlayground design


JoCa96:
### Design

[Figma](https://www.figma.com/design/YfEUBOHk4J4nYrk04geswG/Onyx-Component-Library?node-id=7409-19220&t=XHlm692KNqLFvSum-4)

### Acceptance criteria

- [x] The new playground design is implemented
- [x] tooltip texts have been updated
- [x] grid element designs have been updated
- [x] The pink notes in figma have been considered

### Definition of Done

- Should be covered by tests:
  - [x] functional tests (Playwright or unit test)
  - [x] visual tests (Playwright screenshots)

- Make sure, that
  - [x] follow-up tickets were created if necessary -> #2908 
  - [x] updated version + documentation is deployed
  - [x] Storybook can show the feature
  - [x] Storybook code snippet of new/changed examples are checked that they are generated correctly
  - [x] Namings are aligned with Figma

#### Approval

[Storybook](https://storybook.onyx.schwarz/?path=/story/examples-gridplayground--default)

- Checked and approved by
  - [x] designer 
  - [x] dev @larsrickert

larsrickert:
Review feedback:
- the 16 column select option does not have any effect for me, 20 works
- the gap between the "Grid customization" headline and the text above it is larger in Figma, should we align this?
- the layout breaks on smaller screens (top section overflows + middle section gets pressed together):

![Image](https://github.com/user-attachments/assets/f3eba9bc-2256-4fba-a1ca-05414de165c5)

What about there notes in Figma?

![Image](https://github.com/user-attachments/assets/23d3e5d3-4aa3-4819-9443-8cad14f5c66a)

![Image](https://github.com/user-attachments/assets/fa1f3e82-6f4e-4d2a-9be0-e8f46c79ee98)

Rest looks good to me, well done!

PR #2829, opened 2025-02-25
feat(OnyxIcon): added `inline` mode for use with text

JoCa96:
Relates to #2691 

In some cases we want to use `OnyxIcon` in or with text.
This PR adds support for an `inline` prop, that sets the size and alignment accordingly.


packages/sit-onyx/src/components/OnyxIcon/OnyxIcon.vue
```diff
@@ -39,6 +40,11 @@ const props = defineProps<OnyxIconProps>();
       display: block;
     }
 
+    &--inline {
```
larsrickert, line 43: The icon already has `display: inline-block`. Is there any reason why we don't always set the vertical align middle?

packages/sit-onyx/src/components/OnyxIcon/OnyxIcon.stories.ts
```diff
@@ -51,5 +51,22 @@ export const WithColor = {
...
     color: "success",
+    style: "--icon-size: 1em; vertical-align: middle;",
   },
 } satisfies Story;
+
+/**
+ * This example shows that and how the OnyxIcon can be used with text.
+ */
+export const Inline = {
+  args: {
```
larsrickert, line 62:
```suggestion
  tags: ["new:feature"],
  args: {
```

packages/sit-onyx/src/components/OnyxIcon/OnyxIcon.stories.ts
```diff
@@ -51,5 +51,22 @@ export const WithColor = {
   args: {
     ...Default.args,
     color: "success",
+    style: "--icon-size: 1em; vertical-align: middle;",
```
larsrickert, line 54: Is this change intended? Thats the same as setting inline to true

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxIcon/OnyxIcon.vue
JoCa96: good point, doesn't seem to make a difference for the use-cases I checked

packages/sit-onyx/src/components/OnyxIcon/OnyxIcon.stories.ts
JoCa96: nope, thanks

larsrickert (approved):
(optional) add screenshot test for this

Issue #2828, opened 2025-02-25
Fix missing sorting indicator after flyout selection


mj-hof:
No indicator is shown after choosing sorting in the flyout menu of OnyxDataGrid Column -> different to UX concept

![Image](https://github.com/user-attachments/assets/bece5087-52b5-427b-94b5-a4d8b22bd322)

PR #2862, opened 2025-02-28
Fix(OnyxDataGrid(sorting): fix missing sorting indicator after flyout selection

ChristianBusshoff:
Relates to #2828

- add sorting indicator


JoCa96:
please link the relevant ticket

packages/sit-onyx/src/components/OnyxDataGrid/features/index.ts
```diff
@@ -178,7 +178,7 @@ export type DataGridFeature<
      */
     actions?: (column: PublicNormalizedColumnConfig<TEntry>) => {
       iconComponent?: Component | { iconComponent: Component; position?: string };
```
JoCa96, line 180: Consider renaming and changing type:
Also please always add a tsdoc.
```suggestion
      iconComponent?: Component | 
        { 
          iconComponent: Component;
          /**
           * Will force the iconcomponent to be always shown in the header and not be put into the menu
           */
          alwaysShowInHeader?: boolean
        };
```

JoCa96 (approved):

Issue #2814, opened 2025-02-21
Fix OnyxDataGrid issues


larsrickert:
There are currently a few data grid bugs / inconsistencies:

1. null and undefined values are rendered as "null" / "undefined" but should be not rendered at all
  => should render as `-` like the number formatter currently does, #2910 
2. Feature behavior for setting `enabled` column config is inconsistent. E.g. for filtering, all columns are enabled by default when passing a config so individual columns are disabled. For sorting, all are disabled by default and only specified columns are enabled => enable by default for all columns, add a new prop to the config type to disable/enable for all by default. #2917

![Image](https://github.com/user-attachments/assets/14b845a3-409b-4f15-ba3b-cbef0c777179)

PR #2910, opened 2025-03-06
fix(OnyxDataGrid): improve string, number and fallback formatters

larsrickert:
Relates to #2814

- string/fallback formatter now renders "-" if undefined/null is passed. Also arrays are separated with comma and the array values are recursively formatted (undefined/null values are filtered out). Objects are formatted with `JSON.stringify()` instead of showing "[object Object]"
- number formatter: null is formatted as "-" instead of "0".

## Checklist


packages/sit-onyx/src/components/OnyxDataGrid/features/renderer.ts
```diff
@@ -7,13 +7,45 @@ import "./renderer.scss";
...
+
   const locale = injectI18n().locale;
   const formatter = new Intl.NumberFormat(locale.value);
 
   // We format the given value as Number. In case it renders as NaN, we replace it with `-`.
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any -- The typing is incorrect, the `format` method accepts any value
-  return formatter.format(value as any).replace("NaN", "-");
+  // The typing is incorrect, the `format` method accepts any value
+  return formatter
+    .format(value as Parameters<typeof formatter.format>[0])
```
JoCa96, line 31: If we want to use a complex type like this, then I would prefer to extract it into a explicit, named type.
Also we can expect the type of the Intl API to be very stable, so IMHO we can also just do:

```suggestion
    .format(value as number)
```
or

```suggestion
    .format(value as (number | bigint))
```

JoCa96 (approved):

Issue #2502, opened 2025-01-14
Implement column width adjustability


mj-hof:
### Why?

We want to offer freedom to the DataGrid/Simple Table users if the table view has fixed column width to adjust the width individually.

### Open questions / TODOs

- Is this feature only supported if the table shows vertical borders? No, should be independent
- can only be header border be dragged or also the cell border? Only the header border can be dragged

### Design

[Figma](https://www.figma.com/design/YfEUBOHk4J4nYrk04geswG/Onyx-Component-Library?node-id=4435-44962&t=6s316hl73ffVx8tm-4)

### Acceptance criteria

- [x] a new data grid feature is implemented for allowing adjusting the width
- [x] the column width can be changed by dragging the border
- [x] by default all columns are adjustable but they can be configured to enable/disable only specific columns
- [x] a resize cursor is shown when hovering the header

#### Out of scope

- [ ] double clicking the border will adjust the column width to max content width:
     - [ ] #2746 

### Implementation details

- if possible, prefer a CSS solution (e.g. [resize](https://developer.mozilla.org/en-US/docs/Web/CSS/resize) property

### Reference implementations

- [Element plus](https://element-plus.org/en-US/component/table.html#table-with-border)
- [PrimeVue](https://primevue.org/datatable/#column_resize)

### Definition of Done

- The following component requirements are implemented:
  - [ ] density
  - [ ] example usage added to `apps/demo-app/src/views/HomeView.vue`

- Should be covered by tests:
  - [ ] functional tests (Playwright or unit test)
  - [ ] visual tests (Playwright screenshots)

- Make sure, that
  - [x] follow-up tickets were created if necessary -> #2972 
  - [x] updated version + documentation is deployed
  - [x] Storybook can show the feature
  - [x] Storybook code snippet of new/changed examples are checked that they are generated correctly
  - [x] Namings are aligned with Figma

#### Approval

[Storybook](https://storybook.onyx.schwarz/?path=/story/data-datagrid-features--resizing)

- Checked and approved by
  - [x] designer 
  - [x] dev @larsrickert

larsrickert:
Approval feedback:

1. Can we use the new `type: "date"` for the birthday column of the Storybook example?
2. It might not be clear to the user that he can/should drag the column borders. Maybe we can add a little info text in the example?
3. Please use the new aligned API for enabling the feature, see #2917, so:
```ts
/**
 * The configuration options for the resizing feature in the OnyxDataGrid component.
[...]

PR #2939, opened 2025-03-12
feat(OnyxDataGrid): Implement column resizing

MajaZarkova:
Relates to #2502 

- Add new resizing feature for the OnyxDataGrid

## Checklist


packages/sit-onyx/src/components/OnyxDataGrid/features/resizing/resizing.ts
```diff
@@ -0,0 +1,97 @@
...
+          key: column.key,
+          type: column.key,
+          tdAttributes: {
+            class:
+              headerBeingResized.value !== null &&
+              Object.keys(headerBeingResized.value).includes(column.key as string)
+                ? "resize-border"
+                : "",
+          },
+        };
```
JoCa96, line 58: We have to be careful with not overwriting other column configuration, so we should merge as careful as possible:
```suggestion
        const tdAttributes = {
            class:
                headerBeingResized.value !== null &&
                Object.keys(headerBeingResized.value).includes(column.key as string)
                ? "resize-border"
                : "",
        };
        return {
          ...column,
          tdAttributes: mergeVueProps(tdAttributes, column.tdAttributes),
        };
```

packages/sit-onyx/src/components/OnyxDataGrid/features/resizing/resizing.ts
```diff
@@ -0,0 +1,97 @@
...
+          horizontalScrollOffset + ev.clientX - headerBeingResized.value[column.key].offsetLeft;
+
+        headerBeingResized.value[column.key].style.width = Math.max(min, width) + "px";
+      });
+
+    // Clean up event listeners, classes, etc.
+    const onMouseUp = () => {
+      window.removeEventListener("mousemove", onMouseMove);
+      window.removeEventListener("mouseup", onMouseUp);
+      headerBeingResized.value[column.key].classList.remove("header--being-resized");
```
JoCa96, line 28: Modifying html elements directly is error-prone, as Vue will not be aware of these changes and might overwrite them randomly.
We must interact with the DOM only via Vue's API.

The `window` eventlistener are fine like this, as the `window` object is never managed by Vue

To solve this issue here:
- in `initResize` store the column.key of the column being resized in a `ref`
- in `onMouseMove` store the wanted width per column in a reactive Map or object
- in `modifyColumns` add the `header--being-resized` via the thAttributes based on the ref column.key
- in `modifyColumns` set the wanted width of the column

JoCa96 (changes requested):

packages/sit-onyx/src/components/OnyxDataGrid/features/resizing/resizing.ts
```diff
@@ -0,0 +1,97 @@
...
+
+export const RESIZING_FEATURE = Symbol("Resizing");
+export const useResizing = createFeature(
+  <TEntry extends DataGridEntry>(options?: ResizingOptions<TEntry>) => {
+    let column: Readonly<InternalColumnConfig<TEntry>>;
+    const min = 70;
+    const headerBeingResized = ref<Record<PropertyKey, HTMLElement>>({});
+
+    const onMouseMove = (ev: { clientX: number }) =>
+      requestAnimationFrame(() => {
```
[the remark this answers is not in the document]
MajaZarkova: I tought It will allow a smoother movement. But I guess there is no difference, I'll remove it.

JoCa96 (approved):

JoCa96 (approved):

Issue #2890, opened 2025-03-05
Implement min height with scrollbar for OnyxFlyoutMenu


YifeiGS:
### Description

# What do we need?
I am currently working an internal Schwarz project using Onyx(v1.0.0-beta.92). In the project, we render a list of countries in the NavBar component. 
This list contains a large amount of items when user has certain roles. 
Currently, the list is rendered without a scrollbar and it breaks the look&feel user experience. 
 
<img width="369" alt="Image" src="https://github.com/user-attachments/assets/cb8986ad-92d6-40d5-8cbd-62e8be921891" />

### Confirmations

- [x] I have checked that there is not already a feature request for this feature.
- [x] I confirm that are no sensitive information included in this feature request. e.g. credentials, internal project details (for employees of the Schwarz Group) etc.

### Approvals

[...]

PR #2943, opened 2025-03-12
fix(OnyxFlyoutMenu): fix max-height and scrollable behavior for OnyxFlyoutMenu

JoCa96:
Relates to #2890 

- Limites the mumber of visible menu items in the `OnyxFlyoutMenu` to 8, as defined in Figma.
  - added tests for this behavior
- added test for hover behavior

### DIFF
--- a/packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxFlyoutMenu/OnyxFlyoutMenu.ct.tsx
+++ b/packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxFlyoutMenu/OnyxFlyoutMenu.ct.tsx
@@ -110,6 +110,10 @@ test("should display correctly and allow scrolling for many options", async ({ p
 
   // ACT
   await component.click();
+  // Only in Firefox and in the test: There is a weird bug where the menu doesn't initially have the correct height
+  // TODO: check if this "fix" is still necessary in later versions
+  await component.click();
+  await component.click();
 
   // ASSERT
   await expect(menu).toBeInViewport();
--- a/packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxFlyoutMenu/OnyxFlyoutMenu.ct.tsx
+++ b/packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxFlyoutMenu/OnyxFlyoutMenu.ct.tsx
@@ -1,6 +1,7 @@
 import { menuButtonTesting } from "@sit-onyx/headless/playwright";
 import { expect, test } from "../../../../playwright/a11y";
 import TestWrapperCt from "./TestWrapper.ct.vue";
+import TestWrapperManyCt from "./TestWrapperMany.ct.vue";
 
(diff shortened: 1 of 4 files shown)


packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxFlyoutMenu/OnyxFlyoutMenu.vue
```diff
@@ -99,17 +99,36 @@ const {
...
 
     &__wrapper {
       padding: 0;
+      /**
+       * The last option should only be half visible:
+       * 7.5 * OnyxListItem, where OnyxListItem => 2 * padding + line-height of OnyxListItem 
+       */
+      max-height: calc(
+        (var(--onyx-flyout-menu-visible-item-count, 7) + 0.5) *
+          (2 * var(--onyx-density-xs) + 1.5rem)
```
larsrickert, line 129:
```suggestion
          (2 * var(--onyx-density-xs) + 1lh)
```

Can't we use the `lh` unit here for the line height?

packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxFlyoutMenu/OnyxFlyoutMenu.vue
```diff
@@ -70,7 +68,9 @@ const {
         <slot name="options"></slot>
       </ul>
 
-      <slot name="footer"></slot>
+      <div class="onyx-flyout-menu__list-footer">
```
larsrickert, line 71: Do we still want to render the footer and header if they are empty? We could use a v-if here

packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxFlyoutMenu/OnyxFlyoutMenu.ct.tsx
```diff
@@ -65,3 +66,63 @@ test("should open on click", async ({ page, mount }) => {
     await expect(item).toBeEnabled();
   }
 });
+
+test("should open on hover", async ({ page, mount }) => {
+  // ARRANGE
+  const component = await mount(TestWrapperCt, {
+    props: { label: "Choose application language", trigger: "hover" },
+  });
+  const menu = page.locator("ul");
```
larsrickert, line 75:
```suggestion
  const menu = page.getByLabel("Choose application language")
```

Can't we use the label instead?

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxFlyoutMenu/OnyxFlyoutMenu.vue
JoCa96: we need them to render the padding/whitespace when they are empty, added a comment

larsrickert (approved):
Please add screenshot tests in either this or a follow up PR :)

Issue #2171, opened 2024-11-27
Implement|Fix|Define: Implement nested OnyxNavItems


JoCa96:
### Why?

- 

### Open questions / TODOs

<-- Add questions if there are any, use `@` to tag people. -->

### Implementation details

<-- Add technical details / implementation proposals if necessary, e.g.:
  - API proposal
  - relevant web APIs
  - learnings from other implementations
  - code snippets
  -->

### Definition of Done

- The following component requirements are implemented:
  - [ ] skeleton
  - [ ] density
  - [ ] example usage added to `apps/demo-app/src/views/HomeView.vue`

- covered by tests
  - [ ] functional tests (Playwright or unit test)
  - [ ] visual tests (Playwright screenshots)

- [ ] follow-up tickets were created if necessary - [ ] updated version + documentation is deployed - [ ] Storybook can show the feature - [ ] Storybook code snippet of new/changed examples are checked that they are generated correctly - [ ] Namings are aligned with Figma

#### Approval

- [Storybook]() - [Documentation]() - [ ] approved by designer  - [ ] approved by dev

larsrickert:
Please create a follow up with:

1. when clicking a nav item, is remains in an active state with some styles. Please check with UX if this is intended, I guess that was now the case previously

![Image](https://github.com/user-attachments/assets/fd64a4e4-525c-488b-abc2-d11be1f00ea6)

2. Not all top level items in the Storybook examples have links, should we add them so the user might not be confused?

PR #3072, opened 2025-04-02
docs(OnyxNavBar): updated storybook

JoCa96:
Relates to #2171 

- Updated NavBar story example and documentation
- fix `focus` state of `OnyxNavItem`

### DIFF
--- a/packages/sit-onyx/src/components/OnyxNavBar/OnyxNavBar.stories.ts
+++ b/packages/sit-onyx/src/components/OnyxNavBar/OnyxNavBar.stories.ts
@@ -85,7 +85,6 @@ export const Default = {
     appName: "App name",
     default: () => [
       h(OnyxNavItem, { label: "Router Link", link: "#router-link" }),
-      h(OnyxNavItem, { label: "Button", onClick: action("button clicked") }),
       h(
         OnyxNavItem,
         { label: "Nesting" },
--- a/packages/sit-onyx/src/components/OnyxNavBar/OnyxNavBar.stories.ts
+++ b/packages/sit-onyx/src/components/OnyxNavBar/OnyxNavBar.stories.ts
@@ -17,6 +17,19 @@ import { Default as OnyxUserMenuDefault } from "./modules/OnyxUserMenu/OnyxUserM
 import OnyxUserMenu from "./modules/OnyxUserMenu/OnyxUserMenu.vue";
 import OnyxNavBar from "./OnyxNavBar.vue";
 
+/**
+ * ## Structure
+ *
+ * ### Desktop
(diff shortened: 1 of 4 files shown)


packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxNavItem/OnyxNavItem.stories.ts
```diff
@@ -4,7 +4,9 @@ import OnyxBadge from "../../../OnyxBadge/OnyxBadge.vue";
 import OnyxNavItem from "./OnyxNavItem.vue";
 
 /**
- * The nav item is used to build the navigation bar component and not intended to be used standalone.
+ * The `OnyxNavItem` is used to build the navigation bar component and not intended to be used standalone.
+ * Based on the context and if the `link` property is provided, it will either be rendered as a button or anchor element.
+ * For rendering links, the [`OnyxRouterLink`](/docs/support-routerlink--docs) support component is used, which integrates with the provided router.
```
larsrickert, line 9: I wouldn't mention the OnyxRouterLink here because people might think that they need to place the router link inside the nav item. Maybe just mention that the nav item is integrated with the router for internal links?

packages/sit-onyx/src/components/OnyxNavBar/OnyxNavBar.stories.ts
```diff
@@ -71,19 +84,19 @@ export const Default = {
     logoUrl: "/onyx-logo.svg",
     appName: "App name",
     default: () => [
-      h(OnyxNavItem, { label: "Item 1", link: "https://it.schwarz/" }),
-      h(OnyxNavItem, { label: "Item 2", onClick: action("button clicked") }),
+      h(OnyxNavItem, { label: "Router Link", link: "#router-link" }),
+      h(OnyxNavItem, { label: "Button", onClick: action("button clicked") }),
```
larsrickert, line 88: Parent level items should always have a link. Buttons are only used if they have nested children and do not have a link themselves :)

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxNavBar/OnyxNavBar.stories.ts
JoCa96: Not necessarily, e.g. it could also open a dialog or trigger some action on the same route

packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxNavItem/OnyxNavItem.stories.ts
JoCa96: done

packages/sit-onyx/src/components/OnyxNavBar/OnyxNavBar.stories.ts
larsrickert: If I understood Jannick correctly thats not what nav items are for. They are mainly used for navigation. Actions are put into the context are like icon buttons etc.

larsrickert: If I understood Jannick correctly thats not what nav items are for. They are mainly used for navigation. Actions are put into the context are like icon buttons etc.

JoCa96: Yes, ideally and mainly 😅 But I removed the line. (Your comment here is totally valid, but please consider that approving and commenting is also an option for minor topics)

larsrickert (approved):

PR #3102, opened 2025-04-08
feat(OnyxTooltip): Implement popover api

ChristianBusshoff:
- added popover api
- added anchor()
- added fallback for browers which doesn't support anchors


packages/sit-onyx/src/components/OnyxTooltip/OnyxTooltip.vue
```diff
@@ -63,6 +67,93 @@ const { densityClass } = useDensity(props);
...
+        break;
+    }
+
+    tooltip.style.position = "absolute"; // wichtig
+    tooltip.style.top = `${top}px`;
+    tooltip.style.left = `${left}px`;
+  }
+};
+const checkAnchorSupport = () => {
+  supportsAnchor.value = CSS.supports("anchor-name: --test");
```
JoCa96, line 142: The result of this function is (globally) static and synchronous (the result is returned immediately, e.g. no `await` is necessary).

Therefore:
1. Because it is **static** and will **not change**, there is no need to use a *Ref*. It can be directly assigned as a `const`. The naming convention for constants like this is *ALL_UPPERCASE*.
2. Because it is **globally** static we should move this constant as far up (outside of nested functions/scopes and potentially outside of components) as possible.

packages/sit-onyx/src/composables/useTooltipPositioningWithoutAnchorSupport.ts
```diff
@@ -0,0 +1,84 @@
...
+  toolTipPosition: ComputedRef<TooltipPosition>,
+  alignment: ComputedRef<WedgePosition>,
+  alignsWithEdge: ComputedRef<boolean>,
+  fitParent: ComputedRef<boolean>,
+) => {
+  const wedgeSize = 8;
+  const leftPosition = ref("-1000px");
+  const topPosition = ref("-1000px");
+
+  const updateTooltipPositioningWithoutAnchorSupport = () => {
```
JoCa96, line 17: Unfortunately the calculation doesn't consider scrolled elements/pages into account. Can be done in a follow-up :)

packages/sit-onyx/src/composables/useTooltipPositioningWithoutAnchorSupport.ts
```diff
@@ -0,0 +1,84 @@
+import type { TooltipPosition } from "src/components/OnyxTooltip/types";
+import { ref, type ComputedRef, type ShallowRef } from "vue";
+import type { WedgePosition } from "./useWedgePosition";
+
+export const useHandleTooltipPositioningWithoutAnchorSupport = (
```
JoCa96, line 5: We can make it a bit more generic and call it `useAnchorPositionPolyfill`:
- rename `tooltipRef` to `positionedRef`
- rename `wrapperRef` to `targetRef`
- rename `toolTipPosition` to `positionArea`
- make `wedgeSize` configurable and call it `offset`

packages/sit-onyx/src/composables/useAnchorPositionPolyfill.ts
```diff
@@ -0,0 +1,85 @@
...
+
+export const useAnchorPositionPolyfill = (
+  positionedRef: Readonly<ShallowRef<HTMLElement | null>>,
+  targetRef: Readonly<ShallowRef<HTMLElement | null>>,
+  positionArea: ComputedRef<TooltipPosition>,
+  alignment: ComputedRef<WedgePosition>,
+  alignsWithEdge: ComputedRef<boolean>,
+  fitParent: ComputedRef<boolean>,
+  offset: number,
+) => {
```
JoCa96, line 13: (optional) to make the API more dev-friendly consider using an object as parameter and using `MaybeRefOrGetter`:

```suggestion
type UseAnchorPositionPolyfillOptions = {
  positionedRef: MaybeRefOrGetter<HTMLElement | null>;
  targetRef: MaybeRefOrGetter<HTMLElement | null>;
  positionArea: MaybeRefOrGetter<TooltipPosition>;
  alignment: MaybeRefOrGetter<WedgePosition>;
  alignWithEdge: MaybeRefOrGetter<boolean>;
  fitParent: MaybeRefOrGetter<boolean>;
  offset: MaybeRefOrGetter<number>;
};

export const useAnchorPositionPolyfill = ({
    positionedRef,
    targetRef,
    positionArea,
    alignment,
    alignWithEdge,
    fitParent,
    offset,
}: UseAnchorPositionPolyfillOptions) => {
```

JoCa96 (approved):

PR #3103, opened 2025-04-08
fix(OnyxDataGrid): fix sorting buttons of `useSorting` feature using column key instead of label 

JoCa96:
some minor changes:
- column `label` is now resolved as part of the normalization
- `header.actions` now uses the available `InternalColumnConfig` type instead of the less accurate `PublicNormalizedColumnConfig`
- OnyxDataGrid: The column label is now used instead of the key for the sorting buttons

### DIFF
--- a/packages/sit-onyx/src/components/OnyxDataGrid/features/index.ts
+++ b/packages/sit-onyx/src/components/OnyxDataGrid/features/index.ts
@@ -80,7 +80,7 @@ export type InternalColumnConfig<
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   TColumnGroup extends ColumnGroupConfig = any,
   TTypes extends PropertyKey = PropertyKey,
-> = {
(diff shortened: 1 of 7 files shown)


packages/sit-onyx/src/components/OnyxDataGrid/features/index.ts
```diff
@@ -89,7 +89,8 @@ export type InternalColumnConfig<
    * Attributes that should be set on all `th` elements
    */
   thAttributes?: ThHTMLAttributes;
-} & PublicNormalizedColumnConfig<TEntry, TColumnGroup, TTypes>;
+  label: string;
```
larsrickert, line 92: Is this new required prop user-facing or only internal?

JoCa96: added a jsdoc :)

PR #3177, opened 2025-04-23
FoUp(OnyxToolTip): add popover API

ChristianBusshoff:
added posistion updating on scroll.

### DIFF
--- a/packages/sit-onyx/src/composables/useAnchorPositionPolyfill.ts
+++ b/packages/sit-onyx/src/composables/useAnchorPositionPolyfill.ts
@@ -32,7 +32,6 @@ export const useAnchorPositionPolyfill = ({
   const topPosition = ref("-1000px");
 
   const { isIntersecting: targetVisible } = useIntersectionObserver(targetRef);
-  const { isIntersecting: positionedVisible } = useIntersectionObserver(positionedRef);
 
   const updateAnchorPositionPolyfill = () => {
     const positionedEl = getTemplateRefElement(positionedRef.value);
@@ -99,20 +98,9 @@ export const useAnchorPositionPolyfill = ({
   };
 
   watchEffect(() => {
-    if (targetVisible.value) {
-      const positionedEl = positionedRef.value;
-      if (positionedEl) {
-        window.addEventListener("scroll", updateAnchorPositionPolyfill, true);
-      }
+    if (targetVisible.value && positionedRef.value) {
+      window.addEventListener("scroll", updateAnchorPositionPolyfill, true);
     } else {
-      window.removeEventListener("scroll", updateAnchorPositionPolyfill, true);
-      leftPosition.value = "-1000px";
-      topPosition.value = "-1000px";
-    }
-  });
-
-  watchEffect(() => {
-    if (!positionedVisible.value) {
       window.removeEventListener("scroll", updateAnchorPositionPolyfill, true);
       leftPosition.value = "-1000px";
       topPosition.value = "-1000px";
--- a/packages/sit-onyx/src/components/OnyxTooltip/OnyxTooltip.vue
+++ b/packages/sit-onyx/src/components/OnyxTooltip/OnyxTooltip.vue
@@ -232,7 +232,7 @@ $wedge-size: 0.5rem;
 
 .onyx-tooltip {
   @include layers.component() {
-    position: absolute;
+    position: fixed;
     min-width: var(--onyx-spacing-3xl);
     width: max-content;
     max-width: 19rem;


packages/sit-onyx/src/composables/useAnchorPositionPolyfill.ts
```diff
@@ -27,11 +30,62 @@ export const useAnchorPositionPolyfill = ({
...
+      } else {
+        scrollParent.value?.removeEventListener("scroll", updateAnchorPositionPolyfill);
+        leftPosition.value = "-1000px";
+        topPosition.value = "-1000px";
+      }
+    });
+  });
+
   const getElement = (refOrGetter: MaybeRefOrGetter<HTMLElement | null>): HTMLElement | null => {
     const element = unref(refOrGetter);
```
JoCa96, line 68: Use [`toValue`](https://vuejs.org/api/reactivity-utilities.html#tovalue) instead of `unref` for resolving `MaybeRefOrGetter` types. This will also handle getters.
```suggestion
    const element = toValue(refOrGetter);
```

packages/sit-onyx/src/composables/useAnchorPositionPolyfill.ts
```diff
@@ -27,11 +30,62 @@ export const useAnchorPositionPolyfill = ({
   const leftPosition = ref("-1000px");
   const topPosition = ref("-1000px");
 
+  let targetVisible = false;
+  const scrollParent = computed(() => getScrollParent(getElement(targetRef)));
```
JoCa96, line 34: Unfortunately there are two issues with this approach: 
1. If an element has multiple scroll-parents only the first parent will be considered (so e.g. when the whole page is scrolled this might be missed)
2. We would need an event-listener for every scroll parent, because scroll events don't bubble.

Fortunately there is an easy solution:
Even though not all events bubble, all events do have a [capture phase](https://developer.mozilla.org/en-US/docs/Web/API/Event/eventPhase#event.capturing_phase).
We can use this to have a single, global event listener that is informed of all scroll events.

packages/sit-onyx/src/composables/useAnchorPositionPolyfill.ts
```diff
@@ -27,11 +30,62 @@ export const useAnchorPositionPolyfill = ({
...
+
+  const targetObserver = new IntersectionObserver((entries) => {
```
JoCa96, line 36: *Nice-to-have:* It would be nice to have an `useIntersectionObserver`, which works similar to the `useResizeObserver`.

packages/sit-onyx/src/composables/useAnchorPositionPolyfill.ts
```diff
@@ -27,11 +30,62 @@ export const useAnchorPositionPolyfill = ({
...
+      return node;
+    } else {
+      return getScrollParent(node.parentElement);
+    }
+  };
+
+  nextTick(() => {
+    const observerTarget = getElement(targetRef);
+    if (observerTarget) targetObserver.observe(observerTarget);
+  });
```
JoCa96, line 87: `nextTick` is always the last resort, for this use `watch` or `watchEffect` instead. (similar to [watch](https://github.com/SchwarzIT/onyx/blob/30b1d1ec4e17ec78e1e74f7503514a2d8f90feb0/packages/sit-onyx/src/composables/useResizeObserver.ts#L49))

packages/sit-onyx/src/composables/useAnchorPositionPolyfill.ts
```diff
@@ -27,11 +30,62 @@ export const useAnchorPositionPolyfill = ({
   const leftPosition = ref("-1000px");
   const topPosition = ref("-1000px");
```
JoCa96, line 32: In the polyfill we also need to consider the `unmounted` lifecycle hook.

packages/sit-onyx/src/composables/useAnchorPositionPolyfill.spec.ts
```diff
@@ -11,7 +11,21 @@ describe("useAnchorPositionPolyfill", () => {
   const alignment = ref<WedgePosition>("center");
   const alignsWithEdge = ref(false);
   const fitParent = ref(false);
-  const offset = ref(10);
+
+  beforeEach(() => {
```
JoCa96, line 15: As we really heavily on a real DOM for this composable, it might be worth to consider testing this in a real browser using playwright.

JoCa96 (changes requested):

JoCa96 (approved):

PR #3191, opened 2025-04-24
Joca96/fix template ref warning issue

JoCa96:
- added no-shadow-template-ref rule
- fix warning from more list
- added test:eslint-plugins to `test:all` script, didn't seem to be run before in the CI


packages/eslint-plugin/package.json
```diff
@@ -14,6 +14,8 @@
     "eslint": ">= 8"
   },
   "devDependencies": {
-    "@types/eslint": "^9.6.1"
+    "@eslint/core": "^0.13.0",
```
larsrickert, line 17:
```suggestion
    "@eslint/core": "~0.13.0",
```

Should we use a stricter scope since its a 0.x.x version?

larsrickert (approved):

packages/eslint-plugin/package.json
JoCa96: the [caret range indicator](https://github.com/npm/node-semver?tab=readme-ov-file#caret-ranges-123-025-004) only allows patch-level udpates, so we should be fine here:

> ### Caret Ranges ^1.2.3 ^0.2.5 ^0.0.4
> Allows changes that do not modify the left-most non-zero element in the [major, minor, patch] tuple. In other words, this allows patch and minor updates for versions 1.0.0 and above, patch updates for versions 0.X >=0.1.0, and no updates for versions 0.0.X.

larsrickert: Yes but the caret allows minor updates like 0.1.0 to 0.2.0.
For zero-based versions, the middle number is considered to allow breaking changes so when using `~` to only allow 0.0.x updates, we should be safer for breaking changes

PR #3193, opened 2025-04-24
fix(OnyxTab): prevent call stack size exceed error in Nuxt

### LINKED ISSUE #3190: [Bug]: `OnyxTabs` causing stack overflow if using with nuxt
### Version

1.0.0-beta.193

### Description

adding `OnyxTabs` component in nuxt based application throws `Maximum call stack size exceeded` error 

Image

### Reproduction steps

- run example 
- check console
- try to reload page multiple times
- to verify that `OnyxTabs` causing an issue replace component with `<div>test</div>` and reload the page

https://stackblitz.com/edit/onyx-tabs-nuxt?file=app.vue

### Confirmations

larsrickert:
closes #3190

Fix maximum call stack size exceed error in Nuxt which is caused by the `<Teleport>`.

## Checklist

### DIFF
--- a/.changeset/late-mangos-greet.md
+++ b/.changeset/late-mangos-greet.md
@@ -1,0 +1,5 @@
+---
+"sit-onyx": patch
+---
+
+fix(OnyxTab): prevent call stack size exceed error in Nuxt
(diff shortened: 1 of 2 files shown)


JoCa96 (approved):

JoCa96:
@larsrickert do you know why `disabled` and `v-if` behaves differently on a teleport?

larsrickert:
> @larsrickert do you know why `disabled` and `v-if` behaves differently on a teleport?

- disabled will still render the component but in-place instead of teleporting it
- v-if will not render the content at all

But tbh I don't know why disabled does not work here. I guess it might be some Nuxt-specific detail that somehow the component can not be renderer inline or something. So using v-if to not render it at all fixes the bug

Issue #2228, opened 2024-12-04
Implement more list into nav bar


larsrickert:
### Why?

We want to integrate the existing [OnyxMoreList](https://storybook.onyx.schwarz/?path=/docs/support-morelist--docs) component in the nav bar so that a "more" nav item is shown if not all nav items fit into the available width so the user can still access all nav items.

### Depends on

- #2171

### Design

Same as OnyxNavButton

### Acceptance criteria

- [x] if not all nav items fit into the available width, remaining items should be grouped in a e.g. " +3more" nav item that shows the remaining nav items
- [ ] ~if hidden elements contain children, they should still be shown in the more indicator~ FoUp: #2997
- [x] custom nav item content (.e.g badges) are still visible in the more indicator
- [x] `mobileBreakpoint` property of the OnyxNavBar has been updated to allow boolean seetings

*Known Bugs:*
- Fix MoreList in OnyxNavBar not considering original order
[#3205](https://github.com/SchwarzIT/onyx/issues/3205)

### Definition of Done

- Should be covered by tests:
  - [x] functional tests (Playwright or unit test)
  - [x] visual tests (Playwright screenshots)

- Make sure, that
  - [x] follow-up tickets were created if necessary
  - [x] updated version + documentation is deployed
  - [x] Storybook can show the feature
  - [x] Storybook code snippet of new/changed examples are checked that they are generated correctly
  - [x] Namings are aligned with Figma

#### Approval
- [Storybook](https://storybook.onyx.schwarz/?path=/story/navigation-navbar--with-more-list-item) 
- [x] Checked and approved by designer 
- [x] Checked and approved by dev @larsrickert

PR #3196, opened 2025-04-24
feat(OnyxNavBar): update mobileBreakpoint prop

JoCa96:
Relates to #2228 

- rename prop `mobileBreakpoint` to `mobile`: behaves as before, but also supports boolean type
- renamed so the naming matches the new behaviour better

### DIFF
--- a/packages/sit-onyx/src/components/OnyxNavBar/OnyxNavBar.vue
+++ b/packages/sit-onyx/src/components/OnyxNavBar/OnyxNavBar.vue
@@ -22,7 +22,7 @@ import {
 } from "./types";
 
 const props = withDefaults(defineProps<OnyxNavBarProps>(), {
-  isMobile: "sm",
+  mobile: "sm",
 });
 
 const emit = defineEmits<{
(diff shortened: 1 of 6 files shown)


packages/sit-onyx/src/components/OnyxNavBar/OnyxNavBar.vue
```diff
@@ -71,15 +71,20 @@ const {
 const isBurgerOpen = ref(false);
 const isContextOpen = ref(false);
 
-const isMobile = computed(() => {
-  const mobileWidth =
-    typeof props.mobileBreakpoint === "number"
-      ? props.mobileBreakpoint
-      : ONYX_BREAKPOINTS[props.mobileBreakpoint];
-  return width.value !== 0 && width.value < mobileWidth;
+const isMobileWidth = (mobileWidth: number) => width.value !== 0 && width.value < mobileWidth;
```
larsrickert, line 74: The reactive `width` is used here so this should be a computed, otherwise it will not be re-computed when the width changes :)

larsrickert (changes requested):

Issue #3093, opened 2025-04-08
Implement more actions and icon for OnyxNotificationCard


larsrickert:
### Why?

We want to extend our implementation of the OnyxNotificationCard to support a slot for showing more actions inside a "..." flyout menu as well as an optional icon next to the headline.

### Depends on

- #3087

### Design

[Figma (Icon)](https://www.figma.com/design/ingOjglfzscsp6zaUAg2gV/Touchpoint-Template?node-id=150-6844&m=dev)
[Figma (flyout menu)](https://www.figma.com/design/ingOjglfzscsp6zaUAg2gV/Touchpoint-Template?node-id=150-7118&m=dev)

### Acceptance criteria

- [x] a slot is provided where a custom flyout menu with more actions can be provided by the user
- [x] the more actions are only visible on hover/keyboard focus
- [x] a new `icon` property is supported where an arbitrary icon can be displayed

### Definition of Done

- The following component requirements are implemented:
  - [x] density

- Should be covered by tests:
  - [x] functional tests (Playwright or unit test)
  - [x] visual tests (Playwright screenshots)

- Make sure, that
  - [x] updated version + documentation is deployed
  - [x] Storybook can show the feature
  - [x] Storybook code snippet of new/changed examples are checked that they are generated correctly
  - [x] Namings are aligned with Figma

#### Approval

[Storybook](https://storybook.onyx.schwarz/?path=/story/notifications-notificationcard--actions)

- Checked and approved by
  - [x] designer @jannick-ux 
  - [x] dev @ChristianBusshoff

PR #3209, opened 2025-04-28
fix(OnyxNotificationCard): always show more actions if device does not support hover

larsrickert:
Relates to #3093

Ticket approval feedback from Jannick and Flo:
- show background color hover effect
- always show more actions if the users devices does not support hover (e.g. touch devices)

## Checklist


packages/sit-onyx/src/components/OnyxNotificationCard/OnyxNotificationCard.vue
```diff
@@ -141,9 +146,13 @@ const { timeAgo } = useRelativeTimeFormat({
...
-      .onyx-notification-card__more-actions {
-        @include visibility.visually-hidden();
+    // for devices that support hover, we are only showing the more actions on hover
+    // for all other devices (like mobile devices), they are always shown
+    @media (hover: hover) {
+      &:not(&:hover, &:focus-within) {
+        .onyx-notification-card__more-actions {
+          @include visibility.visually-hidden();
+        }
       }
```
JoCa96, line 156: I think we should be fine to just use opacity here, might make the CSS a bit simpler:

```suggestion
    @media (hover: hover) {
        opacity: 0;
        
        &:hover, &:focus-within {
          opacity: 1;
        }
      }
```

JoCa96 (approved):

packages/sit-onyx/src/components/OnyxNotificationCard/OnyxNotificationCard.vue
larsrickert: If using opacity, the space is reserved which will lead to an unintended gap when there is also the "unread" badge dot visible so I decided to use visually-hidden here so its visually hidden but still focusable with keyboard

Issue #2726, opened 2025-02-11
Implement basic OnyxSidebar


JoCa96:
### Why?

As a designer, I want to be able to use a sidebar to provide additional navigation and other options.

The goal of this ticket is to implement the Sidebar as a container for content. 

### Design

[Figma Sidebar](https://www.figma.com/design/EBi98T51T9fGfjrYYfsulg/onyx---Archive?node-id=4169-89075&m=dev)
[Figma Sidebar Content](https://www.figma.com/design/YfEUBOHk4J4nYrk04geswG/Onyx-Component-Library?m=auto&node-id=10152-53976)

### Acceptance criteria

- [x] custom content can be passed in using a `header`, `body` or `footer` slot
- [x] the OnyxPageLayout supports slots for positioning the sidebar left and right
- [x] optionally, the sidebar can be adjusted by dragging the right border
- the sidebar can
  - [x] either be placed directly into the page ("parallel"), where it shrinks the page grid
  - [x] or act as an "overlay"
- [x] when used as overlay, the backdrop can optionally be removed

### Implementation details

- The `OnyxPageLayout` needs to be updated

### Applicable ARIA Pattern

- https://www.w3.org/WAI/ARIA/apg/patterns/landmarks/examples/complementary.html
- https://www.w3.org/WAI/ARIA/apg/patterns/landmarks/

### Definition of Done

- The following component requirements are implemented:
  - [x] density
  - [x] example usage added to `apps/demo-app/src/views/HomeView.vue`

- Should be covered by tests:
  - [x] functional tests (Playwright or unit test)
  - [x] visual tests (Playwright screenshots)

- Make sure, that
  - [x] follow-up tickets were created if necessary
    - #3340
  - [x] updated version + documentation is deployed
  - [x] Storybook can show the feature
  - [x] Storybook code snippet of new/changed examples are checked that they are generated correctly
  - [x] Namings are aligned with Figma

#### Approval

[Storybook](https://storybook.onyx.schwarz/?path=/docs/navigation-sidebar--docs)

- Checked and approved by
  - [x] designer @jannick-ux 
  - [x] dev @ChristianBusshoff

PR #3282, opened 2025-05-08
feat(OnyxSidebar): implement `resizable` property

larsrickert:
Relates to #2726

- Support sidebar to be resizable by the user
- create a new OnyxResizeHandle support component that is also used in the data grid to unify the resize behavior

## Checklist


packages/sit-onyx/src/components/OnyxResizeHandle/OnyxResizeHandle.vue
```diff
@@ -0,0 +1,149 @@
...
+  if (event.key !== "Escape") return;
+  if (previousWidth.value) emit("updateWidth", previousWidth.value);
+  abort();
+};
+</script>
+
+<template>
+  <button
+    tabindex="-1"
+    type="button"
```
JoCa96, line 93: `presentation` will make this element behave like text content, which makes it inaccessible.
Interestingly, our use-case is described in the `separator` role: 
- https://w3c.github.io/aria/#separator
- https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/separator_role 

So, would be nice to have this in our headless patterns.

```suggestion
```

packages/sit-onyx/src/components/OnyxResizeHandle/OnyxResizeHandle.vue
```diff
@@ -0,0 +1,149 @@
...
+    role="presentation"
+    :class="['onyx-component', 'onyx-resize-handle', isActive ? 'onyx-resize-handle--active' : '']"
+    :aria-label="t('resizeHandle.label')"
+    @mousedown="handleMousedown"
+    @dblclick="handleDoubleClick"
+  ></button>
```
JoCa96, line 99: `OnyxVisuallyHidden` should be able to work here, too and would be preferable over an `aria-label`:
```suggestion
    @mousedown="handleMousedown"
    @dblclick="handleDoubleClick"
  ><OnyxVisuallyHidden>{{ t('resizeHandle.label') }}</OnyxVisuallyHiddenProps></button>
```

packages/sit-onyx/src/components/OnyxResizeHandle/OnyxResizeHandle.vue
```diff
@@ -0,0 +1,149 @@
...
+  );
+});
+
+const handleDoubleClick = () => {
+  emit("autoSize");
+};
+
+const onMouseMove = (event: MouseEvent) => {
+  if (!currentElement.value || !isActive.value) return;
+  const width = event.clientX - currentElement.value.getBoundingClientRect().left;
```
JoCa96, line 72: This will only work for resizing on the right side of an element, correct? Which is fine, but please mention this in the documentation.

JoCa96 (changes requested):

packages/sit-onyx/src/components/OnyxResizeHandle/OnyxResizeHandle.vue
JoCa96: Is also fine to do it an follow-up PR or ticket :)

larsrickert: I created a follow up ticket for this: #3312

JoCa96 (approved):

Issue #2783, opened 2025-02-19
Implement support for custom options of typeRenderers for OnyxDataGrid


JoCa96:
### Why?

Follow-up from #2370:
Features are able to define custom `typeRenderer` which expose custom cell types for columns.
The `columns` configuration should also be able to pass typed configuration to the specific `typeRenderer`.
The configuration type must be defined by the individual typeRenderer.

e.g.:

```ts
columns: [
      // options should be typed correctly
      { key: "age", type: "number", options: { formatter: new Intl.NumberFormat() } },  // no error
      { key: "age", type: "number", options: { formatter: "string for some reason" } },  // should give ts error
    ],
```

### Depends on

- #2370

### Acceptance criteria

- [x] The return type of `createFeature` allows the `OnyxDataGrid` to infer the type of the options for the column `type` in `OnyxDataGrid`s `columns` prop
- [x] This is used to allow passing in of a custom `Intl.NumberFormat` for the `number` type
- [x] Nice to have: Check if its easily possible to use this API: ` { key: "age", type: { name: "number", formatter: new Intl.NumberFormatter() } }`

### Implementation details
Either the type is inferred from the Component Props or we need to add "virtual" properties using a helper function:

```ts
type NumberRendererProps = {
  formatter?: Intl.NumberFormat;
}
```

```ts
typeRenderer: {
  myTypeRenderer: {
// Can we infer the Prop type and use it?
    component: (props: NumberRendererProps) => {}
  }
}
```

```ts
typeRenderer: {
  myTypeRenderer: createTypeRenderer<NumberRendererProps>({
    component: (props) => {}
  })
}
```

### Definition of Done

- Should be covered by tests:
  - [x] functional tests (Playwright or unit test)
  - [x] visual tests (Playwright screenshots)

- Make sure, that
  - [x] follow-up tickets were created if necessary
  - [x] updated version + documentation is deployed
  - [x] Storybook can show the feature
  - [x] Storybook code snippet of new/changed examples are checked that they are generated correctly
  - [x] Namings are aligned with Figma
  - [x] Storybook sidebar badge for "New" is added on component or story level if needed. See: https://github.com/Sidnioulz/storybook-addon-tag-badges/?tab=readme-ov-file

#### Approval

[Storybook]()
[Documentation]()

- Checked and approved by
  - [ ] designer 
  - [ ] dev

PR #3284, opened 2025-05-08
feat(OnyxDataGrid): Support for column type options

JoCa96:
Relates to #2783


packages/sit-onyx/src/components/OnyxDataGrid/features/index.ts
```diff
@@ -49,15 +64,17 @@ export type TypeRenderer<TEntry extends DataGridEntry> = {
...
 
 /**
  * ColumnConfig as it can be defined by the user.
  */
 export type ColumnConfig<
   TEntry extends DataGridEntry,
-  TColumnGroup extends ColumnGroupConfig,
-  TTypes,
+  TColumnGroup extends ColumnGroupConfig = Record<string, never>,
+  TTypes extends ColumnConfigTypeOption<PropertyKey, unknown> = RenderTypesFromFeature<
```
larsrickert, line 75: Can we merge this type with the default types? Otherwise, when defining custom types like in our "CustomColumnTypes" example, all the default types will be removed
image

packages/sit-onyx/src/components/OnyxDataGrid/features/renderer.ts
```diff
@@ -19,95 +45,119 @@ const numberFormatter = <TEntry extends DataGridEntry>(
...
+  fallback?: string;
+};
+
+export const stringFormatter = <TEntry extends DataGridEntry>(
   value: TEntry[keyof TEntry] | undefined,
+  opts?: StringCellOptions,
 ): string => {
   // using loose "==" here to catch both undefined and null
-  if (value == undefined) return FALLBACK_RENDER_VALUE;
+  if (value == undefined) return opts?.fallback ?? FALLBACK_RENDER_VALUE;
```
larsrickert, line 79: Can/should we somehow unify this fallback logic? Currently its duplicated for every type renderer

packages/sit-onyx/src/components/OnyxDataGrid/features/base/base.ts
larsrickert, line 1: Would prefer naming this something like "TypeRendererFeature" or do you plan to add also other stuff to this in the future?

larsrickert (changes requested):

larsrickert (changes requested):
Also please add a changeset :)

packages/sit-onyx/src/components/OnyxDataGrid/features/renderer.ts
```diff
@@ -19,95 +45,119 @@ const numberFormatter = <TEntry extends DataGridEntry>(
...
 };
 
-const NUMBER_RENDERER = Object.freeze({
+export const DATE_RENDERER = createTypeRenderer<DateCellOptions>({
+  header: { component: HeaderCell },
   cell: {
-    component: (props) => numberFormatter(props.modelValue),
-    tdAttributes: { class: "onyx-data-grid-number-cell" },
+    component: (props) =>
+      dateFormatter(props.modelValue, { format: "date", ...props.metadata?.typeOptions }),
```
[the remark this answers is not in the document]
JoCa96: Not sure if I understand you... If the user provides custom typeOptions they are used, otherwise `format: "date"` is the default..?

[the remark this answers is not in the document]
larsrickert: I mean there are other options beside format which behave differently so basically my suggestion is:

```ts
dateFormatter(props.modelValue, props.metadata?.typeOptions ?? { format: "date" })
```

So that our default `format: "date"` is only applied if the user does not provide any options

packages/sit-onyx/src/components/OnyxDataGrid/features/base/base.ts
JoCa96: Yes, I would like to use it for some built-in functionalities, "eat-your-own-dog-food" style :)

packages/sit-onyx/src/components/OnyxDataGrid/features/renderer.ts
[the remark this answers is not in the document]
larsrickert: Lets keep it like this for now then

larsrickert (approved):

PR #3315, opened 2025-05-13
Fixed OnyxTooltip showing outside the dialog.

ChristianBusshoff:
removed translate from OnyxDialog to not interfere with the anchor of the tooltip.
#3311


packages/sit-onyx/src/components/OnyxDialog/OnyxDialog.vue
```diff
@@ -103,13 +104,18 @@ watch(
...
       background-color: var(--onyx-color-component-opacity-backdrop);
     }
+    margin: auto;
+    align-self: anchor-center;
+    &--dont-support-anchor {
+      top: 50%;
+      left: 50%;
+      transform: translate(-50%, -50%);
+      margin: 0;
+    }
```
JoCa96, line 119: we should keep things simple and use an approach that doesn't require the anchor API. There is this cool article, which describes how this can be done: https://levelup.gitconnected.com/dont-use-transform-to-center-element-b378587ad1db


```suggestion
    position: fixed;
    top: 0;
    left: 0;
    right: 0;
    bottom: 0;
    margin: auto;
    width: fit-content;
    height: fit-content;
```

JoCa96 (changes requested):

JoCa96:
Please 
- add a changeset
- link the ticket in the PR

PR #3345, opened 2025-05-15
feat(OnyxDrawer): add new `modal` property

larsrickert:
Relates to #2726

Support disabling the backdrop for the drawer (which is also used in the sidebar) so you can create a drawer or sidebar without the backdrop

## Checklist


packages/sit-onyx/src/components/OnyxDrawer/OnyxDrawer.vue
```diff
@@ -35,7 +36,12 @@ const slots = defineSlots<{
 </script>
 
 <template>
-  <OnyxModalDialog class="onyx-drawer" v-bind="props" @close="emit('close')">
+  <OnyxModalDialog
+    class="onyx-drawer"
+    v-bind="props"
+    :modal="!props.noBackdrop"
```
JoCa96, line 42: Should we consider calling the prop `modal` (and invert the default), so we have less differences and it is clear to the dev, what behavior is here?

At minimum I would like it to be mentioned/described in the tsdoc.

JoCa96 (approved):

Issue #2701, opened 2025-02-05
Implement|Fix|Define Implement flag library


larsrickert:
### Why?

We want to offer flags for onyx that can be used to visually showcase countries.

### Depends on

- #405 

### Design

[Figma](https://www.figma.com/design/YfEUBOHk4J4nYrk04geswG/Onyx-Component-Library?node-id=5817-991&p=f&t=y2eCryzM1386ANyT-0)

### Acceptance criteria

- [ ] a new `@sit-onyx/flags` library is released that contains
    - [ ] flags as SVG
    - [ ] metadata like continent, country code, country name
- [ ] the flags are imported from Figma via our [Figma utils CLI](https://onyx.schwarz/development/packages/figma-utils.html)
- [ ] a pipeline is run daily to check and update the flags from Figma + auto generated changelog and versioning
- [ ] our docs are updated to showcase all flags

### Implementation details

- we can propably re-use our [icon import pipeline/code](https://github.com/SchwarzIT/onyx/blob/main/.github/workflows/import-figma-icons.yml)
- the API should be the same as for our icons so they can be used mixed / in all components that already support icons

### Definition of Done

- Should be covered by tests:
  - [ ] functional tests (Playwright or unit test)
  - [ ] visual tests (Playwright screenshots)

- Make sure, that
  - [ ] follow-up tickets were created if necessary
  - [ ] updated version + documentation is deployed
  - [ ] Storybook can show the feature
  - [ ] Storybook code snippet of new/changed examples are checked that they are generated correctly
  - [ ] Namings are aligned with Figma

#### Approval

[Storybook]()
[Documentation]()

- Checked and approved by
  - [ ] designer 
  - [ ] dev

PR #3346, opened 2025-05-15
feat: release `@sit-onyx/flags` npm package for flag library

larsrickert:
Relates to #2701

Create a new npm package `@sit-onyx/flags` for our flag library. The flags are imported from Figma (the same way as the icons).

I will update the docs to showcase all flags (similar to icons) in a follow up PR.

## Checklist


packages/figma-utils/src/commands/import-flags.ts
```diff
@@ -0,0 +1,78 @@
...
+    options.fileKey,
+    parsedFlags.map(({ id }) => id),
+    options.token,
+  );
+
+  console.log("Optimizing and writing flag files...");
+
+  if (!(await isDirectory(outputDirectory))) {
+    await mkdir(outputDirectory, { recursive: true });
+  }
```
JoCa96, line 62: When the `recursive` flag is used, `mkdir` will not throw an error when the directory already exists.
```suggestion
  await mkdir(outputDirectory, { recursive: true });
```

.github/workflows/import-figma-icons.yml
JoCa96, line 1: We should consider to rename the workflow file also, e.g.: `import-figma-assets`

JoCa96 (approved):

PR #3392, opened 2025-05-20
chore: add typerenderer custom option Story example

JoCa96:
Relates to #2783 

- added usage example for `createTypeRenderer`
- added example for base type renderer options

### DIFF
--- a/packages/sit-onyx/src/components/OnyxDataGrid/examples/CustomColumnTypes.vue
+++ b/packages/sit-onyx/src/components/OnyxDataGrid/examples/CustomColumnTypes.vue
@@ -1,24 +1,16 @@
 <script setup lang="ts">
 import forward from "@sit-onyx/icons/forward.svg?raw";
 import { h } from "vue";
-import {
-  createFeature,
-  OnyxDataGrid,
-  OnyxSystemButton,
-  type ColumnConfig,
-  type ColumnGroupConfig,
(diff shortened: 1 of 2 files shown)


packages/sit-onyx/src/components/OnyxDataGrid/examples/CustomColumnTypes.vue
```diff
@@ -60,12 +51,20 @@ const withCustomType = createFeature(() => ({
...
+  } satisfies TypeRenderMap<Entry>,
 }));
 
 const features = [withCustomType()];
 </script>
 
 <template>
-  <OnyxDataGrid :columns :data :features />
+  <OnyxDataGrid
+    :columns="[
```
larsrickert, line 62: Would prefer to keep the columns in the `<script>` since thats it most likely the way how the project would define them. So we can show an example how to type it correctly

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxDataGrid/examples/CustomColumnTypes.vue
JoCa96: Hi, I created #3394 to follow up on this, as there is no *nice* way to pass the feature types to the ColumnConfig.

PR #3408, opened 2025-05-22
Larsrickert/3398 fix flyou hydration error

### LINKED ISSUE #3398: fix: `OnyxFlyoutMenu` hydration missmatch if using with nuxt
### Version

1.0.0-beta.225

### Description

Using `OnyxFlyoutMenu` or `OnyxUserMenu` with nuxt causing hydration missmatch error.

### Reproduction steps

- open https://stackblitz.com/edit/onyx-flyout-menu-issue?file=app.vue
- check console warnings

Image

- mobile version should not be displayed first after page is loaded
- click on `English` - nothing happend
- hover `UserMenu` - nothing happend
- reduce windows size to make mobile version visible
- resize again to the normal size
- flyout menu is activated user can see options

### Confirmations

larsrickert:
closes #3398

Fix hydration errors in Nuxt when using the OnyxFlyout (or any component that uses the OnyxFlyout internally) or OnyxTooltip components.

The issue is that we are currently immediately checking for the Anchor API support which is executed on server side when using Nuxt so it will always return false. I changed it to only check it on client side. Also using `v-bind` in CSS caused hydration errors so I moved it to inline styles instead

## Checklist


packages/sit-onyx/src/composables/useAnchorPositionPolyfill.ts
```diff
@@ -38,6 +31,24 @@ type UseAnchorPositionPolyfillOptions = {
...
+  onBeforeMount(() => {
+    USERAGENT_SUPPORTS_ANCHOR_API.value =
+      "CSS" in globalThis &&
+      typeof CSS !== "undefined" &&
+      CSS.supports("anchor-name: --test") &&
+      CSS.supports("position-area: top");
+  });
+
+  return { USERAGENT_SUPPORTS_ANCHOR_API };
+};
```
JoCa96, line 50: Wouldn't it be nicer to merge this with the `useAnchorPositionPolyfill`? The composables always seem to be used together.

packages/sit-onyx/src/composables/useAnchorPositionPolyfill.ts
```diff
@@ -38,6 +31,24 @@ type UseAnchorPositionPolyfillOptions = {
...
+
+  onBeforeMount(() => {
+    USERAGENT_SUPPORTS_ANCHOR_API.value =
+      "CSS" in globalThis &&
+      typeof CSS !== "undefined" &&
+      CSS.supports("anchor-name: --test") &&
+      CSS.supports("position-area: top");
+  });
+
+  return { USERAGENT_SUPPORTS_ANCHOR_API };
```
JoCa96, line 49: Only (global) constants should be capitalized

```suggestion
  return { useragentSupportAnchorApi };
```

JoCa96 (approved):

PR #3409, opened 2025-05-22
fix(OnyxNavBar): fix runtime error when OnyxNavItems were mounted asynchronously

JoCa96:
**Cause:** Built-in `Teleport` is used to teleport the `OnyxNavItem` content to the `OnyxMoreList` `options` slot (usually when there is not enough horizontal space to display the item). So far the teleport matched the target via an static id and the [`defer` property](https://vuejs.org/guide/built-ins/teleport.html#deferred-teleport) was used to ensure that the teleport is performed after everything is rendered. Now, when `OnyxNavItems` were added async, the teleport was tried and the `OnyxMoreList` didn't yet render the `options` slot the `Teleport` component crashed with the error `can't access property "type", instance is null`.

**Fix:** Instead of using a static id, we now make use of a TemplateRef. This ref will only be defined when it's elment exists. Additionally we disable the teleport as long as the ref is `undefined`


packages/sit-onyx/src/components/OnyxNavBar/OnyxNavBar.ct.tsx
```diff
@@ -452,6 +453,32 @@ test("should display More Items correctly", async ({ mount, page }) => {
   });
 });
 
+test("should work with async nav items", async ({ mount, page }) => {
+  // ARRANGE
+  await page.setViewportSize({ width: ONYX_BREAKPOINTS.sm, height: 400 }); // set small viewport, so that the last element ends up in the more menu
+  const component = await mount(AsyncTestCase);
+  await expect(component.getByRole("menuitem").count()).resolves.toBe(0);
```
larsrickert, line 460:
```suggestion
  await expect(component.getByRole("menuitem")).toHaveCount(0);
```

Same for the other expect below

larsrickert (approved):

Issue #3212, opened 2025-04-29
Implement skeleton for table and data grid


larsrickert:
### Why?

We want to natively support skeleton for the OnyxTable and OnyxDataGrid.

### Depends on

- #3210

### Design

[Figma](https://www.figma.com/design/YfEUBOHk4J4nYrk04geswG/Onyx-Component-Library?node-id=4435-40137&m=dev), must be updated in dependent #3210

### Acceptance criteria

- a new skeleton property is supported for the following components:
    - [ ] OnyxTable
    - [ ] OnyxDataGrid
- [ ] the prop can be either set to a boolean (shows a pre-defined number of skeleton rows) or a number to define the row count
- [ ] if unset, the skeleton prop is inherited by the parent/context (e.g. from the parent OnyxPageLayout)

### Implementation details

<-- Add technical details / implementation proposals if necessary, e.g.:
  - API proposal
  - relevant web APIs
  - learnings from other implementations
  - code snippets
  -->

### Definition of Done

- The following component requirements are implemented:
  - [ ] skeleton
  - [ ] density
  - [ ] example usage added to `apps/demo-app/src/views/HomeView.vue`

- Should be covered by tests:
  - [ ] functional tests (Playwright or unit test)
  - [ ] visual tests (Playwright screenshots)

- Make sure, that
  - [ ] follow-up tickets were created if necessary
  - [ ] updated version + documentation is deployed
  - [ ] Storybook can show the feature
  - [ ] Storybook code snippet of new/changed examples are checked that they are generated correctly
  - [ ] Namings are aligned with Figma
  - [ ] Storybook sidebar badge for "New" is added on component or story level if needed. See: https://github.com/Sidnioulz/storybook-addon-tag-badges/?tab=readme-ov-file

#### Approval

[Storybook]()

- Checked and approved by
  - [ ] designer 
  - [ ] dev

jannick-ux:
Please remove the hover effect on the skeleton state. Then I will approve

PR #3415, opened 2025-05-23
Feat(OnyxTable, DataGrid): implement skeleton

ChristianBusshoff:
Relates to #3212 

added skeleton in Table & DataGrid

### DIFF
--- a/packages/sit-onyx/src/components/OnyxDataGrid/OnyxDataGridRenderer/OnyxDataGridRenderer.vue
+++ b/packages/sit-onyx/src/components/OnyxDataGrid/OnyxDataGridRenderer/OnyxDataGridRenderer.vue
@@ -21,10 +21,18 @@ defineSlots<{
   empty?(): unknown;
 }>();
 
+// defauft skeletonCount if not provided same in table
+const skeletonRowCount = 5;
+
 const columnStyle = computed(() => {
   return {
     "--onyx-data-grid-column-count": props.columns.length,
-    "--onyx-data-grid-row-count": Math.max(props.rows.length + 2, 3),
+
+    "--onyx-data-grid-row-count": !props.skeleton
+      ? Math.max(props.rows.length + 2, 3)
+      : typeof props.skeleton === "number"
+        ? props.skeleton + 2
+        : skeletonRowCount + 2,
     "--onyx-data-grid-template-columns": props.columns
       .map(({ key, width }) => {
         const name = `--onyx-data-grid-column-${CSS.escape(String(key))}`;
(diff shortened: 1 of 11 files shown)


packages/sit-onyx/src/components/OnyxDataGrid/OnyxDataGridRenderer/OnyxDataGridRenderer.vue
```diff
@@ -43,18 +51,16 @@ const columnStyle = computed(() => {
...
-          v-for="column in props.columns"
-          :key="column.key"
-          v-bind="column.thAttributes"
-          :ref_for="false"
-          scope="col"
-        >
-          <component :is="column.component" v-bind="column.props" />
-        </th>
-      </tr>
+      <!-- We set `ref_for` to false, so the refs are not passed as an array  -->
```
larsrickert, line 54: Why did you remove the `<tr>` here?

packages/sit-onyx/src/components/OnyxTable/OnyxTable.vue
```diff
@@ -61,6 +71,13 @@ const style = computed(() => ({
   "--onyx-table-observed-height": `${height.value}px`,
   "--onyx-table-observed-width": `${width.value}px`,
 }));
+
+const columnCount = computed(() => {
```
larsrickert, line 75: Hmm I am not sure if we should iterate the vnodes here...for the data grid we know the columns because they are passed as prop.

For the OnyxTable we might consider requiring the user to pass the column count as well. We can talk about it in the daily

packages/sit-onyx/src/components/OnyxTable/OnyxTable.vue
```diff
@@ -330,5 +355,9 @@ $border: var(--onyx-1px-in-rem) solid var(--onyx-color-component-border-neutral)
       color: var(--onyx-color-text-icons-primary-intense);
     }
   }
+  &-skeleton {
+    height: 0.5rem;
+    border-radius: 0.5rem;
```
larsrickert, line 360:
```suggestion
    border-radius: var(--onyx-radius-full);
```

packages/sit-onyx/src/components/OnyxDataGrid/types.ts
```diff
@@ -81,6 +82,10 @@ export type OnyxDataGridProps<
    * The data that should be used to fill the datagrid.
    */
   data: TEntry[];
+  /**
+   * Whether to show a skeleton.
```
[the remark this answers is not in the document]
larsrickert: How does it get the row count then? When using skeleton, there is no data available so we can not use the data property of the OnyxDataGrid for this.

I think the current example is only not really working in projects then, because the data flow is:
1. Show skeleton (no data available yet)
2. Load data from backend
3. After some time, the data is loaded
4. Show actual data

PR #3456, opened 2025-05-31
Feat(OnyxSelect): Implement auto grouping of selected options

ChristianBusshoff:
Relates to #3397

- added grouping and ordering of sected options.

### DIFF
--- a/packages/sit-onyx/src/components/OnyxSelect/OnyxSelect.vue
+++ b/packages/sit-onyx/src/components/OnyxSelect/OnyxSelect.vue
@@ -325,28 +325,20 @@ const {
 });
 
 const getOptionsWithGroupForSelected = () => {
-  let options = filteredOptions.value;
-  if (modelValue.value) {
-    if (
-      !props.keepSelectionOrder &&
-      Array.isArray(modelValue.value) &&
-      modelValue.value.length > 0
-    ) {
-      const selectedValues = new Set(modelValue.value as TValue[]);
-
-      options = filteredOptions.value.map((option) => {
-        if (selectedValues.has(option.value)) {
-          return {
-            ...option,
-            group: t.value("selections.selectGroup"),
-          };
-        }
-        return option;
-      });
-    }
-    return options;
+  if (
+    props.keepSelectionOrder ||
+    !modelValue.value ||
+    !Array.isArray(modelValue.value) ||
+    modelValue.value.length == 0
+  ) {
+    return filteredOptions.value;
   }
-  return filteredOptions.value;
+  const selectedValues = new Set(modelValue.value as TValue[]);
+
+  return filteredOptions.value.map((option) => ({
+    ...option,
+    group: selectedValues.has(option.value) ? t.value("selections.selectGroup") : option.group,
+  }));
 };
 
 const groupedOptions = ref<{ name: string; items: SelectOption<TValue>[] }[]>();


packages/sit-onyx/src/components/OnyxSelect/OnyxSelect.vue
```diff
@@ -366,6 +402,24 @@ const selectInputProps = computed(() => {
...
+  },
+  { deep: true },
+);
+onMounted(() => {
+  if (props.keepSelectionOrder) {
+    groupedOptions.value = groupByKey(filteredOptions.value, "group");
+  } else {
+    groupedOptions.value = groupByKey(sortOptions(), "group", t.value("selections.selectGroup"));
+  }
+});
```
JoCa96: There  is an `immediate` option for `watch`, which will run the watcher callback immediately (like onMounted):

```suggestion
  { deep: true, immediate: true },
);
```

packages/sit-onyx/src/components/OnyxSelect/OnyxSelect.vue
```diff
@@ -309,7 +316,36 @@ const {
...
+        return option;
+      });
+    } else {
+      const modelValueOption = options.find(
+        (option) => option.value === (modelValue.value as unknown),
+      );
+      options = options.filter((option) => {
+        return option.value !== (modelValue.value as unknown);
+      });
+      if (modelValueOption) options = [modelValueOption, ...options];
```
JoCa96: IMHO we shouldn't reorder for single select.

packages/sit-onyx/src/utils/objects.ts
```diff
@@ -27,6 +28,23 @@ export const groupByKey = <TValue extends { [key in TKey]?: string }, TKey exten
...
+    sortedGrouped[preferredGroup] = grouped[preferredGroup];
+  }
+
+  for (const k of keys) {
+    if (k !== preferredGroup) {
+      sortedGrouped[k] = grouped[k];
+    }
+  }
+
+  return sortedGrouped;
```
JoCa96: We should consider the order of properties in an object to be random (see examples in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys#using_object.keys), therefore we shouldn't try to order them.
For cases in which order is important an array should be used.

packages/sit-onyx/src/utils/objects.ts
```diff
@@ -17,16 +17,28 @@ export const areObjectsFlatEqual = (obj1: FlatObject, obj2: FlatObject): boolean
...
-  return objects.reduce(
-    (acc, currOpt) => {
-      const groupName = currOpt[key] ?? "";
-      acc[groupName] = acc[groupName] || [];
-      acc[groupName].push(currOpt);
-      return acc;
-    },
-    {} as Record<string, TValue[]>,
-  );
+  preferredGroup?: string,
```
JoCa96: Please move the ordering out, into a dedicated function.

packages/sit-onyx/src/components/OnyxSelect/OnyxSelect.vue
```diff
@@ -309,7 +318,28 @@ const {
...
+        }
+        return option;
+      });
+    }
+    return options;
+  }
+  return filteredOptions.value;
+};
+
+const groupedOptions = ref();
```
JoCa96: Please add a type

packages/sit-onyx/src/components/OnyxSelect/OnyxSelect.vue
```diff
@@ -309,7 +318,28 @@ const {
   onSelect,
 });
 
-const groupedOptions = computed(() => groupByKey(filteredOptions.value, "group"));
+const sortOptions = () => {
```
JoCa96:
```suggestion
const getOptionsWithGroupForSelected = () => {
```

packages/sit-onyx/src/components/OnyxSelect/OnyxSelect.vue
```diff
@@ -366,6 +396,20 @@ const selectInputProps = computed(() => {
...
+watch(
+  [filteredOptions],
+  () => {
+    if (props.keepSelectionOrder) {
+      groupedOptions.value = groupByKey(filteredOptions.value, "group");
+    } else {
+      groupedOptions.value = transformGroupedData<SelectOption<TValue>, "group">(
+        groupByKey<SelectOption<TValue>, "group">(sortOptions(), "group"),
+        t.value("selections.selectGroup"),
+      );
```
JoCa96: Please put this in multiple lines to make it easier to understand and read.
```suggestion
      groupedOptions.value = transformGroupedData<SelectOption<TValue>, "group">(
        groupByKey<SelectOption<TValue>, "group">(sortOptions(), "group"),
        t.value("selections.selectGroup"),
      );
```

JoCa96 (changes requested):

JoCa96 (approved):

Issue #2239, opened 2024-12-04
Implement the setup for the themes as enterprise packages


JoCa96:
- Brand fonts are included in each brand/theme package (see #3027)
- A mono-repo in AzureDevOps:
  - Create single package for each theme: digits, lidl, kaufland, prezero, scos

### Acceptance criteria
- [x] Setup Repo/Pipelines/Artifactory up via ODJ
- [x] Setup basic Monorepo
- [x] Setup screenshot tests and accessibility tests with the [ComponentShowcase](https://storybook.onyx.schwarz/?path=/story/examples-component-showcase--default) component
- [ ] The pipeline publishes the theming packages to our private artifactory

PR #3458, opened 2025-06-02
feat(themes): exporting ComponentShowcase

ChristianBusshoff:
Relates to #2239 

- exporting OnyxComponentShowcase to use it in themes


JoCa96 (changes requested):
You'll need a changeset for it to be released as a version

JoCa96 (changes requested):
You'll need a changeset for it to be released as a version

JoCa96 (approved):

Issue #3200, opened 2025-04-25
Implement async feature support for OnyxDataGrid


larsrickert:
### Why?

We want to support async data/features for the OnyxDataGrid, e.g. backend sorting, filtering, pagination etc.

We already discussed multiple approaches for this in #2994 and decided to implement the "Option 1".

### Depends on

- #2994

### Acceptance criteria

- [x] every existing data grid feature that can be integrated with async backend data supports a new option to disable the default data transformation. When set, no data should be modified. Instead just the feature state should be accessible so it can be integrated with custom watchers. As time of writing, these features are:
    - [x] sorting
    - [x] filtering
- [x] the above mentioned features keep the relevant state up-to-date, so e.g. `DataGridFeatures.useSorting({ sortState })`
- [x] a new "Async" Storybook example is created that demonstrates how to use it. The example can be based on https://github.com/SchwarzIT/onyx/issues/2994#issuecomment-2826682564

**Fo-Up:**
- #3541
- #3542

### Definition of Done

- The following component requirements are implemented:
  - ~[ ] example usage added to `apps/demo-app/src/views/HomeView.vue`~ n/a

- Should be covered by tests:
  - [x] functional tests (Playwright or unit test)
  - ~[ ] visual tests (Playwright screenshots)~ n/a

- Make sure, that
  - [x] follow-up tickets were created if necessary
  - [x] updated version + documentation is deployed
  - [x] Storybook can show the feature
  - [x] Storybook code snippet of new/changed examples are checked that they are generated correctly
  - [x] Namings are aligned with Figma
  - [x] Storybook sidebar badge for "New" is added on component or story level if needed. See: https://github.com/Sidnioulz/storybook-addon-tag-badges/?tab=readme-ov-file

#### Approval

[Storybook](https://storybook.onyx.schwarz/?path=/story/data-datagrid-features--async)

- Checked and approved by
  - [x] designer 
  - [x] dev @larsrickert

PR #3523, opened 2025-06-13
feat(OnyxDataGrid): Support async (disabling) data transformation

JoCa96:
Relates to #3200 

- BREAKING CHANGE update for type `DataGridFeature`:  to support the passing of context from the `OnyxDataGrid` itself a feature is now a function.
- features are now provided with `async` context, so that they can disable data transformation as necessary
- added `asArray` util function

### DIFF
--- a/packages/sit-onyx/src/components/OnyxDataGrid/examples/AsyncExample.vue
+++ b/packages/sit-onyx/src/components/OnyxDataGrid/examples/AsyncExample.vue
@@ -1,8 +1,6 @@
 <script setup lang="ts">
 import { ref, watch } from "vue";
 import { DataGridFeatures, normalizedIncludes, OnyxDataGrid, type ColumnConfig } from "../../..";
-import type { FilterState } from "../features/filtering/filtering";
-import type { SortState } from "../features/sorting/types";
 
 type Entry = {
   id: number;
@@ -27,8 +25,8 @@ const columns: ColumnConfig<Entry>[] = [
   { key: "birthday", label: "Birthday", type: "date" },
 ];
 
-const filterState = ref({} as FilterState<Entry>);
-const sortState = ref({} as SortState<Entry>);
+const filterState = ref<DataGridFeatures.FilterState<Entry>>({});
+const sortState = ref<DataGridFeatures.SortState<Entry>>({ column: undefined, direction: "none" });
 const skeleton = ref(false);
 
 const withFiltering = DataGridFeatures.useFiltering<Entry>({ filterState });
--- a/packages/sit-onyx/src/components/OnyxDataGrid/features/index.ts
+++ b/packages/sit-onyx/src/components/OnyxDataGrid/features/index.ts
@@ -154,7 +154,13 @@ export type PublicNormalizedColumnConfig<
  * Context that is passed to a feature when it is set up by the `useDataGridFeatures` composable.
  */
 export type DataGridFeatureContext = {
+  /**
+   * Ref for the `async` state of the `OnyxDataGrid`. If `true` data mutations should be skipped, if they are expected to be handled by a backend.
+   */
   async: Readonly<Ref<boolean>>;
+  /**
+   * The `i18n` object, which can be used to access messages, formatters and the current locale.
+   */
   i18n: OnyxI18n;
 };
 
@@ -281,7 +287,7 @@ export type DataGridFeatureOptions<
 } & (TWithAsync extends true
   ? {
       /**
-       * When async is `true`, then the internal data transformations are disabled and have to be performed by the user.
+       * When async is `true`, then the internal data transformations of this feature  are disabled and have to be performed manually.
        */
       async?: boolean;
     }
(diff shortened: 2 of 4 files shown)


packages/sit-onyx/src/components/OnyxDataGrid/examples/AsyncExample.vue
```diff
@@ -0,0 +1,83 @@
...
+  { key: "age", label: "Age" },
+  { key: "birthday", label: "Birthday", type: "date" },
+];
+
+const filterState = ref({} as FilterState<Entry>);
+const sortState = ref({} as SortState<Entry>);
+const skeleton = ref(false);
+
+const withFiltering = DataGridFeatures.useFiltering<Entry>({ filterState });
+const withSorting = DataGridFeatures.useSorting<Entry>({ sortState });
```
larsrickert: Wouldn't it be more convient if the state is returned from the feature itself so the user does not have to create the ref and type it accoringly?

packages/sit-onyx/src/components/OnyxDataGrid/examples/AsyncExample.vue
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * Simulate asynchronous data update from a backend.
+ * This function simulates a delay when fetching data and applying filters and sorting.
+ * It also shows a skeleton while the data is being fetched.
+ *
+ * Note: This is a simplified example, and real-world applications need to consider error handling and pagination.
+ */
+const simulateAsyncUpdate = () => {
+  skeleton.value = true;
```
larsrickert, line 45: Usually this is called "isLoading" :) Should we align this? So its easier for the user to migrate this code snippet to their projects when using libs like vueuse, nuxt etc

packages/sit-onyx/src/components/OnyxDataGrid/examples/AsyncExample.vue
```diff
@@ -0,0 +1,83 @@
+<script setup lang="ts">
+import { ref, watch } from "vue";
+import { DataGridFeatures, normalizedIncludes, OnyxDataGrid, type ColumnConfig } from "../../..";
+import type { FilterState } from "../features/filtering/filtering";
+import type { SortState } from "../features/sorting/types";
```
larsrickert: There imports must also be imported from "../../../" so the Storybook code snippet is replaced correctly

apps/demo-app/src/views/DataGridView.vue
```diff
@@ -100,7 +100,7 @@ const columns: ColumnConfig<Entry, ColumnGroupConfig, never>[] = [
   "job",
 ];
 
-const dummyFeature = createFeature(() => ({
+const dummyFeature = createFeature(() => () => ({
```
larsrickert: This syntax looks strange to me because there are two arrow functions. We could consider just using one and always passing the context as first parameter, followed by custom parameters of the feature, what do you think?

packages/sit-onyx/src/components/OnyxDataGrid/features/index.ts
```diff
@@ -141,20 +143,39 @@ export type PublicNormalizedColumnConfig<
...
    */
   columnGroupKey?: keyof TColumnGroup;
 };
 
+/**
+ * Context that is passed to a feature when it is set up by the `useDataGridFeatures` composable.
+ */
+export type DataGridFeatureContext = {
+  async: Readonly<Ref<boolean>>;
+  i18n: OnyxI18n;
```
larsrickert: Please add some JSDocs here :)

packages/sit-onyx/src/components/OnyxDataGrid/types.ts
```diff
@@ -71,6 +74,11 @@ export type OnyxDataGridProps<
    * Check the Storybook examples (e.g. [Sorting](/?path=/story/data-datagrid-features--sorting)) for more details on how the features are used and configured.
    */
   features?: TFeatures;
+  /**
+   * When `async` is `true`, the data transformation of supported features is disabled and externalized.
+   * This allows for backend handling of data fetching and data transformation.
+   */
+  async?: boolean;
```
larsrickert, line 81: Are you sure we want to offer a global property for this in addition to the per-feature property? Might be confusing for the users

larsrickert (changes requested):

apps/demo-app/src/views/DataGridView.vue
JoCa96: We cannot really avoid it, if it should be possible to pass context/parameters from two places. But we can make the composable part optional. I updated the types.

packages/sit-onyx/src/components/OnyxDataGrid/examples/AsyncExample.vue
JoCa96: IMHO it is the best to keep all inputs/parameters in one place and reserve the Return type for out internal API

packages/sit-onyx/src/components/OnyxDataGrid/types.ts
JoCa96: Yes, I want to allow users to opt-out of specific features.

larsrickert: Lets keep this for now and wait for user feedback :D

larsrickert (approved):

Issue #2978, opened 2025-03-19
Fix Flyout alignment


ChristianBusshoff:
### Why?

The flyout is currently cut off if there is insufficient space on its right.
<img width="522" alt="Image" src="https://github.com/user-attachments/assets/217c764c-f6b3-4fe8-b671-fc2d501f5438" />
Instead, it should adjust to the available space.
Additionally, the user should have the option to manually control the alignment (left | right).

### Acceptance criteria

- [x] the issue should be fixed for all tooltips and flyouts (basically all popovers)
- [x] Check if there is enough space, otherwise position it on the other side.
- [x] Add an alignment prop with options: auto, left, and right to control the positioning manually.

### Implementation details

check if https://developer.mozilla.org/en-US/docs/Web/CSS/position-try is usful, if necessary only implement in the anchor polyfill

### Reference implementations

For the Tooltip there is already a similar composable to handle the alignment (useWedgePosition) to orientate on.

SebastianSchmidtInovex:
Unfortunately, it seems that the placement of the Flyout does not work as expected in every situation.
In our product (SIT PAVA) the Flyout menu of the HideColumns feature is rendered to the right of + Icon and gets cut off by the browsers right border when the browser window size is reduced after a column was hidden.

![Image](https://github.com/user-attachments/assets/09616d46-b0c5-433a-8682-54af3ca2d7bd)

We are using:
"@sit-onyx/icons": "1.0.0-beta.16",
"@sit/vue-core": "^3.1.2",
"sit-onyx": "1.0.0-beta.222",

larsrickert:
> Unfortunately, it seems that the placement of the Flyout does not work as expected in every situation. 

@SebastianSchmidtInovex Hey, is it possible for to create a reproduction using our [playground](https://playground.onyx.schwarz) so we can check the issue?

SebastianSchmidtInovex:
> > Unfortunately, it seems that the placement of the Flyout does not work as expected in every situation.
> 
> [@SebastianSchmidtInovex](https://github.com/SebastianSchmidtInovex) Hey, is it possible for to create a reproduction using our [playground](https://playground.onyx.schwarz) so we can check the issue?

Hi, I am trying to do this.
Could you please advice how I can import the `@sit-onyx/icons` module and import an icon into the `App.vue` file inside the Playgroud?

[...]

PR #3537, opened 2025-06-16
bug(onyxFlyout): fixed flyout opening

ChristianBusshoff:
- [issue](https://github.com/SchwarzIT/onyx/issues/2978#issuecomment-2975248738)

- changed default opening in hidecolumn
- fixed alignment calc for "left" and "right" as default in useOpenAlignemnt


.changeset/great-beers-own.md
```diff
@@ -0,0 +1,5 @@
+---
+"sit-onyx": minor
+---
+
+bug(useOpenAlignment): fixed alignment calc for "right" and "left" as default
```
JoCa96:
```suggestion
fix(useOpenAlignment): fixed alignment calc for "right" and "left" as default
```

packages/sit-onyx/src/composables/useOpenAlignment.ts
```diff
@@ -29,7 +29,8 @@ export const useOpenAlignment = (
       return;
     }
 
-    const minSpace = (tooltipElementRect.width - wrapperRect.width + minMargin * 2) / 2;
+    const minSpaceCenter = (tooltipElementRect.width - wrapperRect.width + minMargin * 2) / 2;
+    const minSpace = tooltipElementRect.width - wrapperRect.width + minMargin;
 
     const parentLeft = overflowParentRect?.left ?? window.visualViewport?.pageLeft ?? 0;
     const parentRight =
```
JoCa96: We should be able to remove the whole `overflowParentRect`, as we now use the popover API and the elements cannot be cut off anymore by their parents.

JoCa96:
Is there a relevant ticket/issue?

Issue #3340, opened 2025-05-15
Implement layout grid for OnyxSidebar


larsrickert:
### Why?

As a designer, I want to be able to use a sidebar to provide additional navigation and other options.
The sidebar should provide its own grid implementation that differs a little bit from the page wide grid.

### Depends on

- #2726 

### Design

[Figma](https://www.figma.com/design/EBi98T51T9fGfjrYYfsulg/onyx---Archive?node-id=4169-89075&m=dev)

### Acceptance criteria

- [x] the sidebar provides a custom grid, as described in Figma

### Implementation details

- for the grid content, the existing CSS classes in `packages/sit-onyx/src/styles/grid.scss` can be reused.
- the sidebar grid should be implemented using [container queries](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_containment/Container_queries)
- the page grid needs to be adjusted to use a container query
- https://onyx.schwarz/development/breakpoints.html#container-queries

### Definition of Done

- The following component requirements are implemented:
  - [x] density
  - [x] example usage added to `apps/demo-app/src/views/HomeView.vue`

- Should be covered by tests:
  - [ ] functional tests (Playwright or unit test)
  - [ ] visual tests (Playwright screenshots)

- Make sure, that
  - [x] follow-up tickets were created if necessary
  - [x] updated version + documentation is deployed
  - [x] Storybook can show the feature
  - [x] Storybook code snippet of new/changed examples are checked that they are generated correctly
  - [x] Namings are aligned with Figma
  - [x] Storybook sidebar badge for "New" is added on component or story level if needed. See: https://github.com/Sidnioulz/storybook-addon-tag-badges/?tab=readme-ov-file

#### Approval

[Storybook](https://storybook.onyx.schwarz/?path=/story/navigation-sidebar--grid)
[Documentation](https://onyx.schwarz/development/grid.html)

- Checked and approved by
  - [x] designer 
  - [x] dev

PR #3613, opened 2025-06-27
Feat(OnyxSidebar): implement grid layout

ChristianBusshoff:
Relates to #3340 

- add grid layout for sidebar

### DIFF
--- a/apps/demo-app/src/views/GridDemo.vue
+++ b/apps/demo-app/src/views/GridDemo.vue
@@ -1,4 +1,5 @@
 <script lang="ts" setup>
+import placeholder from "@sit-onyx/icons/placeholder.svg?raw";
 import {
   OnyxBottomBar,
   OnyxButton,
@@ -6,6 +7,8 @@ import {
   OnyxImage,
   OnyxInput,
   OnyxPageLayout,
+  OnyxSidebar,
+  OnyxSidebarItem,
   OnyxSwitch,
 } from "sit-onyx";
 import { useGridStore } from "../stores/grid-store";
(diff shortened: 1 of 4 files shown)


packages/sit-onyx/src/styles/grid.scss
larsrickert, line 1: Would prefer if the sidebar grid is use the same way as the regular grid, so e.g. using `onyx-grid` and `onyx-grid-container` instead of `onyx-sidebar-grid` and `onyx-sidebar-grid-container`, what do you think?

We could just change/override the CSS variables that are used internally when used inside the onyx-sidebar

larsrickert (changes requested):

packages/sit-onyx/src/styles/grid.scss
ChristianBusshoff: Good point, I changed it

apps/demo-app/src/views/GridDemo.vue
```diff
@@ -23,6 +26,32 @@ const gridStore = useGridStore();
...
+              <OnyxSidebarItem v-for="i in 6" :key="i" :link="`#link-${i}`">
+                <OnyxIcon :icon="placeholder" />
+                Item {{ i }}
+              </OnyxSidebarItem>
+            </div>
+            <div class="onyx-grid second-sidebar-grid">
+              <p class="onyx-grid-md-span-2 onyx-grid-lg-span-4 onyx-grid-xl-span-6">Half width</p>
+              <p class="onyx-grid-md-span-2 onyx-grid-lg-span-4 onyx-grid-xl-span-6">Half width</p>
+              <p class="onyx-grid-span-full">Full width</p>
+            </div>
```
larsrickert: For simplicy we can remove this I think

apps/demo-app/src/views/GridDemo.vue
```diff
@@ -23,6 +26,32 @@ const gridStore = useGridStore();
       :height="256"
       loading="eager"
     />
+    <template #sidebar>
+      <OnyxSidebar label="Example sidebar" resizable>
+        <template #header> Header content </template>
+
+        <div class="sidebar__content">
```
larsrickert: sidebar__content is unused

apps/demo-app/src/views/GridDemo.vue
```diff
@@ -23,6 +26,32 @@ const gridStore = useGridStore();
...
+              <p class="onyx-grid-md-span-2 onyx-grid-lg-span-4 onyx-grid-xl-span-6">Half width</p>
+              <p class="onyx-grid-span-full">Full width</p>
+            </div>
+          </div>
+        </div>
+
+        <template #footer>
+          <OnyxButton color="neutral" label="Button" />
+          <OnyxButton label="Button" />
+        </template>
```
larsrickert: The footer is not relevant for the example, right?

packages/sit-onyx/src/styles/grid.scss
```diff
@@ -2,57 +2,104 @@
...
+   "lg" "md"        8       1.5rem 2rem,
+   "xl" "lg"        12      1.5rem 4rem,
+);
+
+/*#sidebar breakpoints*/
+
+$sidebar-breakpoints: (
+  xs: 0px,
+  sm: 200px,
+  tm: 200px,
```
larsrickert: whats tm? Is this a leftover since its the same as sm?

packages/sit-onyx/src/styles/grid.scss
larsrickert, line 1: Is it inteded that there are a lot of newlines added to this file? Makes the git diff a little harder to read :)

packages/sit-onyx/src/styles/grid.scss
```diff
@@ -126,4 +189,75 @@ $gridVariants: (
...
+      }
+    }
+    /**
+* Grid element class for spanning the full row width.
+*/
+
+    .onyx-grid-span-full {
+      grid-column: 1 / span var(--onyx-sidebar-grid-columns);
+    }
+  }
```
larsrickert: We shouldn't need to override/copy+paste all of this. In the generated properties below, you can just use the same variable names as for the regular grid so they will be overriden:

e.g. `onyx-grid-columns` instead of `onyx-sidebar-grid-columns`

larsrickert: So imo the only variables we need to set/override here are:
```css
--onyx-grid-columns
--onyx-grid-gutter
--onyx-grid-margin
```

apps/demo-app/src/views/GridDemo.vue
larsrickert, line 1: Please also extend the docs and the Storybook example with this :) Can also be done in a follow up PR

larsrickert (approved):

Issue #2996, opened 2025-03-20
Implement `drilldown` mode `internal` for OnyxFlyoutMenu


JoCa96:
### Why?

As a designer I want to be able to have flyouts with drill-down lists, so that nested options can be defined.

### Design

[Figma](https://www.figma.com/design/EBi98T51T9fGfjrYYfsulg/onyx---Archive?node-id=3427-53313&m=dev)

![Image](https://github.com/user-attachments/assets/47f8c5aa-455c-461b-aa63-765ca1cdf199)

### Acceptance criteria

- [x] Nested items are passed via the `children` slot of the `OnyxMenuItem` / `OnyxNavItem` component
- [x] The HTML/DOM structure reflects the hierachy of the items
- [x] Initially only the parent element is shown in the list
- Clicking a parent displays:
  - [x] the parents children inside of the flyout
  - [x] a "<- back" button
- [x] Nesting can be done recursively without a limit
- [x] when the flyout is closed, the open state of the nested items is kept

**out-of-scope:**
- animation

### Reference implementations

[Treeview pattern](https://www.w3.org/WAI/ARIA/apg/patterns/treeview/examples/treeview-1a/)

### Applicable ARIA Pattern

[Menubar](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/examples/menubar-navigation/)
[MenuButton](https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/)

### Definition of Done

- The following component requirements are implemented:
  - [x] density

- Should be covered by tests:
  - [x] functional tests (Playwright or unit test)
  - [x] visual tests (Playwright screenshots)

- Make sure, that
  - [x] updated version + documentation is deployed
  - [x] Storybook can show the feature
  - [x] Storybook code snippet of new/changed examples are checked that they are generated correctly
  - [x] Namings are aligned with Figma

#### Approval

[FlyoutMenu](https://storybook.onyx.schwarz/?path=/story/navigation-flyoutmenu--nested)
[Nested nav bar items](https://storybook.onyx.schwarz/?path=/story/navigation-navbar--nested)

- Checked and approved by
  - [x] designer @flubnau 
  - [x] dev @ChristianBusshoff

PR #3616, opened 2025-06-30
feat: support nested items for OnyxFlyoutMenu and OnyxNavItem

larsrickert:
Relates to #2996

Support nested children using an internal drilldown for the flyout menu and nav item.

## Checklist


packages/headless/src/composables/helpers/useOutsideClick.ts
```diff
@@ -22,19 +28,42 @@ export type UseOutsideClickOptions = {
...
+      // the activeElement
+      await new Promise((resolve) => setTimeout(resolve));
+
+      if (isOutsideClick(document.activeElement)) {
+        onOutsideClick(event as Parameters<typeof onOutsideClick>[0]);
+      }
+    };
+
+    useGlobalEventListener({ type: "keydown", listener: keydownListener, disabled });
+  }
```
JoCa96, line 68: Would prefer to use the focusout (or similar) event, so that we don't have a need for `setTimeout`

```suggestion
  if (checkOnTab) {
    const focusoutListener = async (event: FocusEvent) => {
      // check if the new focus target is inside or outside
      if (inside.every(el => !el.contains(event.relatedTarget))) {
        onOutsideClick(event as Parameters<typeof onOutsideClick>[0]);
      }
    };

    useGlobalEventListener({ type: "focusout", listener: focusoutListener, disabled });
  }
```

JoCa96 (changes requested):

packages/headless/src/composables/helpers/useOutsideClick.ts
larsrickert: The issue with focusout is that it is also triggered when clicking on an item which leads to the flyout being closed because the clicked item is no longer inside of the flyout (because the nested children have opened, so the parent is hidden)

larsrickert: I guess the only way to get this working would be to use @focusout.stop inside the OnyxMenuItem but when we are kind of relying on external implementation for the useOutsideClick to work correctly. What do you think / prefer?

larsrickert: As discussed, I created a follow up ticket for this so this PR can be merged: #3633

PR #3626, opened 2025-07-01
fixed font-sizes and weights

ChristianBusshoff:
- removed font-weight onyx-bold
- fixed font-sizes


larsrickert (changes requested):
The bold CSS variable should also be removed, right?
Ideally re-run the Figma import pipeline and make sure that bold is also removed in Figma

larsrickert (approved):

Issue #3435, opened 2025-05-28
Define target types of "sit-onyx" distributables


JoCa96:
### Why?

There are multiple kinds in which the components can and could be consumed/used:
- Through an bundler/compiler
  - with typescript
  - with javascript
- Directly at runtime through an CDN
- In PROD mode
- In DEV mode

There are benefits for providing the components in different formats, e.g.:
- For DEV mode it would be beneficial to provide source-maps/non-minified, DEV mode only warnings
- For an CDN a minified dist would be useful

E.g. right now we provide everything minified in PROD mode, which means: there are no source-maps, onyx code is hard to debug and warnings compiled out (`if (import.meta.env.DEV) {...}`).

See also #2947 

### Acceptance criteria

- [x] Research how other Vue component libraries handle this
- [x] Define and document a proposal how and what would be the best way to provide the `dist` files
- [x] The benefits/disadvantages are weighed

### Implementation details

- It might be necessary to run multiple vite builds.

### Definition of Done

- Make sure, that
  - [x] follow-up tickets were created if necessary
  - [x] updated version + documentation is deployed

#### Approval

- Checked and approved by
  - [x] dev @larsrickert

larsrickert:
@JoCa96  Could you create a follow up ticket to also change the other npm packages?

PR #3641, opened 2025-07-01
feat(dist)!: add dedicated entrypoint for bundlers

JoCa96:
Relates to #3435 

- Updated `sit-onyx` vite build:
  - Use [`vite-plugin-dts`][4] for type generation
  - Add `bundler` build, which provides better developer experience by providing logging, better readable source-code and source-maps
  - Drop commonjs support and switch to publish a [Pure ESM package][1].
  This also requires to add the `.js` file extension for all relative imports. Because 
    1. **[esm loader requires explicit file-extensions][2]**
    2. **[tsc doesn't rewrite file extensions][3]**
  - This also fixes the current publint warning
  - Added `userConsole` util to allow logging and warning user, but is also tree-shaken in production builds.

[1]: https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c
[2]: https://nodejs.org/docs/v22.17.0/api/packages.html#modules-loaders
[3]: https://www.typescriptlang.org/docs/handbook/modules/theory.html#module-specifiers-are-not-transformed
[4]: https://www.npmjs.com/package/vite-plugin-dts


packages/sit-onyx/build/index.js
```diff
@@ -0,0 +1,89 @@
...
+import path from "node:path";
+import { register } from "tsx/esm/api";
+import { build, mergeConfig } from "vite";
+
+async function emptyDirectory(dirPath) {
+  try {
+    // create the directory if it doesn't exist. `recursive` ensures that the command doesn't fail when it already exists.
+    await fsPromises.mkdir(dirPath, { recursive: true });
+
+    // empty the directory without removing it
```
larsrickert, line 14: Wouldn't it be simpler / more performant to remove the whole directory and then create a new empty one?

packages/sit-onyx/build/index.js
```diff
@@ -0,0 +1,89 @@
...
+      build: {
+        minify: false,
+        sourcemap: true,
+        emptyOutDir: false,
+      },
+    },
+    true,
+  );
+  bundlerConfig.build.lib.formats = ["es"];
+  bundlerConfig.build.lib.fileName = () => "index.esm-bundler.js";
```
larsrickert: Can't we use two vite.config.ts files instead of modifying the config here?

packages/sit-onyx/src/components/OnyxDataGrid/OnyxDataGridRenderer/OnyxDataGridRenderer.vue
```diff
@@ -56,7 +55,7 @@ const columnStyle = computed(() => {
     </tr>
 
     <template v-for="(slot, slotName) in slots" :key="slotName" #[slotName]="slotProps">
-      <slot :name="slot?.name" v-bind="slotProps">
+      <slot :name="slot?.name as 'headline'" v-bind="slotProps">
```
larsrickert, line 59: Why is this needed?

larsrickert (approved):
Do we also have to adjust our contribution guide / code examples to include the .js extension?

packages/sit-onyx/build/index.js
JoCa96: In my experience it's more reliable to keep the `dist` directory, as some tools/scripts are "bound" (e.g. watching) to the specifc directory.
E.g. When you want to preview: The server would need to be restarted after each build, because it is still serving the deleted dist folder.

packages/sit-onyx/src/components/OnyxDataGrid/OnyxDataGridRenderer/OnyxDataGridRenderer.vue
JoCa96: added a comment

packages/sit-onyx/build/index.js
JoCa96: done

JoCa96:
@larsrickert pls check again

larsrickert (approved):

Issue #3565, opened 2025-06-18
Implement OnyxFloatingActionButton component


larsrickert:
### Why?

We want to support a new floating action button component that can be used to indicate a global action.

### Open questions / TODOs

- should the component be named OnyxFab oder OnyxFloatingActionButton?

### Design

[Figma](https://www.figma.com/design/YfEUBOHk4J4nYrk04geswG/Onyx-Component-Library?node-id=14188-118079&m=dev)

### Acceptance criteria

- [ ] one or multiple actions can be defined
- every action supports:
    - [ ] label (required)
    - [ ] optional icon
    - [ ] optional link. If a link is provided, the component is rendered using the OnyxRouterLink, otherwise as button.
- [ ] if more than 1 action is passed, the fab extends to a "..." button which will toggle all actions on click
- [ ] is always sticky, keeps position when scrolling
- [ ] supports left/right alignment
- [ ] if possible, the fab can be positioned relative to a container (e.g. sidebar) so that it is positioned fixed relative to its container

### Implementation details

- uses z-index `var(--onyx-z-index-sticky-content)`
- does not use the popover API, dialogs, flyouts etc. should be positioned above the fab

### Definition of Done

- The following component requirements are implemented:
  - [ ] skeleton
  - [ ] density
  - [ ] example usage added to `apps/demo-app/src/views/HomeView.vue`

- Should be covered by tests:
  - [ ] functional tests (Playwright or unit test)
  - [ ] visual tests (Playwright screenshots)

- Make sure, that
  - [ ] follow-up tickets were created if necessary
  - [ ] updated version + documentation is deployed
  - [ ] Storybook can show the feature
  - [ ] Storybook code snippet of new/changed examples are checked that they are generated correctly
  - [ ] Namings are aligned with Figma
  - [ ] Storybook sidebar badge for "New" is added on component or story level if needed. See: https://github.com/Sidnioulz/storybook-addon-tag-badges/?tab=readme-ov-file

#### Approval

[Storybook]()
[Documentation]()

- Checked and approved by
  - [ ] designer 
  - [ ] dev

PR #3656, opened 2025-07-03
feat: implement new OnyxFab component

larsrickert:
Relates to #3565

Implement floating action button component.

## Checklist

### DIFF
--- a/packages/sit-onyx/src/components/OnyxFab/types.ts
+++ b/packages/sit-onyx/src/components/OnyxFab/types.ts
@@ -1,4 +1,4 @@
-import type { Nullable } from "../../types.js";
+import type { Nullable } from "../../types/index.js";
 import type { OnyxFabButtonProps } from "../OnyxFabButton/types.js";
 
 export type OnyxFabProps = OnyxFabButtonProps & {


packages/sit-onyx/src/components/OnyxFabItem/OnyxFabItem.stories.ts
```diff
@@ -0,0 +1,38 @@
...
+import { defineIconSelectArgType } from "../../utils/storybook";
+import OnyxFabItem from "./OnyxFabItem.vue";
+
+const meta: Meta<typeof OnyxFabItem> = {
+  title: "Support/FabItem",
+  component: OnyxFabItem,
+  tags: ["new:component"],
+  argTypes: {
+    ...withNativeEventLogging(["onClick"]),
+    icon: defineIconSelectArgType(),
```
JoCa96, line 13: As the icon is optional, it might be nice to give an `none` option to select in the dropdown, e.g.:

```ts
export const defineIconSelectArgType = (optional = false) => {
  const ALL_ICONS = import.meta.glob("../../node_modules/@sit-onyx/icons/src/assets/*.svg", {
    query: "?raw",
    import: "default",
    eager: true,
  });

  /**
   * Mapping between icon SVG content (key) and icon name (value).
   * Needed to display a labelled dropdown list of all available icons.
   */
  const iconLabels = Object.entries(ALL_ICONS).reduce<Record<string, string>>(
    (labels, [filePath, content]) => {
      labels[content as string] = filePath.split("/").at(-1)!.replace(".svg", "");
      return labels;
    },
    optional ? { "": "none" } : {},
  );

  return {
    options: Object.keys(iconLabels),
    control: {
      type: "select",
      labels: iconLabels,
    },
  } satisfies ArgTypes[number];
};
```

packages/sit-onyx/src/components/OnyxFab/OnyxFab.stories.ts
```diff
@@ -0,0 +1,82 @@
...
+
+export const Options = {
+  args: {
+    label: "Example label",
+    hideLabel: true,
+    default: () => [
+      h(OnyxFabItem, { label: "Action 3", icon: placeholder }),
+      h(OnyxFabItem, { label: "Action 2", icon: placeholder }),
+      h(OnyxFabItem, { label: "Action 1", icon: placeholder }),
+    ],
```
JoCa96, line 61: There seems to be a bug, where sometimes the toggeling stops working..? Not sure if the issue is caused by Storybook?
Reproduction steps (not 100% reliable):
1. open the story
2. change window width
3. click the fab
4. => seems to immediately close again:

Screenshot 2025-07-07 at 13 46 24

larsrickert: Clicking on the outer edge of the button works (not clicking the icon). When clicking in the center on the icon, the bug exists because the useOutsideClick utility will think that you clicked outside because the icon is changed and therefore the previous icon is no longer in the DOM...I will check how to fix it

larsrickert: @JoCa96 I fixed it like this, what do you think about this approach?: https://github.com/SchwarzIT/onyx/pull/3656/commits/e8c2072c9910a9f1ab31b4b9f8df8f30cb26b2b0

JoCa96: I am a little bit worried that this approach might produce unexpected edge-cases. E.g. when outside is clicked, it doesn't close because of the same condition.

Some other ideas to resolve this:
1. Use mousedown
2. Use capture phase
3. Determine outside-click on mousedown, but only apply on click/mouseup

But I this is a very minor issue for now, so creating a low-prio follow-up ticket is also fine

larsrickert: Mousedown seems to work fine 👍  Thanks for the hint!

JoCa96 (approved):

Issue #3638, opened 2025-07-01
FoUp: Implement layout grid for OnyxSidebar


ChristianBusshoff:
### Why?

- change the normal grid form media-query to container-query. // Breakting Change
   So the normal grid changes on the actual space and not on the viewport
- Remove Margin-Top and -Bottom from Sidebar grid-container
- add tests 

### Out od scope
- Add to grid-playground 

### Open questions / TODOs
@SchwarzIT/onyx-dev  setting container-query on pagelayout or on the parent of each grid/grid-container

### Relates to
#3340 

### Definition of Done

  - [x] example usage added to `apps/demo-app/src/views/HomeView.vue`

- Should be covered by tests:
  - [x] functional tests (Playwright or unit test)
  - [x] visual tests (Playwright screenshots)

- Make sure, that
  - [x] follow-up tickets were created if necessary
  - [x] updated version + documentation is deployed
  - [x] Storybook can show the feature
  - [x] Storybook code snippet of new/changed examples are checked that they are generated correctly
  - [x] Namings are aligned with Figma
  - [x] Storybook sidebar badge for "New" is added on component or story level if needed. See: https://github.com/Sidnioulz/storybook-addon-tag-badges/?tab=readme-ov-file

#### Approval

[Storybook]()
[Documentation]()

- Checked and approved by
  - [x] designer 
  - [x] dev

PR #3665, opened 2025-07-04
feat(Grid-Layout): add container-query for main grid

ChristianBusshoff:
- changed from media-query to container-query for the main grid
- added tests for the sidebar grid
- adjusted grid-container margin for the sidebar grid and main grid to be the same for top and bottom on each breakpoint.
- changed onyx-grid-container to onyx-grid-layout and added a onyx-grid-container class as an wrapper to make the container-query work.

Relates to #3638


packages/sit-onyx/src/components/OnyxPageLayout/examples/GridExample.vue
```diff
@@ -0,0 +1,59 @@
...
+      </template>
+
+      <!-- page content -->
+      <OnyxHeadline is="h1">Page content</OnyxHeadline>
+      <p>
+        Breakpoints for the grid dynamically adjust based on the width of the main page content area
+        or the sidebar.
+      </p>
+
+      <div class="onyx-grid-container">
```
larsrickert: the page layout already applies the grid container so this is duplicated here

apps/docs/src/development/grid.md
```diff
@@ -92,6 +92,14 @@ The same principles of span definition apply here: elements within the sidebar g
 
 The grid is configured via specific CSS classes which you can find below.
 
+#### Grid within a Page Layout
+
+When integrating the grid within a standard page layout, the breakpoints are automatically calculated based on the width of the `onyx-page__main` element. This ensures the grid adapts consistently with the main content area of your page.
+
+#### Standalone Grid Usage
+
+For scenarios where the grid is used independently, outside of a `onyx-page__main` context, you can define the breakpoint reference by applying the `onyx-grid-wrapper` class to the component that should determine the grid's width. This allows you to control the responsive behavior precisely, even in custom or isolated layouts.
```
larsrickert:
```suggestion
#### Grid within a OnyxPageLayout

When integrating the grid within a standard [OnyxPageLayout](https://storybook.onyx.schwarz/?path=/docs/layout-pagelayout--docs) component, the breakpoints are automatically calculated based on the width of the content. This ensures the grid adapts consistently with the main content area of your page.

#### Usage without OnyxPageLayout

For scenarios where the grid is used independently, without the OnyxPageLayout component, you must define the container manually by applying the `onyx-grid-wrapper` class to the component that should determine the grid's width. This allows you to control the responsive behavior precisely, even in custom or isolated layouts.
```

apps/docs/src/development/grid.md
```diff
@@ -92,6 +92,14 @@ The same principles of span definition apply here: elements within the sidebar g
...
+
```
larsrickert, line 104: Also I am not sure about the name `onyx-grid__wrapper` since the CSS name for it is "container" but since we already have the onyx-grid-container, the name clashes.

Maybe we should just recommend that the user should set the `container-type: inline-size;` CSS property?

packages/sit-onyx/src/styles/grid.scss
```diff
@@ -126,19 +127,40 @@ $sidebar-breakpoints: (
...
+ * Initializes the grid layout on the `.page__main` element.
+ * Additionally, users may apply the `.onyx-grid-wrapper` class
+ * to explicitly activate the grid. The layout will adapt
+ * to the width of this wrapper accordingly.
+ */
+  .onyx-page__main,
+  .onyx-grid-wrapper {
+    container-type: inline-size;
+
+    @each $name, $maxVariant, $columns, $gutter, $margin in $gridVariants {
```
larsrickert: Please move this outside of the selector. The grid should always adjust to its container, independent on which class the container is set

packages/sit-onyx/src/styles/grid.scss
```diff
@@ -155,7 +177,7 @@ $sidebar-breakpoints: (
         @include generate-grid-properties(
           $columns,
           $gutter,
-          $margin,
+          var(--onyx-spacing-xs) $margin var(--onyx-spacing-xs) $margin,
```
larsrickert: Usually the grid applies margin to all sides, why should this be changed? If needed by the user, he can override it manually.

But by default when applying the gird-container in the sidebar and passing content, it should look pretty with margins all around

ChristianBusshoff: Jannick told me to change it.

larsrickert: Alright, shouldn't it be the same as in the OnyxSidebar then? So var(--onyx-density-md)?

After a quick alignment with Jannick, the page grid should also have a static top+bottom margin, can you also change this?

packages/sit-onyx/playwright/snapshots/components/OnyxBreadcrumb/grid-edge-linux.png
larsrickert, line 1: Something is off here

packages/sit-onyx/playwright/snapshots/components/OnyxNavBar/grid-default-lg-edge-linux.png
larsrickert, line 1: Somehow the background color is changed here

packages/sit-onyx/playwright/snapshots/components/examples/GridPlayground/default-edge-linux.png
larsrickert, line 1: the grid is broken here

packages/sit-onyx/playwright/snapshots/components/OnyxBottomBar/grid-max-center-xl-edge-linux.png
larsrickert, line 1: The bottom does not consider the grid max width

.changeset/fifty-paws-fix.md
```diff
@@ -0,0 +1,6 @@
+---
+"sit-onyx": major
+"docs": major
+---
+
+feat(Grid-Layout): - changed from media-query to container-query for the main grid and added Tests for the sidebar grid.
```
larsrickert: We should describe more detailled what changed for the user and what he needs to change in his project when updating to this breaking change :)

packages/sit-onyx/playwright/snapshots/components/OnyxNavBar/grid-max-center-xl-edge-linux.png
larsrickert, line 1: The nav bar does not consider the grid max width

.changeset/fifty-paws-fix.md
larsrickert: So if I see correctly:

- renamed onyx-grid-container to onyx-grid-layout
- added new onyx-grid-container class
- changed padding of onyx-grid-layout to have fixed top/bottom spacings
- update grid to consider the current container instead of the viewport for applying breakpoints

packages/sit-onyx/src/components/OnyxBottomBar/OnyxBottomBar.ct.tsx
```diff
@@ -73,12 +73,12 @@ Object.entries(ONYX_BREAKPOINTS).forEach(([breakpoint, width]) => {
 
     /* eslint-disable playwright/no-conditional-in-test, playwright/no-conditional-expect -- prevent useless screenshots for breakpoints that don't have a max width */
     if (width > ONYX_BREAKPOINTS.md) {
-      const app = page.locator(".onyx-app");
+      const gridLayout = page.locator(".onyx-grid-layout");
 
-      await app.evaluate((element) => element.classList.add("onyx-grid-max-md"));
+      await gridLayout.evaluate((element) => element.classList.add("onyx-grid-max-md"));
```
larsrickert: Are the onyx-grid-max-md and onyx-grid-center classes only working when applied to the onyx-grid-layout? Usually they are defined once on app level so it should work to apply them onyx the root element :)

packages/sit-onyx/src/components/OnyxBottomBar/OnyxBottomBar.vue
```diff
@@ -49,6 +49,7 @@ const { densityClass } = useDensity(props);
     bottom: 0;
     width: 100%;
     background-color: var(--onyx-color-base-background-blank);
+    container-type: inline-size;
```
larsrickert, line 52: Shouldn't onyx-grid-container already define the container-type?

packages/sit-onyx/src/components/OnyxPageLayout/OnyxPageLayout.vue
```diff
@@ -49,9 +49,9 @@ provideSkeletonContext(props);
     </div>
 
     <main class="onyx-page__main">
-      <slot v-if="props.noPadding"></slot>
+      <slot v-if="props.noPadding" class="onyx-grid-container"></slot>
```
larsrickert: Why is the grid container only applied when the noPadding os set? Shouldn't we always set it in the onyx-page__main?

Also setting classes on slot does not work since they will be replaced by the slot content :)

packages/sit-onyx/src/components/OnyxPageLayout/examples/GridExample.vue
```diff
@@ -0,0 +1,57 @@
...
+        <div class="onyx-grid-example onyx-grid-md-span-4">md span 4</div>
+        <div class="onyx-grid-example onyx-grid-span-full">full span</div>
+      </div>
+    </OnyxPageLayout>
+  </OnyxAppLayout>
+</template>
+
+<style lang="scss" scoped>
+.onyx-grid-example {
+  background-color: var(--onyx-color-component-cta-default);
```
larsrickert: Jannick requested if we can make to color more subtle so they don't look like buttons :)
Also for the OnyxSidebar example

packages/sit-onyx/src/styles/grid.scss
```diff
@@ -40,14 +40,43 @@ $sidebar-breakpoints: (
...
+    .onyx-grid-container.onyx-grid-xl-20 {
+      --onyx-grid-columns: 20;
+    }
+
+    .onyx-grid.onyx-grid-xl-16,
+    .onyx-grid-layout.onyx-grid-xl-16,
+    .onyx-grid-container.onyx-grid-xl-16 {
+      --onyx-grid-columns: 16;
+    }
+  }
```
larsrickert:
```suggestion
  @include container(min, lg) {
    .onyx-grid,
    .onyx-grid-layout,
    .onyx-grid-container {
      &.onyx-grid-xl-20,
      &.onyx-grid-lg-16 {
        --onyx-grid-columns: 16;
      }
    }
  }

  @include container(min, xl) {
    .onyx-grid,
    .onyx-grid-layout,
    .onyx-grid-container {
      &.onyx-grid-xl-20 {
        --onyx-grid-columns: 20;
      }
    }

    .onyx-grid,
    .onyx-grid-layout,
    .onyx-grid-container {
      &.onyx-grid-xl-16 {
        --onyx-grid-columns: 16;
      }
    }
  }
```

This should do the same + is a little better to read in my opinion :)

packages/sit-onyx/src/styles/grid.scss
```diff
@@ -40,14 +40,43 @@ $sidebar-breakpoints: (
...
   * Page class to apply padding, max-width and optional centered.
   * Is usually placed at the root element of every page.
   */
-  .onyx-grid-container {
+  .onyx-grid-layout {
     // Note: The --onyx-grid-margin is implemented as padding here so that is also included in e.g. custom background colors
     // but the term "margin" comes from UX/Figma so we decided to keep naming it margin to be aligned
-    padding: var(--onyx-grid-margin);
+    padding: var(--onyx-density-3xl) var(--onyx-grid-margin) var(--onyx-density-3xl)
+      var(--onyx-grid-margin);
```
larsrickert:
```suggestion
    padding: var(--onyx-density-3xl) var(--onyx-grid-margin);
```

packages/sit-onyx/src/styles/grid.scss
```diff
@@ -126,13 +135,31 @@ $sidebar-breakpoints: (
     }
   }
 
+  /**
+ * Initializes the grid layout on the `.page__main` element. The layout will adapt
+ * to the width of this wrapper accordingly.
+ */
+  .onyx-page__main {
+    container-type: inline-size;
```
larsrickert: (optional) Would prefer to define this inside the OnyxPageLayout itself :)

packages/sit-onyx/src/styles/grid.scss
```diff
@@ -148,6 +175,10 @@ $sidebar-breakpoints: (
   * SIDEBAR GRID
   */
   .onyx-sidebar {
+    .onyx-grid-layout {
+      padding: var(--onyx-density-md) var(--onyx-grid-margin) var(--onyx-density-md)
+        var(--onyx-grid-margin);
```
larsrickert:
```suggestion
      padding: var(--onyx-density-md) var(--onyx-grid-margin);
```

larsrickert: Open question: Should we add a new CSS variable for the top/bottom padding so its easier to change it if needed?
Maybe something like `--onyx-grid-margin-vertical`?

packages/sit-onyx/src/styles/grid.scss
JoCa96: Is there a benefit to require it to appear together with the `onyx-grid-*` classes?
If so, an explanatory comment would be nice :)

```suggestion
  @include container(min, lg) {
    .onyx-grid-xl-20,
    .onyx-grid-lg-16 {
      --onyx-grid-columns: 16;
    }
  }

  @include container(min, xl) {
    .onyx-grid-xl-20 {
      --onyx-grid-columns: 20;
    }

    .onyx-grid-xl-16 {
      --onyx-grid-columns: 16;
    }
  }
```

packages/sit-onyx/src/styles/grid.scss
```diff
@@ -40,14 +40,43 @@ $sidebar-breakpoints: (
...
+      --onyx-grid-columns: 16;
+    }
+  }
+
   /**
   * Page class to apply padding, max-width and optional centered.
   * Is usually placed at the root element of every page.
   */
-  .onyx-grid-container {
+  .onyx-grid-layout {
```
JoCa96: As a follow-up: We also need to update the [Grid System docs](https://onyx.schwarz/development/grid.html) (in `apps/docs/src/development/grid.md`)

packages/sit-onyx/src/styles/grid.scss
```diff
@@ -126,13 +135,31 @@ $sidebar-breakpoints: (
...
+
   @each $name, $maxVariant, $columns, $gutter, $margin in $gridVariants {
     @if $name == "2xs" {
-      @include generate-grid-properties($columns, $gutter, $margin);
+      @include generate-grid-properties(
+        $columns,
+        $gutter,
+        $margin,
+        ".onyx-grid, .onyx-grid-layout, .onyx-grid-container"
+      );
```
JoCa96: It would be preferable to move the selector around the `@include generate-grid-properties`, instead of passing it as a parameter. That way it is closer to CSS and only parameters are passed that are directly used by the function

```suggestion
      @include generate-grid-properties(
        $columns,
        $gutter,
        $margin,
        ".onyx-grid, .onyx-grid-layout, .onyx-grid-container"
      );
```

packages/sit-onyx/src/styles/grid.scss
```diff
@@ -167,4 +198,21 @@ $sidebar-breakpoints: (
...
+      @include generate-grid-properties($columns, $gutter, $margin);
+    } @else {
+      @include screen(min, $name) {
+        @include generate-grid-properties($columns, $gutter, $margin);
+        @if $maxVariant != none {
+          @include generate-grid-max($name, $maxVariant);
+        }
+      }
+    }
+  }
```
JoCa96: We can skip `2xs` here, as it is mostly covered by the generated CSS for the container query (see the `2xs` condition). Only the root class needs to be added there.
```suggestion
    @each $name, $maxVariant, $columns, $gutter, $margin in $gridVariants {
    @if $name != "2xs" {
      @include screen(min, $name) {
        @include generate-grid-properties($columns, $gutter, $margin);
        @if $maxVariant != none {
          @include generate-grid-max($name, $maxVariant);
        }
      }
    }
  }
```

packages/sit-onyx/src/styles/grid.scss
```diff
@@ -40,14 +40,43 @@ $sidebar-breakpoints: (
     gap: var(--onyx-grid-gutter);
   }
 
+  /**
+  * Grid container classes for "xl" and "lg" breakpoint variants with 16 or 20 columns
+  */
+  @include container(min, lg) {
```
JoCa96, line 48: There is no definition for the `onyx-grid-container` class.
I assume it should be the wrapper for the `onyx-grid`?
So we need something like:

```css
.onyx-grid-container {
    container-type: inline-size;
}
```

JoCa96 (changes requested):

packages/sit-onyx/src/components/OnyxBottomBar/OnyxBottomBar.vue
ChristianBusshoff: No, the container-type has to be a parent of the onyx-grid-container. Otherwise the vars aren't working

packages/sit-onyx/src/components/OnyxPageLayout/OnyxPageLayout.vue
ChristianBusshoff: The main has the onyx-grid-layout.

packages/sit-onyx/src/styles/grid.scss
```diff
@@ -41,47 +41,70 @@ $sidebar-breakpoints: (
...
+    // but the term "margin" comes from UX/Figma so we decided to keep naming it margin to be aligned
+    --onyx-grid-margin-vertical: var(--onyx-density-3xl);
+    padding: var(--onyx-grid-margin-vertical) var(--onyx-grid-margin);
+  }
 
   /**
   * Utility class to allow optional centering for pages with max width.
   * Requires a "onyx-grid-max-*" class to be set.
+  *
+  *  Must be placed inside the desired container. E.g. on Onyx-grid-layout/Onyx-grid-container
```
larsrickert:
```suggestion
  *  Must be placed inside the desired container. E.g. on onyx-grid-layout or onyx-grid-container
```

packages/sit-onyx/src/styles/grid.scss
```diff
@@ -119,6 +140,8 @@ $sidebar-breakpoints: (
 
   /**
   * Grid container class for defining breakpoint specific max width
+  *
+  *  Must be placed inside the desired container. E.g. on Onyx-grid-layout/Onyx-grid-container
```
larsrickert:
```suggestion
  *  Must be placed inside the desired container. E.g. on onyx-grid-layout or onyx-grid-container
```

larsrickert (approved):

Issue #3570, opened 2025-06-18
FoUp: Import Font Variables from Figma


ChristianBusshoff:
### Why?
- provide Font Families
- Font Families: 'Variable' is Missing in Figma Export
  - discussing if the Font Families should be changed in Figma or in Code.
- Letter Spacing must be in px or rem not in %
- Font Weights should have unique names 
- check if variable names must be adjusted due to Figma changes

### Depends on

#3440 
#3215 

### Definition of Done

- Make sure, that
  - [x] follow-up tickets were created if necessary
  - [x] updated version + documentation is deployed
  - [x] Storybook can show the feature
  - [x] Storybook code snippet of new/changed examples are checked that they are generated correctly
  - [x] Namings are aligned with Figma
  - [x] Storybook sidebar badge for "New" is added on component or story level if needed. See: https://github.com/Sidnioulz/storybook-addon-tag-badges/?tab=readme-ov-file

#### Approval

[Storybook]()
[Documentation]()

PR #3689, opened 2025-07-11
Fo up 3570/import font variables from figma

ChristianBusshoff:
Relates to #3570 

- ran figma-pipline with correct varialbes.


larsrickert (approved):
Please also add a changeset for this so its released :)

PR #3696, opened 2025-07-14
chore: migrate all packages to be pure esm packages

JoCa96:
Relates to #3716

- Update all typescript packages to drop common.js build and use `node16` moduleResolution
- Exluded "*.e2e.files" files from nuxt "npm  pack"
- Moved `illustrationScreenshotHooks` to playwright folder, so it's excluded from the build
- Updated deprecated vitest config key `workspace`

### DIFF
--- a/.changeset/shaky-carrots-lay.md
+++ b/.changeset/shaky-carrots-lay.md
@@ -1,0 +1,17 @@
+---
+"@sit-onyx/playwright-utils": major
+"@sit-onyx/storybook-utils": major
+"@sit-onyx/vitepress-theme": major
+"@sit-onyx/chartjs-plugin": major
+"@sit-onyx/figma-utils": major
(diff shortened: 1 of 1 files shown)


apps/demo-app/package.json
```diff
@@ -3,6 +3,11 @@
   "private": true,
   "version": "1.0.0-beta.269",
   "type": "module",
+  "author": "Schwarz IT KG",
+  "license": "Apache-2.0",
+  "engines": {
+    "node": ">=20"
```
larsrickert, line 9: I thought its node >= 18?

JoCa96: We had one package which already required "node >= 20", so imho it's simpler to just have the same everywhere. Also current LTS is v22, so it is reasonable as a requirement :)

Issue #3029, opened 2025-03-27
Implement default sorting function for different column types in OnyxDataGrid


JoCa96:
### Why?

As a user, I expect the default column types to have correct sorting behaviour.
Currently sorting is always done alpha-numerically, if not explicitly overriden.

### Acceptance criteria

- [x] all [`DefaultSupportedTypes`](https://github.com/SchwarzIT/onyx/blob/c795e4a40469e64caab00e719bb0a072a6b8cf2e/packages/sit-onyx/src/components/OnyxDataGrid/features/index.ts#L63) are sorted as expected, e.g.
  - `number` is sorted numerically, so `100` > `9`
  - `date` is numerically by its timestamp, so `new Date("2020-02-01")` > `new Date("2020-01-01")`
- [x] like in the [renderer.ts](https://github.com/SchwarzIT/onyx/blob/c795e4a40469e64caab00e719bb0a072a6b8cf2e/packages/sit-onyx/src/components/OnyxDataGrid/features/renderer.ts#L104) it is required through typing, that every possible `DefaultSupportedTypes` has a defined sorting
- [x] the sortFunction to use is determined by the following priority list, the first defined one is used:
    1. `sortFunc` defined via columnConfig
    2. `sortFunc` defined via custom type[Renderer]Sortings (see *nice-to-have*)
    3. `sortFunc` defined from the default types, if matching
    4. fallback sort function

*nice-to-have*
- [ ] ~the `useSorting`s `SortOptions` is extended to allow the users to define sorting based on a `TypeRenderer` type~

### Implementation details

- The default sortings should located in the `OnyxDataGrid/features/sorting` directory/files.
- the generic [TTypeRenderer](https://github.com/SchwarzIT/onyx/blob/c795e4a40469e64caab00e719bb0a072a6b8cf2e/packages/sit-onyx/src/components/OnyxDataGrid/types.ts#L38) type is passed through to the 

### Definition of Done

- Should be covered by tests:
  - [x] functional tests (Playwright or unit test)
  - [x] visual tests (Playwright screenshots)

- Make sure, that
  - [x] follow-up tickets were created if necessary
  - [x] updated version + documentation is deployed
  - [x] Storybook can show the feature
  - [x] Storybook code snippet of new/changed examples are checked that they are generated correctly
  - [x] Namings are aligned with Figma

#### Approval

[Storybook](https://storybook.onyx.schwarz/?path=/story/data-datagrid-features--sorting)
[...]

PR #3708, opened 2025-07-15
Joca96/3029 feat default sortings data grid

JoCa96:
Relates to #3029 

- implement default sortings for our default types
- add extra `TIME_COMPARE` function, which ignores the date and just compares time

### DIFF
--- a/packages/sit-onyx/src/components/OnyxDataGrid/features/sorting/defaults.ts
+++ b/packages/sit-onyx/src/components/OnyxDataGrid/features/sorting/defaults.ts
@@ -1,6 +1,6 @@
-import type { DataGridEntry } from "../../types.ts";
-import type { DefaultSupportedTypes } from "../index.ts";
-import type { Compare } from "./types.ts";
+import type { DataGridEntry } from "../../types.js";
+import type { DefaultSupportedTypes } from "../index.js";
+import type { Compare } from "./types.js";
 
 export const STRING_COMPARE = (a: unknown, b: unknown, collator: Intl.Collator) =>
   collator.compare(String(a), String(b));
--- a/packages/sit-onyx/src/components/OnyxDataGrid/features/sorting/defaults.ts
+++ b/packages/sit-onyx/src/components/OnyxDataGrid/features/sorting/defaults.ts
@@ -7,12 +7,33 @@ export const STRING_COMPARE = (a: unknown, b: unknown, collator: Intl.Collator) 
 
 export const NUMBER_COMPARE = (a: unknown, b: unknown) => Number(a) - Number(b);
 
+const toComparableTime = (date: Date) =>
+  new Date(
+    Date.UTC(
+      0,
+      0,
+      1,
+      date.getUTCHours(),
+      date.getUTCMinutes(),
+      date.getUTCSeconds(),
+      date.getUTCMilliseconds(),
+    ),
+  );
+export const TIME_COMPARE = (a: unknown, b: unknown, collator: Intl.Collator) => {
+  if (!(a instanceof Date && b instanceof Date)) {
+    return STRING_COMPARE(a, b, collator);
+  }
+  const dateA = toComparableTime(a);
+  const dateB = toComparableTime(b);
+  return NUMBER_COMPARE(dateA, dateB);
+};
+
 export const DEFAULT_COMPARES: Record<PropertyKey, Compare<unknown>> = Object.freeze({
   string: STRING_COMPARE,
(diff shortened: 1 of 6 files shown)


packages/sit-onyx/src/components/OnyxDataGrid/features/sorting/defaults.ts
```diff
@@ -0,0 +1,39 @@
+import type { DataGridEntry } from "../../types.js";
+import type { DefaultSupportedTypes } from "../index.js";
+import type { Compare } from "./types.js";
+
+export const STRING_COMPARE = (a: unknown, b: unknown, collator: Intl.Collator) =>
+  collator.compare(String(a), String(b));
+
+export const NUMBER_COMPARE = (a: unknown, b: unknown) => Number(a) - Number(b);
+
+const toComparableTime = (date: Date) =>
```
larsrickert, line 10: Why is this needed instead of comparing date.getTime() as numbers?

packages/sit-onyx/src/components/OnyxDataGrid/features/sorting/types.ts
```diff
@@ -48,4 +48,9 @@ export type SortOptions<TEntry extends DataGridEntry> = DataGridFeatureOptions<
    * The currently applied sorting. Will be updated by the data grid, can be used for reading, updating and watching the applied sorting.
    */
   sortState?: MaybeRef<SortState<TEntry>>;
+  /**
+   * The `Intl.Collator` used for the default (string-based) sorting.
+   * Defaults to `new Intl.Collator(i18n.locale.value, { numeric: true })` where `i18n` is the OnyxI18n instance.
+   */
+  collator?: MaybeRef<Intl.Collator>;
```
larsrickert, line 57: For which use cases would you want to override the collator?

packages/sit-onyx/src/components/OnyxDataGrid/features/sorting/sorting.ts
```diff
@@ -98,10 +109,17 @@ export const useSorting = <TEntry extends DataGridEntry>(options?: SortOptions<T
...
-      watch: [sortState, intlCompare],
+      watch: [sortState, i18n.locale],
       mutation: {
         func: sortData,
       },
+      modifyColumns: [
+        {
+          // save final column config
+          func: (_finalConfig) => (finalConfig = _finalConfig),
+          order: Infinity,
```
larsrickert: The higher the order, the earlier its executed. Don't we want to save the config as late as possible so all column modifications are already applied? So negative infinity in this case

packages/sit-onyx/src/components/OnyxDataGrid/features/sorting/sorting.ts
```diff
@@ -98,10 +109,17 @@ export const useSorting = <TEntry extends DataGridEntry>(options?: SortOptions<T
 
     return {
       name: SORTING_FEATURE,
-      watch: [sortState, intlCompare],
+      watch: [sortState, i18n.locale],
```
larsrickert:
```suggestion
      watch: [sortState, collator],
```

Don't we want to watch the collator instead because it could also be passed as ref by the project

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxDataGrid/features/sorting/defaults.ts
JoCa96: Added a comment; Basically, we only show the time part, sowe only sort by the time part here.

packages/sit-onyx/src/components/OnyxDataGrid/features/sorting/types.ts
JoCa96: There are a lot of tweaking options for the collator: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator/Collator . E.g. You can consider casing, punctuation or accents. Gives people an easy way to customize the sorting.

packages/sit-onyx/src/components/OnyxDataGrid/features/sorting/sorting.ts
JoCa96: done

packages/sit-onyx/src/components/OnyxDataGrid/features/sorting/sorting.ts
JoCa96: done

PR #3761, opened 2025-07-21
feat(OnyxDataGrid-Filtering): removed custom filter config from Storybook example

ChristianBusshoff:
- removed custom filter config

Relates to #3714


.changeset/free-poets-run.md
```diff
@@ -0,0 +1,5 @@
+---
+"sit-onyx": minor
+---
+
+fix(OnyxDataGrid-Filtering): removed filting config from storybook example
```
larsrickert: The change only affects the docs so no changeset is needed here because there is nothing to release :)

larsrickert (approved):

PR #3774, opened 2025-07-22
 fix(background): Fix missing background color on fullscreen stories and sometimes on general page

JoCa96:


packages/sit-onyx/.storybook/manager.ts
```diff
@@ -2,6 +2,10 @@ import type { Badge, TagBadgeParameters } from "storybook-addon-tag-badges";
 import { addons } from "storybook/internal/manager-api";
 
 addons.setConfig({
+  toolbar: {
```
larsrickert, line 5: Is this redundant? Since the same config is also applied in the storybook-utils preview.ts

JoCa96: Good point, seems like the `parameters.toolbar.hidden` one was redundant, when this is in the manager is set, too.

larsrickert: Alright, isn't this technically a breaking change then? Because the project needs to manually set this config too, otherwise the background toolbar icon will be shown, right?

JoCa96: done

Issue #907, opened 2024-04-15
Implement improved `isPrintableCharacter` util


JoCa96:
```ts
const GRAPHEME_SEGMENTER = new Intl.Segmenter();

/**
 * Check if the `key` property of a KeyboardEvent is a printable character.
 *
 * There is no standardized or specified algorithm to check for [named keys](https://www.w3.org/TR/uievents-key/#named-key-attribute-values) vs printable characters.
 * For this check we assume that any `key` value that is a single Grapheme is a printable characters.
 * This way we can ensure that combining characters and emojis are correctly recognized without need to keep a list of all named key.
 */
export const isPrintableCharacter = (key: string) =>
  [...GRAPHEME_SEGMENTER.segment(key)].length === 1;

```

PR #3803, opened 2025-07-25
feat(isPrintableCharacter): implement Segmenter Intl API 

ChristianBusshoff:
Relates to #907

- implement Segmenter Intl API


packages/headless/src/utils/keyboard.ts
```diff
@@ -21,331 +21,14 @@ export const wasKeyPressed = (event: KeyboardEvent, key: PressedKey) => {
   );
 };
 
+const GRAPHEME_SEGMENTER = new Intl.Segmenter();
```
JoCa96: Just to be safe: We should add the locale explicitly, so it behaves the same, and doesn't depend on the users locale setting.
```suggestion
const GRAPHEME_SEGMENTER = new Intl.Segmenter("en-US");
```

.changeset/cold-mangos-count.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sit-onyx/headless": minor
+---
+
+feat(isPrintableCharacter): implement Segmenter Intl API
```
JoCa96: We should consider the readers perspective:
- Is the changeset description understandable without having to check the PR or source code?
- Does the description convey the impact of the change for the reader?

.changeset/cold-mangos-count.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sit-onyx/headless": minor
+---
+
+feat(isPrintableCharacter): implemented Segmenter Intl API for detection of printable keyboard characters, replacing the static NAMED_KEYS_SET lookup.
```
JoCa96:
```suggestion
feat(isPrintableCharacter): making now use of the Segmenter Intl API for detection of printable keyboard characters, replacing the static NAMED_KEYS_SET lookup.
```

JoCa96 (approved):

Issue #3271, opened 2025-05-07
Implement aligned OnyxModalDialog


JoCa96:
### Why?

We currently have multiple components with similar names, props and principles.

### Open Questions

- [x] Decide on the structure:
https://css-tricks.com/clarifying-the-relationship-between-popovers-and-dialogs/

### Acceptance criteria

- [x] The support component `OnyxDialog` is renamed to `OnyxBasicDialog`
- [x] Rename OnyxAlertDialog to OnyxAlertModal
- [x] Rename OnyxModalDialog to OnyxModal
- [x] Rename OnyxDialog to OnyxBasicDialog
- [x] Rename OnyxPopover to OnyxBasicPopover

#### Popover

- menu
- listbox
- tree
- grid
- dialog

##### dialog

- Modal: A dialog with an backdrop and focus trapping
- Non-Modal: A dialog with neither an overlay nor focus trapping
- Alert Dialog: A dialog that alerts screen readers when shown. It can be either modal or non-modal.

### Definition of Done

- Make sure, that
  - [x] follow-up tickets were created if necessary
  - [x] updated version + documentation is deployed
  - [x] Storybook can show the feature
  - [x] Storybook code snippet of new/changed examples are checked that they are generated correctly
  - [x] Namings are aligned with Figma

#### Approval

Storybooks:
- https://storybook.onyx.schwarz/?path=/docs/support-basicdialog--docs
- https://storybook.onyx.schwarz/?path=/docs/support-basicpopover--docs
- https://storybook.onyx.schwarz/?path=/docs/feedback-modal--docs
- https://storybook.onyx.schwarz/?path=/docs/feedback-alertmodal--docs

- Checked and approved by
  - [x] designer @jannick-ux 
  - [x] dev @larsrickert

larsrickert:
My suggestion would instead be:

- rename current `OnyxDialog` to `OnyxSupportDialog`
- rename current `OnyxModalDialog` to `OnyxDialog` + support new `modal` prop that is false by default and can be enabled to behave like the previous OnyxModalDialog
- OnyxAlertDialog is kept as it is

PR #3811, opened 2025-07-28
feat!: Rename Popover, Dialog, Modal components for better separation and better parity with designs

JoCa96:
Relates to #3271 

- Renamed `OnyxPopover` to `OnyxBasicPopover`.
- Renamed `OnyxDialog` to `OnyxBasicDialog`.
- Renamed `OnyxModalDialog` to `OnyxModal`.
- Renamed `OnyxAlertDialog` to `OnyxAlertModal`.
- Renamed `disableClosingOnBackdropClick` prop to `nonDismissible` for `OnyxBasicDialog` and `OnyxModal`.
- Added `role` prop to `OnyxBasicPopover` (Defaults to `dialog`).
- Removed most instances of the `close` emit, where it was used in combination with the `open` prop. Instead the `open` prop now has `v-model` support.


packages/sit-onyx/src/components/OnyxBasicDialog/OnyxBasicDialog.vue
```diff
@@ -77,17 +77,17 @@ useOutsideClick({
...
       densityClass,
       'onyx-truncation-multiline',
-      props.alignment !== 'center' ? `onyx-dialog--${props.alignment}` : '',
+      props.alignment !== 'center' ? `onyx-basic-dialog--${props.alignment}` : '',
     ]"
     :aria-modal="props.modal"
     :aria-label="props.label"
     :role="props.alert ? 'alertdialog' : undefined"
-    @cancel.prevent="emit('close')"
+    @cancel.prevent="props.nonDismissible || emit('close')"
```
larsrickert: Should we really not emit the close when pressing Escape when nonDismissible is set?

Imho its the same as just not listening to the close event in the project. I only see the use case where we do not emit the close event on backdrop click.

packages/sit-onyx/src/components/OnyxBasicDialog/OnyxBasicDialog.vue
```diff
@@ -96,22 +96,22 @@ useOutsideClick({
 <style lang="scss">
 @use "../../styles/mixins/layers.scss";
 
-.onyx-dialog {
+.onyx-basic-dialog {
```
larsrickert, line 112: Question: Should we remove all default styles here and make the dialog really "basic"?

I mean we can still keep the max size but maybe we want to remove the other visual styles?

packages/sit-onyx/src/components/OnyxBasicDialog/OnyxBasicDialog.vue
JoCa96: I would like to enforce a consistent behavior here. Imho if you are not able to close it by backdrop click, it should also not close on escape.
Anyways maybe, we can streamline the API here more: What about removing the `close` event and provide `open` as a v-model?

larsrickert: Sounds good 👍

JoCa96: Done

packages/sit-onyx/src/components/OnyxBasicDialog/OnyxBasicDialog.vue
JoCa96: @larsrickert I would prefer the minimal styles we currently have over the default browser styles:

Screenshot 2025-08-06 at 16 11 06
Screenshot 2025-08-06 at 16 10 57
Screenshot 2025-08-06 at 16 10 02
Screenshot 2025-08-06 at 16 09 47

packages/sit-onyx/src/components/OnyxBasicDialog/OnyxBasicDialog.stories.ts
```diff
@@ -20,22 +21,52 @@ const meta: Meta<typeof OnyxDialog> = {
 };
 
 export default meta;
-type Story = StoryObj<typeof OnyxDialog>;
+type Story = StoryObj<typeof OnyxBasicDialog>;
 
 export const Default = {
```
larsrickert, line 26: Should we create a .vue file for the Storybook example instead?

packages/sit-onyx/src/components/OnyxModal/examples/DefaultExample.vue
```diff
@@ -9,7 +9,7 @@ const isOpen = ref(false);
   <div>
     <OnyxButton label="Show modal" @click="isOpen = true" />
 
-    <OnyxModalDialog label="Example modal dialog" :open="isOpen" @close="isOpen = false">
+    <OnyxModal v-model:open="isOpen" label="Example modal dialog" non-dismissible>
```
larsrickert:
```suggestion
    <OnyxModal v-model:open="isOpen" label="Example modal dialog">
```

Shouldn't we support dismissable for the default example? Otherwise users might think that it is not supported

packages/sit-onyx/src/components/OnyxSidebar/OnyxSidebar.vue
```diff
@@ -86,6 +85,9 @@ const isModalOpen = computed<boolean>({
   },
   set: (newVal: boolean) => {
     _isModalOpen.value = newVal;
+    if (!newVal) {
+      emit("close");
```
larsrickert, line 89: Should we also switch to v-model:open for the Sidebar? Can be done in a follow up PR ofc

larsrickert (approved):

Issue #3778, opened 2025-07-23
Implement refactoring of icon and flag lib utility functions


larsrickert:
Preparation for #1802 because its a breaking change so we want to implement it before version 1.

Implement for the icon and the flag libary

Issue #1802, opened 2024-08-28
Implement icons as JavaScript exports


larsrickert:
### Open Questions/To-dos

- Should we sanitize icon names to valid JavaScript variable names (e.g. `_24h` instead of `24h`) or should they be renamed in Figma instead so the name is consistent between UX and DEV?
- Should we still offer the "raw" .svg files in the npm package or just the JavaScript exports (bundle size)?

### Depends on

- #939

### Why?

Currently the user has to go to our [icon docs](https://onyx.schwarz/icons.html) and copy the `?raw` import for every icon which is not the best developer experience.

We want to support importing all our icons as JavaScript variables which has the following benefits:

- autocompletion / easier import
- easier to import all icons (no glob needed)
- other departments depend on this (e.g. STACKIT) in order to migrate to our `@sit-onyx/icons` package

### Acceptance criteria

- [ ] all available icons (SVG content) can be imported as JavaScript variables, e.g. `import { trash } from "@sit-onyx/icons";`
- [ ] it is possible to import all icons, e.g. using `import * as ALL_ICONS from "@sit-onyx/icons";`
- [ ] the existing exports like `groupIconsByCategory` etc. are moved to `@sit-onyx/icons/utils` so the main package only includes icons
- [ ] `@sit-onyx/icons` supports both ESM and CJS

### Definition of Done

- covered by tests
  - [ ] functional tests (Playwright or unit test)
- [ ] follow-up tickets were created if necessary
- [ ] updated version + documentation is deployed
- [ ] Storybook can show the feature
- [ ] Storybook code snippet of new/changed examples are checked that they are generated correctly
- [ ] Namings are aligned with Figma

**Approval**

- Storybook 
- Deployed Docs 
- [ ] approved by designer @jannick-ux
- [ ] approved by dev 

### Implementation details

We could build `@sit-onyx/icons` with Vite and implement a custom Vite plugin + virtual module for converting the icons, e.g.:

```ts
// vite.config.ts
import { fileURLToPath } from "node:url";
import { defineConfig } from "vite";
import svgToTsPlugin from "./plugins/svg-to-ts";

export default defineConfig({
  plugins: [svgToTsPlugin()],
  build: {
    lib: {
      entry: getFilePath("./src/index.ts"),
      formats: ["es", "cjs"],
      fileName: "index",
    },
  },
});

/** Gets the given path while ensuring cross-platform and correct decoding */
function getFilePath(path: string) {
[...]

PR #3830, opened 2025-07-30
feat: support icons and flags as JavaScript exports

larsrickert:
Relates to #1802, #3778

- add vite plugin that is used when building the icon and flag library that supports all icons as JavaScript exports
- the exported icons will be prefixed with "icon", followed by the pascal case icon name
- the exported flags will be prefixed with "flag", followed by the uppercase flag name

I will update the docs and Storybook code snippet generation in a follow up PR

## Checklist


packages/icons/vite.config.ts
```diff
@@ -0,0 +1,44 @@
...
+import { defineConfig } from "vite";
+import dts from "vite-plugin-dts";
+import { viteStaticCopy } from "vite-plugin-static-copy";
+import { capitalize } from "./src/utils.js";
+
+// https://vitejs.dev/config
+export default defineConfig({
+  ...VITE_BASE_CONFIG,
+  plugins: [
+    vue(),
```
JoCa96: There doesn't seem to be any need for the vue plugin:

```suggestion
import { VITE_BASE_CONFIG } from "@sit-onyx/shared/vite.config.base";
import svg from "@sit-onyx/vite-plugin-svg";
import { fileURLToPath, URL } from "node:url";
import { DiagnosticCategory } from "typescript";
import { defineConfig } from "vite";
import dts from "vite-plugin-dts";
import { viteStaticCopy } from "vite-plugin-static-copy";
import { capitalize } from "./src/utils.js";

// https://vitejs.dev/config
export default defineConfig({
  ...VITE_BASE_CONFIG,
  plugins: [
```

packages/icons/src/index.ts
```diff
@@ -1,63 +1,2 @@
...
-};
-
-/**
- * Capitalizes the first character of the given string.
- */
-export const capitalize = (value: string) => {
-  return value.charAt(0).toUpperCase() + value.slice(1);
-};
+// this virtual import is handled by the @sit-onyx/vite-plugin-svg and will export all SVG icons as JavaScript exports
+export * from "virtual:vite-plugin-svg";
```
JoCa96, line 2: Nice approach with the vite plugin :) But I was wondering if an less opinionated API would be possible using vite query. E.g.:

```ts
import iconCalendar from "./calendar.svg?ts-inline";
```

Or for this use-case here:

```ts
const icons = import.meta.glob("./assets/*.svg", {
  eager: true,
  query: "?ts-inline",
});

export default icons;
```
This way vite would also take care of the file loading/resolving and it would be usable in any vite project.

packages/vite-plugin-svg/src/index.ts
```diff
@@ -0,0 +1,100 @@
...
+
+          return { fileName, exportName };
+        });
+
+      const exports = await Promise.all(
+        files.map(async ({ fileName, exportName }) => {
+          const svgContent = await this.fs.readFile(path.join(options.input, fileName), {
+            encoding: "utf8",
+          });
+          return `export const ${exportName} = \`${svgContent}\``;
```
JoCa96: I would have hoped that we use this ticket to switch to an approach, that doesn't require `v-html` to be used. 😅

JoCa96 (changes requested):

packages/vite-plugin-svg/src/index.ts
larsrickert: Whats the issue with using v-html? :)
To be honest, I'd prefer passing the icons as full "svg" since this is a very flexible API so the user can use any icon library or custom icons (e.g. exported from their Figma design), place them in their project and use seamlessly use them with all onyx components

packages/icons/src/index.ts
larsrickert: Them main issue with this is that there is no type information for the icons when using the glob import, so intellisense does not work which is one of the main reasons why we want to offer the icons as JS exports imo :)

But we can use the import query inside the vite plugin so we don't have to read the file content ourselves

packages/vite-plugin-svg/src/index.ts
JoCa96: `v-html` is susceptible to XSS, so I would like to avoid it's usage.

packages/icons/src/index.ts
JoCa96: The example above is only about the api here, for the icon type generation :) so the plugin would still take care of the type generation. So the output would stay identical to the current one.

JoCa96: But with the proposed API change, it would also possible to use the plugin independently and for single imports.

JoCa96 (approved):

PR #3832, opened 2025-07-31
fix(nuxt): always provide onyx plugin

larsrickert:
Previously, the onyx plugin was only registered if Nuxt I18n was used.
This caused the issue that the router integration did not work when not using i18n.

## Checklist

### DIFF
--- a/packages/nuxt/src/runtime/plugins/onyx.ts
+++ b/packages/nuxt/src/runtime/plugins/onyx.ts
@@ -2,32 +2,30 @@ import { computed, type ComputedRef } from "#imports";
 import type { LocaleObject } from "@nuxtjs/i18n";
 import { defineNuxtPlugin, useNuxtApp, useRouter } from "nuxt/app";
 import { createOnyx, type TranslationFunction } from "sit-onyx";
-import type { Composer } from "vue-i18n";
+import { type Composer } from "vue-i18n";
 
 type I18n = Composer & { localeProperties: ComputedRef<LocaleObject> };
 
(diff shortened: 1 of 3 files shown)


packages/nuxt/src/runtime/plugins/onyx.ts
```diff
@@ -1,33 +1,36 @@
...
+import { createOnyx, type ProvideI18nOptions, type TranslationFunction } from "sit-onyx";
 import type { Composer } from "vue-i18n";
 
 type I18n = Composer & { localeProperties: ComputedRef<LocaleObject> };
 
 export default defineNuxtPlugin({
   name: "onyx:plugin",
-  // @ts-ignore: dependency comes from @nuxtjs/i18n
-  dependsOn: ["i18n:plugin"],
+  dependsOn: [],
```
markbrockhoff, line 11: If using i18n is optional removing it from the dependencies should be fine. However, we need to be careful not to run into race conditions later on. Nuxt already allowed for plugins to be initialized in parallel. So far sequential initialization is still the default but without the dependency things could break if it's switched to parallel by default. E.g. if the i18n plugin is applied after the onyx one (just due to the given race condition).

I'd suggest to explicitly set `parallel: false` to avoid running into this situation in the future. For more information see: https://nuxt.com/docs/4.x/guide/directory-structure/app/plugins#loading-strategy

larsrickert: Thanks for the hint!

PR #3843, opened 2025-08-01
chore(playwright): add emit test helper

JoCa96:
- Added `createEventSpy` and `expectEventCall` for easier palywright testing of emits/events.


packages/playwright-utils/src/emits/types.ts
```diff
@@ -0,0 +1,31 @@
...
+ * ```ts
+ * import MyInput from "./MyInput.vue";
+ * import type { ComponentExposed } from "vue-component-type-helpers";
+ *
+ * type MyInputPublicProps = ComponentExposed<typeof MyInput>["$props"];
+ * type MyInputVModelEmits = VModelEmits<MyInputPublicProps>; // => { 'onUpdate:modelValue': (newValue: string) => void }
+ *
+ * ```
+ */
+export type PickEmitsFromProps<T> = {
```
larsrickert, line 18: Isn't there also a ComponentEmits type from vue-component-type-helpers?

packages/playwright-utils/src/emits/index.ts
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * Asserts the emits recorded from the spy created by `createEmitSpy`.
+ * Expects the spy to have recorded `n` calls, and `matches` to match the arguments of the last call.
+ * @see `createEmitSpy` documentation for example usage.
+ */
+export const expectEmit = <Handler extends { [EMIT_SPY_SYMBOL]: unknown[][] }>(
+  emitSpy: Handler,
+  n: number,
+  matches: unknown[] | Record<string, unknown>,
```
larsrickert:
```suggestion
  n: number,
  matches: unknown[] | Record<string, unknown>,
```

Can't we combine the second and third parameter? So that you just pass in an array of expected data and the length of the array is the expected n?

packages/playwright-utils/src/emits/index.ts
larsrickert, line 1: Isn't this file the same as packages/sit-onyx/src/playwright/events.ts?

larsrickert (changes requested):

packages/playwright-utils/src/emits/index.ts
JoCa96: yes, decided to move it to the `playwright-utils` package after I created PR. Removed it from `sit-onyx` now.

packages/playwright-utils/src/emits/types.ts
JoCa96: Good question, will add a comment about that!
Basically `ComponentEmit` (from *vue-component-type-helpers*) provides the emits as intersection, which does not allow to extract single types from the intersection.
The custom `ComponentEmitHandler` extracts the emits from the Props:
```ts
type A = ComponentEmit<typeof OnyxColorSchemeDialog>;
/* 
   ((event: "update:modelValue", value: ColorSchemeValue) => void) 
   & ((event: "update:open", value: Nullable<boolean>) => void) 
*/
type B = ComponentEmitHandler<typeof OnyxColorSchemeDialog>;
/* 
{ 
  "onUpdate:modelValue"?: ((value: ColorSchemeValue) => any) | undefined;
  "onUpdate:open"?: ((value: Nullable<boolean>) => any) | undefined;
}
*/
```

packages/playwright-utils/src/emits/index.ts
JoCa96: Done

packages/playwright-utils/src/emits/index.ts
JoCa96: I like this approach, because it allows readable step-by-step testing, e.g.:

```tsx
// ACT
  await clickOption("Light");
  await component.getByRole("button", { name: "Apply" }).click();

  // ASSERT
  expect(updateModelValueEvents).toStrictEqual(["dark", "auto", "light"]);
  expectEmit(onUpdateOpen, 3, [false]);

  // ACT
  await component.getByRole("button", { name: "Cancel" }).click();

  // ASSERT
  expectEmit(onUpdateOpen, 4, [false]);

  // ACT
  await page.keyboard.press("Escape");

  // ASSERT
  expectEmit(onUpdateOpen, 5, [false]);
```

PR #3845, opened 2025-08-03
chore: setup trusted publishing for npm packages

JoCa96:
Relates to [3844](https://github.com/SchwarzIT/onyx/issues/3844)

The **npm** registry now supports [Trusted publishing for npm packages](https://github.blog/changelog/2025-07-31-npm-trusted-publishing-with-oidc-is-generally-available/).
This allows packages to be published using oidc, without the need for long-living API tokens.
Also each new published version is then directly coupled to a pipeline run and commit, which is also shown on a new tile in the npm UI.

See also the [official docs](https://docs.npmjs.com/trusted-publishers).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,6 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     environment: ${{ github.ref_name == 'main' && 'prod' || 'dev' }}
     permissions:
+      id-token: write
       contents: write
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
larsrickert, line 1: Don't we still need the node-setup step to install node, pnpm and dependencies?

larsrickert: We can just remove the NODE_AUTH_TOKEN parts from .github/templates/node-setup/action.yml, right?

JoCa96: good point! Don't know what I was thinking 😅

PR #3851, opened 2025-08-04
chore: Fix release pipeline failing auth

JoCa96:
Use latest version of npm. *pnpm* uses npm under the hood for publishing.
For publishing securely always the latest npm version should be used.

### DIFF
--- a/.github/templates/node-setup/action.yml
+++ b/.github/templates/node-setup/action.yml
@@ -13,6 +13,8 @@ runs:
         node-version-file: .node-version
         cache: pnpm
 
+    # *pnpm* uses *npm* under the hood for publishing.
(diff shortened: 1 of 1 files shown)


.github/templates/node-setup/action.yml
```diff
@@ -13,6 +13,10 @@ runs:
         node-version-file: .node-version
         cache: pnpm
 
+    - name: Keep npm up-to-date
```
larsrickert, line 18: Should we add a small comment why this is needed so we are aware of it in the future?

larsrickert (approved):

.github/templates/node-setup/action.yml
JoCa96: done

PR #3857, opened 2025-08-05
fix(playground): Fix "failed to resolve module specifier" error caused by missing import-map entries

JoCa96:


apps/playground/src/composables/useStore.ts
```diff
@@ -53,6 +53,8 @@ export const useStore = () => {
         mergeImportMap(importMap.value, {
           imports: {
             "sit-onyx": `https://cdn.jsdelivr.net/npm/sit-onyx@${onyxVersion.value}/dist/index.js`,
+            "@sit-onyx/icons": `https://cdn.jsdelivr.net/npm/@sit-onyx/icons@beta/dist/index.js`,
```
larsrickert, line 56: We could also add the flags lib here so users can also use flags in the Playground

larsrickert (approved):

apps/playground/src/composables/useStore.ts
JoCa96: done

Issue #3865, opened 2025-08-06
Implement i18n integration for Nuxt module documentation template


larsrickert:
### Why?

The Nuxt documentatiopn template feature set and developer experience should be improved.

### Acceptance criteria

- [x] supports i18n
- [x] allow and document how to override any component using Nuxt layer imports

### Definition of Done

- Should be covered by tests:
  - [x] functional tests (Playwright or unit test)

- Make sure, that
  - [x] updated version + documentation is deployed

#### Approval

[i18n documentation](https://onyx.schwarz/development/packages/nuxt-docs.html#i18n)
[Customization documentation](https://onyx.schwarz/development/packages/nuxt-docs.html#customization)

- Checked and approved by
  - [x] designer @jannick-ux 
  - [x] dev @ChristianBusshoff

PR #3866, opened 2025-08-06
feat(nuxt)!: auto register onyx locales

larsrickert:
Relates to #3865

Remove the `registerOptions` property from the Nuxt module as it involved manual user effort although it can be automated.
The module will now detect which locales the project is using/defining and register the corresponding onyx locales accordingly.

## Checklist


packages/nuxt/src/module.ts
```diff
@@ -72,13 +59,33 @@ export default defineNuxtModule<ModuleOptions>({
...
+
+          const language = locale.language ?? locale.code;
+
+          try {
+            await stat(resolve(`./runtime/locales/${language}.js`));
+            onyxLocalesToRegister[locale.code] = language;
+          } catch {
+            // noop, file does not exist in onyx, so we don't need to register it
+          }
+        }
```
JoCa96: The filesystem access is currently performed synchronously (because of the await in the for-loop). For a better performance this could be done asynchronously, e.g.:

```suggestion
      const registerOnyxLocales: NuxtI18nModuleHooks["i18n:registerModule"] = async (register) => {
        // auto-detect onyx locales to register
        const onyxLocalesToRegister: Record<string, string> = {};

        const localRegistrations =
          nuxt.options.i18n?.locales?.map(async (projectLocale) => {
            const locale =
              typeof projectLocale === "string"
                ? { code: projectLocale, language: projectLocale }
                : projectLocale;

            const language = locale.language ?? locale.code;

            await stat(resolve(`./runtime/locales/${language}.js`));

            onyxLocalesToRegister[locale.code] = language;
          }) ?? [];

        await Promise.allSettled(localRegistrations);
```

Issue #3828, opened 2025-07-30
Implement file cards for the OnyxFileUpload


larsrickert:
### Why?

The new [OnyxFileCard](https://storybook.onyx.schwarz/?path=/docs/cards-filecard--docs) component should be integrated in the [OnyxFileUpload](https://storybook.onyx.schwarz/?path=/docs/form-elements-fileupload--docs) so that the selected files of the upload are shown in a list.

### Design

[Figma](https://www.figma.com/design/j7Hy7Zk8MiCmJBJf0ArQRm/Onyx-Documentation?node-id=16307-41970&m=dev)

### Acceptance criteria

- [x] the currently selected file(s) of the upload are shown with file cards
- [x] by default, each file card shows a. remove button that removes the file from the selection
- [x] the file cards and actions can be fully customized/overridden by the project if needed
- the file card list supports the following size behaviors:
    - [x] all cards are shown beneath each other without height limitation (default)
    - [x] max height can be set so the list becomes scrollable
    - [x] a "Show/hide all files" system button can be shown (hidden by default) which can be clicked to show/hide the file list
    - [x] a system button can be shown to open a custom modal with the file list (responsibility of the project!)

#### Out of scope

- [ ] file card supports showing a progress bar
- [ ] file cards are considered in form validations, meaning if a file is invalid (e.g. too large), the form should not be submittable (same behavior as e.g. the input)

### Implementation details

<-- Add technical details / implementation proposals if necessary, e.g.:
  - API proposal
  - relevant web APIs
  - learnings from other implementations
  - code snippets
  -->

### Definition of Done

- The following component requirements are implemented:
  - [x] skeleton
  - [x] density
  - [x] example usage added to `apps/demo-app/src/views/HomeView.vue`

- Should be covered by tests:
  - [x] functional tests (Playwright or unit test)
  - [x] visual tests (Playwright screenshots)

- Make sure, that
  - [x] follow-up tickets were created if necessary
  - [x] updated version + documentation is deployed
  - [x] Storybook can show the feature
  - [x] Storybook code snippet of new/changed examples are checked that they are generated correctly
  - [x] Namings are aligned with Figma
  - [x] Storybook sidebar badge for "New" is added on component or story level if needed. See: https://github.com/Sidnioulz/storybook-addon-tag-badges/?tab=readme-ov-file

#### Approval

[Storybook](https://storybook.onyx.schwarz/?path=/docs/form-elements-fileupload--docs)

- Checked and approved by
[...]

PR #3894, opened 2025-08-11
feat(OnyxFileUpload): implement file cards 

ChristianBusshoff:
Relates to #3828 

- add a file-card-area where all file Cards are displayed.

### DIFF
--- a/packages/sit-onyx/src/components/OnyxFileUpload/OnyxFileUpload.ct.tsx
+++ b/packages/sit-onyx/src/components/OnyxFileUpload/OnyxFileUpload.ct.tsx
@@ -113,6 +113,7 @@ test("should select a single file", async ({ mount, page }) => {
 
   // ASSERT
   await expect(() => expect(file).toBeDefined()).toPass();
+  await expect(component).toHaveScreenshot("one-file-selected.png");
 });
 
 test("should select multiple files", async ({ mount, page }) => {
@@ -120,15 +121,21 @@ test("should select multiple files", async ({ mount, page }) => {
   let files: File[] = [];
 
   const component = await mount(
-    <OnyxFileUpload multiple onUpdate:modelValue={(newFiles) => (files = newFiles)} />,
+    <OnyxFileUpload
+      multiple
+      maxHeight={"12rem"}
+      onUpdate:modelValue={(newFiles) => (files = newFiles)}
+    />,
   );
   const button = component.getByRole("button", { name: "Click to select" });
+  const removeFirstFileButton = component.getByRole("button", { name: "Delete file" }).first();
 
   // ACT
   selectFiles(page, button, 2);
 
   // ASSERT
   await expect(() => expect(files).toHaveLength(2)).toPass();
+  await expect(component).toHaveScreenshot("multiple-file-selected.png");
 
   // ACT
   files = [];
@@ -141,6 +148,13 @@ test("should select multiple files", async ({ mount, page }) => {
 
   // ASSERT
   await expect(() => expect(files).toHaveLength(4)).toPass();
+  await expect(component).toHaveScreenshot("max-height.png");
+
+  // ACT
+  await removeFirstFileButton.click();
+
+  // ASSERT
+  await expect(() => expect(files).toHaveLength(3)).toPass();
 });
 
 test("should replace files", async ({ mount, page }) => {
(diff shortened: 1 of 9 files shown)


packages/sit-onyx/src/components/OnyxFileUpload/OnyxFileUpload.ct.tsx
```diff
@@ -198,3 +212,71 @@ test("should not support drag and drop when disabled", async ({ mount, page }) =
   // ASSERT
   await expect(() => expect(file).not.toBeDefined()).toPass();
 });
+
+test("should has hide button", async ({ mount, page }) => {
```
larsrickert:
```suggestion
test("should have hide button", async ({ mount, page }) => {
```

packages/sit-onyx/src/components/OnyxFileUpload/OnyxFileUpload.vue
```diff
@@ -182,6 +231,70 @@ const handleDragEnter = () => {
...
+                  () => {
+                    removeFile(file);
+                  }
+                "
+              />
+            </template>
+          </OnyxFileCard>
+        </template>
+      </slot>
+    </div>
```
larsrickert:
```suggestion
    <div
      class="onyx-file-area"
      :class="[
        'onyx-file-upload__list',
        { 'onyx-file-upload__list--max-height': props.listType === 'max-height' },
      ]"
    >
      <OnyxSystemButton
        v-if="props.listType === 'button' && currentFiles.length"
        class="onyx-file-upload__list-button"
        :label="hideFiles ? t('fileUpload.revealFilesButton') : t('fileUpload.hideFilesButton')"
        @click="hideFiles = !hideFiles"
      />

      <template v-if="!hideFiles">
        <template v-for="(file, index) in currentFiles" :key="file.name">
          <slot :file :status="fileStatuses[index]">
            <OnyxFileCard :filename="file.name" :size="file.size" :status="fileStatuses[index]">
              <template #actions>
                <OnyxIconButton
                  color="danger"
                  :icon="iconTrash"
                  label="Delete file"
                  @click="removeFile(file)"
                />
              </template>
            </OnyxFileCard>
          </slot>
        </template>
      </template>
    </div>
```

I'd suggest a little simpler API so basically applying these changes to the current API:
- add property `listType` (or some other speaking name) which can e.g. be "list", "max-height" or "button" to differentiate the multiple display styles
- put the slot only around the OnyxFileCard itself + remove the OnyxFileCard props from the OnyxFileUpload since they are file individual. If the user wants to add custom actions or add any custom file card prop, he should use the slot to render the OnyxFileCard manually
- translate the "Remove file" label
- for the max height I'd prefer a CSS variable over applying inlines styles. Ideally, we have a `--onyx-file-upload-max-files` CSS variable which is a numeric count of how many files should be shown without scrolling. Then we internally use `calc()` to calculate the max-height accordingly :) So its easy for the user to control the height
- nice to have / optional for this PR: It would be nice if we add a default link to the file card (can be a data URL) that opens the file content. Since the files are all JavaScript "File" objects, we should be able to get the file content and create a URL / link out of it

packages/sit-onyx/src/components/OnyxFileUpload/types.ts
```diff
@@ -1,16 +1,18 @@
...
 export type OnyxFileUploadProps<TMultiple extends boolean> = DensityProp &
-  Pick<SharedFormElementProps, "name"> & {
+  Pick<SharedFormElementProps, "name"> &
+  OnyxFileArea & {
     /**
      * Currently selected file(s).
      * If `multiple` property is enabled, this value is an array, otherwise a single file.
      */
-    modelValue?: TMultiple extends true ? File[] : File;
+    modelValue?: TMultiple extends true ? File[] : File | null;
```
larsrickert: We can use our Nullable type here :)

packages/sit-onyx/src/components/OnyxFileUpload/OnyxFileUpload.vue
```diff
@@ -3,12 +3,12 @@
   setup
   generic="
     TMultiple extends boolean = false,
-    TModelValue extends TMultiple extends true ? File[] : File = TMultiple extends true
+    TModelValue extends TMultiple extends true ? File[] : File | null = TMultiple extends true
       ? File[]
-      : File
+      : File | null
```
larsrickert: We can use our Nullable type here :)

packages/sit-onyx/src/components/OnyxFileUpload/OnyxFileUpload.vue
```diff
@@ -59,7 +73,32 @@ const modelValue = useVModel<TModelValue, "modelValue", typeof props, undefined>
 
 const input = useTemplateRef<HTMLInputElement>("inputRef");
 
-const currentFiles = computed<File[]>(() => asArray(modelValue.value));
+const currentFiles = computed<File[]>(() => asArray(modelValue.value ?? []) as unknown as File[]);
```
larsrickert: Null and undefined values should be filtered out instead if using the type cast

packages/sit-onyx/src/components/OnyxFileUpload/OnyxFileUpload.vue
```diff
@@ -59,7 +73,32 @@ const modelValue = useVModel<TModelValue, "modelValue", typeof props, undefined>
...
+const hideFiles = ref(false);
+
+const fileStatuses = computed((): (FileCardStatus | undefined)[] => {
```
larsrickert, line 87: We should use `useFileSize` for formatting the size inside the translated message so they are human readable strings instead of big binary numbers

Issue #3898, opened 2025-08-11
Implement number formatting for OnyxStepper


JoCa96:
### Why?

To increase readability for big numbers, the numbers should be formatted according to the users locale.

### Depends on

- https://github.com/SchwarzIT/onyx/discussions/3593

### Acceptance criteria

- [x] When the `OnyxStepper` input is blurred, the modelValue number is displayed formatted
  - [x] If decimal places are support, they are shown (e.g. `10,00`)
  - [x] If decimal places are not supported, they are not shown (e.g. `10`)
  - [x] For large numbers thousand separators are shown according to the locale (e.g. `10.000,00`)
- [x] While the `OnyxStepper` is focused the number is **not** formatted
- [x] You can still focus he input by 
  - [x] tabbing
  - [x] clicking
- [x] also supports custom formatting

### Implementation details

Easiest solution is probably to overlay the formatted number over the input when it's **not** focused.
When focused, we hide the overlay for the time of the focusing.

Otherwise we would need to switch the `type=number` to `type=text` for the input, which will drastically increase the effort necessary.
So the actual value in the input should not be touched, instead we show the formatted number over it.

The input needs to stay focusable by clicking, for this `pointer-events: none;` can be used.

For ARIA we should make this new overlay `aria-hidden="true"`

### Definition of Done

- Should be covered by tests:
  - [x] functional tests (Playwright or unit test)
  - [x] visual tests (Playwright screenshots)

- Make sure, that
  - [x] follow-up tickets were created if necessary
  - [x] updated version + documentation is deployed
  - [x] Storybook can show the feature
  - [x] Storybook code snippet of new/changed examples are checked that they are generated correctly

#### Approval

[Storybook](https://storybook.onyx.schwarz/?path=/story/form-elements-stepper--formatted-number)

- Checked and approved by
  - [x] designer @jannick-ux 
  - [x] dev @JoCa96

PR #3939, opened 2025-08-18
Feat(OnyxStepper): implement number formatting

ChristianBusshoff:
Relates to #3898

- added formatNumber prop to the stepper. Can set to true (1000 => 1.000) or can be a custom function


packages/sit-onyx/src/components/OnyxStepper/OnyxStepper.vue
```diff
@@ -95,17 +119,20 @@ const handleClick = (direction: "stepUp" | "stepDown") => {
   const stepValue = (direction === "stepUp" ? 1 : -1) * props.stepSize;
   const newValue = parseFloat(getFormattedValue.value(currentValue + stepValue));
   modelValue.value = applyLimits(newValue, props.min, props.max);
+  displayValue.value = getDisplayValue.value(applyLimits(newValue, props.min, props.max));
 };
 
 const handleChange = () => {
   wasTouched.value = true;
-  if (inputValue.value == undefined) {
+  if (inputValue.value === undefined || inputValue.value === null || inputValue.value === "") {
```
JoCa96:
```suggestion
  if (!inputValue.value) {
```

packages/sit-onyx/src/components/OnyxStepper/OnyxStepper.vue
```diff
@@ -95,17 +119,20 @@ const handleClick = (direction: "stepUp" | "stepDown") => {
   const stepValue = (direction === "stepUp" ? 1 : -1) * props.stepSize;
   const newValue = parseFloat(getFormattedValue.value(currentValue + stepValue));
   modelValue.value = applyLimits(newValue, props.min, props.max);
+  displayValue.value = getDisplayValue.value(applyLimits(newValue, props.min, props.max));
```
JoCa96:
```suggestion
  const limitedValue = applyLimits(newValue, props.min, props.max);
  modelValue.value = limitedValue;
  displayValue.value = getDisplayValue.value(limitedValue);
```

packages/sit-onyx/src/components/OnyxStepper/OnyxStepper.vue
```diff
@@ -84,9 +84,33 @@ const getFormattedValue = computed(() => {
...
+    }
+    return String(value);
+  };
+});
 
 watchEffect(() => {
-  inputValue.value = getFormattedValue.value(modelValue.value);
+  const formattedNumber = getFormattedValue.value(modelValue.value);
+  inputValue.value = formattedNumber;
+  if (formattedNumber) displayValue.value = getDisplayValue.value(parseFloat(formattedNumber));
```
JoCa96: We always seem to update `displayValue` after modelValue is updated/changed - Is there a reason why, we can't use a `computed`?

```ts
const displayValue = computed(() => getDisplayValue.value(modelValue.value));
```

This would also allow us to simplify the logic around applying `displayValue`

packages/sit-onyx/src/components/OnyxStepper/OnyxStepper.ct.tsx
```diff
@@ -555,3 +555,21 @@ test("should format zero with precision", async ({ mount }) => {
...
+      formatNumber={true}
+      style={{ width: "16rem" }}
+    />,
+  );
+  const input = component.getByLabel("Label");
+  const display = component.locator(".onyx-stepper__display");
+  // // ASSERT
+  await expect(input).toHaveValue("1000000");
+  await expect(display).toHaveText("1,000,000");
+  await expect(page).toHaveScreenshot("formatNumber.png");
```
JoCa96: The screenshots contain a lot of whitespace, we can reduce the size by only considering the component itself for the screenshot.
```suggestion
  await expect(component).toHaveScreenshot("formatNumber.png");
```

packages/sit-onyx/src/components/OnyxStepper/OnyxStepper.ct.tsx
```diff
@@ -555,3 +555,21 @@ test("should format zero with precision", async ({ mount }) => {
   await expect(input).toHaveValue("0.00");
   await expect(() => expect(modelValue).toBe(0)).toPass();
 });
+
+test("should format with formatNumber", async ({ page, mount }) => {
```
JoCa96: Please add also a test for decimal numbers

packages/sit-onyx/src/components/OnyxStepper/OnyxStepper.vue
```diff
@@ -84,9 +84,33 @@ const getFormattedValue = computed(() => {
...
+    if (value == undefined) return "";
+    if (props.formatNumber) {
+      if (typeof props.formatNumber === "boolean") {
+        if (props.precision) {
+          return value.toLocaleString(undefined, {
+            minimumFractionDigits: props.precision,
+            maximumFractionDigits: props.precision,
+          });
+        }
+        return value.toLocaleString();
```
JoCa96: We should use the `locale` from `injectI18n` to use the user language:

```suggestion
          return value.toLocaleString(locale.value, {
            minimumFractionDigits: props.precision,
            maximumFractionDigits: props.precision,
          });
        }
        return value.toLocaleString(locale.value);
```

JoCa96 (changes requested):

packages/sit-onyx/src/components/OnyxStepper/OnyxStepper.vue
ChristianBusshoff: Can't use !inputValue.value because 0 is a falsy value and would not pass the validation.

JoCa96 (approved):

PR #3965, opened 2025-08-21
fix: use correct types for emitted v-model events

larsrickert:
Relates to #

Some v-model "update:" events where typed incorrectly which also allowed null / undefined values.
This made it harder to work the emitted values because nullish checks had to be implemented by the user, although the actually emitted value is always defined.

The following events are updated:

| Component         | Event name            | New type | Old type           |
| ----------------- | --------------------- | -------- | ------------------ |
| OnyxAccordion     | "update:modelValue"   | TValue[] | Nullable<TValue[]> |
| OnyxBasicDialog   | "update:open"         | boolean  | Nullable<boolean>  |
| OnyxCheckboxGroup | "update:modelValue"   | TValue[] | Nullable<TValue[]> |
| OnyxFAB           | "update:open"         | boolean  | Nullable<boolean>  |
| OnyxFilterTag     | "update:active"       | boolean  | Nullable<boolean>  |
| OnyxInput         | "update:modelValue"   | string   | Nullable<string>   |
| OnyxMiniSearch    | "update:modelValue"   | string   | Nullable<string>   |
| OnyxFlyoutMenu    | "update:open"         | boolean  | Nullable<boolean>  |
| OnyxMenuItem      | "update:open"         | boolean  | Nullable<boolean>  |
| OnyxNavItem       | "update:open"         | boolean  | Nullable<boolean>  |
| OnyxUserMenu      | "update:flyoutOpen"   | boolean  | Nullable<boolean>  |
| OnyxProgressSteps | "update:highestValue" | number   | Nullable<number>   |
| OnyxSwitch        | "update:modelValue"   | boolean  | Nullable<boolean>  |
| OnyxTextarea      | "update:modelValue"   | string   | Nullable<string>   |

## Checklist


packages/sit-onyx/src/composables/useVModel.ts
```diff
@@ -41,17 +51,23 @@ export type UseVModelOptions<
...
+    | undefined
+    | (TValue extends PrimitiveType | undefined ? TValue : () => TValue),
+  TComputed extends TDefaultValue extends undefined | null
+    ? TValue
+    : NonNullable<TValue> = TDefaultValue extends undefined | null ? TValue : NonNullable<TValue>,
+  TEmit extends (evt: `update:${TKey}`, value: TComputed) => void = (
     evt: `update:${TKey}`,
-    value: TValue,
+    value: TComputed,
   ) => void,
```
JoCa96: (optional): If you want to reduce the length and duplication a bit, we could make the extends part less specific, but keep the defaults.
This will work the same, as the defaults are the important part, as they infer the types for us.

Also for line 5.

```suggestion
  TProps extends object,
  TKey extends keyof TProps & string,
  TValue extends TProps[TKey] = TProps[TKey],
  TDefaultValue = undefined | (TValue extends PrimitiveType | undefined ? TValue : () => TValue),
  TComputed extends Nullable<TValue> = TDefaultValue extends undefined | null
    ? TValue
    : NonNullable<TValue>,
  TEmit extends (...args: any[]) => void = (evt: `update:${TKey}`, value: TComputed) => void,
```

larsrickert: Nice :) Done!

Issue #2765, opened 2025-02-14
Implement|Fix|Define Implement OnyxSegmentedControl


larsrickert:
### Why?

We want to offer a component to switch between different values (similar to tabs but with a different design/purpose)

### Design

[Figma](https://www.figma.com/design/YfEUBOHk4J4nYrk04geswG/Onyx-Component-Library?node-id=10299-104743&m=dev)

### Acceptance criteria

- [ ] the design is implemented
- [ ] text, text + icon or only icon can be displayed per item

### Implementation details

- the component should technically be a radio button group

### Applicable ARIA Pattern

[Radio group pattern](https://www.w3.org/WAI/ARIA/apg/patterns/radio/)

### Definition of Done

- The following component requirements are implemented:
  - [ ] skeleton
  - [ ] density
  - [ ] example usage added to `apps/demo-app/src/views/HomeView.vue`

- Should be covered by tests:
  - [ ] functional tests (Playwright or unit test)
  - [ ] visual tests (Playwright screenshots)

- Make sure, that
  - [ ] follow-up tickets were created if necessary
  - [ ] updated version + documentation is deployed
  - [ ] Storybook can show the feature
  - [ ] Storybook code snippet of new/changed examples are checked that they are generated correctly
  - [ ] Namings are aligned with Figma

#### Approval

[Storybook]()
[Documentation]()

- Checked and approved by
  - [ ] designer 
  - [ ] dev

mj-hof:
@jannick-ux do we need icon+text as figma component?

PR #3989, opened 2025-08-25
feat(OnyxSegmentedControl): implement onyx segmented control

ChristianBusshoff:
Relates to #2765

- added Segmented Control
- added Segmented Control Element
- added tests


packages/sit-onyx/src/components/OnyxSegmentedControlElement/OnyxSegmentedControlElement.vue
larsrickert, line 1: We should implement this component as [radio group](https://www.w3.org/WAI/ARIA/apg/patterns/radio/) which is the correct aria pattern :)

Like this we also don't need the active state as JavaScript since we can use the CSS native `:checked` state

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxSegmentedControlElement/OnyxSegmentedControlElement.vue
larsrickert: Also I'd prefer to use a similar API as for the OnyxRadioGroup so where the options are passed as property

.changeset/weak-papers-clean.md
```diff
@@ -0,0 +1,5 @@
+---
+"sit-onyx": minor
+---
+
+feat(OnyxSegmentedControl): implemented OnyxSegmentedControl & ONyxSegmentedCOntrolElement
```
JoCa96:
```suggestion
feat(OnyxSegmentedControl): implemented new components: OnyxSegmentedControl and OnyxSegmentedControlElement
``

File: packages/sit-onyx/src/components/OnyxSegmentedControl/types.ts [hunk 2]
```diff
@@ -0,0 +1,42 @@
...
+  /**
+   * unique identifier for the option
+   */
+  value: string;
+  /**
+   * Label for the option.
+   */
+  label?: string;
+  /**
+   * Icon for the option. If an icon widthout a label is provided the width of tie control will be `fit-content`
```
JoCa96:
```suggestion
   * Icon for the option. If an icon without a label is provided the width of the control will be `fit-content`.
```

packages/sit-onyx/src/components/OnyxSegmentedControl/OnyxSegmentedControl.vue
```diff
@@ -0,0 +1,119 @@
...
+
+const skeleton = useSkeletonContext(props);
+const { densityClass } = useDensity(props);
+
+const modelValue = computed({
+  get: () => props.modelValue,
+  set: (value: string | symbol) => {
+    emit("update:modelValue", value);
+  },
+});
```
JoCa96: We can improve the behavior here by using our `useVModel` composable:

```suggestion
const modelValue = useVModel<OnyxSegmentedControlProps, "modelValue">({
  props,
  emit,
  key: "modelValue",
});
```

packages/sit-onyx/src/components/OnyxSegmentedControl/OnyxSegmentedControl.vue
```diff
@@ -0,0 +1,119 @@
...
+      &:has(.onyx-segmented-control-element__input:checked) {
+        background-color: var(--onyx-color-base-background-blank);
+        font-weight: var(--onyx-font-weight-semibold);
+      }
+      &:has(.onyx-segmented-control-element__input:focus-visible) {
+        outline: var(--onyx-outline-width) solid var(--outline-color);
+      }
+      &:has(.onyx-segmented-control-element__input:disabled) {
+        color: var(--onyx-color-text-icons-neutral-soft);
+        cursor: default;
```
JoCa96: default `cursor` is not applied, because the cursors are applied to different elements

packages/sit-onyx/src/components/OnyxSegmentedControl/types.ts
```diff
@@ -0,0 +1,42 @@
...
+   */
+  options: OnyxSegmentedControlOption[];
+  /**
+   * Same as the native `name` attribute of `<input>`.
+   * Used to reference the input in JavaScript or in submitted form data.
+   */
+  name?: string;
+};
+
+export type OnyxSegmentedControlOption = {
```
JoCa96: Please define the type similar to `packages/sit-onyx/src/components/OnyxRadioButton/types.ts`. 
More importantly, we should always require the `label`, but also have a `hideLabel` property for a11y.

packages/sit-onyx/src/components/OnyxSegmentedControl/OnyxSegmentedControl.vue
```diff
@@ -0,0 +1,119 @@
...
+      box-sizing: border-box;
+
+      border-radius: var(--onyx-radius-sm);
+      color: var(--onyx-color-text-icons-neutral-medium);
+      font-weight: var(--onyx-font-weight-regular);
+      font-family: var(--onyx-font-family);
+      line-height: var(--onyx-font-line-height-md);
+      width: 100%;
+
+      &__label {
```
JoCa96: according to figma, a gap is missing between label and icon

JoCa96 (approved):

PR #4029, opened 2025-08-27
Feat(Segmented Control): made modelValue required

ChristianBusshoff:
- made modelValue required


.changeset/sad-guests-study.md
```diff
@@ -0,0 +1,5 @@
+---
+"sit-onyx": patch
```
larsrickert:
```suggestion
"sit-onyx": major
```

This is a breaking change since the user must provide the value

packages/sit-onyx/src/components/OnyxSegmentedControl/types.ts
```diff
@@ -1,13 +1,12 @@
...
 import type { SkeletonInjected } from "../../composables/useSkeletonState.js";
 import type { BaseSelectOption } from "../../types/components.js";
-import type { Nullable } from "../../types/utils.js";
 
 export type OnyxSegmentedControlProps = DensityProp & {
   /**
    * value of the active control element
    */
-  modelValue?: Nullable<string>;
+  modelValue: string;
```
larsrickert, line 9: Not related to this PR but I noticed that the emits are not typed correctly for the OnyxSegmentedControl, it should be:

```ts
const emit = defineEmits<{
  /**
   * Emitted when the selected value changes.
   */
  "update:modelValue": [value: string];
}>();
```

larsrickert (changes requested):

larsrickert (approved):

Issue #4013, opened 2025-08-27
[Bug]: OnyxFileUpload not validating file type


ProgrammingFox:
### Version

1.0.0-beta.306

### Description

When including an OnyxFileUpload with the "accept" attribute inside an OnyxForm the file type will not be validated. One can simply drag and drop a file onto the OnyxFileUpload and it will be shown without issue. When submitting the form no validation errors are shown and the form will count as valid.

Expected would be a validation error from the component that the file does not match the given files types specified in the "accept" attribute of the OnyxFileUpload and the OnyxForm will not validate.

### Reproduction steps

In the following playground drag and drop a file not specified in the valid types (.jpg for example) and submit. The form will validate and the submit variable will switch to true.

[...]

PR #4036, opened 2025-08-28
feat(OnyxFileUpload): add basic file type validation

JoCa96:
Relates to #4013


packages/sit-onyx/src/i18n/locales/en-US.json
larsrickert, line 1: Please also add German translations :)

packages/sit-onyx/src/utils/file.ts
```diff
@@ -0,0 +1,39 @@
...
+ * @param file The file to validate.
+ * @param accept The accepted file types as array.
+ * @returns `true` if the file type is accepted, `false` otherwise.
+ */
+export const validateFileType = (
+  { name: _name, type: _type }: File,
+  accept?: FileType[],
+): boolean => {
+  // While "*/*" is not a valid value for the `accept` attribute, we still handle it here, just to be sure.
+  if (!accept || accept.length === 0 || accept.some((a: string) => a.trim() === "*/*")) {
```
larsrickert, line 16: Why do we check for `*/*` if it can never exist?

larsrickert (approved):

packages/sit-onyx/src/i18n/locales/en-US.json
JoCa96: done

packages/sit-onyx/src/utils/file.ts
JoCa96: It's a thing I have seen used sometimes and I wanted to be nice to the JS users :)

PR #4043, opened 2025-08-28
chore(docs): Updated technical guidelines and .browserlistrc

JoCa96:
- revisited technical guidelines, updated based on discussions and experiences
- removed some points, because they are not relevant anymore, have been implemented or tickets exists for them
- updated `.browserlistrc` based on our actual target browsers
- updated stylelint ignores that don't apply anmyore
- updated vitepress cache function in `apps/docs/src/cached.ts` to use a hash instead of cleaned url, because the Url became too long


.browserslistrc
```diff
@@ -1,4 +1,17 @@
 # Browsers that we support
+# The big three, up to 1 year old
 
-> 1%, last 2 versions, Firefox ESR,
-not dead, not op_mini all, not KaiOS < 4.0, not and_qq > 0, not and_uc > 0
+# CHROME - monthly release
```
larsrickert, line 4: Should/can we also include chromium browsers? Or popular ones like edge?

apps/docs/src/cached.ts
```diff
@@ -1,20 +1,15 @@
 /* eslint-disable no-console -- we want to be able to log output to the console */
 import { mkdir, readFile, writeFile } from "fs/promises";
+import * as crypto from "node:crypto";
```
larsrickert:
```suggestion
import { hash } from "node:crypto";
```

We only need the hash function right?

apps/docs/src/principles/technical-guidelines.md
```diff
@@ -36,86 +23,76 @@ Our technical vision describes the goals that we as the development team of **on
...
 1. 📜 The DX _(Developer Experience)_ of our users is our top priority
-2. 📜 We only support modern technologies [(ES202x)](https://tc39.es/ecma262/)
-3. 📜 **onyx** components work in all of the three big browsers: Chromium, Firefox and Safari
-4. 📜 We offer excellent typing support
-5. 🔮 We only use Web APIs that have a ["Can I Use"](https://caniuse.com/) score of > X%
+2. 📜 We offer excellent typing support
+3. 📜 We only support modern technologies [(ES202x)](https://tc39.es/ecma262/)/ESM
+4. 📜 **onyx** components must work in the three big browser (engines): Chromium, Firefox and Safari
+   - Support is limited to versions up to one year old
+5. 📜 We only use Web APIs that match our [browserlist](https://github.com/SchwarzIT/onyx/blob/joca96/chore-update-principles/.browserslistrc)
```
larsrickert:
```suggestion
5. 📜 We only use Web APIs that match our [browserlist](http://localhost:5173/development/#browser-support)
```

We have a nicer representation of the browserlistrc in our docs :) Or at least we should point to the main branch instead of your feature branch

apps/docs/src/principles/technical-guidelines.md
```diff
@@ -36,86 +23,76 @@ Our technical vision describes the goals that we as the development team of **on
...
 ### Naming conventions
 
 1. 📜 All (public) components are prefixed with `Onyx`, e.g. `OnyxButton`
 2. 📜 We use [v-model](https://vuejs.org/guide/components/v-model.html) for two-way-data-binding wherever possible
 3. 📜 Component variants are named after what they are, not where they are used (e.g. a link variant "fullWidth" instead of "sideBar")
-4. 📜 Event names won't get an `onyx` prefix
-5. 📜 Boolean properties are prefixed with `is-` or `has-`. Exceptions are made to align with default html properties, e.g. `required`
-6. 📜 All CSS class names are prefixed with `onyx-`
-7. 📜 All CSS variables are prefixed with `--onyx-`
+4. 📜 All CSS class names are prefixed with `onyx-`
```
larsrickert, line 63: Should we move the first bullet point here too? So that all "are prefixed with" bullet points are grouped?

stylelint.config.js
larsrickert, line 1: Not necessarily this PR but:
What about the other ignored rules? Are we able to implement them differently?

larsrickert (changes requested):

apps/docs/src/principles/technical-guidelines.md
JoCa96: Does vitepress replace the `http://localhost:5173/` 😅 ? Probably not, right?

larsrickert: No but you can just set it to `/development/#browser-support` and it will then always link correctly :) I forgot this in the suggestion sorry

JoCa96: no worries^^

.browserslistrc
JoCa96: Differences between Chrome and Chromium are almost non-existent (I think). But, Edge is very popular, especially in enterprise contexts, so seems like a good idea to have it: Increases our browserlist 
*Audience coverage* and is also shown in the docs :)

JoCa96: Done

stylelint.config.js
JoCa96: Afaik most of these are only minor visual things, but would need to investigate in detail to tell you more

larsrickert (approved):

Issue #3982, opened 2025-08-22
Implement user detail page for demo application


larsrickert:
### Why?

Our demo application should be extended with a user detail page.

### Depends on

#3981 

### Design

[Figma](https://www.figma.com/design/YG8K4K7VnH6Dux2WeSnCS2/Onyx-Demo-Application?node-id=39-12544&m=dev)

### Acceptance criteria

- a sidebar is displayed that links to the following sub-views
   -  Personal data (default)
        - [x] user details are displayed
        - [x] a form is shown to edit the user data (does not actually edit anything / does not save the data)
        - [x] the form uses some validations (required, min/max length) to showcase form validation support
        - [x] uses leading and trailing slots of inputs (e.g. phone icon etc.)
   - Roles
        - [x] Displays several user roles as cards with radio buttons that can be checked
    - Documents
         - [x] shows a data grid with already uploaded files (shows some dummy files initially). Supports pagination, filtering and sorting
         - [x] a file upload is supported to upload new files (just adds the file to the data grid, does not actually upload)
         - [x] when selecting files, a upload is simulated by showing a increasing progress bar for x seconds. When the progress is done, the files are added to the data grid and removed from the file upload

### Definition of Done

- Make sure, that
  - [x] updated version + documentation is deployed

#### Approval

[Demo app](https://demo.onyx.schwarz/forms)

- Checked and approved by
  - [x] designer @flubnau 
  - [x] dev @ChristianBusshoff

PR #4054, opened 2025-08-29
feat(demo-app): implement user profile form page

larsrickert:
Relates to #3982

- Implement user detail page with a form to enter personal data

## Checklist


packages/flags/vite.config.ts
```diff
@@ -33,6 +33,9 @@ export default defineConfig({
       formats: ["es"],
       fileName: (format, entryName) => `${entryName}.js`,
     },
+    // using terser instead of esbuild because it has export name conflicts when used inside Nuxt
+    // because inside utils.js, a variable is named "h" which conflicts with the Nuxt auto import from "vue"
+    minify: "terser",
```
JoCa96: While this fixes the issue, it doesn't seem like this solves the underlying issue.
The minifier/bundler shouldn't matter, as the produced code should act the same for all of them.

`h` itself is not exported and scoped to the module, so it can not interfere with the global scope.

Based on the following error message, it seems like `utils.js` is not loaded as a module, but as a path. Maybe this is configurable somehow? Or could be a weird interaction between pnpm workspaces and nuxt.
```
demo-app:build:  ERROR  Nuxt Build Error: ../../packages/flags/dist/utils.js (1:515): Identifier "h" has already been declared
```

larsrickert: I found the issue cause + fix, see: https://github.com/SchwarzIT/onyx/pull/4054/files#diff-7af1e14a4c9f8c5c99264f2d2c8c39117e55b2eae37407812a6d7544313b44cc

JoCa96 (approved):

Issue #4003, opened 2025-08-26
[Bug]: Fix jumping line height for data grid header


larsrickert:
### Version

1.0.0-beta.304

### Description

The table / data grid header height jumps around when showing/hiding a system button. The height should not change.

We could change the line height to md instead

#3678 

### Reproduction steps

https://storybook.onyx.schwarz/?path=/story/data-datagrid--default

### Confirmations

- [x] I have checked that there is not already a bug report for my issue.
- [x] I confirm that are no sensitive information included in this bug report. e.g. credentials, internal project details (for employees of the Schwarz Group) etc.

PR #4057, opened 2025-09-01
fix(OnyxTable): increase header min height

larsrickert:
Relates to #4003

- slightly increased the tables min height so the header height does not change when e.g. a system button or icon is shown/hidden inside the header.
- center header content vertically

## Checklist


ChristianBusshoff (approved):

JoCa96 (changes requested):
sec

packages/sit-onyx/src/components/OnyxTable/OnyxTable.vue
```diff
@@ -202,6 +202,8 @@ $border: var(--onyx-1px-in-rem) solid var(--onyx-color-component-border-neutral)
 
 .onyx-table {
   @include layers.component() {
+    --onyx-table-padding-vertical: var(--onyx-density-xs);
```
JoCa96: The counterpart of `inline` is `block`:

```suggestion
    --onyx-table-padding-block: var(--onyx-density-xs);
```

packages/sit-onyx/src/components/OnyxTable/OnyxTable.vue
```diff
@@ -223,7 +225,7 @@ $border: var(--onyx-1px-in-rem) solid var(--onyx-color-component-border-neutral)
     th,
     td {
       position: relative;
-      padding: var(--onyx-density-xs) var(--onyx-density-md);
+      padding: var(--onyx-table-padding-vertical) var(--onyx-table-padding-inline);
```
JoCa96: imho, a simpler solution would suffice:
- no need for CSS variables
- no need for `calc()`

```suggestion
      position: relative;
      height: 2.5rem;
      padding: 0 var(--onyx-density-md);
      align-content: center;
      vertical-align: middle;
```

JoCa96 (changes requested):

packages/sit-onyx/src/components/OnyxTable/OnyxTable.vue
larsrickert: But this does not scale with densities or custom paddings :)
Thats why I used calc instead so the height adapts automatically for all densities

JoCa96: Ah, good point! I think in this case we can just use `box-sizing: content-box`:
```css
      padding: var(--onyx-density-xs) var(--onyx-density-md);
      box-sizing: content-box;
      height: var(--onyx-spacing-lg);
      align-content: center;
```

JoCa96 (approved):

packages/sit-onyx/src/components/OnyxTable/OnyxTable.vue
larsrickert: Nope that doesn't work (see http://localhost:6006/?path=/story/data-datagrid-features--selection) since `th` is not a regular box/block element.

I'd stick to the calc solution if thats also fine for you :)

Issue #3983, opened 2025-08-22
Implement chart page for demo application


larsrickert:
### Why?

Our demo application should be extended with a page that showcases some charts using chart.js and the onyx chart.js plugin.

### Depends on

- #3969 

### Design

[Figma](https://www.figma.com/design/YG8K4K7VnH6Dux2WeSnCS2/Onyx-Demo-Application?node-id=42-9929&m=dev)

### Acceptance criteria

- [ ] some KPI cards are shown
- [ ] a start / end date can be selected to show the data for
- [ ] some example charts are displayed (mock data). Example chart types can be found in the [onyx docs](https://onyx.schwarz/development/packages/chartjs-plugin.html#example-charts) or [chart.js docs](https://www.chartjs.org/docs/latest/samples/information.html)

### Definition of Done

- Should be covered by tests:
  - [ ] functional tests (Playwright or unit test)
  - [ ] visual tests (Playwright screenshots)

- Make sure, that
  - [ ] follow-up tickets were created if necessary
  - [ ] updated version + documentation is deployed

#### Approval

[Documentation](https://demo.onyx.schwarz/charts)

- Checked and approved by
  - [x] designer 
  - [x] dev @larsrickert

PR #4061, opened 2025-09-01
feat(demo-App): implement chart page

ChristianBusshoff:
Relates to #3983

- implemented cart page

### DIFF
--- a/apps/demo-app/app/components/KPICard.vue
+++ b/apps/demo-app/app/components/KPICard.vue
@@ -1,0 +1,59 @@
+<script lang="ts" setup>
+import { iconExpandWindow } from "@sit-onyx/icons";
+import { OnyxCard, OnyxIcon, OnyxTag, type OnyxColor } from "sit-onyx";
+
+const props = defineProps<{
+  headline: string;
+  value: number;
+  color: OnyxColor;
+  percent: number;
+  icon: string;
+}>();
+</script>
+
+<template>
+  <OnyxCard class="card" clickable>
+    <div class="card__header">
+      {{ props.headline }}
+      <OnyxIcon :icon="iconExpandWindow" />
+    </div>
+
+    <div class="card__content">
+      <span class="card__value"> {{ props.value }} </span>
+      <OnyxTag
+        :label="props.percent + ' %'"
+        :icon="props.icon"
+        density="compact"
+        :color="props.color"
+      />
+    </div>
+  </OnyxCard>
+</template>
+
+<style lang="scss" scoped>
+.card {
+  width: 16rem;
+
+  &__header {
+    color: var(--onyx-color-text-icons-neutral-medium);
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+    gap: var(--onyx-card-gap);
+    width: 100%;
+  }
+
+  &__content {
+    display: flex;
+    align-items: center;
+    flex-wrap: wrap;
+    gap: var(--onyx-density-md);
+  }
+
+  &__value {
+    font-size: 3rem;
+    font-weight: var(--onyx-font-weight-semibold);
+    line-height: 2.75rem;
+  }
+}
+</style>
--- a/apps/demo-app/app/pages/charts.vue
+++ b/apps/demo-app/app/pages/charts.vue
@@ -1,7 +1,204 @@
-<script lang="ts" setup></script>
+<script lang="ts" setup>
+import { registerOnyxPlugin } from "@sit-onyx/chartjs-plugin";
+import { iconArrowSmallDown, iconArrowSmallRight, iconArrowSmallUp } from "@sit-onyx/icons";
+import type { ChartData } from "chart.js";
+import { Chart, registerables } from "chart.js";
+import { OnyxHeadline } from "sit-onyx";
+import { Bar, Line, Pie } from "vue-chartjs";
+import { useI18n } from "vue-i18n";
+
+Chart.register(...registerables);
+registerOnyxPlugin(Chart);
+
+const { t } = useI18n();
+
+/**
+ * Charts
+ */
+
+const barChartData: ChartData<"bar"> = {
+  labels: ["01.01.2024", "01.02.2024", "01.03.2024", "01.04.2024", "01.05.2024"],
+  datasets: [
+    {
+      label: t("charts.productLines.a"),
+      data: [15000, 18000, 16500, 20000, 19500],
+      borderWidth: 2,
+    },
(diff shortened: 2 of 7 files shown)


apps/demo-app/package.json
```diff
@@ -18,14 +18,17 @@
     "@nuxtjs/color-mode": "^3.5.2",
     "@nuxtjs/i18n": "^10.0.6",
     "@pinia/nuxt": "~0.11.2",
+    "@sit-onyx/chartjs-plugin": "1.0.0-beta.299",
```
larsrickert:
```suggestion
    "@sit-onyx/chartjs-plugin": "workspace:^",
```

You also need to run pnpm instsall afterwards to update the lock file

apps/demo-app/i18n/locales/de-DE.json
```diff
@@ -13,7 +13,42 @@
   "action": "Aktion | Aktionen",
   "dataGrid": "Data Grid / Tabelle",
   "forms": "Formulare",
-  "charts": "Diagramme",
+  "charts": {
```
larsrickert, line 15: The charts translation key is missing on the home page
image

apps/demo-app/app/components/KPICard.vue
```diff
@@ -0,0 +1,59 @@
...
+        density="compact"
+        :color="props.color"
+      />
+    </div>
+  </OnyxCard>
+</template>
+
+<style lang="scss" scoped>
+.card {
+  width: 16rem;
```
larsrickert: Since we are using a grid layout we should remove the custom width so they adapt correctly

apps/demo-app/app/pages/charts.vue
```diff
@@ -1,7 +1,210 @@
...
+Chart.register(...registerables);
+registerOnyxPlugin(Chart);
+
+const { t } = useI18n();
+
+/**
+ * Charts
+ */
+
+const barChartData: Ref<ChartData<"bar">> = computed(() => {
```
larsrickert:
```suggestion
const barChartData = computed<ChartData<"bar">>(() => {
```

We should use the generic for better typing : ) Same for the other computeds in this file

apps/demo-app/app/pages/charts.vue
```diff
@@ -1,7 +1,210 @@
...
+]);
+const chartOptions = { maintainAspectRatio: false };
+</script>
 
 <template>
-  <OnyxHeadline is="h1">{{ $t("charts") }}</OnyxHeadline>
+  <OnyxHeadline is="h1">{{ t("charts.pageName") }}</OnyxHeadline>
+
+  <OnyxHeadline is="h3" class="headline">{{ t("charts.rangeAndType") }}</OnyxHeadline>
+  <OnyxForm>
```
larsrickert:
```suggestion
  <OnyxForm @submit.prevent>
```

Prevent submit so no full page reload is triggered when clicking the submit button

apps/demo-app/app/pages/charts.vue
```diff
@@ -1,7 +1,210 @@
...
+      </OnyxCard>
+    </div>
+  </div>
 </template>
 
-<style lang="scss" scoped></style>
+<style lang="scss" scoped>
+.headline {
+  margin-top: var(--onyx-density-xl);
+  margin-bottom: var(--onyx-density-sm);
```
larsrickert: Would prefer a flex layout with a gap on the page root instead of margins :) We might also want to use the `var(--onyx-grid-gutter)` CSS variable here

apps/demo-app/app/pages/charts.vue
```diff
@@ -1,7 +1,210 @@
...
+}
+.content {
```
larsrickert: Class seems to be unused

apps/demo-app/app/pages/charts.vue
```diff
@@ -1,7 +1,210 @@
...
+  display: flex;
+  flex-direction: column;
+  gap: var(--onyx-grid-gutter);
+}
+.onyx-form {
```
larsrickert, line 184: Would prefer a regular "onyx-grid" layout with column spans instead

apps/demo-app/app/pages/charts.vue
```diff
@@ -1,7 +1,210 @@
...
+            :label="t('charts.fullStats')"
+          />
+        </OnyxHeadline>
+
+        <div class="chart">
+          <Bar :data="barChartData" :options="chartOptions" />
+        </div>
+      </OnyxCard>
+    </div>
+    <div class="onyx-grid-span-4 onyx-grid-md-span-6 chart-wrapper">
```
larsrickert: The charts become very small when resizing the screen, maybe we can already wrap them below each other at a bigger breakpoint?

larsrickert (changes requested):

apps/demo-app/app/pages/charts.vue
ChristianBusshoff: because the bottom margin of the Headlines should be smaller, it's way more code using gap. Or am I wrong?

apps/demo-app/app/pages/charts.vue
```diff
@@ -1,7 +1,208 @@
...
+  <OnyxHeadline is="h1">{{ t("charts.pageName") }}</OnyxHeadline>
+  <OnyxHeadline is="h3" class="headline">{{ t("charts.rangeAndType") }}</OnyxHeadline>
+  <OnyxForm class="onyx-grid" @submit.prevent>
+    <OnyxDatePicker class="onyx-grid-span-3" :label="t('charts.range')" />
+    <OnyxSelect
+      class="onyx-grid-span-3"
+      :label="t('charts.type')"
+      list-label="List label"
+      :options="options"
+      placeholder="Placeholder..."
```
larsrickert: Please translate the list label and placeholder :)

apps/demo-app/i18n/locales/de-DE.json
```diff
@@ -12,7 +12,42 @@
...
+      "private": "Privatkunden"
+    },
+    "metrics": {
+      "revenue": "Umsatz (in Tsd. €)",
+      "netProfit": "Nettogewinn (in Tsd. €)"
+    },
+    "kpi": {
+      "title": "KPI",
+      "roi": "Return on Investment",
+      "clv": "Customer Lifetime Value"
```
larsrickert:
```suggestion
      "roi": "Kapitalrendite",
      "clv": "Kundenertragswert"
```

Issue #4019, opened 2025-08-27
Fix inconsistent (error) props between form elements


JoCa96:
### Why?

The `OnyxInput` currently has a `customError` prop, as well as an `error` prop.
They are supposed to serve the same purpose, but only `customError` works.
To align and improve the APIs, we prefer to only have the `error` prop (we already have the props `message`and 
`success` which serve a similar purpose.)

### Acceptance criteria

- [ ] All FormElements have the same API for (*custom-*)errors.
  - [ ] `customError` prop is removed the component APIs
  - [ ] Instead the `error` prop exists and works the same

### Implementation details

The `useCustomValidity` composable is currently neither "general purpose" nor "onyx specific", therefor this composable should changed:

- [ ] `useCustomValidity` is updated to work indepenedently (without onyx)
- [ ] a new composable `useFormElementError` is created, which:
  - [ ] consumes the components props and
  - [ ] uses `useCustomValidity` internally and extends it with **onyx** specifics.
  - [ ] uses `useErrorClass` internally.

### DEV handover checklist

[Storybook](ADD_LINK_HERE)

DEV handover done with: @ADD_DEV_NAME_HERE

**Testing**
- [ ] covered by visual screenshot tests
- [ ] covered by functional tests (Playwright or unit test)

**Documentation**
- [ ] auto-generated Storybook code snippets are checked
- [ ] example usage added to `apps/demo-app/src/views/HomeView.vue`
- [ ] component status has been updated in the [docs overview](https://onyx.schwarz/#components)
- [ ] updated version is released
- [ ] updated documentation is deployed

**Other**
- [ ] follow-up tickets were created if necessary (add links below)

**Figma**
- [ ] Figma component is published

**Documentation**
- [ ] Storybook category is aligned with Figma
- [ ] Storybook examples are checked

PR #4063, opened 2025-09-02
bug(FormElements): Fix inconsistent error props

ChristianBusshoff:
Relates to #4019

- changed customError to error
- removed duplicates


packages/sit-onyx/src/utils/validity.spec.ts
```diff
@@ -4,7 +4,7 @@ import { getFirstInvalidType } from "./validity.js";
 const getTestState = (overrideData: Partial<ValidityState>): ValidityState => ({
   valid: false,
   badInput: false,
-  customError: false,
+  error: false,
```
JoCa96: In the native ValidityState the error is still called `customError`, so it shouldn't be changed here :)

```suggestion
  customError: false,
```

JoCa96 (approved):

JoCa96:
Please add a changeset with details about the breaking change(s)

PR #4065, opened 2025-09-03
feat(demo-app): implement roles page

larsrickert:
Relates to #3982

- Implement the role page of our demo app
- move existing form page into a child route so we can use a parent route for the sidebar

## Checklist

### DIFF
--- a/apps/demo-app/app/pages/forms/roles.vue
+++ b/apps/demo-app/app/pages/forms/roles.vue
@@ -20,6 +20,11 @@ const roles = computed(() => {
     value: role,
   }));
 });
+
+const updateRoleSelection = (checked: boolean, role: string) => {
+  if (checked) selectedRoles.value.add(role);
+  else selectedRoles.value.delete(role);
+};
 </script>
 
 <template>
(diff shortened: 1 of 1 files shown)


apps/demo-app/app/pages/forms.vue
```diff
@@ -1,262 +1,27 @@
...
+      <NuxtPage />
+    </div>
 
     <template #footer>
-      <OnyxBottomBar>
-        <OnyxButton :label="$t('reset')" type="reset" color="neutral" mode="plain" :form="formId" />
-        <OnyxButton :label="$t('save')" type="submit" :form="formId" />
-      </OnyxBottomBar>
+      <div ref="footerRef"></div>
     </template>
```
JoCa96, line 25: Would that be an idea for an PageLayout/AppLayout Feature?

JoCa96 (approved):

apps/demo-app/app/pages/forms.vue
larsrickert: Do you mean the teleport? The OnyxPageLayout already supports a footer slot. We have a little special case here because we use the page layout inside the parent view but the bottom bar is placed inside the child (which is only the page content and does not use a page layout on its own)

JoCa96: If it's a common pattern (it's in the demo app, so probably?) we could consider adding some out-of-the-box support. E.g. an `actions`/`buttons` slot for the OnyxForm, that can teleport there (based on another prop maybe)

Issue #3987, opened 2025-08-25
Implement notification center for demo application


larsrickert:
### Why?

Our onyx demo application should include a global [notification center](https://storybook.onyx.schwarz/?path=/story/notifications-notificationcenter--default)

### Depends on

- #3969 

### Design

[Figma](https://www.figma.com/design/YG8K4K7VnH6Dux2WeSnCS2/Onyx-Demo-Application?node-id=56-14477&m=dev)

### Acceptance criteria

- [x] the [notification center](https://storybook.onyx.schwarz/?path=/story/notifications-notificationcenter--default) is implemented globally
- [x] example notifications can be shown/triggered
- [x] initially, a few notifications are displayed after showing a skeleton for X seconds

### Definition of Done

- The following component requirements are implemented:
  - [x] skeleton
  - [x] density

- Should be covered by tests:
  - [x] functional tests (Playwright or unit test)
  - [x] visual tests (Playwright screenshots)

- Make sure, that
  - [x] follow-up tickets were created if necessary
  - [x] updated version + documentation is deployed

#### Approval

[Documentation](https://demo.onyx.schwarz/)

- Checked and approved by
  - [x] designer 
  - [x] dev @larsrickert

PR #4071, opened 2025-09-04
feat(demoApp): implement  notification center

ChristianBusshoff:
Relates to #3987

- add Notification Center
- add FAB to add new Notifications

##Questions

- Are the positions, I inserted the Notification Center & FAB right?

## Checklist


apps/demo-app/app/components/NavBar.vue
```diff
@@ -24,7 +27,14 @@ const localePath = useLocalePath();
       <LocaleSwitch />
 
       <OnyxSeparator orientation="vertical" />
-
+      <OnyxNotificationDot :hidden="!store.unreadNotifications.length">
+        <OnyxIconButton
+          label="Notifications"
```
larsrickert: Please translate the label :)

apps/demo-app/app/components/NavBar.vue
```diff
@@ -24,7 +27,14 @@ const localePath = useLocalePath();
...
+          color="neutral"
+          :icon="iconBell"
+          @click="store.isSidebarOpen = true"
```
larsrickert: The sidebar open state should be a local state of the component instead of a global state in the pinia store :)

Otherwise we mix up local component UI-specific state with the global state that should only contain the actual notifications data

(optional) We can also move this code to a separate NotificationIconButton.vue component like we also did with e.g. the locale switch so the nav bar code is more minimal

apps/demo-app/app/stores/notification-store.ts
```diff
@@ -0,0 +1,32 @@
...
+      }));
+    },
+    /**
+     * Adds a new unread notification.
+     */
+    add(notification: Omit<MyNotification, "unread">) {
+      this.notifications.unshift({ ...notification, unread: true });
+    },
+  },
+});
```
larsrickert:
```suggestion
export const useNotificationStore = defineStore("notification", () => {
  const notifications = ref<MyNotification[]>([]);

  const unreadNotifications = computed(() => notifications.value.filter(({ unread }) => unread));
  const readNotifications = computed(() => notifications.value.filter(({ unread }) => !unread));

  /**
   * Marks all existing notifications as read.
   */
  const markAllAsRead = () => {
    notifications.value = notifications.value.map((notification) => ({
      ...notification,
      unread: false,
    }));
  };

  /**
   * Adds a new unread notification.
   */
  const add = (notification: Omit<MyNotification, "unread">) => {
    notifications.value.unshift({ ...notification, unread: true });
  };

  return { notifications, unreadNotifications, readNotifications, markAllAsRead, add };
});
```

Nowadays, you typically use this setup style way of defining the pinia store because its almost identical to writing composables or Vue components :)

This difference is similar to Vue's Options API vs. Composition API where we want to always use the composition API

apps/demo-app/i18n/locales/de-DE.json
```diff
@@ -48,6 +48,16 @@
       }
     }
   },
+  "notification": {
+    "fab": "Benachrichtigung auslösen",
+    "headline": "Benachrichtigungen",
+    "description": "Sehen Sie hier alle Benachrichtigungen von allen Kontaktpunkten.",
```
larsrickert:
```suggestion
    "description": "Hier sind alle Benachrichtigungen zu finden.",
```

The German description sounds a little weird :D My suggestion is just a proposal, if you find a better wording, feel free!

apps/demo-app/i18n/locales/en-US.json
```diff
@@ -48,6 +48,16 @@
       }
     }
   },
+  "notification": {
+    "fab": "trigger Notification",
```
larsrickert:
```suggestion
    "fab": "Trigger notification",
```

apps/demo-app/layers/blueprint/app/app.vue
```diff
@@ -11,6 +11,7 @@ defineSlots<{
 
 <template>
   <OnyxAppLayout v-bind="props" class="onyx-grid-max-lg onyx-grid-center">
+    <OnyxNotifications />
```
larsrickert: We already include the toast and global FAB within the OnyxAppLayout by default, should we also do the same for the notifications to keep it aligned? So the user does not have to add it manually

apps/demo-app/layers/blueprint/app/layouts/default.vue
```diff
@@ -1,5 +1,8 @@
 <script lang="ts" setup>
-import type { OnyxPageLayoutProps } from "sit-onyx";
+import { iconBellRing, iconCircleAttention } from "@sit-onyx/icons";
+import { OnyxFAB, useNotification, type OnyxPageLayoutProps } from "sit-onyx";
```
larsrickert:
```suggestion
import {useNotification, type OnyxPageLayoutProps } from "sit-onyx";
```

In Nuxt, the components are auto imported so we don't need to import them here

apps/demo-app/layers/blueprint/app/layouts/default.vue
```diff
@@ -1,5 +1,8 @@
...
+import type { MyNotification } from "../../../../app/components/NotificationCenter.vue";
+import NotificationCenter from "../../../../app/components/NotificationCenter.vue";
```
larsrickert: We shouldn't mix the demo-apps specific code with the blueprint layer. The blueprint layer is intended to just contain basic general code like the nav bar etc. which is probably the same in many applications.

The idea is to extract this blueprint layer in a separate npm package in the future that we can offer standalone so projects can use it as quick start.

To refactor this code, you can create a new `app/layouts/default.vue` file in the demo app folder which will then basically replace the default layout of the blueprint layer

apps/demo-app/app/components/NotificationCenter.vue
```diff
@@ -0,0 +1,158 @@
...
+  OnyxBottomBar,
+  OnyxEmpty,
+  OnyxHeadline,
+  OnyxNotificationCard,
+  OnyxSidebar,
+  type OnyxNotificationCardProps,
+} from "sit-onyx";
+import { ref } from "vue";
+
+export type MyNotification = OnyxNotificationCardProps & {
```
larsrickert: Should we move this type to the pinia store?

apps/demo-app/app/components/NotificationCenter.vue
```diff
@@ -0,0 +1,205 @@
...
+  () => isSidebarOpen,
+  (isNewValue) => {
+    if (isNewValue) {
+      skeleton.value = store.unreadNotifications.length;
+      setTimeout(() => {
+        skeleton.value = 0;
+      }, 1000);
+    }
+  },
+);
```
larsrickert:
```suggestion
watch(isSidebarOpen, async (open) => {
  if (!open) return;
  skeleton.value = 3;
  await promiseTimeout(1000);
  skeleton.value = 0;
});
```

For this demo, I'd just always show the skeleton when opening the sidebar

apps/demo-app/app/components/NotificationCenter.vue
```diff
@@ -0,0 +1,205 @@
...
+  const notification: MyNotification = {
+    headline: `${t("notification.notificationTitle")} ${store.notifications.length + 1}`,
+    createdAt: Date.now(),
+    icon,
+    description:
+      "Lorem ipsum dolor sit amet consectetur. Dui purus quisque est varius vulputate. Ut odio dui diam pulvinar velit mollis cursus eu ut.",
+  };
+
+  store.add(notification);
+  show({ ...notification, icon });
```
larsrickert:
```suggestion
  show(notification);
```

The icon is already part of the notification

apps/demo-app/app/components/NotificationCenter.vue
```diff
@@ -0,0 +1,205 @@
...
+
+const id = useId();
+
+onMounted(() => {
+  globalFAB.add(
+    computed(() => ({
+      id,
+      label: t("notification.fab"),
+      icon: iconBellRing,
+      onClick: () => addExampleNotification(),
```
larsrickert:
```suggestion
      onClick: addExampleNotification,
```

We can also pass the function directly so we don't need to anonymous arrow function

apps/demo-app/app/components/NotificationCenter.vue
```diff
@@ -0,0 +1,205 @@
...
+          v-bind="notification"
+        >
+          {{ notification.description }}
+
+          <template #actions>
+            <OnyxButton label="Button" color="neutral" />
+            <OnyxButton label="Button" />
+          </template>
+        </OnyxNotificationCard>
+      </OnyxAccordionItem>
```
larsrickert: (optional) we could extract this code into a separate NotificationAccordionItem which we can reuse for the read and unread messages

But you don't have to do it in this PR

apps/demo-app/app/components/NotificationCenter.vue
```diff
@@ -0,0 +1,205 @@
...
+        skeleton.value = 0;
+      }, 1000);
+    }
+  },
+);
+
+const openAccordions = ref(["unread"]);
+
+// Add Global Fab
+const { show } = useNotification();
```
larsrickert, line 35: Currently nothing is shown, I think we need to add the OnyxNotifications component to the OnyxAppLayout (+ add a changeset for this)

larsrickert (changes requested):

larsrickert (approved):

PR #4080, opened 2025-09-05
feat(demo-app): implement file upload page

larsrickert:
Relates to #3982

- demo app: implement file upload page
- fix(OnyxDataGrid): prevent "Cannot use in to search for 'ResizeObserver' / 'CSS' in undefined" errors
- fix(OnyxDataGrid): preventSR hydration errors when using SSR / Nuxt
- fix(OnyxSidebar): check if should be collapsed only onMounted to support SSR / prevent hydration errors
- OnyxFileCard: BREAKING CHANGE - remove default slot prop `status` in favor of new `props` that also contains other useful file card properties
- export `useFileSize` composable that can be used to format a file size in a user-friendly format

## Checklist

### DIFF
--- a/packages/sit-onyx/src/components/OnyxDataGrid/features/stickyColumns/stickyColumns.ts
+++ b/packages/sit-onyx/src/components/OnyxDataGrid/features/stickyColumns/stickyColumns.ts
@@ -42,7 +42,7 @@ export const useStickyColumns = <TEntry extends DataGridEntry>(
           resizeObserver.disconnect();
           Object.values(elementsToStyle.value).forEach((el) => resizeObserver.observe(el));
         },
-        { deep: true },
+        { deep: true, immediate: true },
       );
       onUnmounted(() => resizeObserver.disconnect());
     });


packages/sit-onyx/src/utils/dom.ts
```diff
@@ -1,8 +1,11 @@
...
+ * Escapes the given key for save usage in CSS (e.g. as part of a CSS variable name or selector).
  * Will use the native "CSS.escape()" if available in the browser, otherwise a fallback will be used.
  */
-export const escapeCSS = (name: string) => {
-  if ("CSS" in window && window.CSS) return CSS.escape(name);
+export const escapeCSS = (key: PropertyKey) => {
+  // to prevent server side hydration errors, we are using the description for symbols so the name does not contain "(" or ")"
+  // which would be escaped differently on server side (where window.CSS) is not available than on client side
+  const name = typeof key === "symbol" && key.description ? key.description : String(key);
+  if (window && "CSS" in window && window.CSS) return CSS.escape(name);
```
JoCa96, line 9: This will still cause hydration errors, because `window.CSS` is never available on the server

apps/demo-app/app/components/FileDataGrid.vue
```diff
@@ -0,0 +1,183 @@
...
+  } satisfies TypeRenderMap<FileEntry>,
+}));
+
+const features = [
+  withFiltering,
+  withPagination,
+  withHideColumns,
+  withSorting,
+  withStickyColumns,
+  withResizing,
```
JoCa96, line 148: Would probably remove the resizing as it is a bit broken when used with the others 
```suggestion
```

packages/sit-onyx/src/utils/dom.ts
larsrickert: The hydration errors are fixed (at least the ones we had for the data grid) by checking if the key is a symbol.

Yes "window" is never available on the server but usually this method is called onMounted so its no problem. But there are cases where its called server side so to prevent critial errors, we check if window is defined

apps/demo-app/app/components/FileDataGrid.vue
larsrickert: What is broken here?

Issue #3163, opened 2025-04-16
Implement|Fix|Define Implement docs page layout for Nuxt docs template


larsrickert:
### Why?

Implement the "docs" page layout for our [Nuxt documentation template](https://onyx.schwarz/development/packages/nuxt-docs.html).

### Open questions / TODOs

- is the design final? It seems not 100% ready yet

### Depends on

- #2615

### Design

[Figma](https://www.figma.com/design/EBi98T51T9fGfjrYYfsulg/onyx---Archive?node-id=4571-28462&t=fzqtVFuNXG54Nrr4-4)

### Acceptance criteria

- [ ] the "docs" layout is the default layout so if no layout is explicitly defined, the docs layout is used
- [ ] a sidebar is shown that:
    - [ ] supports showing nav items
    - [ ] optionally supports grouping nav items with accordions. The initial open state can be configured per accordion
    - [ ] by default, the sidebar is auto-filled using the [queryCollectionNavigation](https://content.nuxt.com/docs/utils/query-collection-navigation) composable from the Nuxt content module
    - [ ] the sidebar content can be overriden inside the `app.config.ts` file
- [ ] a "last updated" timestamp is shown that is auto-detected based on the git/file changes of the corresponding content
- [ ] a "next" and "previous" page button is displayed. There are auto-generated using the [queryCollectionItemSurroundings](https://content.nuxt.com/docs/utils/query-collection-item-surroundings) composable

#### Out of scope

- table of content

### Implementation details

<-- Add technical details / implementation proposals if necessary, e.g.:
  - API proposal
  - relevant web APIs
  - learnings from other implementations
  - code snippets
  -->

### Definition of Done

- The following component requirements are implemented:
  - [ ] skeleton
  - [ ] density
  - [ ] example usage added to `apps/demo-app/src/views/HomeView.vue`

- Should be covered by tests:
  - [ ] functional tests (Playwright or unit test)
  - [ ] visual tests (Playwright screenshots)

- Make sure, that
  - [ ] follow-up tickets were created if necessary
  - [ ] updated version + documentation is deployed
  - [ ] Storybook can show the feature
  - [ ] Storybook code snippet of new/changed examples are checked that they are generated correctly
  - [ ] Namings are aligned with Figma
  - [ ] Storybook sidebar badge for "New" is added on component or story level if needed. See: https://github.com/Sidnioulz/storybook-addon-tag-badges/?tab=readme-ov-file

#### Approval

[Storybook]()
[...]

PR #4082, opened 2025-09-08
feat(nuxt-docs): implement sidebar page layout

larsrickert:
Relates to #3163

- implement basic sidebar layout for the nuxt docs that auto-generated a sidebar navigation based on the content structure
- nuxt-docs: add default margins to prose components so the markdown renderer page has pretty spacings by default
- fix: if a router is provided, the active state of links will only be set to true, if the link matches exactly. This fixes the bug that when using nested routes, the parent(s) is/are always active when a child view is currently open

## Checklist


packages/sit-onyx/src/composables/useLink.spec.ts
```diff
@@ -24,9 +24,9 @@ describe("useLink", () => {
     { current: "/test/", link: "/test", active: true },
     { current: "/test", link: "/test/", active: true },
     { current: "/test   ", link: "   /test", active: true },
-    // nested links: parent should be active if child page is currently active
+    // nested links: only exact match should be active, parent should NOT be active if child page is currently active
```
JoCa96, line 27: When we want to support nested navigation menus, we will some way to support this

larsrickert: We also support nested navigation menus and they still work fine :) Marking the parent visually as active in such cases can by done by CSS but marking the whole route chain as active per default mostly leads to unexpected behavior, a few examples:

- we have a sidebar with nested children: /, /bar, /bar/baz
- locale prefixed routes with i18n where e.g. all routes are prefixed with /de
- nav item with children where the parent is part of the flyout itself, e.g. /admin and /admin/users. In this case both nav items would be active when visiting the /admin/users route

Issue #4101, opened 2025-09-10
Implement webhook for automate deployment of internal demo app


larsrickert:
### Why?

Currently, the internal demo app is not deployed automatically when we make changes to the main branch of the public demo app in GitHub.

We can use a webhook to automate this

### Design

[Figma](ADD_FIGMA_LINK_HERE)

### Acceptance criteria

- [ ] when changes are pushed to the main branch of the public GitHub repository, the internal demo app is deployed as well
- [ ] we might want to update the onyx dependencies of the internal demo app to the latest version before building / deploying it. Otherwise the build might fail if we've added e.g. a new function in sit-onyx which we are using in the public demo app

### Implementation details

<-- Add technical details / implementation proposals if necessary, e.g.:
  - API proposal
  - relevant web APIs
  - learnings from other implementations
  - code snippets
  -->

### DEV handover checklist

[Storybook](ADD_LINK_HERE)

DEV handover done with: @ADD_DEV_NAME_HERE

**Testing**
- [ ] covered by visual screenshot tests
- [ ] covered by functional tests (Playwright or unit test)

**Documentation**
- [ ] "New" badge in Storybook is added, see [storybook-addon-tag-badges](https://github.com/Sidnioulz/storybook-addon-tag-badges/?tab=readme-ov-file#-usage)
- [ ] auto-generated Storybook code snippets are checked
- [ ] component status has been updated in the [docs overview](https://onyx.schwarz/#components)
- [ ] updated version is released
- [ ] updated documentation is deployed

**Other**
- [ ] follow-up tickets were created if necessary (add links below)

### UX handover checklist

UX handover done with: @ADD_UX_NAME_HERE

**Naming is aligned with Figma**
- [ ] component name
- [ ] which support components are used and how they are named
- [ ] all property names and types

**Implementation is checked**
- [ ] all features of Figma are implemented or a follow-up ticket is refined
- [ ] skeleton
- [ ] densities
- [ ] light and dark mode
- [ ] truncation
- [ ] responsiveness / wrapping behavior
- [ ] min / max width

**Figma**
- [ ] Figma component is published

**Documentation**
- [ ] Storybook category is aligned with Figma
- [ ] Storybook examples are checked

**Other**
- [ ] follow-up tickets were created if necessary (add links below)

PR #4119, opened 2025-09-11
chore(demo-app): setup webhook for automated deployment of internal demo app

larsrickert:
Relates to #4101

I've set up a Azure webhook (with secret so it can not be called by anyone) which we call during the public demo app deployment in the GitHub action which will also trigger a deployment of the Schwarz internal demo app

### DIFF
--- a/.github/workflows/deploy-websites.yml
+++ b/.github/workflows/deploy-websites.yml
@@ -104,6 +104,16 @@ jobs:
     if: ${{ (inputs.environment || inputs.environment) != '' }}
     environment: ${{ inputs.environment || inputs.environment }}
     steps:
+      - name: 🪝 Invoke internal demo app webhook
+        uses: distributhor/workflow-webhook@v3
+        # since the webhook is an external dependency, we still want to continue deploying the public demo app
+        # we just call it here already so the internal demo app deployment can run in parallel
+        continue-on-error: true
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -172,6 +172,15 @@ jobs:
...
     steps:
+      - name: 🪝 Invoke internal demo app webhook
+        uses: distributhor/workflow-webhook@v3
+        # since the webhook is an external dependency, we still want to continue deploying the public demo app
+        # we just call it here already so the internal demo app deployment can run in parallel
+        continue-on-error: true
+        with:
+          webhook_url: ${{ secrets.AZURE_DEMO_APP_WEBHOOK_URL }}
+          webhook_secret: ${{ secrets.AZURE_DEMO_APP_WEBHOOK_SECRET }}
+
```
JoCa96: Whats the reason that we don't use the native github feature? 
- https://github.com/SchwarzIT/onyx/settings/hooks
- https://docs.github.com/en/webhooks/using-webhooks/creating-webhooks

larsrickert: Because there we can only subscribe to very generic events like any code push to any branch. We'd need to implement the conditions then in our Azure pipeline.

This approach would be simpler since we can just invoke the webhook where and when we need it

Which option do you prefer?

JoCa96 (approved):

PR #4144, opened 2025-09-17
feat(OnyxNotifications): return notification id after creation

bergser:
Return notification `id` after it is created

## Checklist

### DIFF
--- a/.changeset/tough-jeans-fly.md
+++ b/.changeset/tough-jeans-fly.md
@@ -1,0 +1,5 @@
+---
+"sit-onyx": minor
+---
+
+feat(OnyxNotifications): return notification id after it is created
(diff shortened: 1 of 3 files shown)


packages/sit-onyx/src/components/OnyxNotifications/useNotification.ts
```diff
@@ -99,7 +102,10 @@ export const useNotification = () => {
     () => {
       return {
         notifications: computed(() => []),
-        show: logWarning,
+        show: () => {
+          logWarning();
+          return 0;
```
larsrickert:
```suggestion
          return -1;
```

Would prefer to return -1 here so its clearer that something is used wrong in this case :)

.changeset/tough-jeans-fly.md
```diff
@@ -0,0 +1,5 @@
+---
+"sit-onyx": minor
+---
+
+feat(OnyxNotifications): return notification id after it is created
```
larsrickert:
```suggestion
feat(useNotification): return notification id when showing a new notification
```

larsrickert (changes requested):
Thanks for the contribution @bergser! I only have two small comments :)

Issue #4079, opened 2025-09-05
FoUp: OnyxDialog stays in viewport 


ChristianBusshoff:
### Why?

Immediately hiding the dialog seems weird, we should avoid to immediately hide it.
Dialog stays in the viewport when scrolled with a distance of i.e. 16px to the screen border.

### Acceptance criteria

- [ ] find and implement a solution

**Other**
- [x] follow-up tickets were created if necessary (add links below)

**Documentation**
- [x] Storybook examples are checked

PR #4197, opened 2025-09-29
feat(OnyxDialog) Popover stays in viewport while scrolling

ChristianBusshoff:
- added clipping

Relates to #4079

### DIFF
--- a/packages/sit-onyx/src/components/OnyxBasicPopover/OnyxBasicPopover.ct.tsx
+++ b/packages/sit-onyx/src/components/OnyxBasicPopover/OnyxBasicPopover.ct.tsx
@@ -46,6 +46,33 @@ test.describe("OnyxBasicPopover", () => {
     // ASSERT
     await expect(popover).toBeHidden();
   });
+
+  test("should stay in view if clipping is active", async ({ mount, page }) => {
+    const component = await mount(OnyxBasicPopoverTestCase, {
+      props: {
+        label: "Popover for testing",
+        open: true,
+        clipping: true,
+      },
+    });
+
+    const popover = component.getByText("Popover Content");
+
+    // ASSERT
+    await expect(popover).toBeVisible();
+
+    // ACT
+    await page.evaluate(() => window.scrollTo(0, document.body.scrollHeight));
+
+    // ASSERT
+    await expect(popover).toBeVisible();
+    const box = await popover.boundingBox();
+    expect(box).not.toBeNull();
+    const viewportHeight = await page.evaluate(() => window.innerHeight);
+    if (box) {
+      expect(box.y + box.height).toBeLessThanOrEqual(viewportHeight);
+    }
+  });
 });
 test.describe("OnyxBasicPopover  Screenshot Tests", () => {
   test.describe("Alignment screenshot tests", () => {
--- a/packages/sit-onyx/src/components/OnyxBasicPopover/OnyxBasicPopover.vue
+++ b/packages/sit-onyx/src/components/OnyxBasicPopover/OnyxBasicPopover.vue
@@ -20,6 +20,7 @@ const props = withDefaults(defineProps<OnyxBasicPopoverProps>(), {
   position: "auto",
   alignment: "auto",
   role: "dialog",
+  clipping: false,
 });
 
 defineSlots<{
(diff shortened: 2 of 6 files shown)


packages/sit-onyx/src/components/OnyxBasicPopover/OnyxBasicPopover.ct.tsx
```diff
@@ -46,6 +46,33 @@ test.describe("OnyxBasicPopover", () => {
...
+    await page.evaluate(() => window.scrollTo(0, document.body.scrollHeight));
+
+    // ASSERT
+    await expect(popover).toBeVisible();
+    const box = await popover.boundingBox();
+    expect(box).not.toBeNull();
+    const viewportHeight = await page.evaluate(() => window.innerHeight);
+    if (box) {
+      expect(box.y + box.height).toBeLessThanOrEqual(viewportHeight);
+    }
```
larsrickert:
```suggestion
    const box = (await popover.boundingBox())!;
    const viewportHeight = await page.evaluate(() => window.innerHeight);
    expect(box.y + box.height).toBeLessThanOrEqual(viewportHeight);
```

We can assume the box is always defined :)

packages/sit-onyx/src/components/OnyxBasicPopover/OnyxBasicPopover.vue
```diff
@@ -278,6 +289,11 @@ const popoverStyles = computed(() => {
       &--dont-support-anchor {
         margin: 0;
       }
+      &--clipping {
+        margin: 0;
```
larsrickert, line 296: Should there be a min. gap to the screen border?

packages/sit-onyx/src/components/OnyxDialog/OnyxDialog.vue
```diff
@@ -59,7 +60,12 @@ const isExpanded = useVModel({
 </script>
 
 <template>
-  <OnyxBasicPopover v-bind="props" v-model:open="isExpanded" :class="['onyx-dialog', densityClass]">
+  <OnyxBasicPopover
+    v-bind="props"
+    v-model:open="isExpanded"
+    :class="['onyx-dialog', densityClass]"
+    :clipping="props.clipping"
```
larsrickert:
```suggestion
```

The clipping prop is already bound using `v-bind="props"`

packages/sit-onyx/src/composables/useClipping.ts
```diff
@@ -0,0 +1,92 @@
+import { useGlobalEventListener } from "@sit-onyx/headless";
+import { computed, nextTick, onMounted, ref, type Ref } from "vue";
+
+export function useClipping(
+  popoverRef: Ref<HTMLElement | null>,
+  popoverWrapperRef: Ref<HTMLElement | null>,
+  popoverPosition: Ref<string>,
+  isVisible: Ref<boolean>,
+  clipping: Ref<boolean>,
+) {
```
larsrickert:
```suggestion
export type UseClippingOptions = {
  popoverRef: Ref<HTMLElement | null>;
  popoverWrapperRef: Ref<HTMLElement | null>;
  popoverPosition: Ref<AnchorPosition | OpenDirection>;
  isVisible: Ref<boolean>;
  clipping: Ref<boolean>;
};

export function useClipping(options: UseClippingOptions) {
```

Would prefer to use an options object here

packages/sit-onyx/src/composables/useClipping.ts
```diff
@@ -0,0 +1,92 @@
...
+  const clippingStyles = ref("");
```
larsrickert:
```suggestion
 const clippingStyles = ref<CSSProperties>({});
```

Instead of replying on string concatenation, I'd prefer to use an object type for this

packages/sit-onyx/src/composables/useClipping.ts
```diff
@@ -0,0 +1,92 @@
...
+  const scrolledOut = ref<null | "top" | "bottom">(null);
```
larsrickert:
```suggestion
  const scrolledOut = ref<"top" | "bottom">(null);
```

The default value will always be added to the type union automatically. Also is there a reason why we use null here instead of undefined?

packages/sit-onyx/src/composables/useClipping.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    if (isClipping.value) {
+      const popoverRect = popoverRef.value.getBoundingClientRect();
+      const wrapperRect = popoverWrapperRef.value.getBoundingClientRect();
+      const requiredHeight = popoverRect.height + MIN_DISTANCE_TO_BORDER + MARGIN;
+
+      if (
+        popoverPosition.value === "top" ||
+        popoverPosition.value === "top left" ||
+        popoverPosition.value === "top right"
```
larsrickert:
```suggestion
        popoverPosition.value.includes("top")
```

Shouldn't this be the same?

larsrickert: Same for bottom below

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxBasicPopover/OnyxBasicPopover.vue
ChristianBusshoff: Yes but it's set via top/bottom.

PR #4206, opened 2025-09-30
chore: update playwright.yml pipeline to allow checking of only changed files

JoCa96:
- Remove unused artifact uploads from `check.yml` pipeline
- Add optional 'only-changed' property for `playwright.yml` pipeline, which makes use of the identically named [playwright flag](https://playwright.dev/docs/release-notes#--only-changed-cli-option) to only check files affected by changes
- Updated artifact merge logic of the `playwright-result` job in the `playwright.yml` pipeline. Artifacts are not deleted anymore after a merge, so that they are still available for later attempts of the same run.

### DIFF
--- a/packages/shared/src/playwright.config.base.ts
+++ b/packages/shared/src/playwright.config.base.ts
@@ -53,7 +53,7 @@ export const PLAYWRIGHT_BASE_CONFIG: PlaywrightTestConfig = {
    */
   timeout: 20 * 1000,
   forbidOnly: !!process.env.CI, // fail build on CI if we left test.only in the source code
-  retries: process.env.CI ? 1 : 0, // retry on CI only
+  retries: process.env.CI ? 3 : 0, // retry on CI only
 
   /**
    * REPORTERS


.github/workflows/playwright.yml
```diff
@@ -33,7 +39,7 @@ jobs:
         run: pnpm exec playwright install-deps
 
       - name: 🔎 Run Playwright tests
-        run: pnpm run test:playwright:all
+        run: pnpm run test:playwright:all --concurrency 1 ${{ inputs.only-changed != '' && format('-- --only-changed {0}', inputs.only-changed) }}
```
larsrickert: Does this also correctly run the test if we e.g. use the OnyxSystemButton inside the OnyxDataGrid and we change something inside the OynxSystemButton, will also the OnyxDataGrid tests run?

.github/workflows/check.yml
```diff
@@ -74,20 +74,10 @@ jobs:
...
-          name: documentation
-          path: apps/docs/src/.vitepress/dist
-
   screenshots:
     needs: build
     name: Component tests
     uses: ./.github/workflows/playwright.yml
     secrets: inherit
+    with:
+      only-changed: ${{ github.event.pull_request.base.ref && format('origin/{0}', github.event.pull_request.base.ref) }}
```
larsrickert: So this parameter is only set for PRs, right? For the merge queue all tests should be run

.github/workflows/playwright.yml
JoCa96: Yes, it is able to determine dependencies between files. But I don't think it is able to detect cross-package dependencies.

.github/workflows/check.yml
JoCa96: Correct: This way we are sure that all tests run, but should still be able to merge faster.

larsrickert (approved):

PR #4225, opened 2025-10-06
Larsrickert/4222 datepicker empty string

### LINKED ISSUE #4222: Fix OnyxDatePicker emits empty string when value is de-selected
### Version

1.1.0

### Description

The OnyxDatePicker emits an empty string for its v-model when the user clears the selected date.
This is unintuitive and might cause other errors in the application when trying to parse an empty string as date.
Instead, undefined should be emitted.

### Reproduction steps

1. Open [Playground link](https://playground.onyx.schwarz/?onyxVersion=1.1.0#eNp9kk9P3DAUxL+K5RNIbHJoT9ul6j8O5cCi0qOlKk3eBoNjW/ZLuqso371jZxcCQlyi+M3Y8/Mko/zqfTH0JNdyE+ugPQtT2fZSSY5Kikjc+8/K6s67wGIUW3vY/6iYbnX9SEFMYhdcJ5SMmlcOmpKfFu5Au2cLUrKqbO1sZNHgGHGZPGfnym7KOR5hWDB13kDHSohNowdRmypGYKWQVRt0szLVwfWsZPbA9QrNVH/JYEMa4SbDqnNNHqRcDMrjvqSvxTiK67vtTRE5aNvq3eEs2c7FNGWCEgjwb8oFmLyQ80VXXeWLh+gsShyTXR0FNIiT55hlRWus7pl9XJdl3VjsBZkeQmGJS+u78mT9kqIiIx0PbRvaw6tkOhFcEwg4osydbl/l167z2lDYetYo+wVHZYz7d51nHHq6OM3re6of35g/xCPybaBIYUB5TxpXoSV8gyRf3d3QHu9PIgrvDdzviL8oOtMnxtn2rbcNsBe+TPsz14kP8zte7ZlsPF0qgeY2sj//Y9/fufoz7ofi46LFPwOFdCYKnBuX03/5qwp7)
2. Select any date
3. Open the date picker and click "Clear"
4. See that the modelValue is an empty string

### Confirmations

larsrickert:
closes #4222

fix(OnyxDatePicker): empty undefined instead of empty string when value is cleared

Also the type for the `update:modelValue` event has been fixed to be `string | undefined` instead of `DateValue | undefined` since its always a string timestamp if a date is selected.

## Checklist

### DIFF
--- a/packages/sit-onyx/src/components/OnyxDatePicker/OnyxDatePicker.vue
+++ b/packages/sit-onyx/src/components/OnyxDatePicker/OnyxDatePicker.vue
@@ -9,7 +9,6 @@ import {
   useSkeletonContext,
 } from "../../composables/useSkeletonState.js";
 import { useVModel } from "../../composables/useVModel.js";
-import type { Nullable } from "../../types/index.js";
 import { useRootAttrs } from "../../utils/attrs.js";
 import { isValidDate } from "../../utils/date.js";
 import { FORM_INJECTED_SYMBOL, useFormContext } from "../OnyxForm/OnyxForm.core.js";
@@ -32,7 +31,7 @@ const emit = defineEmits<{
   /**
    * Emitted when the current value changes. Will be a ISO timestamp created by `new Date().toISOString()`.
    */
-  "update:modelValue": [value?: Nullable<DateValue>];
+  "update:modelValue": [value?: string];
   /**
    * Emitted when the validity state of the input changes.
    */
@@ -84,7 +83,7 @@ const value = computed({
   get: () => getNormalizedDate.value(modelValue.value),
   set: (value) => {
     const newDate = new Date(value ?? "");
-    modelValue.value = isValidDate(newDate) ? newDate.toISOString() : "";
+    modelValue.value = isValidDate(newDate) ? newDate.toISOString() : undefined;
   },
 });
 const input = useTemplateRef("inputRef");
--- a/.changeset/thick-glasses-create.md
+++ b/.changeset/thick-glasses-create.md
@@ -3,3 +3,5 @@
 ---
 
 fix(OnyxDatePicker): empty undefined instead of empty string when value is cleared
+
+Also the type for the `update:modelValue` event has been fixed to be `string | undefined` instead of `DateValue | undefined` since its always a string timestamp if a date is selected.
--- a/.changeset/thick-glasses-create.md
+++ b/.changeset/thick-glasses-create.md
@@ -1,0 +1,5 @@
+---
+"sit-onyx": patch
+---
+
+fix(OnyxDatePicker): empty undefined instead of empty string when value is cleared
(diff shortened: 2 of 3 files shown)


packages/sit-onyx/src/components/OnyxDatePicker/OnyxDatePicker.vue
```diff
@@ -32,7 +31,7 @@ const emit = defineEmits<{
   /**
    * Emitted when the current value changes. Will be a ISO timestamp created by `new Date().toISOString()`.
    */
-  "update:modelValue": [value?: Nullable<DateValue>];
+  "update:modelValue": [value?: string];
```
JoCa96, line 34: We actually have a more complicated type issue here and we will need to change the modelValue behaviour here:
Check out [this playground](https://playground.onyx.schwarz/?onyxVersion=1.1.0#eNp9U01v2zAM/SuETi2Q2IftlLnFvgqsO6zF1qOBQZMZR60sCRLtJAj830cpX25QNIfYIh/F9x7pnfjifTH0KBaiiipoT2CkbW9qQbEWEJF6f1tb3XkXCHYQcAkjLIProBZcV4tJ8sFuNz9QNkZbnOXTd0n4qNULhnNV1DR3nKvFp9rWVjkbCRoGwk26/sriGlLd1fX1MautJi1Nij5tfQISP9wylxWDND3yXVW5V8B8+UDYecNpPgFUjR5AGRkjK0vN523QzdzIreupFrkC+FddkDbyHxouubeEQSqCtaYVdCwbhnnnmpxMJDhQ5k4AmeGBW/ZLR6gUY293uylvGMeqzPHZUaDZguYeclJwKf1UVGRdJQvjvlU5kcvHSFuTXhMSdgmpnHFhAW1AtOzVmO06oMRM7Gc476QvnqOzvA65qj4keBcW+3tS7DzBBZ9WRD4uylI1lmvZEj2EwiKV1nflEfo5cYvEdPlP2wY3jE3LA8BURmZAkWe91O1Ff+U6rw2GB0+ad+EVDzbMrX/mGIUeZ8e4WqF6eSP+HA+UHwNGDANP7ZQjGVrkVUjpuz+/cMPvpyRPujeMfif5G6MzfeK4h33tbcO0J7jM9j7bqW37FO82hDYeRSWi2Y2Mz9/Wt3ekn+l+KD5OXPw7YEh3soF7x8X4H2ByVB8=) you can see that the ref type changes and is then not matching the typescript type.

JoCa96: We need to ensure, that this cannot happen, so one solution would be to check the runtime type of "modelValue " and emit with the type that is passed in.

larsrickert: I'd say thats a different issue. While the date picker allows any valid value accepted by `new Date()` to be passed, the datepicker itself always emits as string so its a unified behavor.

I don't think we should detect / change the emit type

JoCa96: > I'd say thats a different issue.

Kind of, but then we have two PRs that address the emit type. Just need to make sure to put both in the same release and update the Changelog accordingly.

larsrickert: So what exactly is your proposal now? Imo we can't / shouldn't check the runtime value and emit this type because usually the initial value is undefined so we cant know which type the user has typed its ref.

So I rather always emit a string for consistency

Issue #2875, opened 2025-03-04
Implement|Fix|Define Implement OnyxSlider component


larsrickert:
### Why?

We want to offer a slider component where the user can selected either a single number or a range between two numbers.

### Design

[Figma](https://www.figma.com/design/Vgm8AscnSKiy0njEqEhZG5/slider?node-id=11159-46960&m=dev)
[Archive](https://www.figma.com/design/EBi98T51T9fGfjrYYfsulg/onyx---Archive?node-id=4322-26258&t=NdB5mGZ3bIheAHB9-4)

### Acceptance criteria

- [ ] a single number can be selected
- [ ] a range between the min and max value can be selected
- [ ] optionally a min value can be set, defaults to 0
- [ ] optionally a max value can be set, defaults to 100
- [ ] optionally a step size can be set which determines the next/previous value of the slider based on the current value, can also be set to "any". defaults to 1
- [ ] the default value is set to to the half of min and max, unless max is lower than min, then its set to the min value
- [ ] optionally, the steps can be shown visually (small dots inside the track)
- [ ] if steps are shown, an optional tick mark (label) can be displayed for each step
- [ ] the grabber shows the current value on hover and focus inside a tooltip
- [ ] supports everything of the OnyxFormElement (labels etc.) except: required, readonly, maxlength, counter, placeholder
- the following indicator types / input controls are supported
    - [ ] none (does not show any inputs/controls)
    - [ ] icon (shows a +/- icon button) to adjust the value
    - [ ] value (shows non interactive min and max value)
    - [ ] input (shows an input to change the value. For single select left or right position can be set, for range always two inputs are shown)
- [ ] optionally, the slider can be disabled (also disables inputs/controls if enabled)
- [ ] error handling is supported
- [ ] a headless composables is implemented for the aria pattern below which also supports the described keyboard interactions

### Implementation details

<-- Add technical details / implementation proposals if necessary, e.g.:
  - API proposal
  - relevant web APIs
  - learnings from other implementations
  - code snippets
  -->

### Reference implementations

- [Element Plus](https://element-plus.org/en-US/component/slider.html)
- [PrimeVue](https://primevue.org/slider/)
- [shadcn/vue](https://www.shadcn-vue.com/docs/components/slider.html)

### Applicable ARIA Pattern

[Slider pattern](https://www.w3.org/WAI/ARIA/apg/patterns/slider/)
[...]

PR #4245, opened 2025-10-12
feat(OnyxSlider): implement comprehensive OnyxSlider component with headless architecture

lovelycentury:
Relates to #2875  

Add full-featured slider component with both headless composable and UI implementation:

**Headless Architecture**
- Create `createSlider` composable in @sit-onyx/headless with complete functionality
- Implement builder pattern following existing onyx conventions
- Add comprehensive TypeScript types and JSDoc documentation

**Core Features**
- Single and multi-thumb range slider support
- Horizontal and vertical orientation modes
- Step-based and discrete (mark-only) value constraints
- Track modes: default, inverted, and disabled (false)
- Visual marks system with optional labels
- Auto-generated marks based on step values

**Accessibility & Interaction**
- Full WAI-ARIA compliance with proper roles and attributes
- Complete keyboard navigation (arrows, page up/down, home/end)
- Touch and mouse interaction with passive event listeners
- Focus-visible detection for keyboard vs mouse navigation
- Screen reader support with aria-label and value announcements

**Design System Integration**
- CSS custom properties for comprehensive theming
- BEM methodology for consistent class naming
- Rem-based sizing for accessibility and scaling
- Disabled and inverted state styling
- Hover, focus, and active state management

**Technical Implementation**
- Multi-touch support with touch identifier tracking
- Intentional drag detection to prevent accidental movements
- Value clamping and step rounding with decimal precision
- Mark positioning with edge overflow prevention

## Checklist

- [] The added / edited code has been documented with [JSDoc](https://jsdoc.app/about-getting-started)
- [] The following component requirements are implemented: skeleton, density
- [] Covered by tests: functional tests (Playwright or unit test)
- [] Storybook can show the feature

### DIFF
--- a/packages/sit-onyx/src/components.spec.ts
+++ b/packages/sit-onyx/src/components.spec.ts
@@ -264,6 +264,12 @@ const COMPONENTS: Components = {
       value: "option-1",
     },
   },
+  OnyxUnstableSlider: {
+    props: {
+      label: "Slider",
+      modelValue: 50,
+    },
+  },
 };
 
 describe("components", () => {


packages/sit-onyx/src/components/OnyxSlider/OnyxSlider.vue
```diff
@@ -0,0 +1,725 @@
...
+import { useVModel } from "../../composables/useVModel.js";
+import { useRootAttrs } from "../../utils/attrs.js";
+import { FORM_INJECTED_SYMBOL, useFormContext } from "../OnyxForm/OnyxForm.core.js";
+import OnyxFormElement from "../OnyxFormElement/OnyxFormElement.vue";
+import OnyxIconButton from "../OnyxIconButton/OnyxIconButton.vue";
+import OnyxSkeleton from "../OnyxSkeleton/OnyxSkeleton.vue";
+import OnyxTooltip from "../OnyxTooltip/OnyxTooltip.vue";
+import type { OnyxSliderProps } from "./types.js";
+
+const clamp = (value: number, min: number, max: number) => Math.min(Math.max(value, min), max);
```
larsrickert: There is already a `applyLimits` utility for this that you can re-use / import from "utils/numbers.ts"

packages/sit-onyx/src/components/OnyxSlider/OnyxSlider.vue
```diff
@@ -0,0 +1,725 @@
...
+  return `${percentage}%`;
+};
+
+const props = withDefaults(defineProps<OnyxSliderProps>(), {
+  min: 0,
+  max: 100,
+  step: 1,
+  shiftStep: 10,
+  marks: false,
+  modelValue: (props) => [props.min ?? 0] as number[],
```
larsrickert:
```suggestion
  modelValue: (props) => {
    const { min = 0, max = 100 } = props;
    return [(max - min) / 2];
  },
```

The default value should be in the middle of min/max :)
@JoCa96 Since sliders logically always require a value, should we instead of defining a default value ourselves just make the modelValue property required? Otherwise we might set a default value internally that the user is not aware of which might cause errors in some edge cases 

Also the min/max prop could change dynamically so we wouldn't need to make sure that our default value updates when the min/max properties are changed

packages/sit-onyx/src/components/OnyxSlider/types.ts
```diff
@@ -0,0 +1,92 @@
...
+  marks?: SliderMark[] | boolean;
+  /**
+   * The track style to use.
+   * - `default`: The track fills the area between the minimum value and the active thumb(s).
+   * - `inverted`: The track fills the area between the active thumb(s) and the maximum value.
+   * - `false`: No track is rendered.
+   *
+   * @default "default"
+   */
+  trackMode?: OnyxSliderTrackMode;
```
larsrickert: I double checked the trackMode with UX and we came to the conclusion that we would like to not offer this feature since it can be confusion for the user. A typical slider always has the "default" style of visualizing the selection which the users are familiar with :)

Did you have any special use case in mind for this?

packages/sit-onyx/src/components/OnyxSlider/types.ts
```diff
@@ -0,0 +1,92 @@
...
+  /**
+   * When to show the tooltip with the current value over the thumb.
+   *
+   * - `auto`: Show tooltip when the thumb is focused or being dragged.
+   * - `always`: Always show the tooltip.
+   * - `never`: Never show the tooltip.
+   *
+   * @default "auto"
+   */
+  tooltipDisplay?: OnyxSliderTooltipDisplay;
```
larsrickert: After talking with UX, we decided to change the tooltip display a little bit from whats currently defined in Figma:
By default, the tooltip should behave like the current "auto" implementation. Optionally, the tooltip can be disabled. But we don't want to provide an option to always show it since tooltips are only shown on interaction so we don't see a use case for always showing it.

tl;dr: My suggestion would be to rename to `tooltipDisplay` property to something like `disableTooltip?: boolean` (or noTooltip, hideTooltip...) which is undefined by default (so it behaves like the current "auto" implementation.

packages/sit-onyx/src/components/OnyxSlider/types.ts
```diff
@@ -0,0 +1,92 @@
...
+   * - `value`: shows min and max value labels.
+   * - `icon`: shows icon buttons to increment/decrement the value. The buttons increment/decrement by the shiftStep value. Available only for 1 or 2 thumbs.
+   */
+  control?: OnyxSliderControl;
+  /**
+   * Orientation of the slider.
+   *
+   * @default "horizontal"
+   */
+  orientation?: SliderOrientation;
```
larsrickert:
```suggestion
  orientation?: Orientation;
```

For onyx components we usually use this type imported from `import type { Orientation } from "../../types/index.js";`

The reason is a current limitation that types imported from @sit-onyx/headless do not exists when building the library and using it in projects since its (intentionally) not a dependency of `sit-onyx`. Therefore, no user-facing types must be exposed from `@sit-onyx/headless`. If you need a type from there, please copy it to `sit-onyx` for now :)

packages/sit-onyx/src/components/OnyxSlider/types.ts
```diff
@@ -0,0 +1,92 @@
...
+   *
+   * @default 1
+   */
+  step?: number | null;
+  /**
+   * Specifies the value increment applied when adjusting the slider using Page Up/Page Down or Shift + Arrow keys.
+   *
+   * @default 10
+   */
+  shiftStep?: number;
```
larsrickert: @larsrickert Check this with UX

packages/sit-onyx/src/components/OnyxSlider/types.ts
```diff
@@ -0,0 +1,92 @@
...
+  /**
+   * Highest possible number.
+   */
+  max?: number;
+  /**
+   * Step size.
+   *
+   * @default 1
+   */
+  step?: number | null;
```
larsrickert:
```suggestion
  step?: number;
```

Since the prop is already optional, we don't need to additionally support null :)

packages/sit-onyx/src/components/OnyxSlider/types.ts
```diff
@@ -0,0 +1,92 @@
...
+  | "minlength"
+  | "maxlength"
+  | "placeholder"
+  | "readonly"
+  | "required"
+  | "requiredMarker"
+  | "withCounter"
+> & {
+  /**
+   * Currently selected value.
```
larsrickert:
```suggestion
   * Currently selected value. Must be between `min` and `max`.
```

packages/sit-onyx/src/components/OnyxSlider/OnyxSlider.vue
```diff
@@ -0,0 +1,725 @@
...
+   */
+  validityChange: [validity: ValidityState];
+  /**
+   * Emitted when the slider changes
+   */
+  "update:modelValue": [value: number[]];
+  /**
+   * Emitted when the slider is committed (e.g. on blur, or on pressing enter)
+   */
+  "update:committed": [value: number[]];
```
larsrickert:
```suggestion
```

To align with other form elements, the modelValue should always be updated immediately. For listening on blur, the user can listen to the native blur event that bubbles up from inside the component

packages/sit-onyx/src/components/OnyxSlider/OnyxSlider.vue
larsrickert: Since a slider must logically always have a value, I'd prefer to make the modelValue property required instead. This way, we don't need to deal with edge cases and its more explicit for the user.

We should then document in the JSDoc that the default value (if no other logically value is defined by the users requirement) should be the middle of mix and max.

packages/sit-onyx/src/components/OnyxSlider/OnyxSlider.vue
```diff
@@ -0,0 +1,725 @@
...
+
+const modelValue = useVModel({
+  props,
+  emit,
+  key: "modelValue",
+  default: () => [props.min ?? 0] as number[],
+});
+
+const customError = computed(() => props.customError);
+const { vCustomValidity, errorMessages } = useCustomValidity({ props, emit, customError });
```
larsrickert:
```suggestion
const { vCustomValidity, errorMessages } = useCustomValidity({ props, emit });
```

The composable already uses the props.customError by default so we don't need to pass it separately

packages/sit-onyx/src/components/OnyxSlider/OnyxSlider.vue
```diff
@@ -0,0 +1,725 @@
...
+const messages = computed(() => getFormMessages(props.message));
+
+defineOptions({ inheritAttrs: false });
+
+const { rootAttrs, restAttrs } = useRootAttrs();
```
larsrickert:
```suggestion
```

Since there is no single logical component that the slider is wrapping, I don't think we need to split up the attributes here and can just use's Vue's defaults :)

Unlike the OnyxInput or OnyxButton for example, there it is clear that the rootAttrs should be forwarded to the native HTMl input/button

packages/sit-onyx/src/components/OnyxSlider/OnyxSlider.vue
```diff
@@ -0,0 +1,725 @@
...
+
+const isValueControl = computed(() => props.control === "value");
+/**
+ * Icon control works only when there is a single thumb.
+ */
+const isIconControl = computed(() => props.control === "icon" && modelValue.value.length === 1);
+</script>
+
+<template>
+  <div
```
larsrickert, line 164: To keep things simple here and to be aligned with other form elements like inputs etc, we only want to show a single skeleton "block" instead of multiple fine-tuned skeletons

Image

packages/sit-onyx/src/components/OnyxSlider/OnyxSlider.vue
```diff
@@ -0,0 +1,725 @@
...
+    v-bind="rootAttrs"
+    class="onyx-component onyx-slider"
+    :class="[
+      {
+        'onyx-slider--vertical': props.orientation === 'vertical',
+        'onyx-slider--inverted': props.trackMode === 'inverted',
+        'onyx-slider--disabled': disabled,
+        'onyx-slider--is-dragging': isDragging,
+        'onyx-slider--is-focused': focusedThumbIndex !== -1,
+        'onyx-slider--is-active': activeThumbIndex !== -1,
```
larsrickert: Are those 3 classes needed? Can't we use native CSS selectors like `:focus-within` or `:active`?

packages/sit-onyx/src/components/OnyxSlider/OnyxSlider.vue
```diff
@@ -0,0 +1,725 @@
...
+      class="onyx-slider__form-element"
+      :message="messages"
+      :error-messages="errorMessages"
+    >
+      <template #default="{ id: inputId }">
+        <div
+          class="onyx-slider__container"
+          :class="{
+            'onyx-slider__container--with-marks': marksList.length > 0,
+          }"
```
larsrickert: We could replace this with the `:has` selector to reduce JavaScript based conditional classes

```css
.onyx-slider__container:has(.onyx-slider__mark)
```

packages/sit-onyx/src/components/OnyxSlider/OnyxSlider.vue
```diff
@@ -0,0 +1,725 @@
...
+          <div v-if="isIconControl" class="onyx-slider__icon-control-container">
+            <OnyxIconButton
+              :disabled="disabled || (modelValue[0] ?? props.min) <= props.min"
+              :label="props.label || inputId"
+              color="neutral"
+              :icon="iconMinusSmall"
+              tabindex="0"
+              @click="handleDecreaseByIcon"
+            />
+          </div>
```
larsrickert, line 219: Idea: Since the control rendering is quiet similar for the minus and plus variant, we could extract this into a support component OnyxSliderControl which renders accordingly depending on the control type

You don't have to do this in this PR, we can merge it like this just fine and the onyx team will do the refactoring later :)

packages/sit-onyx/src/components/OnyxSlider/OnyxSlider.vue
```diff
@@ -0,0 +1,725 @@
...
+
+          <span class="onyx-slider__root" v-bind="root" @touchstart.passive="root.onTouchstart">
```
larsrickert, line 222: Question: Why is the additional touchstart event needed here? The `root` binding already includes a touchstart event listener. Shouldn't this be enough?

If it is needed, a explainatory comment would be great here!

packages/sit-onyx/src/components/OnyxSlider/OnyxSlider.vue
larsrickert, line 1: When setting value 0 and then increasing the min value to e.g. 6, the thumb is moving too far to the left. Do you have an idea what could be the issue here?

Image

packages/sit-onyx/src/components/OnyxSlider/OnyxSlider.vue
```diff
@@ -0,0 +1,725 @@
...
+                v-if="markItem.label"
+                class="onyx-slider__mark-label"
+                v-bind="markLabel({ value: markItem.value })"
+              >
+                {{ markItem.label }}
+              </span>
+            </template>
+
+            <span
+              v-for="(value, index) in modelValue"
```
larsrickert: I talked with UX on this (supporting more than 2 thumbs). Sliders are intended to represent either a single numeric value or a defined range within a min/max (with 2 thumbs). Having 3+ thumbs would drift away from the UX purpose of a slider since it would allow to select some kind of position within the selected range (like choosing CSS gradients) which does not fit the purpose we intend the slider for :)

Therefore, we would prefer for now to only allow single or range mode (with 2 thumbs). This also allows us to better type the modelValue property generically so that its typed similar to this:
```ts
type SliderType = "single" | "range";

export type OnyxSliderProps<TType extends SliderType> = {
  /**
   * Whether to select a single value or a range of two values.
   */
  type?: SliderType;
  /**
   * Currently selected slider value.
   */
  modelValue?: TType extends "single" ? number : [number, number];
};
```

Still, thanks for thinking about such features!

packages/sit-onyx/src/components/OnyxSlider/OnyxSlider.stories.ts
```diff
@@ -0,0 +1,316 @@
...
+      options: ["horizontal", "vertical"],
+    },
+    control: {
+      control: { type: "radio" },
+      options: ["value", "input", "icon"],
+    },
+    tooltipDisplay: {
+      control: { type: "radio" },
+      options: ["auto", "always", "never"],
+    },
```
larsrickert:
```suggestion
```

Storybook auto generates those controls so we don't need to explicitly define them 💪🏼

packages/sit-onyx/src/components/OnyxSlider/OnyxSlider.stories.ts
```diff
@@ -0,0 +1,316 @@
...
+    (story) => ({
+      components: { story },
+      template: `<div style="width: 16rem;"> <story /> </div>`,
+    }),
+  ],
+  argTypes: {
+    ...withNativeEventLogging(["onChange"]),
+    trackMode: {
+      control: { type: "radio" },
+      options: ["default", "inverted", false],
```
larsrickert:
```suggestion
      options: SLIDER_TRACK_MODES,
```

We can import and use the existing const for this

packages/headless/src/utils/dom.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Checks if an element has focus-visible state (keyboard focus).
+ * Falls back to false if :focus-visible is not supported.
+ */
+export const isFocusVisible = (element: Element): boolean => {
```
larsrickert, line 5: To check if an element is the currently focused element on the page, I'd prefer to use:

```ts
element === document.activeElement;
```

packages/headless/src/utils/dom.spec.ts
```diff
@@ -0,0 +1,77 @@
+import { describe, expect, test, vi } from "vitest";
+import { isFocusVisible } from "./dom.js";
+
+describe("isFocusVisible", () => {
```
larsrickert: (optional) Maybe a Playwright test would be more appropriate here that e.g. mounts a custom .vue component which is a button and shows a "Is focused: true" label inside (using the isFocusVisible) utility.

We can then Tab the button using Playwright and check if it contains "Is focused: true"

This way, we test be behavior instead of the internal implementation (like mocking and testing the .matches method here)

packages/headless/src/composables/slider/createSlider.ts
```diff
@@ -0,0 +1,861 @@
...
+   *
+   * @default false
+   */
+  marks?: MaybeRef<Nullable<SliderMark[] | boolean>>;
+  /**
+   * Aria label for the slider.
+   *
+   * @default undefined
+   */
+  label?: MaybeRef<string>;
```
larsrickert:
```suggestion
  /**
   * Aria label for the slider.
   */
  label: MaybeRef<string>;
```

The label should be required for accessibility reasons :)

packages/headless/src/composables/slider/createSlider.ts
```diff
@@ -0,0 +1,861 @@
...
+export const createSlider = createBuilder((options: CreateSliderOptions) => {
+  const sliderRef = createElRef<HTMLElement>();
+
+  // Computed properties from options
+  const min = computed(() => unref(options.min) ?? 0);
+  const max = computed(() => unref(options.max) ?? 100);
+  const values = computed(() => unref(options.values) ?? [min.value ?? 0]);
+  const step = computed(() =>
+    typeof unref(options.step) === "undefined" ? 1 : unref(options.step),
+  );
```
larsrickert:
```suggestion
  const step = computed(() => unref(options.step) ?? 1);
```

packages/headless/src/composables/slider/createSlider.ts
```diff
@@ -0,0 +1,861 @@
...
+   * Reason: lets the slider track and finish the drag with the *same finger*
+   * only—avoiding interference from other touches in multi-touch scenarios and
+   * preventing stuck or incorrect drags.
+   */
+  let touchId: number | null = null;
+  /**
+   * Counts the number of move events during a drag operation.
+   * Reason: lets us differentiate between intentional drags and accidental slight movements,
+   */
+  let movesSinceStart: number = 0;
```
larsrickert: Why are you differentiating between intentional and unintentional changes? Do you have an example use case? I'd think we just always emit the changed values.

This might cause unintentional behavior if we classify a change as unintentional but it was intended by the user.

packages/headless/src/composables/slider/createSlider.ts
```diff
@@ -0,0 +1,861 @@
...
+   * Indicates whether the slider is a range slider (with two or more thumbs) or a single value slider (with one thumb).
+   */
+  const isRange = computed(() => values.value.length > 1);
+
+  /**
+   * If `showMarks` is `true` and `step` is provided, marks will be generated for each step in the range.
+   * If `marks` is an array and `showMarks` is `true`, it will be used as provided.
+   * If `showMarks` is `false`, no marks will be shown.
+   */
+  const marksList = computed<{ value: number; label?: string }[]>(() => {
```
larsrickert:
```suggestion
  const marksList = computed<Exclude<SliderMark, number>[]>(() => {
```

So we don't need to maintain the type in two places :) Alternatively, we can also remove the `| number` from the type itself and define the marks option as `(SliderMark | number)[]`

packages/headless/src/composables/slider/createSlider.ts
```diff
@@ -0,0 +1,861 @@
...
+    if (shouldSetActive) {
+      activeThumbIndex.value = index;
+    }
+  };
+
+  const eventToCoords = (event: MouseEvent | TouchEvent, touchId?: number | null) => {
+    if (touchId !== undefined && (event as TouchEvent).changedTouches) {
+      const touchEvent = event as TouchEvent;
+      for (let i = 0; i < touchEvent.changedTouches.length; i += 1) {
+        const touch = touchEvent.changedTouches[i];
```
larsrickert:
```suggestion
    if (touchId !== undefined && event instanceof TouchEvent) {
      for (let i = 0; i < event.changedTouches.length; i += 1) {
        const touch = event.changedTouches[i];
```

packages/headless/src/composables/slider/createSlider.ts
```diff
@@ -0,0 +1,861 @@
...
+        step: step.value === null && marks.value ? "any" : (step.value ?? undefined),
+        disabled: typeof isDisabled.value === "boolean" ? isDisabled.value : false,
+        onChange: handleHiddenInputChange,
+        onFocus: handleHiddenInputFocus,
+        onBlur: handleHiddenInputBlur,
+        onKeydown: handleHiddenInputKeydown,
+        /**
+         * Visually hidden styles.
+         */
+        style: {
```
larsrickert: The styling should be responsibility of the component, not the headless logic :) onyx has a OnyxVisuallyHidden component that you can use for this

packages/headless/src/composables/slider/createSlider.ts
```diff
@@ -0,0 +1,861 @@
...
+      /**
+       * List of marks to display on the slider.
+       * - If marks option is `true`, marks are generated based on the step value.
+       * - If marks option is an array of numbers or objects, it is used as provided (filtered to be within range).
+       * - If marks option is `false`, no marks are shown.
+       */
+      marksList,
+    },
+
+    internals: {
```
larsrickert: The internals here all use reactive state so they should be a computed to ensure that they update when their reactive dependencies update. For example:

```ts
const valueToPercent = computed(() => {
    return (value: number) => MathUtils.valueToPercent(value, min.value, max.value);
  });
```

packages/sit-onyx/src/components/OnyxSlider/types.ts
larsrickert: Nice addition :) What do you think, should we use a default value that is somehow depending on the min, max or step prop? E.g. that the shift step is always 10% of the available range.

Because propably most users will just change the min, max or step property without noticing that they should also update the shiftStep property.

packages/sit-onyx/src/components/OnyxSlider/OnyxSlider.stories.ts
```diff
@@ -0,0 +1,316 @@
...
+    marks: true,
+    label: "Automatically Generated Marks",
+  },
+} satisfies Story;
+
+/**
+ * This example shows the discrete state of the slider with marks.
+ * The step is set to null, so the thumbs can only be moved to the marks.
+ */
+export const Discrete = {
```
larsrickert, line 178: Question: Whats the API currently to have "discrete" steps?
If I see it correctly the users must pass marks and set the step to null, right? Would be great if we can find an API that also allows auto-generating the marks (which requires the step property).

Maybe we can make this more explicit and add a new property `discrete?: boolean`

packages/headless/src/composables/slider/createSlider.ts
lovelycentury: Good point

My goal here is not to guess user intent but to guard against tiny touches. On touch devices a tap/scroll can fire 1–2 pointermove events and accidentally change the value. I only start emitting after two moves to avoid that.

I will rename INTENTIONAL_DRAG_COUNT_THRESHOLD to DRAG_MOVE_THRESHOLD and update the JSDoc to say “prevent accidental value changes from minor touch movement.” to o make it more clear.

packages/headless/src/utils/dom.ts
lovelycentury: `element === document.activeElement` checks focus for both mouse and keyboard interactions,
while `:focus-visible` only matches when focus comes from the keyboard.
I’d keep them separate since they serve different purposes.

packages/sit-onyx/src/components/OnyxSlider/types.ts
lovelycentury: Good point

I did not have a specific use case in mind. I just added it because I was inspired by another library, and I thought it might be nice to stay consistent with that API.

If UX decided to keep only the default style, I am fine with removing it.

packages/sit-onyx/src/components/OnyxSlider/OnyxSlider.stories.ts
lovelycentury: Let's stay with the auto-generated ones

packages/sit-onyx/src/components/OnyxSlider/OnyxSlider.stories.ts
lovelycentury: I'll add `discrete: boolean` to `createSlider` and `OnyxSlider`. And in addition to this, I will remove the nullable `step`.
 
To generate marks automatically, all you need to do is pass `marks: true`. This wasn't properly documented, but I'll update the documentation.

packages/sit-onyx/src/components/OnyxSlider/types.ts
lovelycentury: Will be removed.

Previously, `null` was an indicator of discrete behavior. Now, the `discrete: boolean` prop will be used for this purpose.

packages/sit-onyx/src/components/OnyxSlider/types.ts
lovelycentury: That’s a brilliant idea ✨.

Making it 10% of the min–max range sounds perfectly logical! It would make the behavior much more intuitive when those values change.

packages/sit-onyx/src/components/OnyxSlider/OnyxSlider.vue
lovelycentury: Good catch :)

I added it explicitly because I wanted the listener to be passive, but Vue doesn’t let you set `{ passive: true } ` through v-bind event bindings.

That’s why I had to define it manually here. I’ll add a short comment to clarify that.

packages/sit-onyx/src/components/OnyxSlider/OnyxSlider.vue
lovelycentury: You’re right about keeping modelValue updated immediately to align with other form elements.
That said, a “committed” moment is a very common slider pattern across libraries (used to defer heavy updates, fire analytics, etc.).

Compromise proposal: I’ll remove the public update:committed` event from the UI component, but keep the commit logic inside the headless implementation.
This way we don’t expose extra API on the component, while advanced users can still access commit semantics via the headless layer when needed.

If that works for you, I’ll update the PR accordingly.

packages/sit-onyx/src/components/OnyxSlider/OnyxSlider.vue
lovelycentury: Thanks for the suggestion! I’ll keep it as-is in this PR.

In a follow-up, we can extract an `OnyxSliderControl` helper component that also covers control: 'input'.
If that sounds good, let's plan it for a separate PR.

packages/sit-onyx/src/components/OnyxSlider/OnyxSlider.vue
lovelycentury: Not really,  `onyx-slider--is-dragging` and `onyx-slider--is-focused` are really unnecessary. 
And on a desktop the usual would be enough `:focus-within`, but because of the mobile drag it doesn't always work correctly and for this we need `onyx-slider--is-active`

packages/sit-onyx/src/components/OnyxSlider/OnyxSlider.vue
lovelycentury: Good point

There's no normalization of values ​​when changing min, max, or step prop dynamically. I'll add it.

packages/headless/src/composables/slider/createSlider.ts
lovelycentury: Unfortunately, we can't check it in Safari like `event instanceof TouchEvent`, because TouchEvent constructor doesn't exist in Safari. If you don't like the type assertion, I could add a guard for this purpose, like

`const isTouchEvent = (event: Event): event is TouchEvent =>
  "touches" in event || "changedTouches" in event || "targetTouches" in event;
`

packages/sit-onyx/src/components/OnyxSlider/types.ts
lovelycentury: `disableTooltip?: boolean` will be provided

packages/headless/src/composables/slider/createSlider.ts
larsrickert: Good catch :) Both is fine for me, if you have time to implement a type guard that would be ideal!

larsrickert:
@lovelycentury Thanks a lot for taking your time to address the comments! We can merge the PR as it is now :)
Since you also added Playwright screenshot tests, can you run the "[Update Playwright screenshots](https://github.com/SchwarzIT/onyx/actions/workflows/playwright-screenshots.yml)" pipeline for your branch in your fork?
It will generate the screenshots and create a new PR to your branch to merge the screenshots

Let me know if the pipeline does not work in your fork. We can force merge the PR then so I can update the screenshots in the onyx repository itself

lovelycentury:
> @lovelycentury Thanks a lot for taking your time to address the comments! We can merge the PR as it is now :) Since you also added Playwright screenshot tests, can you run the "[Update Playwright screenshots](https://github.com/SchwarzIT/onyx/actions/workflows/playwright-screenshots.yml)" pipeline for your branch in your fork? It will generate the screenshots and create a new PR to your branch to merge the screenshots
> 
> Let me know if the pipeline does not work in your fork. We can force merge the PR then so I can update the screenshots in the onyx repository itself

Can I generate them myself on my branch by disabling these two settings? :)

```  // we don't want to update snapshots on the local machine of each developer.
  // if you want to update snapshots for your branch, use the corresponding GitHub action:
  // https://github.com/SchwarzIT/onyx/actions/workflows/playwright-screenshots.yml
  ignoreSnapshots: !process.env.CI,
  updateSnapshots: process.env.PW_UPDATE_SNAPSHOTS === "true" ? "changed" : "none",```

larsrickert:
> Can I generate them myself on my branch by disabling these two settings? :)

Unfortunately not, they must be generated in GitHub because if you run them on your machine, they are generated for your operating system (mostly Mac or Windows) so when comparing them, they will guarantee to fail because text rendering etc. differ in different OS

PR #4251, opened 2025-10-13
fix(OnyxDataGrid): Fix issue where row height didn't adapt to content height on webkit

### LINKED ISSUE #4250: Fix OnyxDataGrid row height doesn't adapt on Webkit
### Version

1.1.0

### Description

The problem appears when you open the datagrid example in safari and display the webpage as iPhone SE size. It is important to open it up in safari as other browsers behave differently.
Due to limited screen size the value in "Birthday" gets will break into multiple lines however as a user I have no chance to see the hidden parts of the text as the columns aren't scrollable.
The same behaviour seems to appear if you use an actual iPhone SE so at least it should be reproducible using the simulation mode in safari on a mac.

(reproducible on iphone 16/17 as well)

### Reproduction steps

https://playground.onyx.schwarz/?onyxVersion=1.1.0#eNqFVW1P2zAQ/isnf9qkhr4AG2SFqXTVBB9gAr4RNDmJ2xgcO7IdaFX1v+/spEnKKKhVVZ/Pzz139/i8JpOiOHgpGQnJ2CSaFxYMs2UBgsrFWUSsich5JHleKG1hDTdytfxFLf2tedoDuyoYTJUoczlVcs4XsIG5VjlExHAbKHSOyI9IRtJ73s+k1Ss4g3UkAXgagizzmGn0AJA0ZyEYq7lceANd4LrjEHNts5SuQkACzJu4mSSWv7CfIcRKCUYlmjc+YKKksZAi1bCO+/CIkR/csbWPPezVMSMyETxhEelVMQ8HvU4wyV59wC8RGZ6eDoLBEL8R+dproodgdclg02uxRy32NKNa8A768X70k2AwCoZv0edUmF34wxb+QsUN9OgD6ONg8C0YHn8KfdRC36qYyxb8ZA/4aOBKMgpGo8+qgvS22Fcqa6GPsFz7eH8PBkfB8GQv78dOtxOvRBPuSHJcdf+82/5nhnEi4sg4EoLGTDjDtTc0nGs35LjjNanWTtIexUu0PeZ+YHt2mxYeqOwNysXbnQpv3ZQItevjqMJyTM9tzZXOqXX/3OadXQnvOS+FcPHdx4O9x2RbvP+ZXBqg1TVq9l55ajO3l9NlgMW1TNo3RDG36so19n4fVqqEhErA9igoqDGQlMbiQKiTAKvAZqwaHPHKu+B9Bzyi4ieWWODYSEZTUHOg9TRo0OsKNevt0HiXCnQK1zEifV3aDEuyawWnHqUdmCmThBkcfF2wbVHrZWfRltxpcdyvxigOTVxYlhcCW4UrgHF3eEKGWQouGQ7Z2ZKiG2tM2Muw1jKEboRBHwHG/QaN9HAwY1tQ3gdPRkmc3j6biCQqL7hg+qbKPSJNnqhjIdTrlbe5q1lngGcyljy/Y38yOL1dQf5oZpjuSAffBaoXDCXhtmd312zZygNFo9JSoPcHm7fMYIKOY+V2UcoUaXf8PNtL/+6gBu7NbIkiNNukHNFW6RHBJ2z6Qeot3cODI38O3wms4t8Xph0mFtAV1liy+Qc8cTrT

### Confirmations

JoCa96:
Closes #4250

- Fix issue where row height didn't adapt to content height:
  - Adding `grid-template-rows: subgrid;` for `tr` to resolve the issue without affecting Chrome/Firefox
  - Unset `OnyxTable` height in OnyxDataGrid, as it would otherwise interfere with the height defined in the `OnyxDataGrid`
- Added `truncation` property to table, which allows users to switch to `ellipsis` mode instead of always having `multiline` behaviour.

### DIFF
--- a/.changeset/legal-hounds-eat.md
+++ b/.changeset/legal-hounds-eat.md
@@ -2,4 +2,4 @@
 "sit-onyx": minor
 ---
 
-feat(OnyxTable, OnyxDataGrid): Added `truncation` property which allows to define multiline behaviour of the table rows
+feat(OnyxTable, OnyxDataGrid): Added `truncation` property which allows to define multiline or ellipsis truncation (single line) behaviour of the table rows


.changeset/legal-hounds-eat.md
```diff
@@ -0,0 +1,5 @@
+---
+"sit-onyx": minor
+---
+
+feat(OnyxTable, OnyxDataGrid): Added `truncation` property which allows to define multiline behaviour of the table rows
```
larsrickert:
```suggestion
feat(OnyxTable, OnyxDataGrid): Added `truncation` property which allows to define multiline or ellipsis truncation (single line) behaviour of the table rows
```

packages/sit-onyx/src/components/OnyxTable/OnyxTable.vue
```diff
@@ -266,6 +263,14 @@ $border: var(--onyx-1px-in-rem) solid var(--onyx-color-component-border-neutral)
       line-height: var(--onyx-font-line-height-md);
     }
 
+    &--cell-truncation-ellipsis {
```
larsrickert, line 266: Should we also add styles for the multiline truncation? Or are those not needed? In our truncation utility classes, we define them like so:

```css
  .onyx-truncation {
    &-ellipsis {
      white-space: nowrap;
      overflow: hidden;
      text-overflow: ellipsis;
    }

    &-multiline {
      // needed to wrap the text to a newline even when its one long word,
      // otherwise the max width would be overflown
      overflow-wrap: anywhere;
      hyphens: auto;
    }
  }
```

larsrickert (approved):

packages/sit-onyx/src/components/OnyxTable/OnyxTable.vue
JoCa96: I would prefer not to add it, as it can interfere with the data presentation. E.g. it could add an extra `-` dash inside of an ID string

PR #4274, opened 2025-10-15
fix: Fix extraneous props rendered as attributes and cluttering the DOM

### LINKED ISSUE #3966: Fix additional props appearing as html attributes when `v-bind`ing all props to child component
### Why?

We often extend or modify Props of wrapped child components and then pass these through using `v-bind`.

This can cause html attributes to appear unintentionally, when not all props are also defined by the Child component:

[Check this Playground example](https://play.vuejs.org/#eNqtVMFuGjEQ/ZWpD00iwVJoT2RBatMc2kOL2hx9wLADOF1sy/YSIsS/d2yzsIsQaqXsZe2Z55n3Zjzesc/GZJsK2ZDlbm6l8eDQVwZKoZYjzrzjbMyVXBttPexgIiwqP7HaONjDwuo13GQ9/2rQZd7d3B+hDytZFv0jIm1DpnPMoI0Z1Biu5lo5DybmGsGL9KuvuBBV6d1tgQupMNLIG5TGt3cdIimG4G2FHZgN4UMHVrIoUCUb7O8odt5LWkkZbTyuTSk80g4OAsMqf9fthj/UWqQ7hIIZzkXlcAjJH79+BpwlJGdQaHTqxkMiCgKm6eg06mkcG2TwtEKLC20R/AphGgBTIBL2NaQUxpQSCxC0BOG9lbPKY4jQ7UbG+YHepjuTqqCexYoRh15yk4y60olNCNSiA0IVIRWlPxBR2tdaW1kGl7PkvUYNWYduDfVuIZfZs9OKrtYuROBsrtdGlmh/Gi+pt5wNIXqCT5SlfvkebbF3tX2+wvmfC/Zntw02ziYWHdoNcnb0eWGX6JP78fcP3NL66FzroioJfcX5C50uq8Axwb5UqiDaDVxk+y3eYqmWT+5x61G5WlQgGpD7iOeMbvTDFeknuh+zT/EcV3uq4mlm/nU644k4Cf3z6bw0Us0pahylKbo+I3khN3FxGA3oR2svmc8vw2ms3+6Ruaak/R78p5LBNSU1A9LBFW4j22Br1X2UGktv0EzrEoUiBhBeIlWtZ2hpR+1tH2/qHYWdl6JsdgTeU1XqZ+wQFvbje7b/Cxpt7SE=)

So far we didn't notice any issues resulting from this.
But as the example above shows, it is possible for props to unintentionally break stuff.

### Acceptance criteria

- [x] a new composable exists, that picks only valid props for a child component
- [x] it is reactive (=> computed)
- [x] has unit tests
- [x] is applied everywhere, where props are bound indiscriminently
- [x] *Nice-to-have*: Otherwise Follow-up ticket
  - [ ] ~eslint rule, that ensures, that props are not bound directly to a child~ FoUp: [#4277](https://github.com/SchwarzIT/onyx/issues/4277)

### Implementation details

We can access the `prop` object on most component definitions (always for SFC).
Using that we can map the valid props:

```ts
// TODO: Fix type
// TODO: with computed
// TODO: check if component is ConcreteComponent using typeguard
const useForwardProps = <T extends Data, C extends Component, R = Pick<T, keyof ComponentProps<C>>>(props: T, target: C) => {
  const keys = new Set(Object.keys((target as ConcreteComponent).props));
  return Object.fromEntries(
    Object.entries(props).filter(([key]) => keys.has(key)),
  ) as R;
};

const forward = useForwardProps(props, OnyxChildComponent)
```

### Definition of Done

- Should be covered by tests:
  - [x] functional tests (Playwright or unit test)

- Make sure, that
  - [x] follow-up tickets were created if necessary
  - [x] updated version + documentation is deployed

#### Approval

[Documentation]()

- Checked and approved by
  - [ ] dev

JoCa96:
Closes #3966

- fix: Fix extraneous parent props rendered as attributes and cluttering the DOM
  - implemented `useForwardProps` composable, which allows to only forward relevant props to the child component
  - added extensive description to our *Patterns* doc page
- renamed the `MaybePick` util type to `MaybeUnwrap`
  - added type tests 
- created new util type `MaybePick` which works like the native `Pick` type, but can handle non-existent props.
  - added type tests


packages/sit-onyx/src/components/OnyxAlertModal/OnyxAlertModal.vue
```diff
@@ -14,6 +15,8 @@ const props = withDefaults(defineProps<OnyxAlertModalProps>(), {
   icon: () => ({ icon: iconCircleAttention, color: "danger" }),
 });
 
+const dialogProps = useForwardProps(props, OnyxBasicDialog);
```
larsrickert, line 18: With this approach it would theoretically be possible to use a component here that does not match the actually used component in the template / HTML, right?

Do you know if there is a way to prevent this? Or would this be a job of the custom eslint rule?

packages/sit-onyx/src/utils/props.ts
```diff
@@ -0,0 +1,55 @@
...
+  TProps = ComponentProps<TComponent>,
+  R = MaybePick<T, keyof TProps>,
+>(
+  props: T,
+  target: TComponent,
+) => {
+  // endregion docs
+  const keys = new Set(Object.keys((target as ConcreteComponent).props ?? {}));
+
+  if (!(target as ConcreteComponent).props) {
```
larsrickert, line 46: Why do we want to log an error in this case? If a component does not have any props and I forward the props, the returned object should just be an empty object as expected, no?

packages/sit-onyx/src/components/OnyxAlertModal/OnyxAlertModal.vue
JoCa96: The props are typed, so typescript at least catches mismatching props.
Yes, a eslint rule seems like the best approach.

packages/sit-onyx/src/utils/props.ts
JoCa96: `props` is an object describing the properties that the component can receive.
If it doesn't exist, no props are defined at all.
Trying to forward no props is almost certainly an error, so logging an error makes sense to me here.
An empty object is returned anyways, so it doesn't have any side-effects.

Issue #4258, opened 2025-10-14
Implement individual disabled dates for OnyxCalendar


larsrickert:
### Why?

We want to enable the OnyxCalendar to support disabling only single or multiple specific dates (e.g. holidays).

### Open questions / TODOs

- how should the range selection behave if disabled dates are in between the selected range? Answer: We leave this up to the project to e.g. set a different min/max date, no todo on our side.

### Acceptance criteria

- [ ] one or more individual dates can be disabled

### Implementation details

For flexibility, we can e.g. implement it as callback function so that the user can also pass in some kind of pattern (e.g. disable every Monday) instead of having to explicitly add all dates

```ts
export type OnyxCalendarProps = {
  disabled?: boolean | ((date: Date) => boolean)
}
```

### DEV handover checklist

[Storybook](ADD_LINK_HERE)

DEV handover done with: @ADD_DEV_NAME_HERE

**Testing**
- [ ] covered by visual screenshot tests
- [ ] covered by functional tests (Playwright or unit test)

**Documentation**
- [ ] "New" badge in Storybook is added, see [storybook-addon-tag-badges](https://github.com/Sidnioulz/storybook-addon-tag-badges/?tab=readme-ov-file#-usage)
- [ ] auto-generated Storybook code snippets are checked
- [ ] component status has been updated in the [docs overview](https://onyx.schwarz/#components)
- [ ] updated version is released
- [ ] updated documentation is deployed

**Other**
- [ ] follow-up tickets were created if necessary (add links below)

### UX handover checklist

UX handover done with: @Guergchen 

**Naming is aligned with Figma**
- [x] component name
- [x] which support components are used and how they are named
- [x] all property names and types

**Implementation is checked**
- [x] all features of Figma are implemented or a follow-up ticket is refined
- [x] skeleton
- [x] densities
- [x] light and dark mode
- [x] truncation
- [x] responsiveness / wrapping behavior
- [x] min / max width

**Figma**
- [x] Figma component is published

**Documentation**
- [x] Storybook category is aligned with Figma
- [x] Storybook examples are checked

**Other**
- [x] follow-up tickets were created if necessary (add links below)

PR #4281, opened 2025-10-16
feat(OnyxCalendar): implement individual disabled dates

ChristianBusshoff:
Relates to #4258

- added individual disabled dates
- added range example
- added tests


packages/sit-onyx/src/components/OnyxCalendarCell/OnyxCalendarCell.vue
```diff
@@ -24,6 +24,11 @@ const slots = defineSlots<{
 }>();
 
 const { densityClass } = useDensity(props);
+const handleHover = () => {
+  if (!props.disabled) {
+    emit("hovered");
+  }
+};
```
JoCa96: Prefer to set event-listener dynamically instead of having a conditional in the handler.
That way the the event-listener won't even be created:


```suggestion
const hoverHandler = computed(() =>
  !props.disabled
    ? {
        onMouseenter: () => emit("hovered"),
        onFocusin: () => emit("hovered"),
      } satisfies HTMLAttributes
    : {},
);
```

packages/sit-onyx/src/components/OnyxCalendarCell/OnyxCalendarCell.vue
```diff
@@ -45,8 +50,8 @@ const contentAttributes = computed(() => {
         [`onyx-calendar-cell--range-${props.rangeType}`]: props.rangeType,
       },
     ]"
-    @mouseenter="emit('hovered')"
-    @focusin="emit('hovered')"
+    @mouseenter="handleHover"
+    @focusin="handleHover"
```
JoCa96: See other comment;

```suggestion
    v-bind="hoverHandler"
```

packages/sit-onyx/src/components/OnyxCalendarCell/OnyxCalendarCell.vue
JoCa96, line 1: We don't want to replicate native events using Vue events.
Good indicators, that a native event is replicated are:
- the emit is named like a native event
- the emit is triggered by an event listener without additional business logic
- these event listener are on the root element of the component

Therefore please remove the `hovered` emit and move the event-listener to the parent.

packages/sit-onyx/src/components/OnyxCalendar/types.ts
```diff
@@ -12,7 +12,7 @@ export type OnyxCalendarProps<TSelection extends OnyxCalendarSelectionMode> = De
   /**
    * Whether the calendar is disabled. Disables all interactions and prevents date selection.
    */
```
JoCa96, line 26: Please extend the description regarding the new callback based approach, and maybe add an example

JoCa96 (changes requested):

packages/sit-onyx/src/components/OnyxCalendar/types.ts
```diff
@@ -10,9 +10,17 @@ export type OnyxCalendarProps<TSelection extends OnyxCalendarSelectionMode> = De
...
-   * Whether the calendar is disabled. Disables all interactions and prevents date selection.
+   * Whether the calendar is disabled.
+   * * This can be a simple boolean to globally disable all interactions and selection,
+   * or a **callback function** to disable specific dates individually.
+   * * @example
+   * // Globally disables the calendar
+   * disabled: true
+   * * @example
+   * // Disables only weekends (Saturday and Sunday)
+   * disabled: (date: Date) => date.getDay() === 0 || date.getDay() === 6
```
JoCa96:
```suggestion
   * @example
   * ```ts
   * // Globally disables the calendar
   * { disabled: true }
   * ```
   * @example
   * ```ts
   * // Disables only weekends (Saturday and Sunday)
   * { disabled: (date: Date) => date.getDay() === 0 || date.getDay() === 6 }
   * ```
```

JoCa96 (approved):

PR #4284, opened 2025-10-16
fix(OnyxDatepicker): Fix incorrect `v-model` type and behaviour which could cause runtime type mismatches

JoCa96:
Fix type mismatch between `update:modelValue` and `modelValue`, both are now always using [ISO8601](https://en.wikipedia.org/wiki/ISO_8601) strings.
This mismatch caused the value of a `ref` that is used with `v-model` to be updated as string, even though it could have been typed as `Date` or `number`.

### DIFF
--- a/packages/sit-onyx/src/components/OnyxDatePicker/OnyxDatePicker.ct.tsx
+++ b/packages/sit-onyx/src/components/OnyxDatePicker/OnyxDatePicker.ct.tsx
@@ -76,14 +76,14 @@ test("should emit events", async ({ mount, makeAxeBuilder }) => {
 
   // ASSERT
   await expect(inputElement).toHaveValue("2024-11-25T12:34");
-  expectEmit(onUpdateModelValue, 3, ["2024-11-25T12:34:00Z"]);
+  expectEmit(onUpdateModelValue, 3, ["2024-11-25T11:34:00.000Z"]);
 
   // ACT
   await inputElement.clear();
 
   // ASSERT
   await expect(inputElement).toHaveValue("");
-  expectEmit(onUpdateModelValue, 2, [undefined]);
+  expectEmit(onUpdateModelValue, 4, [undefined]);
 });
 
 test("should show min errors", async ({ mount }) => {


packages/sit-onyx/src/components/OnyxDatePicker/OnyxDatePicker.vue
```diff
@@ -85,7 +80,8 @@ const value = computed({
   get: () => getNormalizedDate.value(modelValue.value),
   set: (value) => {
     const newDate = new Date(value ?? "");
-    modelValue.value = isValidDate(newDate) ? newDate.toISOString() : undefined;
+    // If the type is `datetime-local`, we always use UTC as a timezone to minimize edge-cases for our users.
+    modelValue.value = dateToISOString(newDate, props.type === "date" ? "date" : "datetime-utc");
```
larsrickert, line 84: To be honest I don't think we should transform the date into UTC because it is provided as local ISO string via the props.

So the user would e.g. pass in 11 o'clock but we emit 13. So if the user would now parse the emitted date string again, it would be 15

Or am I getting something wrong here?

packages/sit-onyx/src/utils/date.ts
```diff
@@ -2,9 +2,41 @@
...
+ * @example dateToISOString(new Date("2025-10-16T11:01:09.564Z", "datetime-utc")) // "2025-10-16T11:01:09.564Z"
+ * @example dateToISOString(new Date("2025-10-16T13:01:09.564Z", "datetime-local")) // "2025-10-16T13:01:09.564"
+ * @example dateToISOString(new Date("not-a-date")) // null
+ */
+export const dateToISOString = (
+  date: Date | undefined,
+  type: "date" | "datetime-utc" | "datetime-local",
+): string | null => {
+  if (!isValidDate(date)) {
+    return null;
```
larsrickert: I'd prefer to use undefined since we never emit null in other components, only undefined

packages/sit-onyx/src/components/OnyxDatePicker/OnyxDatePicker.vue
larsrickert: So basically we should enforce / document that we use the same format for the property and the emitted event.

Either we say to always use the local date or we also use UTC. But since we previously used the local date, changing to UTC would be a breaking change

larsrickert: Imo if the user want to have the UTC date, he should transform it on his own:
```
<OnyxDatepicker @update:model-value="value = new Date($event).toUTCString()" />
```

JoCa96: > To be honest I don't think we should transform the date into UTC because it is provided as local ISO string via the props.
> 
> So the user would e.g. pass in 11 o'clock but we emit 13. So if the user would now parse the emitted date string again, it would be 15
> 
> Or am I getting something wrong here?

The timezone doesn't effect the point in time, e.g.
- `new Date(1760622139012)` // UNIX time in ms
- `new Date("2025-10-16T13:42:19.012Z")` // UTC
- `new Date("2025-10-16T15:42:19.012+2")` // UTC+2 (explicit time zone)
- `new Date("2025-10-16T15:42:19.012")` // UTC+2 implicit local time
_You can also try it out in Node.js or the browser console_

All of the above are representing the same point in time; timezone is basically only a accessory to make it look nicer.
But looking nice is not important for data transfer protocols, instead consistency and ease-of-use.

JoCa96: > So basically we should enforce / document that we use the same format for the property and the emitted event.
> 
> Either we say to always use the local date or we also use UTC. But since we previously used the local date, changing to UTC would be a breaking change

The format is still a ISO8601 compliant string, so there is no change to what was supported before.
If you put all of them into a Date constructor they result in the same Date and timezone.

packages/sit-onyx/src/utils/date.ts
JoCa96: Sounds good, done!

PR #4291, opened 2025-10-17
chore: add dependabot-dedupe workflow

JoCa96:
Dependabot [doesn't dedupe dependencies](https://github.com/dependabot/dependabot-core/issues/5830) after updating them.
This bloats up the node_modules unnecessarily, causes additonal manual work and confusion in the past.

This PR addresses this issue by adding a workflow to automatically dedupe and update all dependabot PRs.

Based on: https://github.com/mingjunlu/mingjun.lu/blob/f22bbca2222f65d501fab79dd10333864450ec24/.github/workflows/dependabot-dedupe.yaml

### DIFF
--- a/.github/workflows/dependabot-dedupe.yml
+++ b/.github/workflows/dependabot-dedupe.yml
@@ -5,7 +5,9 @@ on:
     branches:
       - "dependabot/npm_and_yarn/*"
     paths:
-      - "package.json"
+      - "**/package.json"
+      - "pnpm-lock.json"
+      - "pnpm-workspace.json"
 
 permissions:
   contents: write


.github/workflows/dependabot-dedupe.yml
```diff
@@ -0,0 +1,39 @@
+name: Deduplicate Dependabot PRs
+
+on:
+  push:
+    branches:
```
larsrickert, line 5: Doesn't this lead to repeating pipeline runes? Because this pipeline is run on push to the PR, but itself also pushes to the PR so its run again afterwards etc.

JoCa96: Therefore we check for the dependabot in line 19: `if: github.actor == 'dependabot[bot]'` :)

larsrickert: Ah I didn't see that, thanks!

larsrickert (approved):

PR #4293, opened 2025-10-17
chore: re-enable check pipeline for main branch

JoCa96:
- Our [code scanning results](https://github.com/SchwarzIT/onyx/security/code-scanning) are out of date. Seems like the main branch trigger was (accidentally?) removed in https://github.com/SchwarzIT/onyx/commit/4eee8390a651cc76021a6fa24e0f923642599b93

- Added explicit permissions

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -81,6 +81,7 @@ jobs:
           path: coverage
 
   screenshots:
+    if: github.ref != 'refs/heads/main'
     needs: build
     name: Component tests
     uses: ./.github/workflows/playwright.yml


.github/workflows/check.yml
```diff
@@ -1,5 +1,8 @@
 name: Check code quality
 on:
+  push:
+    branches:
+      - main
```
larsrickert, line 5: Can we save pipeline run times by not running playwright tests when on main?
Basically when running on main, only linting should be run, right? Build, tests etc. have all already been run in the merge queue

larsrickert (changes requested):

.github/workflows/check.yml
JoCa96: done

PR #4305, opened 2025-10-20
feat(OnyxModal): add grid example

ChristianBusshoff:
- added Grid example


packages/sit-onyx/src/components/OnyxModal/OnyxModal.stories.ts
```diff
@@ -21,3 +21,4 @@ const meta: Meta<typeof OnyxModal> = {
 export default meta;
 
 export const Default = createAdvancedStoryExample("OnyxModal", "DefaultExample");
+export const Gird = createAdvancedStoryExample("OnyxModal", "GridExample");
```
larsrickert:
```suggestion
export const Grid = createAdvancedStoryExample("OnyxModal", "GridExample");
```

Issue #1714, opened 2024-08-07
Implement OnyxPagination


JoCa96:

JoCa96:
@mj-hof @jannick-ux We need some more input from you before we feel confident estimating and starting this story.

PR #4310, opened 2025-10-21
Feat(OnyxPagination): implement inline pagination

ChristianBusshoff:
Relates to #1714

- added InlinePagination
- added tests 

- didn't implement Keyboard-control via left, right cause of aria-pattern

### DIFF
--- a/packages/sit-onyx/src/components/OnyxPagination/OnyxInlinePagination.vue
+++ b/packages/sit-onyx/src/components/OnyxPagination/OnyxInlinePagination.vue
@@ -86,7 +86,6 @@ const displayPagesNumbers = computed(() => {
     ]"
     role="group"
     :aria-label="t('pagination.label')"
-    :style="{ '--onyx-pagination-character-count': props.modelValue.toString().length }"
   >
     <button
       class="onyx-pagination__navigate-button"
--- a/packages/sit-onyx/src/components/OnyxPagination/OnyxSelectPagination.vue
+++ b/packages/sit-onyx/src/components/OnyxPagination/OnyxSelectPagination.vue
@@ -138,7 +138,12 @@ const hasReachedMax = computed(() => props.modelValue >= props.pages);
 .onyx-pagination-skeleton {
   @include layers.component() {
     height: var(--onyx-pagination-height);
-    width: calc(11.5rem + 4 * var(--onyx-density-sm) + 4 * var(--onyx-pagination-padding-vertical));
+    // 2 large buttons, 4.25rem width +left/right padding
+    // + 2 buttons, each is 1.5rem wide + left/right padding
+    width: calc(
+      2 * (4.25rem + 2 * var(--onyx-density-sm)) + 2 *
+        (1.5rem + 2 * var(--onyx-pagination-padding-vertical))
+    );
   }
 }
 


packages/sit-onyx/playwright/snapshots/components/OnyxPagination/Pagination-inline--edge-linux.png
larsrickert, line 1: Nice to have: Currently the skeleton is defined to have a fixed with in Figma. Imo it would be nice if the skeleton has the same width as the "default" variant here in the screenshot (considering density) so the layout does not jump around. Should we clarify this with UX?

packages/sit-onyx/src/components/OnyxPagination/types.ts
```diff
@@ -18,4 +18,11 @@ export type OnyxPaginationProps = DensityProp & {
    * Whether to show a skeleton pagination.
    */
   skeleton?: SkeletonInjected;
+  /**
+   * The display mode of the pagination
+   * @default select
+   */
+  mode?: OnyxPaginationModes;
```
larsrickert:
```suggestion
  /**
   * The display mode of the pagination
   * @default select
   */
  type?: PaginationType;
```

Would prefer naming it "type" here :) It think thats more aligned with other components + Figma

packages/sit-onyx/src/i18n/locales/de-DE.json
```diff
@@ -169,6 +169,8 @@
     "ofPages": "von 1 Seite | von {n} Seiten",
     "previous": "vorherige Seite",
     "next": "nächste Seite",
+    "buttonLabel": "Seite {page}",
+    "ellipsis": "Weitere Seiten",
```
```suggestion
    "morePages": "Weitere Seiten",
```

Would prefer to name the key after what it describes, not how its displayed visually in the component :)

packages/sit-onyx/src/components/OnyxPagination/OnyxPagination.vue
```diff
@@ -20,9 +22,14 @@ const selectPaginationProps = useForwardProps(props, OnyxSelectPagination);
 </script>
 
 <template>
+  <OnyxInlinePagination
+    v-if="props.mode === 'inline'"
+    v-bind="selectPaginationProps"
```
larsrickert:
```suggestion
    v-bind="paginationProps"
```

We no longer use it only for the select now ;)

packages/sit-onyx/src/components/OnyxPagination/OnyxPagination.ct.tsx
larsrickert, line 1: (optional) It might be cleaner to define separate screenshot tests for select and inline so we don't need to use several conditions to use either one setup or the other :)
Also it might be easier this way if we add the "compact" pagination in the future

packages/sit-onyx/playwright/snapshots/components/OnyxPagination/Pagination-select--edge-linux.png
larsrickert, line 1: Can you also delete the old screenshots (e.g. packages/sit-onyx/playwright/snapshots/components/OnyxPagination/Pagination-edge-linux.png) that are no longer used due to the test renaming?

.changeset/mighty-zoos-sit.md
```diff
@@ -0,0 +1,5 @@
+---
+"sit-onyx": minor
+---
+
+feat(OnyxPagination): implement inline pagination
```
larsrickert: In general, we should always write the changelog in the POV of the user what is new / changed for him :) So in this case, I'd prefer to say that we added a new property XYZ that allows to optionally display the pagination in an inline style

packages/sit-onyx/src/components/OnyxPagination/OnyxInlinePagination.vue
```diff
@@ -0,0 +1,207 @@
...
+  pagesToShow.add(currentPage + 1);
+
+  if (currentPage === 1) pagesToShow.add(3);
+  if (currentPage === totalPages) pagesToShow.add(totalPages - 2);
+
+  const uniqueSortedPages = Array.from(pagesToShow)
+    .filter((page) => page >= 1 && page <= totalPages)
+    .sort((a, b) => a - b);
+
+  const finalPages: Array<number | string> = [];
```
larsrickert:
```suggestion
  const finalPages: (number | "ellipsis")[] = [];
```

We can type this more strictly so it does only allow "ellipsis" and not all strings :) Also you can remove the generic type from the computed here so we don't have to type it two times. TypeScript will then just infer the type from the return value correctly

packages/sit-onyx/src/components/OnyxPagination/OnyxInlinePagination.vue
```diff
@@ -0,0 +1,207 @@
...
+
+<template>
+  <OnyxSkeleton v-if="skeleton" :class="['onyx-pagination-skeleton', 'onyx-text', densityClass]" />
+
+  <div
+    v-else
+    :class="[
+      'onyx-component',
+      'onyx-pagination',
+      'onyx-pagination__inline',
```
larsrickert:
```suggestion
      'onyx-pagination--inline',
```

In BEM, __ refers to another child element but the class is used as modifier of the same element here so we should use -- :)

packages/sit-onyx/src/components/OnyxPagination/OnyxInlinePagination.vue
```diff
@@ -0,0 +1,207 @@
...
+      'onyx-text',
+      densityClass,
+    ]"
+    role="group"
+    :aria-label="t('pagination.label')"
+    :style="{ '--onyx-pagination-character-count': props.modelValue.toString().length }"
+  >
+    <button
+      class="onyx-pagination__inline__button"
```
larsrickert: In BEM naming, its only allowed to have one layer of nesting (so one __). We could rename all classes to onyx-pagination-inline__button etc. or onyx-inline-pagination__button

packages/sit-onyx/src/components/OnyxPagination/OnyxInlinePagination.vue
```diff
@@ -0,0 +1,207 @@
...
+      :aria-label="t('pagination.previous')"
+      type="button"
+      :disabled="props.disabled || hasReachedMin"
+      @click="emit('update:modelValue', props.modelValue - 1)"
+    >
+      <OnyxIcon :icon="iconChevronLeftSmall" />
+    </button>
+    <template v-for="pageNumber in displayPagesNumbers" :key="`page-${pageNumber}`">
```
larsrickert:
```suggestion
    <template v-for="pageNumber in displayPagesNumbers" :key="pageNumber">
```

Is there a reason for the custom key?

packages/sit-onyx/src/components/OnyxPagination/OnyxInlinePagination.vue
```diff
@@ -0,0 +1,207 @@
...
+        @click="emit('update:modelValue', pageNumber)"
+      >
+        {{ pageNumber }}
+      </button>
+      <div
+        v-else
+        :class="['onyx-pagination__inline__ellipsis']"
+        :aria-label="t('pagination.ellipsis')"
+      >
+        <span aria-hidden="true">&hellip;</span>
```
larsrickert:
```suggestion
        <span aria-hidden="true">...</span>
```

Why don't we use regular dots?

packages/sit-onyx/src/components/OnyxPagination/OnyxInlinePagination.vue
```diff
@@ -0,0 +1,207 @@
...
+        <OnyxVisuallyHidden>
```
larsrickert, line 115: Nice 👍

packages/sit-onyx/src/components/OnyxPagination/OnyxInlinePagination.vue
```diff
@@ -0,0 +1,207 @@
...
+      cursor: pointer;
+      color: var(--onyx-color-text-icons-neutral-medium);
+
+      &:first-of-type {
+        border-radius: var(--onyx-radius-sm) 0 0 var(--onyx-radius-sm);
+        border-left: var(--onyx-pagination-border-size) solid
+          var(--onyx-color-component-border-neutral);
+      }
+      &:last-of-type {
+        border-radius: 0 var(--onyx-radius-sm) var(--onyx-radius-sm) 0;
```
larsrickert:
```suggestion
        border-radius: var(--onyx-pagination-border-radius) 0 0 var(--onyx-pagination-border-radius);
        border-left: var(--onyx-pagination-border-size) solid
          var(--onyx-color-component-border-neutral);
      }
      &:last-of-type {
        border-radius: 0 var(--onyx-pagination-border-radius) var(--onyx-pagination-border-radius) 0;
```

Since we use the border radius more than 1 time, I'd prefer to add a new CSS variable for this so its easier to adjust the value in the future

packages/sit-onyx/src/components/OnyxPagination/types.ts
```diff
@@ -18,4 +18,11 @@ export type OnyxPaginationProps = DensityProp & {
...
 };
+
+export type OnyxPaginationModes = "select" | "inline";
```
larsrickert: Imo we should talk with UX about aligning the property values with Figma. I'd prefer using select and inline instead of modern and classic so we name the types after what they are, not how they might be used. Also thats less opinionated about the usage

Image

packages/sit-onyx/src/components/OnyxPagination/OnyxInlinePagination.vue
```diff
@@ -0,0 +1,232 @@
...
+        border-left: none;
+        background-color: inherit;
+        color: inherit;
+        aspect-ratio: 1;
+      }
+
+      & > .onyx-pagination__more-pages {
+        cursor: default;
+      }
+      &:has(.onyx-pagination__navigate-button:disabled) {
```
larsrickert: This way, the whole pagination is being displayed with disabled styles, although only one button is disabled. If e.g. the current page is 1, the previous arrow will be disabled. With this CSS, all other pages/buttons will also be displayed disabled although they are clickable

packages/sit-onyx/src/components/OnyxPagination/OnyxInlinePagination.vue
```diff
@@ -0,0 +1,232 @@
...
+      <OnyxIcon :icon="iconChevronRightSmall" />
+    </button>
+  </div>
+</template>
+
+<style lang="scss">
+@use "../../styles/mixins/layers.scss";
+.onyx-pagination-skeleton.onyx-pagination-skeleton--inline {
+  @include layers.component() {
+    width: calc(10.5rem + 14 * var(--onyx-pagination-padding-vertical));
```
larsrickert:
```suggestion
    // 7 buttons, each is 1.5rem wide + left/right padding
    width: calc(7 * (1.5rem + 2 * var(--onyx-pagination-padding-vertical)));
```

Imo if we define it like this, its clearer where the value comes from

packages/sit-onyx/playwright/snapshots/components/OnyxPagination/Pagination-inline--edge-linux.png
larsrickert, line 1: Here the skeleton width is not the same as the regular component width. But when looking at it in Storybook it works correctly. Are the screenshots up to date?

packages/sit-onyx/src/components/OnyxPagination/OnyxInlinePagination.vue
```diff
@@ -0,0 +1,235 @@
...
+    :class="[
+      'onyx-component',
+      'onyx-pagination',
+      'onyx-pagination--inline',
+      'onyx-text',
+      densityClass,
+    ]"
+    role="group"
+    :aria-label="t('pagination.label')"
+    :style="{ '--onyx-pagination-character-count': props.modelValue.toString().length }"
```
larsrickert: Why is this needed? It does not seem to be used in this file

packages/sit-onyx/src/components/OnyxPagination/OnyxSelectPagination.vue
```diff
@@ -138,14 +138,15 @@ const hasReachedMax = computed(() => props.modelValue >= props.pages);
 .onyx-pagination-skeleton {
   @include layers.component() {
     height: var(--onyx-pagination-height);
-    width: 16rem;
+    width: calc(11.5rem + 4 * var(--onyx-density-sm) + 4 * var(--onyx-pagination-padding-vertical));
```
larsrickert: Does `var(--onyx-density-sm)` has a special meaning here? E.g. is it used as gap? If yes, it would be nice if we create a new CSS variable for it

PR #4313, opened 2025-10-21
chore: implement and enable no-v-bind-props rule

JoCa96:
Relates to #3966

Implemented eslint rule to check for unguarded forwarding of props.
The rule is now implemented without any typed linting required.


ChristianBusshoff (approved):
Looks good.

Issue #4259, opened 2025-10-14
Implement range select improvements and tests for OnyxCalendar


larsrickert:
### Why?

We want to implement some feedback / improvements for the calendar.

### Acceptance criteria

- [x] range selection: if start and end date are the same, display it like single selection
- [x] range selection: if today is in between the range (range-middle), still highlight it visually (with regular today styles)
- Functional Playwright tests are written to:
    - [ ] keyboard navigation
    - [ ] select logic
    - [ ] week start day
    - [ ] min/max date
- Visual Playwright screenshot tests are written for:
    - [ ] single select
    - [ ] multi select
    - [ ] range select
    - [ ] range select (hover)
    - [ ] range select (keyboard navigation)
    - [ ] custom content
    - [ ] custom content + selection

### DEV handover checklist

[Storybook](ADD_LINK_HERE)

DEV handover done with: @ADD_DEV_NAME_HERE

**Testing**
- [ ] covered by visual screenshot tests
- [ ] covered by functional tests (Playwright or unit test)

**Documentation**
- [ ] "New" badge in Storybook is added, see [storybook-addon-tag-badges](https://github.com/Sidnioulz/storybook-addon-tag-badges/?tab=readme-ov-file#-usage)
- [ ] auto-generated Storybook code snippets are checked
- [ ] component status has been updated in the [docs overview](https://onyx.schwarz/#components)
- [ ] updated version is released
- [ ] updated documentation is deployed

**Other**
- [ ] follow-up tickets were created if necessary (add links below)

### UX handover checklist

UX handover done with: @ADD_UX_NAME_HERE

**Naming is aligned with Figma**
- [ ] component name
- [ ] which support components are used and how they are named
- [ ] all property names and types

**Implementation is checked**
- [ ] all features of Figma are implemented or a follow-up ticket is refined
- [ ] skeleton
- [ ] densities
- [ ] light and dark mode
- [ ] truncation
- [ ] responsiveness / wrapping behavior
- [ ] min / max width

**Figma**
- [ ] Figma component is published

**Documentation**
- [ ] Storybook category is aligned with Figma
- [ ] Storybook examples are checked

**Other**
- [ ] follow-up tickets were created if necessary (add links below)

ChristianBusshoff:
<img width="258" height="255" alt="Image" src="https://github.com/user-attachments/assets/71af2658-719b-42d1-9731-ab0202971d1e" />

[...]

PR #4339, opened 2025-10-28
feat(OnyxCalendar): Implement range select improvements and tests

ChristianBusshoff:
Relates to #4259

- added tests
- added style changes

### DIFF
--- a/packages/sit-onyx/src/i18n/locales/de-DE.json
+++ b/packages/sit-onyx/src/i18n/locales/de-DE.json
@@ -227,7 +227,7 @@
   "calendar": {
     "todayButton": { "label": "Heute", "tooltip": "Springe zu Heute" },
     "calendarWeek": "KW",
-    "calendarWeekButtonLabel": "Select calendar week {weekNumber}",
+    "calendarWeekButtonLabel": "Kalenderwoche {weekNumber} auswählen",
     "previousMonthButton": "Vorheriger Monat",
     "nextMonthButton": "Nächster Monat"
   },


packages/headless/src/composables/calendar/createCalendar.testing.ts
```diff
@@ -0,0 +1,202 @@
...
+    "role",
+    "grid",
+  );
+  await expect(calendarGrid).toBeVisible();
+
+  await page.keyboard.press("Tab");
+  await expect(prevMonthButton, "Prev Month Button should be focused first").toBeFocused();
+
+  await page.keyboard.press("Tab");
+  await expect(nextMonthButton, "Next Month Button should be focused second").toBeFocused();
```
larsrickert: The previous/next buttons are not part of the headless implementation since you could also build a calendar without those buttons

packages/headless/src/composables/calendar/createCalendar.testing.ts
```diff
@@ -0,0 +1,202 @@
...
+  const day2Cell = day2.locator("..");
+  await day1.focus();
+  await day1.press("Enter");
+  await day1.press("Enter");
+
+  await expect(day1.locator(".."), "Day 1 must be selected").toHaveAttribute(
+    "aria-selected",
+    "true",
+  );
+  await expect(day1, "Day 1 must have is-selected class").toHaveClass(/is-selected/);
```
larsrickert: This class is not added by the headless calendar, right? If its added by the OnyxCalendar, please remove the check here. Same for the other is-selected class tests below

packages/headless/src/composables/calendar/createCalendar.testing.ts
```diff
@@ -0,0 +1,202 @@
...
+    "true",
+  );
+};
+
+const testRangeSelection = async ({
+  dayButtons,
+  initialFocusDay,
+}: Pick<CalendarTestingOptions, "dayButtons" | "initialFocusDay">) => {
+  const dayStart = initialFocusDay; // Tag 15
+  const dayEnd = dayButtons.filter({ hasText: "24" }).first(); // Tag 24
```
larsrickert:
```suggestion
  const dayStart = initialFocusDay; // day 15
  const dayEnd = dayButtons.filter({ hasText: "24" }).first(); // day 24
```

packages/sit-onyx/src/components/OnyxCalendar/OnyxCalendar.ct.tsx
```diff
@@ -117,3 +117,95 @@ test.describe("Screenshot tests (custom content)", () => {
...
+  await expect(component).toHaveScreenshot("range.png");
+  // Today inside range
+  // ACT
+  await startRange.click();
+  await endRange2.click();
+  await page.mouse.move(0, 0);
+  // ASSERT
+  await expect(startCell).toHaveClass(/onyx-calendar-cell--range-start/);
+  await expect(endCell2).toHaveClass(/onyx-calendar-cell--range-end/);
+  await expect(todayCell).toHaveClass(/onyx-calendar-cell--range-middle/);
```
larsrickert:
```suggestion
  await expect(startCell).toContainClass("onyx-calendar-cell--range-start");
  await expect(endCell2).toContainClass("onyx-calendar-cell--range-end");
  await expect(todayCell).toContainClass("onyx-calendar-cell--range-middle");
```

We can use `.toContainClass()` so we don't need a regex

packages/sit-onyx/src/components/OnyxCalendar/OnyxCalendar.vue
```diff
@@ -142,10 +143,41 @@ const getDayRangeType = computed(() => {
...
+          ? undefined
+          : hoveredDate.value,
     });
   };
 });
+
+const selectWeek = (week: RenderWeek) => {
+  const newRange: DateRange = {
+    start: week.days[0]!.date,
+    end: week.days[6]!.date,
```
larsrickert:
```suggestion
    end: week.days.at(-1)!.date,
```

`.at(-1)` gets the last element :)

packages/sit-onyx/src/components/OnyxCalendar/OnyxCalendar.vue
```diff
@@ -142,10 +143,41 @@ const getDayRangeType = computed(() => {
...
+    } else {
+      return {
+        role: "button",
+        "aria-disabled": false,
+        onClick: () => selectWeek(week),
+      };
+    }
+  }
+  return {};
+};
```
larsrickert:
```suggestion
const getWeekNumberProps = (week: RenderWeek) => {
  if (props.selectionMode !== "range") return {};
  const hasDisabledDate = week.days.some(({ date }) => isDisabled.value(date));

  if (hasDisabledDate) {
    return {
      role: "button",
      tabindex: "-1",
      "aria-disabled": true,
    };
  } else {
    return {
      role: "button",
      "aria-disabled": false,
      onClick: () => selectWeek(week),
    };
  }
};
```

Question: Shoudln't the week selection be also disabled if any date inside the week is disabled (e.g. Wednesday)?

packages/sit-onyx/src/components/OnyxCalendar/OnyxCalendar.vue
```diff
@@ -161,25 +193,32 @@ const getDayRangeType = computed(() => {
   >
     <div class="onyx-calendar__header">
       <div class="control-container time-control-container">
-        <OnyxTag
-          v-if="calendarSize !== 'small'"
+        <OnyxSystemButton
           :label="t('calendar.todayButton.label')"
           class="control-container__today-btn"
           :disabled="disabled === true"
           :clickable="t('calendar.todayButton.tooltip')"
```
larsrickert: clickable prop does not exist on OnyxSystemButton

packages/sit-onyx/src/components/OnyxCalendar/OnyxCalendar.vue
```diff
@@ -161,25 +193,32 @@ const getDayRangeType = computed(() => {
...
           @click="goToToday"
         />
+
+        <OnyxHeadline is="h2" class="control-container__date-display">
+          {{ d(viewMonth, { month: "long", year: "numeric" }) }}
+        </OnyxHeadline>
+        <OnyxTag
+          v-if="showCalendarWeeks && calendarSize === 'big'"
+          color="primary"
+          :label="`${t('calendar.calenderWeek')} ${weeksToRender[0]?.weekNumber} - ${weeksToRender[weeksToRender.length - 1]?.weekNumber} `"
```
larsrickert:
```suggestion
          :label="`${t('calendar.calenderWeek')} ${weeksToRender[0]?.weekNumber} - ${weeksToRender.at(-1)?.weekNumber} `"
```

Maybe we should also extract this to a dedicated computed?

packages/sit-onyx/src/components/OnyxCalendar/OnyxCalendar.vue
```diff
@@ -205,7 +244,7 @@ const getDayRangeType = computed(() => {
 
         <tbody>
           <tr v-for="week in weeksToRender" :key="week.weekNumber">
-            <th v-if="showCalendarWeeks" scope="row">
+            <th v-if="showCalendarWeeks" scope="row" v-bind="getWeekNumberProps(week)">
```
larsrickert, line 246: We should use a `<button>` if its clickable inside the td so its also correctly clickable via keyboard

packages/sit-onyx/src/components/OnyxCalendarCell/types.ts
```diff
@@ -28,6 +28,10 @@ export type OnyxCalendarCellProps = DensityProp & {
    */
   rangeType?: CalendarCellRangeType;
   buttonAttributes?: HTMLAttributes;
+  /**
+   * Optional tooltip text
+   */
+  toolTipText?: string;
```
larsrickert:
```suggestion
  tooltipText?: string;
```

packages/sit-onyx/src/components/OnyxCalendarCell/OnyxCalendarCell.vue
```diff
@@ -52,7 +53,14 @@ const contentAttributes = computed(() => {
     >
       <div class="onyx-calendar-cell__header">
         <div class="onyx-calendar-cell__date-container">
-          <span class="onyx-calendar-cell__date">
+          <OnyxTooltip v-if="props.toolTipText" :text="props.toolTipText" without-wedge>
```
larsrickert:
```suggestion
          <OnyxTooltip v-if="props.toolTipText" :text="props.toolTipText">
```

Imo the tooltip looks a little lost without the wedge, is there a reason why we differ from the default here?

Image

packages/sit-onyx/src/components/OnyxCalendar/OnyxCalendar.vue
ChristianBusshoff: by default it's possible to have a disabled day inside a range. So I would handle the week button the same. Or shall I change it?

packages/sit-onyx/src/components/OnyxCalendar/OnyxCalendar.vue
ChristianBusshoff: couldn't set a button filling the hight of the th, so I removed the button. Also Nadine said it should not have keyboard accessability(because it's an addon and not necessary to navigate).

packages/sit-onyx/src/components/OnyxCalendarCell/OnyxCalendarCell.vue
ChristianBusshoff: Its without an wedge in the design.

packages/sit-onyx/src/components/OnyxCalendar/OnyxCalendar.vue
larsrickert: No lets keep it the same then. But why do we check the start and end date for disabled then? Should we change to to disable to week button when all days in the week are disabled?

packages/sit-onyx/src/components/OnyxCalendar/OnyxCalendar.vue
larsrickert: It should be possible to fill the full height with the button (I mean we do the same for the td cells). Also a button would be semantically correct here (also for things like screen readers

packages/headless/src/composables/calendar/createCalendar.testing.ts
```diff
@@ -0,0 +1,205 @@
...
+  await expect(
+    calendarGrid,
+    "Clicking next month button should change aria-label",
+  ).not.toHaveAttribute("aria-label", initialMonth!);
+
+  await prevMonthButton.click();
+  await expect(
+    calendarGrid,
+    "Clicking previous month button should return to initial month label",
+  ).toHaveAttribute("aria-label", initialMonth!);
```
larsrickert: The previous/next month buttons are also not part of the headless calendar itself, right? So you could also build a calendar without them

packages/headless/src/composables/calendar/TestCalendar.ct.tsx
larsrickert, line 1: You can get rid of the eslint warnings by adding the functions inside eslint.config.js (line 127)

packages/sit-onyx/src/components/OnyxCalendar/OnyxCalendar.vue
```diff
@@ -142,10 +143,45 @@ const getDayRangeType = computed(() => {
...
 });
+
+const selectWeek = (week: RenderWeek) => {
+  const newRange: DateRange = {
+    start: week.days[0]!.date,
+    end: week.days.at(-1)!.date,
+  };
+  modelValue.value = newRange as unknown as typeof modelValue.value;
+};
+const getWeekNumberProps = (week: RenderWeek) => {
```
larsrickert: We should also add an aria label, maybe something like "Select calendar week XX"

packages/sit-onyx/src/i18n/locales/de-DE.json
```diff
@@ -226,7 +226,8 @@
   },
   "calendar": {
     "todayButton": { "label": "Heute", "tooltip": "Springe zu Heute" },
-    "calenderWeek": "KW",
+    "calendarWeek": "KW",
+    "calendarWeekButtonLabel": "Select calendar week {weekNumber}",
```
```suggestion
    "calendarWeekButtonLabel": "Kalenderwoche {weekNumber} auswählen",
```

Issue #4289, opened 2025-10-17
[Bug]: OnyxBadge not displayed correctly in flex layout


larsrickert:
### Version

1.1.0

### Description

Based on #4285

The OnyxBadge does not display correctly when using inside a flex layout.

### Reproduction steps

[First example](https://playground.onyx.schwarz/?onyxVersion=1.1.0#eNqNVE1v00AQ/Suj5QJS7DRpezGmEkU9wIEi6HEltLUnZst619odm0RR/juz66RxUBRxyny8efORt96Kj12XDz2KQpSh8rojMMo2H6SgIAUEpL67k1a3nfMEW3i0m/WTc4Z0N0vOvaob3Js9kbOwg5V3LUgRNGWO41K8l7acj/RMxg5h2xlFyB5AWesBKqNC4LaxIGu8rjOjNq4nKRLmH9TKINPuM5ybjAWEa2LI4jZfXOXLq+XtDBY3xXIxwXPFYQJ4U+NK9SaWbIG8bhr0sDsB7xukVWHInrWt44FGLF+pcsZ5jljsySvDtUf4opy/OtP+85MTjKHJFscor32XlfP4c7LuSH+p9fJc67N80flP0usp6Rmas6OOwjDqGQ3Tjq4U8wuVR/tgnVysDLQx0cijEmAbgbUOnN4UEEOsOIBGdQUMyr/Nkg6zGi1rcpOF9h3nd0mUiUfMWO6Vsyvd5C/BWX4NiVKKyrWdNugfO9LO8pMoxmYxp4xxf76kGPkeZ4d49Qur32fiL4E1W7DxzWNAP6AUrzlSvkEWe0w//PjKEp4kW1f3htEXkt8xONPHGUfYfW9rHnuCS9N+Tu9Y2+YpPKwpnmO/VBw0IncJLwV/Ej5dWP047nV+k+r4nnzFnwP6yMkHjP9UILH7C4jga4o=)
[...]

PR #4368, opened 2025-10-30
fix(OnyxBadge): not displayed correctly in flex layout

ChristianBusshoff:
Relates to #4289

- alined OnyxBadge correctly


.changeset/nine-insects-matter.md
```diff
@@ -0,0 +1,5 @@
+---
+"sit-onyx": minor
+---
+
+fix(OnyxBadge): alined the badge correctly in flex layout.
```
larsrickert:
```suggestion
fix(OnyxBadge): align the badge correctly in a flex layout
```

larsrickert (changes requested):
We should also implement a screenshot test to ensure this no longer occurrs in the future :) Maybe you can use the two examples provided in #4289 as base

larsrickert (approved):

PR #4371, opened 2025-11-02
feat(OnyxSlider): enhance slider controls with input support and improve discrete mode handling

lovelycentury:
Relates to #2875

- New OnyxSliderControl component with icon, value, and input control types
- Input support for direct value entry in sliders (single & range modes)
- Comprehensive Playwright tests for both components
- Improved discrete mode handling in headless logic. Now to activate the discrete mode, you don't need to manually pass   steps: true, only discrete: true.

## Checklist

- [x] A changeset is added with npx changeset add if your changes should be released as npm package (because they affect the library usage)


packages/sit-onyx/src/components/OnyxSlider/OnyxSlider.ct.tsx
```diff
@@ -427,4 +428,110 @@ test.describe("Interaction tests", () => {
...
+      props: {
+        label: "Input control slider",
+        modelValue,
+        control: "input",
+      },
+      on: eventHandlers,
+    });
+
+    const slider = component.getByRole("slider");
+    const input = component.getByRole("spinbutton");
```
larsrickert:
```suggestion
    const input = component.getByLabel("Change value");
```

Would prefer getting the input by its label :) Same for the range test below

packages/sit-onyx/src/components/OnyxSlider/OnyxSlider.ct.tsx
```diff
@@ -427,4 +428,110 @@ test.describe("Interaction tests", () => {
...
+    // ACT - Change value via input
+    await input.fill("75");
+    await input.blur();
+
+    // ASSERT
+    expect(modelValue).toBe(75);
+    await expect(slider).toHaveValue("75");
+
+    // ACT - Change value via input to boundary
+    await input.fill("100");
```
larsrickert: Would be great if we can also test entering a value > max / < min (just like you did for the range mode)

packages/sit-onyx/src/components/OnyxSlider/OnyxSlider.vue
```diff
@@ -142,20 +151,24 @@ const isIconControl = computed(() => props.control === "icon" && props.mode ===
...
+            v-if="props.control === 'icon' && props.mode === 'single'"
+            control="icon"
+            direction="decrease"
+            :shift-step="shiftStep"
+            :model-value="normalizedValues[0]"
+            :disabled="disabled || (normalizedValues[0] ?? props.min) <= props.min"
+            @update:model-value="handleSliderIconControlChange"
+          />
+          <OnyxSliderControl
+            v-if="props.control === 'input' && props.mode === 'range' && !props.discrete"
```
larsrickert: For discrete mode it would still be great to input a value. If the value does not match the discrete steps, we can adjust it automatically :)

packages/sit-onyx/src/components/OnyxSliderControl/OnyxSliderControl.ct.tsx
larsrickert, line 1: Instead of locating the inputs in this file by role spinbutton, I'd prefer to use .getByLabel("") and pass the corresponding label :)

packages/sit-onyx/src/components/OnyxSliderControl/OnyxSliderControl.stories.ts
```diff
@@ -0,0 +1,51 @@
...
+} satisfies StoryObj<typeof OnyxSliderControl>;
+
+export const IconControlIconIncrease = {
+  args: {
+    control: "icon",
+    direction: "increase",
+  },
+} satisfies StoryObj<typeof OnyxSliderControl>;
+
+export const IconControlInput = {
```
larsrickert:
```suggestion
export const Value = {
  args: {
    control: "value",
    value: 0,
  },
} satisfies StoryObj<typeof OnyxSliderControl>;

export const IconDecrease = {
  args: {
    control: "icon",
    direction: "decrease",
  },
} satisfies StoryObj<typeof OnyxSliderControl>;

export const IconIncrease = {
  args: {
    control: "icon",
    direction: "increase",
  },
} satisfies StoryObj<typeof OnyxSliderControl>;

export const Input = {
```

We can simplify the Story names here :)

packages/sit-onyx/src/components/OnyxSliderControl/types.ts
```diff
@@ -0,0 +1,60 @@
+import type { DensityProp } from "../../composables/density.js";
+
+export const SLIDER_CONTROLS = ["icon", "value", "input"] as const;
+export type SliderControl = (typeof SLIDER_CONTROLS)[number];
```
larsrickert, line 4: Don't we already have those in the OnyxSlider so we can import them here?

packages/sit-onyx/src/i18n/locales/en-US.json
```diff
@@ -241,8 +241,9 @@
       "hover": "Hover/Focus to toggle action visibility"
     }
   },
-  "slider": {
-    "decreaseValue": "Decrease value by {n}",
-    "increaseValue": "Increase value by {n}"
+  "sliderControl": {
```
larsrickert:
```suggestion
  "slider": {
```

Would prefer to still name the key after the parent component :)

packages/sit-onyx/src/i18n/locales/de-DE.json
```diff
@@ -238,8 +238,9 @@
       "hover": "Mit der Maus darüberfahren/fokussieren, um die Sichtbarkeit der Aktionen umzuschalten"
     }
   },
-  "slider": {
-    "decreaseValue": "Wert um {n} verringern",
-    "increaseValue": "Wert um {n} erhöhen"
+  "sliderControl": {
+    "decreaseValueBy": "Decrease value by {n}",
+    "increaseValueBy": "Increase value by {n}",
+    "changeValue": "Change value"
```
```suggestion
    "decreaseValueBy": "Wert um {n} verringern",
    "increaseValueBy": "Wert um {n} erhöhen",
    "changeValue": "Wert ändern"
```

Here are the German translations :)

packages/sit-onyx/src/components/OnyxSliderControl/OnyxSliderControl.stories.ts
```diff
@@ -0,0 +1,51 @@
+import { withNativeEventLogging } from "@sit-onyx/storybook-utils";
+import type { Meta, StoryObj } from "@storybook/vue3-vite";
+import OnyxSliderControl from "./OnyxSliderControl.vue";
+
+/**
+ * The slider control component intended to be used alongside the OnyxSlider component to provide additional controls for adjusting the slider value.
+ */
+const meta: Meta<typeof OnyxSliderControl> = {
+  title: "Support/OnyxSliderControl",
```
larsrickert:
```suggestion
  title: "Support/SliderControl",
```

For the Storybook title, we don't prefix the component with "Onyx" so they are easier to read in the sidebar :)

packages/sit-onyx/src/components/OnyxSliderControl/types.ts
```diff
@@ -0,0 +1,60 @@
...
+   */
+  control: SliderControl;
+  /**
+   * Slider control model value
+   */
+  modelValue?: number;
+  /**
+   * The read-only value to display inside the control. In the slider context, this is min and max values.
+   */
+  value?: number;
```
larsrickert: For the control="value" we can also pass the modelValue property so we can get rid of this additional prop :)

packages/sit-onyx/src/components/OnyxSliderControl/types.ts
lovelycentury: I moved `SLIDER_CONTROLS` constant from OnyxSlider to OnyxSliderControl, because it's semantically closer

packages/sit-onyx/src/components/OnyxSlider/OnyxSlider.vue
```diff
@@ -72,9 +67,16 @@ const skeleton = useSkeletonContext(props);
 
 const { min, max, step, marks, label, discrete } = toRefs(props);
 
+/**
+ * During focus value inside of stepper component was not updated.
+ * So, we need to manually trigger an update when the control version changes.
+ */
+const inputVersion = ref<number>(0);
```
larsrickert: Can you extend on why this is needed? I tried removed this and it still worked fine.
We shouldn't need to force re-render components (thats usually a sign of a potential bug)

lovelycentury: Yes sure.

Try to change the value when min = 20 and max = 80 to 100 and then focus `OnyxStepper` once more.
The value is preserved during the focus.

https://github.com/user-attachments/assets/83bf402e-72cb-490d-a7c7-d2c27546829d

lovelycentury: @larsrickert 

Okay, I understand the root cause. The issue is that triggering "update:modelValue" from `OnyxStepper` causes desynchronization when the `modelValue` hasn’t changed in the parent component.

`OnyxStepper` keeps an outdated `inputValue` (local ref) while the correct value remains in the parent, which leads to an incorrect value being shown on focus.

I would fix this by ensuring that the value is synchronized on both `blur` and `focus` or only `blur`, similar to how it’s done in `watchEffect` already in `OnyxStepper`.

What do you think about it?

```
watchEffect(() => {
  inputValue.value = getFormattedValue.value(modelValue.value);
});
```

larsrickert: I see, thanks for the clear explanation! This seems to be a bug in the OnyxStepper then since the `modelValue` is not in sync with the `inputValue`. I like the idea of ensuring the sync on `@blur` like done in the watcher already :)

Do you have time to implement this or should I create a separate PR?

lovelycentury: I will implement in 10 min

.changeset/cute-cats-jog.md
```diff
@@ -0,0 +1,11 @@
+---
+"@sit-onyx/headless": minor
+"sit-onyx": minor
+---
+
+feat(OnyxSlider): add `OnyxSliderControl` component and improve discrete mode
+
+- Now when passing `discrete: true` to `OnyxSlider`, marks will be generated automatically and there is no need to manually pass `marks: true`
+- Add unstable `OnyxSliderControl` component for slider input controls
```
larsrickert, line 9:
```suggestion
- Add unstable `OnyxSliderControl` component for slider input controls
- fix(OnyxStepper): ensure formatted display value is in sync with modelValue on blur
```

larsrickert: Thanks for applying the fix to the stepper!

PR #4385, opened 2025-11-05
refactor(headless): improve useOutsideClick Tab navigation with event-driven approach

### LINKED ISSUE #3633: Implement refactoring of useOutsideClick composable for Tab keyboard press
### Why?

Follow up for #3616.

In #3636, we implemented en enhancement for the useOutsideClick utility that will listen on "Tab" keyboard presses and emit an outside event when the newly focussed element is outside.

This is used e.g. for the OnyxFlyoutMenu so that it is hidden when tabbing out of the menu / last item.

Currently, it is using a `setTimeout()` to ensure that side effects have run earlier which is not the best practice.
In this ticket, we want to refactor this implementation to not reply on setTimeout.

We already tried to use `onFocusout` but it had negative side effects that caused the flyout to close when clicking on an nested item.

### Acceptance criteria

- [ ] setTimeout is no longer used
- [ ] all test run successfully without modification

### Implementation details

<-- Add technical details / implementation proposals if necessary, e.g.:
  - API proposal
  - relevant web APIs
  - learnings from other implementations
  - code snippets
  -->

### Definition of Done

- The following component requirements are implemented:
  - [ ] skeleton
  - [ ] density
  - [ ] example usage added to `apps/demo-app/src/views/HomeView.vue`

- Should be covered by tests:
  - [ ] functional tests (Playwright or unit test)
  - [ ] visual tests (Playwright screenshots)

- Make sure, that
  - [ ] follow-up tickets were created if necessary
  - [ ] Storybook can show the feature
  - [ ] Storybook code snippet of new/changed examples are checked that they are generated correctly
  - [ ] Namings are aligned with Figma
  - [ ] Storybook sidebar badge for "New" is added on component or story level if needed. See: https://github.com/Sidnioulz/storybook-addon-tag-badges/?tab=readme-ov-file

#### Approval

[Storybook]()
[Documentation]()

- Checked and approved by
  - [ ] designer 
  - [ ] dev

lovelycentury:
closes #3633

- Replace setTimeout with focusin/blur/keyup event listeners
- Use AbortController for proper cleanup
- Add FocusEvent support and detailed documentation

## Checklist

### DIFF
--- a/apps/demo-app/app/components/Banner.vue
+++ b/apps/demo-app/app/components/Banner.vue
@@ -1,0 +1,22 @@
+<script setup lang="ts">
+import { OnyxImage } from "sit-onyx";
+
+const props = defineProps<{
+  src: string;
+}>();
+</script>
+
+<template>
+  <OnyxImage class="banner" alt="Hero banner" :src="props.src" :width="1920" :height="280" />
+</template>
+
+<style lang="scss" scoped>
+.banner {
+  width: 100%;
+  display: block;
+
+  :deep(.onyx-image__source) {
+    display: inherit;
+  }
+}
+</style>
--- a/apps/demo-app/app/pages/index.vue
+++ b/apps/demo-app/app/pages/index.vue
@@ -1,8 +1,9 @@
 <script lang="ts" setup>
 import { iconChart, iconTextSelector, iconToolTable } from "@sit-onyx/icons";
 import { useI18n } from "vue-i18n";
-import banner from "~/assets/images/banner.webp";
+import bannerImg from "~/assets/images/banner.webp";
 import onyxLogo from "~/assets/images/onyx-logo.svg?raw";
+import Banner from "~/components/Banner.vue";
 import type { LinkCardProps } from "../components/LinkCard.vue";
 
 definePageMeta({ layout: false });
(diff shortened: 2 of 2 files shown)


packages/headless/src/composables/helpers/useOutsideClick.ts
```diff
@@ -52,16 +52,59 @@ export const useOutsideClick = <TCheckOnTab extends boolean | undefined>({
...
+      /**
+       * Handles the Tab key release event.
+       * Cleans up temporary listeners if Tab navigation is completed.
+       */
+      const onKeyUp = (e: KeyboardEvent) => {
+        if (e.key === "Tab") controller.abort();
+      };
+
+      document.addEventListener("focusin", onFocusIn, {
+        once: true,
```
larsrickert, line 98: Why do we use once here? :) Doesn't that mean that the checks are not done when the event occurs multiple times?

lovelycentury: The listeners are attached only when the Tab key is pressed, so for each Tab keydown there is exactly one fresh set of listeners.

Using `once: true` here ensures that each of them fires only once and then automatically cleans up (together with the AbortController).

This way we avoid accumulating multiple listeners or triggering `onOutsideClick` more than once per Tab navigation step.

larsrickert: I see, thanks for the clarification!

PR #4393, opened 2025-11-06
feat(DemoApp): fix Banner type

ChristianBusshoff:
Relates to #4373

### DIFF
--- a/apps/demo-app/app/components/Banner.vue
+++ b/apps/demo-app/app/components/Banner.vue
@@ -1,9 +1,7 @@
 <script setup lang="ts">
 import { OnyxImage, type OnyxImageProps } from "sit-onyx";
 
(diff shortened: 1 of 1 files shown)


apps/demo-app/app/components/Banner.vue
```diff
@@ -1,8 +1,8 @@
 <script setup lang="ts">
-import { OnyxImage } from "sit-onyx";
+import { OnyxImage, type OnyxImageProps } from "sit-onyx";
 
 const props = defineProps<{
-  src: string;
+  src: OnyxImageProps["src"];
 }>();
```
larsrickert:
```suggestion
const props = defineProps<Pick<OnyxImageProps, "src">>();
```

Issue #4266, opened 2025-10-14
Implement OnyxCodeTabs component


larsrickert:
### Why?

For our new onyx showcase, we need to be able to display code snippets.

### Design

[Figma](https://www.figma.com/design/Jz5vPod9LNB5pH64dKT0oX/Onyx-Nuxt-Templates?node-id=1-56165&m=dev)

### Acceptance criteria

- [x] supports displaying a code block
- [x] optionally supports multiple tabs / files (label + icon)
- [x] shows a copy button to copy the current code to the users clipboard
- [x] displays the code snippet language (e.g. JS, TS, Vue etc.)
- [x] all markdown code blocks are rendered with the code group component, see [Nuxt content prose components](https://content.nuxt.com/docs/components/prose#prosepre)

NpmInstallCodeTabs
- [x] an additional `NpmInstallCodeTabs` component is created to show "npm install" commands
- [x] uses the OnyxCodeTabs component internally
- [x] always displays tabs for pnpm, npm, yarn and bun package manager
- [x] supports regular and dev dependencies
- [x] uses proper icons / logos for the package managers (check copyright / legal terms for this!!)

### Implementation details

- syntax highlighting (coloring) will be done by the Nuxt content module, we won't implement this in the component itself
- implement it only as part of the Nuxt documentation template for now
- [onyx prose components](https://github.com/SchwarzIT/onyx/tree/main/packages/nuxt-docs/app/components/content)

```vue
<!-- ProsePre.vue -->
<script setup lang="ts">
const props = withDefaults(
  defineProps<{
    /**
     * Raw code snippet.
     */
    code?: string;
    /**
     * Code language
     *
     * @example js, ts, html, css, etc.
     */
    language?: string;
    /**
     * Filename.
     */
    filename?: string;
    /**
     * Highlighted code line numbers.
     */
    highlights?: number[];
    meta?: string;
    class?: string;
  }>(),
  {
    code: "",
  },
);

defineSlots<{
  /**
   * Renderer code snippet.
   */
  default(): unknown;
}>();
</script>

<template>
  <pre :class="[props.class, 'pre']">
    <slot />
    </pre>
</template>

<style lang="scss" scoped>
.pre code .line {
  display: block;
}
</style>
```

### Reference implementations

- [Docs](https://github.com/nuxt-content/mdc/blob/main/src/runtime/components/prose/ProsePre.vue) for Nuxt ProsePre component

### DEV handover checklist

[...]

PR #4403, opened 2025-11-07
feat(OnyxCodeTabs): implement skeleton and adjust copy button padding

larsrickert:
Relates to #4266

- implement skeleton
- fix right padding of copy button
- catch error when copying to clipboard

## Checklist

### DIFF
--- a/packages/sit-onyx/src/components/OnyxCodeTabs/OnyxCodeTabs.ct.tsx
+++ b/packages/sit-onyx/src/components/OnyxCodeTabs/OnyxCodeTabs.ct.tsx
@@ -99,7 +99,13 @@ test("should copy code", async ({ mount, page, context, browserName }) => {
   await expect(component).not.toContainText("Copied!");
 });
 
-test("should show error when copy fails", async ({ mount, page, context }) => {
+test("should show error when copy fails", async ({ mount, page, context, browserName }) => {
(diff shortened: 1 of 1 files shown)


packages/sit-onyx/src/components/OnyxCodeTabs/OnyxCodeTabs.vue
```diff
@@ -55,9 +55,15 @@ const isCopied = ref(false);
...
-  isCopied.value = true;
-  setTimeout(() => (isCopied.value = false), 3000);
+
+  try {
+    await navigator.clipboard.writeText(activeTabCode.value);
+    isCopied.value = true;
+    setTimeout(() => (isCopied.value = false), 3000);
+  } catch {
+    // noop, the user might not have granted the permission for the browser / application
+    // to access the clipboard
```
ChristianBusshoff: Shouldn't there be something inside the catch?

larsrickert: Do you have a suggestion for this? The clipboard can only fail if the user did not grant permissions to the browser for accessing the clipboard.

We can talk with UX if we want to display a "Failed" tag with danger color in this case

larsrickert: Done :) We now show a "Failed!" error tag

ChristianBusshoff (approved):

Issue #2066, opened 2024-11-11
Define OnyxKey component


JoCa96:

larsrickert:
Implemented as community contribution in #4429

PR #4429, opened 2025-11-11
feat(OnyxShortcut): implement shortcut component

lovelycentury:
Relates to #2066

## Description

-  Add keyboard shortcut system with `OnyxKey` and `OnyxShortcut` components
- `useShortcut` - composable for detecting multi-step keyboard short sequences
- `useOperatingSystem` - auto-detect OS for appropriate key styling

## Key Features

- Automatic OS detection with platform-specific key symbols
- Support for complex shortcuts like `Ctrl+K → G`
- Highlighting feedback during key press sequences

## Checklist


lovelycentury:
@larsrickert 

This feature introduces fairly uncommon functionality. I’d really appreciate any feedback on the overall concept, API shape, or naming conventions.

If you have ideas for simplifying the internal logic, improving composability, or aligning this with other utilities, please feel free to share them. 

I’d be happy to iterate on this together.

packages/sit-onyx/src/components/OnyxKey/OnyxKey.stories.ts
```diff
@@ -0,0 +1,96 @@
...
+  args: {
+    skeleton: true,
+    keyName: "enter",
+  },
+} satisfies Story;
+
+/**
+ * Example: simple shortcut.
+ */
+export const Shortcut = {
```
larsrickert: For displaying shortcuts, we intend to use the `OnyxShortcut` component right? I'd prefer to remove this example then. Maybe we can update the description at the top to mention that the OnyxShortcut should be used for this?

packages/sit-onyx/src/components/OnyxKey/OnyxKey.stories.ts
```diff
@@ -0,0 +1,96 @@
...
+            :keyName="key"
+          />
+        </div>
+      `,
+    };
+  },
+  args: {
+    variant: "macOS",
+  },
+} satisfies Story;
```
larsrickert:
```suggestion
/**
 * Displays all supported canonical keys from CANONICAL_KEYS.
 * Useful for visual verification of the mapping.
 */
export const AllCanonicalKeys = createAdvancedStoryExample(
  "OnyxKey",
  "CanonicalKeysExample",
) satisfies Story;
```

The downside of using render functions in Storybook is that the generated code snippet in the docs does not reflect the actual code written here. To solve this, we created a `createAdvancedStorybookExample()` utility function. With it, you can create a dedicated .vue file inside e.g. OnyxKey/examples/CanonicalKeysExample.vue and it will then be rendered as story as well as including the correct source code from the vue file as code snippet

packages/sit-onyx/src/components/OnyxKey/OnyxKey.vue
```diff
@@ -0,0 +1,154 @@
...
+const skeleton = useSkeletonContext(props);
+
+/**
+ * Effective OS based on variant / auto-detect
+ */
+const effectiveOs = computed<OperatingSystem>(() => {
+  if (props.variant !== "auto") {
+    return props.variant;
+  }
+  return detectOperatingSystem();
```
larsrickert: It would be great if we can also support SSR (server side rendering), e.g. with Nuxt. Since `detectOperatingSystem` would be executed on server side, the output might not match the actual users OS so a hydration mismatch would exist.

To solve this, we could e.g. only call the method in onBeforeMount / onMounted or change the plain function to be a composable instead which handles this under the hood

packages/sit-onyx/src/components/OnyxKey/OnyxKey.vue
```diff
@@ -0,0 +1,154 @@
...
+  const os = effectiveOs.value;
+
+  if (typeof key === "string" && isCanonicalSpecialKey(key)) {
+    const osMap = MAP_CANONICAL_TO_SCREEN_READER_LABEL_BY_OS[os] ?? {};
+    const genericMap = MAP_CANONICAL_TO_SCREEN_READER_LABEL_BY_OS.generic ?? {};
+
+    return `${osMap[key] ?? genericMap[key] ?? String(props.keyName)} key`;
+  }
+
+  return `${String(props.keyName)} key`;
```
larsrickert:
```suggestion
  if (props.label) return props.label;
  return `${visualLabel.value} key`;
```

Can't we reuse the logic from the visual label? :)
Also ideally we use `const { t } = injectI18n()` to use as translated label

larsrickert: Since we are using `<kbd>` here, do we even need an dedicated aria label? The browser should already know that its a keyboard shrotcut + its already labeled by its text content (visualLabel)

packages/sit-onyx/src/components/OnyxKey/OnyxKey.vue
```diff
@@ -0,0 +1,154 @@
...
+  <OnyxSkeleton v-if="skeleton" :class="['onyx-component', 'onyx-key-skeleton']" />
+  <kbd
+    v-else
+    :class="['onyx-component', 'onyx-key']"
+    :data-os="effectiveOs"
+    :data-pressed="props.pressed || undefined"
+    :aria-label="ariaLabel"
+  >
+    {{ visualLabel }}
+  </kbd>
```
larsrickert, line 98:
```suggestion
 <kbd
    v-else
    :class="['onyx-component', 'onyx-key', { 'onyx-key--pressed': props.pressed }]"
    :aria-label
  >
    {{ visualLabel }}
  </kbd>
```

Would prefer to use a CSS class instead of data attributes :)

packages/sit-onyx/src/components/OnyxKey/OnyxKey.vue
```diff
@@ -0,0 +1,154 @@
...
+    border: 1px solid var(--onyx-key-border-color);
+    border-radius: var(--onyx-key-border-radius);
+    color: var(--onyx-key-color);
+
+    &[data-pressed="true"] {
+      background-color: var(--onyx-key-background-pressed);
+      border-color: var(--onyx-key-border-color-pressed);
+    }
+  }
+}
```
larsrickert, line 131:
```suggestion
.onyx-key-skeleton,
.onyx-key {
  @include layers.component() {
    --onyx-key-border-radius: var(--onyx-radius-sm);
    --onyx-key-size: 1.5rem;
  }
}

.onyx-key {
  @include layers.component() {
    display: inline-flex;
    align-items: center;
    justify-content: center;
    padding-inline: var(--onyx-density-xs);

    height: var(--onyx-key-size);
    min-width: var(--onyx-key-size);

    font-family: var(--onyx-font-family);
    font-size: var(--onyx-font-size-md);
    line-height: var(--onyx-font-line-height-md);

    background-color: var(--onyx-color-base-background-tinted);
    border: var(--onyx-1px-in-rem) solid var(--onyx-color-base-neutral-300);
    border-radius: var(--onyx-key-border-radius);
    color: var(--onyx-color-text-icons-neutral-medium);

    &[data-pressed="true"] {
      background-color: var(--onyx-color-base-neutral-200);
      border-color: var(--onyx-color-base-neutral-400);
    }
  }
}
```

Some style adjustments to match whats defined in Figma :) (Basically the font family and font size changed). Also I'd prefer to remove CSS variables that are only used once to keep it simple, what do you think?

packages/sit-onyx/src/components/OnyxKey/types.ts
```diff
@@ -0,0 +1,188 @@
...
+    "page-up": "PgUp",
+    "page-down": "PgDn",
+    home: "Home",
+    end: "End",
+    delete: "Del",
+    fn: "Fn",
+  },
+};
+
+export type OnyxKeyProps<TKeyName extends string = CanonicalKey> = {
```
larsrickert: Whats the purpose of the generic here? :) It seems like its not functional used so we can get rid of it

packages/sit-onyx/src/components/OnyxKey/types.ts
```diff
@@ -0,0 +1,188 @@
...
+  /**
+   * The key name (can be canonical or any custom string).
+   */
+  keyName: TKeyName;
```
larsrickert: Just an idea, let me know what you think:

For a great developer experience, it might be nice to type the keyName as a union type that includes all characters so you get e.g. autocompletion / autosuggestion in the IDE and we can unify it a little bit to e.g. not allow "Space", "space", " " etc. for the same key.

it would mean maintainance effort on our side but it might be worth the DX. What do you think?
Incomplete example code:
```ts
export type NumericKey = 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 0;
export type AlphabeticKey =
  | "A"
  | "B"
  | "C"
  | "D"
  | "E"
  | "F"
  | "G"
  | "H"
  | "I"
  | "J"
  | "K"
  | "L"
  | "M"
  | "N"
  | "O"
  | "P"
  | "Q"
  | "R"
  | "S"
  | "T"
  | "U"
  | "V"
  | "W"
  | "X"
  | "Y"
  | "Z";
export type SpecialKey = "ArrowUp" | "ArrowDown" | "...more keys";
export type KeyboardKey = NumericKey | AlphabeticKey | SpecialKey;

export type OnyxKeyProps = {
  key: KeyboardKey;
}
```

packages/sit-onyx/src/components/OnyxShortcut/OnyxShortcut.stories.ts
```diff
@@ -0,0 +1,78 @@
+import type { Meta, StoryObj } from "@storybook/vue3-vite";
+import OnyxShortcut from "./OnyxShortcut.vue";
+
+const meta: Meta<typeof OnyxShortcut> = {
+  title: "Feedback/Shortcut",
```
larsrickert:
```suggestion
  title: "Basic/Shortcut",
```

We intended to move the shortcut component to the "Basic" category as its use-case can vary so we found "Basic" the most appropriate one. What do you think?

larsrickert (changes requested):
Very cool contribution, thanks! 🎉  I haven't found the time yet to review the OnyxShortcut component or shortcut composable (I hope to find time at the end of the week) but here is already some smaller feedback for the OnyxKey component :)

I like the overall API / usage of the components so I'm open to it :)
I just think that there is currently a lot of key mapping going on which adds a lot of code and might be harder to maintain in the future 🤔  Maybe we can streamline this a little bit, what do you think? Maybe we might introduce / limit our custom `KeyboardKey` type as proposed in one of my comments to define which keys we are expecting?

@JoCa96 What are your thoughts on this?

packages/sit-onyx/src/components/OnyxShortcut/OnyxShortcut.stories.ts
lovelycentury: I simply consider anything that asks the user to take some action as "feedback".
Anyway I would move it simply to "Basic")

packages/sit-onyx/src/components/OnyxKey/types.ts
lovelycentury: Brilliant idea :)

I was thinking of providing a constant that we would recommend using as a source of truth for keys with the possibility of having some of our own keys and about our own type in the same way as you described as well.

I don't really like the constant, since it's an additional import and an additional variable whose name you always need to keep in mind. So I would avoid using it.

Yes, own type gives a very good DX due to autocomplete, 100% agree. 

But support may indeed be an issue. Despite this, in the context of a component like shortcut, this doesn't sound like the end of the world, the most important keys can be declared in the type.

packages/sit-onyx/src/components/OnyxKey/OnyxKey.vue
lovelycentury: I'm not sure about the code styling, but if you think the absence of specific CSS variables that are used once is better. 
I'm also for it :)

packages/sit-onyx/src/components/OnyxKey/OnyxKey.vue
lovelycentury: Since `<kbd />` already has the correct semantics for screen readers, and both macOS (VoiceOver) and Windows read the key symbols or text correctly, we don’t need to generate an automatic aria-label.

aria-label should only be used when the user provides a custom label, but we can also remove this completely once we switch to our own key type. I added it only because it was mentioned in the requirements.

`https://github.com/SchwarzIT/onyx/issues/2066 icon should be hidden for screenreader, instead key name should be shown (visually hidden)`

lovelycentury:
@JoCa96 

This PR will be heavily modified within a week, so don't waste time reviewing it for now.

lovelycentury:
@larsrickert 
I updated the pull request and addressed all the bugs from the previous version, so it's ready for the review again.

packages/sit-onyx/playwright/snapshots/components/OnyxGlobalSearch/responsive-webkit-linux.png
larsrickert: Do you know why the "Esc" key looks broken here? Seems to only affect webkit/safari

packages/sit-onyx/src/composables/useShortcut.ts
```diff
@@ -0,0 +1,322 @@
...
+   * Targets on which to call `event.stopPropagation()` during keydown event.
+   *
+   * - `sequenceComplete` - stop propagation when the full sequence is completed.
+   * - `stepComplete` - stop propagation when step is completed.
+   * - `keydown` - stop propagation on every keydown event.
+   * - `none` - do not stop propagation on any event.
+   *
+   * @default "stepComplete"
+   */
+  stopPropagationOn?: MaybeRef<ShortcutEventTarget>;
```
larsrickert: Can you explain why this is needed? Imo, keyboard shortcuts should be non-interruptive so they shouldn't prevent default or stop propagation. Is there a specific reason why we need them in order for the "useShortcut()" to work?

packages/sit-onyx/src/composables/useShortcut.ts
```diff
@@ -0,0 +1,322 @@
...
+  /**
+   * Whether the shortcut handling is disabled.
+   */
+  disabled?: MaybeRef<boolean>;
+  /**
+   * Whether to listen to repeated keydown events.
+   *
+   * @default false
+   */
+  listenOnRepeat?: MaybeRef<boolean>;
```
larsrickert: What does this prop do for the user?

packages/sit-onyx/src/composables/useShortcut.ts
lovelycentury: My idea was to provide a mechanism that would handle a case where two handlers are assigned to the same shortcut, but only one should be called.

This could happen e.g. if we want to open a modal/dialog in which we want to execute a specific shortcut that already exists on the background page.

Now I see it could look a bit complicated, and we can remove those 2 props.

packages/sit-onyx/src/composables/useShortcut.ts
lovelycentury: I got inspired by this prop when I looked at shortcut in global search.

There are navigation arrows there, and you want to hold down the button and constantly call the handler for switching the focused element, so that the user does not constantly lift his finger

Use case looks like this:

```
useShortcut({ sequence: [{ any: ["ArrowUp", "ArrowDown"] }], listenOnRepeat: true })
```

lovelycentury:
> Thanks for contributing, nice to have such advanced shortcut functionality in onyx!
> 
> I made some adjustments, could you please re-run the Playwright screenshot pipeline in your forked repo?

@larsrickert Does it make sense to resolve conflicts before running screenshoot tests?

packages/sit-onyx/src/composables/useShortcut.ts
larsrickert: I see, thanks for the explanation! Can't we solve this by always resetting the pressed keys when the shortcut is completed? So that if you press the shortcut multiple times, the event will also be triggered multiple times

larsrickert:
> @larsrickert Does it make sense to resolve conflicts before running screenshoot tests?

Sure, I rebased the branch 👍

packages/sit-onyx/src/composables/useShortcut.ts
larsrickert: Edit: I'd suggest to remove the property for now to keep things simple and wait for user feedback here.
Imo shortcuts are intended for single execution only and not for repeated events (e.g. holding down a key)

lovelycentury: Agree with you, such case with arrows probably batter to handle using OnyxKey + normal event listener instead of shorcut.

In any case, I tried to throw some ideas, maybe you would like some of them  :)

lovelycentury:
> > @larsrickert Does it make sense to resolve conflicts before running screenshoot tests?
> 
> Sure, I rebased the branch 👍

@larsrickert Done! Thanks a lot for your collaboration :)

PR #4471, opened 2025-11-20
fix(OnyxDataGrid): not applying ellipsis truncation

### LINKED ISSUE #4387: [Bug]: Datagrid Truncation ellipsis
### Version

1.3.0

### Description

` <OnyxDataGrid :columns :data truncation="ellipsis" />` should use ellipsis for long values.
It actually just shows multiple lines, see image.

Image

### Reproduction steps

https://playground.onyx.schwarz/?onyxVersion=1.1.0#eNq9U8FO4zAQ/ZWRz9lWu7CXbEGCbrWCA6yAG0YoTYbU4Iwt2ymtqv47Y6cloUIcuUQZv3njNzPPG3Fm7WjZosjFxJdO2QC6oPpEiuClAI+htaeSVGONC7CBa1qv/hah+OdUlUFYW4Sp0W1DU0NPqoYtPDnTgBRehR+Gk6X4I0lSyrybUXBrOIGNJABV5UBtM0fHGQBUNJiDD05RzQfbRCsN+QAVX5jv2PcPzL+PhE2q8DPbMaU406pEFr3NevhXD08XhdPqMOGoT7htLTrw3/hFMq4xrQdtqE46zs38QOBxL/DGzBUdwL97+NIs9ujDYHpl2o/PPyxq0k3zdDjOF1zHMqqSImMXzFHH8CKFcX8x6hY20LAjRQ0faFfpYCBG0mTcOYz9xEHAxuoiIEcwGdoK8p1iyOPiIbiWyiIoQ+xK1FpZr6I3x12d8XshkbFpuWdub/TsDbGnk9GkKE1jlUZ3bWMZJuedBSNWaG1eL9MZ34SprcRZYPnyyfmzZ0/HDv879OiWse09FgpXY+jg2e0Vrvj/HWxM1WrO/gK8Qc+tR41d2nlLFcse5CW1F+k18ju587NVQPL7pqLQmNktRwp+2NMvWu/lHo2OE4/fHU/xcYku1uQBxsH6ILZvWZVk3A==

### Confirmations

JoCa96:
Closes #4387 

**Cause:** The `truncation` prop was not forwarded from the `OnyxDataGrid` component to the internal `OnyxDataGridRenderer` component and therefore not applied.

**Fix:** The `useForwardProps` composable is now used to forward all matching props from the `OnyxDataGrid` to the `OnyxDataGridRenderer`


packages/sit-onyx/src/components/OnyxTable/types.ts
```diff
@@ -31,6 +31,7 @@ export type OnyxTableProps = DensityProp & {
   /**
    * Defines how overflows of text content in cells (`td`-elements) are handled.
    * The cell can either grow to multiple lines (`'multiline'`) or truncate the text (`'ellipsis'`).
+   * The `'ellipsis'` mode will only show an ellipsis, when a (maximum) width has been defined for a cell/column.
```
larsrickert, line 34: Question: If no max width is defined, e.g. because the user has 4 columns all without a width so they all automatically have the same width, is the ellipsis still applied then?

larsrickert (approved):

packages/sit-onyx/src/components/OnyxTable/types.ts
JoCa96: The default width for a column is `minmax(min-content, 1fr)`, so an overflow will be the result.

JoCa96: We could consider changing the default width to 1fr for ellipsis truncation?

larsrickert: Yes imo it would be nice if the ellipsis truncation works "out of the box"

larsrickert (commented):
It seems like there are some changes shown as changed that are coming from other already completed PRs, did something went wrong while rebasing?

The actual data grid changes look fine to me

Issue #4404, opened 2025-11-07
Implement OnyxItemsPerPage component


larsrickert:
### Why?

We want to offer a unified component for selecting the number of items per page for paginated data (e.g. the data grid with pagination)

### Design

[Figma](https://www.figma.com/design/YfEUBOHk4J4nYrk04geswG/Onyx-Components?node-id=19029-56015&m=dev)
[Documentation](https://www.figma.com/design/j7Hy7Zk8MiCmJBJf0ArQRm/Onyx-Documentation?node-id=16307-163365&m=dev)
[Documentation for integration into data grid](https://www.figma.com/design/j7Hy7Zk8MiCmJBJf0ArQRm/Onyx-Documentation?node-id=16307-303380&m=dev)

### Acceptance criteria

- [x] a new OnyxItemsPerPage component exists that
   - [x] shows a default translated label
   - [x] supports right (default) and left alignment of the label
   - [x] can optionally hide the label (but its still applied as aria-label)
   - [x] can pass a list of options that can be selected from
- [x] the existing pagination feature of the data grid is updated:
   - [x] by default, no items per page is displayed
   - [x] the developer can pass itemsPerPage options that then display the OnyxItemsPerPage component in the bottom left slot of the data grid
   - [x] the itemsPerPage is supported for all pagination types (default, lazy and more button loading)

### DEV handover checklist

[Storybook](https://onyx-storybook-dev.apps.01.cf.eu01.stackit.cloud/?path=/docs/data-itemsperpage--docs)
[Data grid pagination feature](https://onyx-storybook-dev.apps.01.cf.eu01.stackit.cloud/?path=/story/data-datagrid-features--pagination)

DEV handover done with: @larsrickert 

**Testing**
- [x] covered by visual screenshot tests
- [x] covered by functional tests (Playwright or unit test)

**Documentation**
- [x] updated version is released

### UX handover checklist

UX handover done with: @flubnau 

**Naming is aligned with Figma**
- [x] component name
- [x] which support components are used and how they are named
- [x] all property names and types

**Implementation is checked**
- [x] all features of Figma are implemented or a follow-up ticket is refined
- [x] skeleton
- [x] densities
- [x] light and dark mode
- [ ] truncation
- [ ] responsiveness / wrapping behavior

**Figma**
- [x] Figma component is published

**Documentation**
- [x] Storybook category is aligned with Figma
- [x] Storybook examples are checked

**Other**
[...]

PR #4481, opened 2025-11-22
feat(OnyxItemsPerPage): implement `OnyxItemsPerPage` component and adjust `OnyxDataGrid` pagination feature with `itemsPerPage` option

lovelycentury:
Relates to #4404

- Add `OnyxItemsPerPage` component for configurable page size selection
- Add `itemsPerPage` option to `OnyxDataGrid` pagination feature to control visible rows per page
- Fix `useOpenDirection `composable to correctly handle viewport boundaries when no overflow parent exists

## Checklist

### DIFF
--- a/packages/sit-onyx/src/components/OnyxItemsPerPage/OnyxItemsPerPage.stories.ts
+++ b/packages/sit-onyx/src/components/OnyxItemsPerPage/OnyxItemsPerPage.stories.ts
@@ -15,8 +15,8 @@ type Story = StoryObj<typeof OnyxItemsPerPage>;
 
 export const Default = {
   args: {
-    modelValue: 5,
-    options: [5, 10, 20, 30, 40, 75],
+    modelValue: 20,
+    options: [10, 20, 30, 40, 50],
   },
 } satisfies Story;
 


lovelycentury:
@JoCa96

In this pull request, I slightly modified the `useOpenDirection` composable. Because the way the position was calculated when there was no parent with `overflow: hidden` was incorrect. 

`getBoundingClientRect` - always returns the position of the element relative to the viewport, not including scrolling. 
So,  this statement `const parentTop = overflowParentRect?.top ?? window.visualViewport?.pageTop ?? 0;` is incorrect, because when there is no parent, we take the initial position, including the scroll, but in fact it should always be `0`. 

For complete clarity, I will provide a video before and after the fix, so that there is understanding.

Before
https://github.com/user-attachments/assets/dfff8820-be16-4be8-800f-bf3b80583f06

After
https://github.com/user-attachments/assets/f1d22f1b-8f83-4a61-a383-f5bfabd859bb

packages/sit-onyx/playwright/snapshots/components/OnyxItemsPerPage/ItemsPerPage-button-states--webkit-linux.png
larsrickert, line 1: Do you know why the page size value is truncated / cut off here?

packages/sit-onyx/src/components/OnyxDataGrid/features/pagination/pagination.ct.tsx
```diff
@@ -141,6 +141,87 @@ test("should hide pagination when empty or only one page exists", async ({ mount
...
+  await component.update({
+    props: {
+      style: { maxHeight: "24rem" },
+      data: getTestData(52),
+      paginationOptions: { type: "lazy", itemsPerPage: [10, 25, 50] },
+    },
+  });
+
+  // ASSERT
+  await expect(itemsPerPage, "should hide items per page for lazy pagination").toBeHidden();
```
larsrickert: It should still be possible to have the itemsPerPage when lazy or button loading is used so the user can optionally change how many pages are lazy/button loaded :) Since the developer needs to explicitly pass the itemsPerPage, in this case he also decides that lazy/button loading should have the itemsPerPage funcitonality

packages/sit-onyx/src/components/OnyxDataGrid/features/pagination/pagination.ts
```diff
@@ -139,6 +140,24 @@ export const usePagination = (options: PaginationOptions = {}) =>
             }),
           ];
         },
+        bottomLeft: () => {
+          if (!isEnabled.value() || type !== "select" || !options.itemsPerPage?.length) return [];
+          const skeleton = ctx.skeleton.value && !shouldShowPagination.value;
```
larsrickert: (optional) We also use this condition for the pagination itself, we could move this to a separated computed so we share the logic there

packages/sit-onyx/src/components/OnyxDataGrid/features/pagination/types.ts
```diff
@@ -37,6 +37,10 @@ export type PaginationOptions = Pick<
    * Should only be used if pagination is async.
    */
   loading?: Ref<boolean>;
+  /**
+   * If defined, `OnyxItemsPerPage` component will be displayed in the bottom left slot of the data grid.
+   */
+  itemsPerPage?: Array<number>;
```
larsrickert:
```suggestion
  itemsPerPage?: number[];
```

Just to align with the code style :D

packages/sit-onyx/src/components/OnyxItemsPerPage/OnyxItemsPerPage.ct.tsx
```diff
@@ -0,0 +1,196 @@
...
+  // ACT
+  await select.click();
+  await component.getByRole("option", { name: "5", exact: true }).click();
+
+  // ASSERT
+  expect(modelValue).toBe(5);
+  await expect(select).toHaveValue("5");
+});
+
+test("should use custom label", async ({ mount }) => {
```
larsrickert: We already have this covered by the screenshot test so we can remove this one :)

packages/sit-onyx/src/components/OnyxItemsPerPage/OnyxItemsPerPage.ct.tsx
```diff
@@ -0,0 +1,196 @@
...
+      modelValue: 10,
+      label: "Custom label text",
+    },
+  });
+
+  // ASSERT
+  await expect(component.getByText("Custom label text")).toBeVisible();
+});
+
+test("should render with custom options", async ({ mount }) => {
```
larsrickert: Isn't this the same as the other tests?

packages/sit-onyx/src/components/OnyxItemsPerPage/OnyxItemsPerPage.vue
```diff
@@ -0,0 +1,156 @@
...
+} from "../../composables/useSkeletonState.js";
+import { useVModel } from "../../composables/useVModel.js";
+import { injectI18n } from "../../i18n/index.js";
+import OnyxSelect from "../OnyxSelect/OnyxSelect.vue";
+import type { SelectOption } from "../OnyxSelect/types.js";
+import OnyxSkeleton from "../OnyxSkeleton/OnyxSkeleton.vue";
+import { DEFAULT_ITEMS_PER_PAGE_OPTIONS, type OnyxItemsPerPageProps } from "./types.js";
+
+const props = withDefaults(defineProps<OnyxItemsPerPageProps>(), {
+  labelAlignment: "left",
```
larsrickert:
```suggestion
  labelAlignment: "right",
```

I'm sorry I added the wrong information in #4404, the default alignment should be right :) I updated it in the ticket as well

packages/sit-onyx/src/components/OnyxItemsPerPage/OnyxItemsPerPage.vue
```diff
@@ -0,0 +1,156 @@
...
+  props.options.map((option) => ({ label: option.toString(), value: option })),
+);
+
+const actualLabel = computed(() => {
+  if (props.label) {
+    return props.label;
+  }
+
+  return t.value("itemsPerPage.label");
+});
```
larsrickert:
```suggestion
const actualLabel = computed(() => props.label || t.value("itemsPerPage.label"));
```

packages/sit-onyx/src/components/OnyxItemsPerPage/types.ts
```diff
@@ -0,0 +1,46 @@
...
+   *
+   * @default false
+   */
+  disabled?: boolean;
+  /**
+   * An optional label for the items per page selector.
+   *
+   * @default "Results per Page"
+   */
+  label?: string;
```
larsrickert: For now we don't want to allow changing the label so its uniform across multiple projects since the use case is pretty specific

packages/sit-onyx/src/components/OnyxItemsPerPage/OnyxItemsPerPage.vue
```diff
@@ -0,0 +1,156 @@
...
+import { useVModel } from "../../composables/useVModel.js";
+import { injectI18n } from "../../i18n/index.js";
+import OnyxSelect from "../OnyxSelect/OnyxSelect.vue";
+import type { SelectOption } from "../OnyxSelect/types.js";
+import OnyxSkeleton from "../OnyxSkeleton/OnyxSkeleton.vue";
+import { DEFAULT_ITEMS_PER_PAGE_OPTIONS, type OnyxItemsPerPageProps } from "./types.js";
+
+const props = withDefaults(defineProps<OnyxItemsPerPageProps>(), {
+  labelAlignment: "left",
+  options: () => DEFAULT_ITEMS_PER_PAGE_OPTIONS,
```
larsrickert: There should not be any default options so the user is required to pass them. Since these propably also depend on what the API supports

packages/sit-onyx/src/i18n/locales/en-US.json
```diff
@@ -267,5 +267,11 @@
       "select": "Select"
     },
     "system": "System"
+  },
+  "itemsPerPage": {
+    "label": "Results per Page",
+    "select": {
+      "listLabel": "Select number of results per Page"
+    }
```
larsrickert:
```suggestion
    "label": "Items per page",
    "select": {
      "listLabel": "Select number of items per page"
    }
```

Would prefer to keep the translations aligned with the component name :)

packages/sit-onyx/src/i18n/locales/de-DE.json
```diff
@@ -264,5 +264,11 @@
       "select": "Auswählen"
     },
     "system": "System"
+  },
+  "itemsPerPage": {
+    "label": "Ergebnisse pro Seite",
+    "select": {
+      "listLabel": "Anzahl der Ergebnisse pro Seite auswählen"
+    }
```
```suggestion
    "label": "Elemente pro Seite",
    "select": {
      "listLabel": "Anzahl der Elemente pro Seite auswählen"
    }
```

packages/sit-onyx/src/composables/useOpenDirection.spec.ts
larsrickert, line 1: Thanks for creating the tests! Somehow, when keeping the tests as they are but reverting the implementation back to the old one, the tests still pass so they seem to not catch/test the scrolling issue that you fixed. Do you have an idea whats the missing part here?

larsrickert (commented):
@lovelycentury I slightly changed the text color and translation to match Figma and give it a better contrast. Could you please re-generate the screenshots? :)

The rest looks good to me!

packages/sit-onyx/src/composables/useOpenDirection.spec.ts
lovelycentury: Good catch :)

I'll update one test that will check that visualViewport.pageTop doesn't affect how composable behaves.
Basically here is a test:

```test("should open top when there is more space above (without overflow parent)", () => {
    // ARRANGE - Create element with more space above
    const element = document.createElement("div");
    // Mock visualViewport.pageTop to ensure that implementation is not relying on it
    Object.defineProperty(window, "visualViewport", {
      value: { height: 1000, pageTop: 10000 },
      configurable: true,
    });
    Object.defineProperty(element, "getBoundingClientRect", {
      value: () => ({
        top: 800,
        bottom: 850,
        left: 0,
        right: 0,
        width: 100,
        height: 50,
      }),
    });

    // Mock getComputedStyle to return no overflow hidden
    vi.spyOn(global, "getComputedStyle").mockReturnValue({
      overflow: "visible",
    } as CSSStyleDeclaration);

    const elementRef = shallowRef(element);
    const { openDirection, updateOpenDirection } = useOpenDirection(elementRef);

    // ACT
    updateOpenDirection();

    // ASSERT
    // Free space below: 1000 - 850 = 150
    // Free space above: 800 - 0 = 800
    // Should open top since 800 > 150
    expect(openDirection.value).toBe("top");
  });
  ```
  
  And here is the check itself

  
  ``` // Mock visualViewport.pageTop to ensure that implementation is not relying on it
    Object.defineProperty(window, "visualViewport", {
      value: { height: 1000, pageTop: 10000 },
      configurable: true,
    });
 ```
    
 Does it works for you?

larsrickert: Yes that seems to work :) Would you mind including it in the PR?

lovelycentury: yes, will include it now and will run action to update screenshots

larsrickert: Nice, thanks! FYI: I migrated your unit tests to Playwright so we don't need to mock all the browsers details

larsrickert (approved):

packages/sit-onyx/src/composables/useOpenDirection.spec.ts
lovelycentury: Using playwright was the best decision :)

Issue #4260, opened 2025-10-14
Implement basic OnyxTimepicker component


larsrickert:
### Why?

We want to offer a timepicker component where users can select a time.

### Design

[Figma](https://www.figma.com/design/YfEUBOHk4J4nYrk04geswG/Onyx-Components?node-id=18375-30874&m=dev)
[Documentation](https://www.figma.com/design/j7Hy7Zk8MiCmJBJf0ArQRm/Onyx-Documentation?node-id=17388-13864&m=dev)

### Acceptance criteria

- a time can be selected in one of the following formats:
    - [ ] hh:mm (default)
    - [ ] hh:mm:ss
    - [ ] mm:ss
    - [ ] hh
- [ ] depending on the format, time segment inputs are display in the flyout
- [ ] in the main input, the user can also manually enter a time with the given format which is parsed automatically and reflected in the timepicker
- [ ] the current value can be cleared with a clear button or by manually deleting the input value
- [ ] when opening the flyout, the first time segment is automatically focused
- [ ] supports all common properties of the OnyxFormElement (label, info tooltips, required, disabled, message etc.)
- [ ] optionally, a info label can be displayed at the bottom of the flyout

#### Out of scope
- [ ] optionally a am / pm select can be displayed. Can be set to a boolean or auto where the users locale is used to determine if 12h or 24h format should be used

### Implementation details

 - we should re-use the OnyxSelect and OnyxFormElement if possible

### Reference implementations

- [RekaUI](https://reka-ui.com/docs/components/time-field)

### Applicable ARIA Pattern

[Combobox](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/)

### DEV handover checklist

[Storybook](ADD_LINK_HERE)

DEV handover done with: @ADD_DEV_NAME_HERE

**Testing**
- [ ] covered by visual screenshot tests
- [ ] covered by functional tests (Playwright or unit test)

**Documentation**
- [ ] "New" badge in Storybook is added, see [storybook-addon-tag-badges](https://github.com/Sidnioulz/storybook-addon-tag-badges/?tab=readme-ov-file#-usage)
- [ ] auto-generated Storybook code snippets are checked
- [ ] component status has been updated in the [docs overview](https://onyx.schwarz/#components)
- [ ] updated version is released
- [ ] updated documentation is deployed

**Other**
- [ ] follow-up tickets were created if necessary (add links below)

### UX handover checklist

[...]

PR #4503, opened 2025-11-29
Feat(OnyxTimepicker): implement component

ChristianBusshoff:
Relates to #4260

- implemented OnyxTimepicker component
- marked as unstable 
- added tests

## Open Question

I think the opening logic is not perfect, but I didn't know how to have it with working while supporting keyboard control.


packages/sit-onyx/src/components/OnyxTimepicker/OnyxTimepickerInput.vue
JoCa96, line 1: This duplicates a lot of code from the `OnyxInput`, and also relies on the styles from it, as it re-uses the classes.
That creates implicit dependencies on the `OnyxInput`. So it would be better to just re-use the component directly and override some of the styles.

packages/sit-onyx/src/components/OnyxTimepicker/types.ts
```diff
@@ -0,0 +1,38 @@
...
+      second?: boolean;
+    };
+    /**
+     * Info label text.
+     */
+    infoLabel?: string;
+    /**
+     * Whether to show the info label.
+     */
+    hideInfoLabelIcon?: boolean;
```
JoCa96, line 18: Please double check with UX, if this intended to be configurable. Seems like it could be an unintended.
If intended we should consider if this can be aligned with the list description from the `OnyxSelect`.

packages/sit-onyx/src/components/OnyxTimepicker/types.ts
```diff
@@ -0,0 +1,38 @@
...
+    /**
+     * Time in Seconds since midnight.
+     */
+    modelValue?: number;
```
JoCa96: > Time in Seconds since midnight

This seems like an unconventional time format and might be annoying to use.
To make it more intuitive and future-proof, I would suggest to use RFC 9557, which the [Temporal API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal/PlainTime#rfc_9557_format) is going to use.

packages/sit-onyx/src/components/OnyxTimepicker/OnyxTimepicker.vue
```diff
@@ -0,0 +1,526 @@
...
+          v-bind="inputProps"
+          v-model="timeString"
+          class="onyx-timepicker__input"
+          :placeholder="placeholderText"
+          type="text"
+          @focus="handleOpen"
+          @click="handleSegmentClick"
+          @input="handleInputFilter"
+          @keydown="handleInputKeyDown"
+        />
```
JoCa96: According to [Figma](https://www.figma.com/design/j7Hy7Zk8MiCmJBJf0ArQRm/Onyx-Documentation?node-id=17388-13889&m=dev):

> Input Field: This is the primary interactive area. It serves both to display the currently selected time and as a button to open the time selection flyout. In its empty state, the field displays a placeholder text of HH:MM:SS to guide the user on the expected format.

IMHO this means that typing directly into the input is not supported, which would align with our current `OnyxSelect` behavior.

packages/sit-onyx/src/components/OnyxTimepicker/OnyxTimepicker.vue
```diff
@@ -0,0 +1,526 @@
...
+};
+
+onMounted(() => {
+  document.addEventListener("mousedown", handleClickOutside);
+});
+
+onUnmounted(() => {
+  document.removeEventListener("mousedown", handleClickOutside);
+});
+
```
JoCa96: Can we use the `useOutsideClick` composable, that is provided by the `@sit-onyx/headless`?

packages/sit-onyx/src/components/OnyxTimepicker/OnyxTimepicker.vue
```diff
@@ -0,0 +1,526 @@
...
+              :label="t('timepicker.labels.hour')"
+              :placeholder="t('timepicker.placeholder.hour')"
+              :disabled="props.disabled"
+              :loading="props.loading"
+              hide-label
+              hide-clear-icon
+              hide-success-icon
+              hide-buttons
+              @click="handleSegmentFocus(hourInputRef)"
+              @focus="handleSegmentFocus(hourInputRef)"
```
JoCa96: IMHO this should not be necessary, which issue is this trying to solve?

packages/sit-onyx/src/components/OnyxTimepicker/OnyxTimepicker.vue
```diff
@@ -0,0 +1,526 @@
...
+              hide-label
+              :disabled="props.disabled"
+              :loading="props.loading"
+              hide-clear-icon
+              hide-success-icon
+              hide-buttons
+              @click="handleSegmentFocus(minuteInputRef)"
+              @focus="handleSegmentFocus(minuteInputRef)"
+              @keydown="(e: KeyboardEvent) => handleInputChange('minute', e)"
+            />
```
JoCa96, line 270: Can we use `min` and `max` here, or will that create issues?

packages/sit-onyx/src/components/OnyxTimepicker/types.ts
ChristianBusshoff: It's intended

packages/sit-onyx/src/components/OnyxTimepicker/OnyxTimepicker.vue
ChristianBusshoff: Will create a layoutShift cause of the errorState

packages/sit-onyx/src/components/OnyxTimepicker/OnyxTimepicker.vue
ChristianBusshoff: Talked with UX, should be possible to edit in there as well.

packages/sit-onyx/src/components/OnyxDatePicker/OnyxDatePicker.vue
JoCa96, line 1: Please revert the Datepicker changes

JoCa96 (approved):

PR #4522, opened 2025-12-03
fix(OnyxLoadingSpinner): fix circle variant not having correct width in some contexts

JoCa96:
The circle loading spinner was around 3pixel smaller than intended.
This caused a content-shift in some cases.


packages/sit-onyx/playwright/snapshots/components/OnyxSwitch/Switch-hidden-label--webkit-linux.png
ChristianBusshoff, line 1: the new spinner looks off.

JoCa96: oh, good catch

JoCa96: I fixed the switch and cross-checked the other components with Figma, should be fine now

ChristianBusshoff (approved):

Issue #4262, opened 2025-10-14
Implement select type for OnyxTimepicker


larsrickert:
### Why?

Optionally, the time picker should be able to be displayed as select with a set of pre-defined time options.

### Depends on

- #4260

### Design

[Figma](https://www.figma.com/design/YfEUBOHk4J4nYrk04geswG/Onyx-Components?node-id=18332-17915&m=dev)
[Documentation](https://www.figma.com/design/j7Hy7Zk8MiCmJBJf0ArQRm/Onyx-Documentation?node-id=17388-13864&m=dev)

### Acceptance criteria

- [ ] the timepicker can be displayed as "select" type with a given set of available time options
- [ ] supports setting a `step` size which generates the options by default. Default step size: 30
- [ ] optionally, custom steps can be passed which override the default generated ones
- [ ] supports / optionally exposes useful existing properties of the OnyxSelect

### Implementation details

- we should use the existing OnyxSelect internally

### DEV handover checklist

[Storybook](ADD_LINK_HERE)

DEV handover done with: @ADD_DEV_NAME_HERE

**Testing**
- [ ] covered by visual screenshot tests
- [ ] covered by functional tests (Playwright or unit test)

**Documentation**
- [ ] "New" badge in Storybook is added, see [storybook-addon-tag-badges](https://github.com/Sidnioulz/storybook-addon-tag-badges/?tab=readme-ov-file#-usage)
- [ ] auto-generated Storybook code snippets are checked
- [ ] component status has been updated in the [docs overview](https://onyx.schwarz/#components)
- [ ] updated version is released
- [ ] updated documentation is deployed

**Other**
- [ ] follow-up tickets were created if necessary (add links below)

### UX handover checklist

UX handover done with: @ADD_UX_NAME_HERE

**Naming is aligned with Figma**
- [ ] component name
- [ ] which support components are used and how they are named
- [ ] all property names and types

**Implementation is checked**
- [ ] all features of Figma are implemented or a follow-up ticket is refined
- [ ] skeleton
- [ ] densities
- [ ] light and dark mode
- [ ] truncation
- [ ] responsiveness / wrapping behavior
- [ ] min / max width

**Figma**
- [ ] Figma component is published

**Documentation**
- [ ] Storybook category is aligned with Figma
- [ ] Storybook examples are checked

**Other**
- [ ] follow-up tickets were created if necessary (add links below)

PR #4538, opened 2025-12-09
feat(OnyxTimepicker): implement select type

ChristianBusshoff:
Relates to #4262

- implemented select type for timepicker

## Checklist


packages/sit-onyx/src/components/OnyxSelectInput/OnyxSelectInput.vue
```diff
@@ -199,16 +206,18 @@ useAutofocus(input, props);
...
+              class="onyx-select-input__button"
+              type="button"
+              :aria-label="t('select.toggleDropDown')"
+              :title="t('select.toggleDropDown')"
+              tabindex="-1"
+              :disabled="disabled || props.readonly || props.loading"
+            >
+              <OnyxIcon :icon="iconChevronDownUp" />
+            </button>
+          </slot>
```
JoCa96: To also support all the button interactions, it would be preferable to only swap the icon.
```suggestion
          <button
            class="onyx-select-input__button"
            type="button"
            :aria-label="t('select.toggleDropDown')"
            :title="t('select.toggleDropDown')"
            tabindex="-1"
            :disabled="disabled || props.readonly || props.loading"
          >
            <slot name="icon">
              <OnyxIcon :icon="iconChevronDownUp" />
            </slot>
          </button>
```

packages/sit-onyx/src/components/OnyxSelectInput/OnyxSelectInput.vue
```diff
@@ -47,6 +47,13 @@ const emit = defineEmits<{
   validityChange: [validity: ValidityState];
 }>();
 
+defineSlots<{
+  /**
+   * Optional slot to override the input icon.
```
JoCa96:
```suggestion
   * Optional slot to override the icon of the toggle button inside the input.
```

packages/sit-onyx/src/components/OnyxSelect/OnyxSelect.vue
```diff
@@ -98,6 +98,11 @@ const slots = defineSlots<{
    * Optional slot to override the option content.
    */
   option?(props: SelectOption<TValue>): unknown;
+
+  /**
+   * Optional slot to override the input icon.
```
JoCa96:
```suggestion
    * Optional slot to override the icon of the toggle button inside the input.
```

packages/sit-onyx/src/components/OnyxSelect/OnyxSelect.vue
```diff
@@ -98,6 +98,11 @@ const slots = defineSlots<{
...
+   */
+  icon?(): unknown;
```
JoCa96: Would prefer to use an property name like `toggleIcon`, as there might exist further icons
```suggestion
  toggleIcon?(): unknown;
```

packages/sit-onyx/src/components/OnyxTimepicker/OnyxTimepicker.vue
```diff
@@ -198,15 +211,65 @@ const inputProps = computed(() => {
     showSeconds: __,
     infoLabel: ___,
     hideInfoLabelIcon: ____,
+    type: _____,
     ...others
   } = props;
   return others;
 });
```
JoCa96: Can't we make use of [`useForwardProps`](https://onyx.schwarz/principles/contributing/patterns.html#shared-child-props-forwarding) here?

packages/sit-onyx/src/components/OnyxTimepicker/types.ts
```diff
@@ -1,8 +1,42 @@
...
+       */
+      endTime?: string;
+      /**
+       * The step size, in minutes, used for generating the time options.
+       * @default 30
+       */
+      stepSize?: number;
+      /**
+       * A function that provides custom time options.
+       * These options will be MERGED with the automatically generated times.
```
JoCa96:
```suggestion
       * These options will be MERGED with the automatically generated times, if defined.
```

packages/sit-onyx/src/components/OnyxTimepicker/types.ts
```diff
@@ -1,8 +1,42 @@
...
+    options?: {
+      /**
+       * The starting time for generating time options (using 24-hour format).
+       * @format HH:MM (e.g., "08:00")
+       */
+      startTime?: string;
+      /**
+       * The ending time for generating time options (using 24-hour format).
+       * @format HH:MM (e.g., "17:30")
+       */
```
JoCa96: Please mention if the endTime is inclusive or exclusive

packages/sit-onyx/src/components/OnyxTimepicker/types.ts
```diff
@@ -1,8 +1,42 @@
...
 
 export type OnyxTimepickerProps = Omit<SharedFormElementProps, "placeholder"> &
   AutofocusProp & {
+    /**
+     * Specifies the type of timepicker input.
+     * 'default': Free text input with validation.
+     * 'select': Displays a dropdown list with pre-generated time options.
+     * @default 'default'
+     */
+    type?: "default" | "select";
```
JoCa96: Can we make it generic like e.g. `OnyxSlider` to not allow options for the `default` mode

packages/sit-onyx/src/components/OnyxTimepicker/types.ts
```diff
@@ -1,8 +1,42 @@
...
+    /**
+     * Configuration options for the timepicker's option generation.
+     * This property only takes effect when the `type` is set to 'select'.
+     */
+    options?: {
+      /**
+       * The starting time for generating time options (using 24-hour format).
+       * @format HH:MM (e.g., "08:00")
```
JoCa96: What about support for seconds? Default type supports seconds.

packages/sit-onyx/src/components/OnyxTimepicker/types.ts
```diff
@@ -1,8 +1,42 @@
...
+       */
+      startTime?: string;
+      /**
+       * The ending time for generating time options (using 24-hour format).
+       * @format HH:MM (e.g., "17:30")
+       */
+      endTime?: string;
```
JoCa96: We could use `min` and `max` instead, then it follows the convention of the native input and `OnyxStepper`. They can also be use the with `default` mode. This also means we need to move these properties from `options` to the root.

packages/sit-onyx/src/components/OnyxTimepicker/types.ts
```diff
@@ -1,8 +1,42 @@
...
+      /**
+       * The step size, in minutes, used for generating the time options.
```
JoCa96: For supporting seconds, we should preferably use seconds as unit.

JoCa96 (changes requested):

packages/sit-onyx/src/components/OnyxTimepicker/OnyxTimepicker.vue
```diff
@@ -198,15 +211,65 @@ const inputProps = computed(() => {
...
+  }
+
+  return options;
+};
 </script>
 
 <template>
   <div ref="rootTemplateRef" class="onyx-component onyx-timepicker">
     <OnyxBasicPopover
+      v-if="props.type === 'default'"
```
JoCa96: We have separate logic and template, we should split these two variants into two internal components (e.g. `TimeInput` and `TimeSelect`). Then the `OnyxTimepicker` does only swap between them.

packages/sit-onyx/src/components/OnyxTimepicker/OnyxTimepicker.vue
```diff
@@ -300,6 +364,35 @@ const inputProps = computed(() => {
         </div>
       </template>
     </OnyxBasicPopover>
+
+    <div v-else>
```
larsrickert: Is the additional div necessary or can we remove it and place the v-else directly on the OnyxSelect?

packages/sit-onyx/src/components/OnyxTimepicker/OnyxTimepicker.vue
```diff
@@ -198,15 +211,65 @@ const inputProps = computed(() => {
...
     hideInfoLabelIcon: ____,
+    type: _____,
     ...others
   } = props;
   return others;
 });
+
+/**************Select Timepicker**************** */
+
+const generateTimeOptions = (): SelectOption<string>[] => {
```
larsrickert: Would prefer to use a computed here instead of a function that returns the value so its guaranteed to be reactive :)

Since computed properties are lazy, the logic is only executed when the value is actually accessed (so type is "select") so there is no overhead when not using the select type

packages/sit-onyx/src/components/OnyxTimepicker/OnyxTimepicker.vue
```diff
@@ -198,15 +211,65 @@ const inputProps = computed(() => {
...
+    startHour === null ||
+    startMinute === undefined ||
+    startMinute === null ||
+    endHour === undefined ||
+    endHour === null ||
+    endMinute === undefined ||
+    endMinute === null
+  ) {
+    return [];
+  }
```
larsrickert:
```suggestion
  if (
    startHour == undefined ||
    startMinute == undefined ||
    endHour == undefined ||
    endMinute == undefined
  ) {
    return [];
  }
```

Using `==` instead of `===` automatically checks for both null and undefined

packages/sit-onyx/src/components/OnyxTimepicker/types.ts
```diff
@@ -1,8 +1,42 @@
...
+       * @default 30
+       */
+      stepSize?: number;
+      /**
+       * A function that provides custom time options.
+       * These options will be MERGED with the automatically generated times.
+       */
+      customTimes?: () => SelectOption<string>[];
```
larsrickert: Is there a specific reason why the custom times are merged? When using custom times, the user most likely wants to have full control over the options so also including auto-generated times might be very rare, what do you think?

Alternatively, we could type the prop like so:
```ts
options?: SelectOption<string>[] | {
      /**
       * The starting time for generating time options (using 24-hour format).
       * @format HH:MM (e.g., "08:00")
       */
      startTime?: string;
      /**
       * The ending time for generating time options (using 24-hour format).
       * @format HH:MM (e.g., "17:30")
       */
      endTime?: string;
      /**
       * The step size, in minutes, used for generating the time options.
       * @default 30
       */
      stepSize?: number;
    }
```

packages/sit-onyx/src/components/OnyxTimepicker/types.ts
```diff
@@ -1,8 +1,42 @@
...
+    /**
+     * Configuration options for the timepicker's option generation.
+     * This property only takes effect when the `type` is set to 'select'.
+     */
+    options?: {
+      /**
+       * The starting time for generating time options (using 24-hour format).
+       * @format HH:MM (e.g., "08:00")
+       */
+      startTime?: string;
```
larsrickert:
```suggestion
      startTime?: `${number}:${number}`;
```

To narrow down the allowed string type, we can type it like this so the user gets a type error if he tries passing in an invalid string :) We could also create a new type for this, e.g. 
```ts
export type TimeString = `${number}:${number}`;
```

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxTimepicker/OnyxTimepicker.stories.ts
```diff
@@ -35,3 +36,66 @@ export const InfoLabel = {
...
+          disabled: true,
+        },
+        {
+          value: "11:00",
+          label: "11:00",
+          group: "Morning Slots (09:00 - 11:30)",
+        },
+        {
+          value: "12:30",
+          label: "12:30 PM – Lunch Break",
```
larsrickert:
```suggestion
          label: "12:30 PM – Lunch Break",
          group: "Noon slots",
```

Personally, I think the lunch break option looks a little lost, what do you think? We could either also add a group to it or remove it completely

packages/sit-onyx/src/components/OnyxTimepicker/OnyxTimeSelect.vue
```diff
@@ -0,0 +1,135 @@
...
+    </template>
+  </OnyxSelect>
+</template>
+
+<style lang="scss">
+@use "../../styles/mixins/layers.scss";
+
+.onyx-timepicker {
+  @include layers.component() {
+    .onyx-icon {
```
larsrickert:
```suggestion
    &__ {
```

Would prefer to use a more specific timepicker related selector since overriding all icons might lead to unintentional styles in nested elements

packages/sit-onyx/src/components/OnyxTimepicker/types.ts
```diff
@@ -1,23 +1,73 @@
...
+export type TimepickerSelectOptions = {
+  /**
+   * The step size, in seconds, used for generating the time options.
+   * @default 1800 // 30 min
+   */
+  stepSize?: number;
+  /**
+   * A function that provides custom time options.
+   */
+  customTimes?: () => SelectOption<string>[];
```
larsrickert: Is there a reason why this is a function instead of just passing an array?

packages/sit-onyx/src/components/OnyxTimepicker/types.ts
```diff
@@ -1,23 +1,73 @@
...
+    /**
+     * Time in Seconds since midnight.
+     */
+    modelValue?: string;
+
+    /**
+     * Minimum time to input in 24-hour format (including the minimum time).
+     * @format HH:MM:SS (e.g., "08:00:00")
+     */
+    min?: `${number}:${number}`;
```
larsrickert: This type does not allow passing seconds 🤔  If we want to use this type it should propably be:
```ts
export type TimeString = `${number}:${number}` | `${number}:${number}:${number}`
```

packages/sit-onyx/src/components/OnyxTimepicker/types.ts
```diff
@@ -1,23 +1,73 @@
...
+     * Maximum time to input in 24-hour format (including the maximum time).
+     * @format HH:MM:ss (e.g., "17:30:00")
+     */
+    max?: `${number}:${number}`;
+
     /**
      * Whether to show the seconds segment (:SS).
      * If true, the format is HH:MM:SS. If false, the format is HH:MM.
      */
     showSeconds?: boolean;
```
larsrickert, line 57: I am not sure if this API is flexible enough for the timepicker. Currently the timepicker is limited to always include hours and minutes (only seconds are optional). But there might also be use cases where only mm:ss should be selected etc.

The timepicker is unstable/experimental anyways at the beginning so its also fine for me to merge it like this for now and then check this later :) But we should create a follow up ticket for it

packages/sit-onyx/src/components/OnyxTimepicker/OnyxTimeInput.vue
```diff
@@ -0,0 +1,394 @@
...
+                hide-clear-icon
+                hide-success-icon
+                hide-buttons
+                @click="handleSegmentFocus(secondInputRef)"
+                @keydown="(e: KeyboardEvent) => handleInputChange('second', e)"
+              />
+            </template>
+          </div>
+
+          <div v-if="props.infoLabel" class="onyx-timepicker__info-label">
```
larsrickert, line 315: The info label looks pretty huge, is this correct or should it be smaller?

Image

packages/sit-onyx/src/components/OnyxSelectInput/OnyxSelectInput.vue
```diff
@@ -207,9 +220,22 @@ useAutofocus(input, props);
             tabindex="-1"
             :disabled="disabled || props.readonly || props.loading"
           >
-            <OnyxIcon :icon="iconChevronDownUp" />
+            <slot name="icon">
+              <OnyxIcon :icon="iconChevronDownUp" />
+            </slot>
+          </button>
+          <button
```
larsrickert: Always including the clear button for the select will break some current components :/ This might even be a breaking change to include it by default. Personally, I'd prefer to implement it in a separate standalone PR instead of putting it in here since its not really related to the timepicker itself.

This would currently break:
- version select in playground
- grid playground
- select pagination

If its already a breaking change for us, it will most likely also be one for all users.

packages/sit-onyx/src/components/OnyxTimepicker/types.ts
ChristianBusshoff: currently we implemented the timepicker with the html default input, which can't do mm:ss

PR #4561, opened 2025-12-17
feat(OnyxItemsPerPage): remove search and improve responsiveness

larsrickert:
Relates to #4404

- remove search (UX feedback) to keep the component simple. It should include only a few options anyway
- add max-width so the component truncates correctly
- when used inside data grid: hide label on small breakpoints (UX feedback)
- update / clean up some Playwright tests
- OnyxMiniSearch and OnyxPagination: ensure correct placeholder min-width in webkit/safari

## Checklist

### DIFF
--- a/packages/sit-onyx/src/components/OnyxDataGrid/features/pagination/pagination.scss
+++ b/packages/sit-onyx/src/components/OnyxDataGrid/features/pagination/pagination.scss
@@ -13,18 +13,7 @@
       }
     }
 
-    // needed to support the container query below
-    &:has(&__items-per-page) {
-      .onyx-table-wrapper__bottom {
-        container-type: inline-size;
-      }
-    }
-
     &__items-per-page {
-      // this element is just a wrapper div that we need for scoping the styles to the class
-      // so we set it to display contents to not have any additional layout
-      display: "contents";
-
       @include breakpoints.container(max, xs) {
         .onyx-items-per-page__label {
           display: none;
--- a/packages/sit-onyx/src/components/OnyxDataGrid/features/pagination/pagination.ts
+++ b/packages/sit-onyx/src/components/OnyxDataGrid/features/pagination/pagination.ts
@@ -145,20 +145,17 @@ export const usePagination = (options: PaginationOptions = {}) =>
           if (!shouldShowPagination.value && !skeleton.value) return [];
 
           return [
-            h(
-              "div",
-              { class: "onyx-data-grid__items-per-page" },
-              h(OnyxItemsPerPage, {
-                modelValue: state.value.pageSize,
-                options: options.itemsPerPage,
-                disabled: isDisabled.value,
-                skeleton: skeleton.value,
-                "onUpdate:modelValue": (newPageSize: number) => {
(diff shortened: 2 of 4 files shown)


packages/sit-onyx/src/styles/mixins/text.scss
```diff
@@ -0,0 +1,8 @@
+/**
+* Function for getting a calculated CSS "ch" unit value that can be used to e.g. set the width of an element based on the text length.
+* Includes a bug fix for webkit/safari so the text is not truncated.
+*/
+@function ch($length) {
+  // the 1px offset is needed to fix a bug in webkit/safari where the text is truncated
+  @return calc($length * 1ch + var(--onyx-1px-in-rem));
+}
```
JoCa96, line 7: Imho this is not a bug but a quirk of the definition of the `ch` unit: It's an size approximation based on the `0` character of the current font, but browsers can always decide to just use `0.5em` instead. So for (not mono-)fonts it will always be kind of inaccurate.

```suggestion
/**
* Function for getting a approximated CSS "ch" length value that can be used to e.g. set the width of an element based on the text length.
* Adds 1px of additional space to ensure no overflow for low character counts.
*/
@function ch($length) {
  @return calc($length * 1ch + var(--onyx-1px-in-rem));
}
```
(For an exact solution we would need to use the trick that we use with the textarea.)

packages/sit-onyx/src/components/OnyxDataGrid/features/pagination/pagination.ts
```diff
@@ -145,16 +145,20 @@ export const usePagination = (options: PaginationOptions = {}) =>
...
+                modelValue: state.value.pageSize,
+                options: options.itemsPerPage,
+                disabled: isDisabled.value,
+                skeleton: skeleton.value,
+                "onUpdate:modelValue": (newPageSize: number) => {
+                  state.value.pageSize = newPageSize;
+                  state.value.current = 1;
+                },
+              }),
+            ),
```
JoCa96: Is the wrapper really necessary?
We could put the class directly on the OnyxItemsPerPage props or reference the `onyx-items-per-page` class directly:
```suggestion
              h(OnyxItemsPerPage, {
                class: "onyx-data-grid__items-per-page",
                modelValue: state.value.pageSize,
                options: options.itemsPerPage,
                disabled: isDisabled.value,
                skeleton: skeleton.value,
                "onUpdate:modelValue": (newPageSize: number) => {
                  state.value.pageSize = newPageSize;
                  state.value.current = 1;
                },
              }),
```

packages/sit-onyx/src/components/OnyxDataGrid/features/pagination/pagination.scss
```diff
@@ -11,5 +12,24 @@
         margin-inline: auto;
       }
     }
+
+    // needed to support the container query below
+    &:has(&__items-per-page) {
+      .onyx-table-wrapper__bottom {
+        container-type: inline-size;
+      }
+    }
```
JoCa96: Would prefer to always set the `container-type`, not just for this feature.

Issue #4457, opened 2025-11-18
Implement stable OnyxSlider


larsrickert:
### Why?

Follow up ticket for the slider to implement some minor slider improvements.

### Acceptance criteria

- [x] remove property `disableTooltip` in favor of `tooltip.hidden`
- [x] new property `tooltip.formatter?: (value: number, index: number) => string`, which allows to customize the value string inside the tooltip
- [x] when the "icon" has `:focus` the tooltip with the value should be shown
- [x] if there is a scala/marks below the tooltip should be shown above instead
- [ ] ~~if possible and easy: remove the required/optional dropdown~~
- [x] Replace Storybook tag `unstable` with `new`

### DEV handover checklist

[Storybook](https://onyx-storybook-dev.apps.01.cf.eu01.stackit.cloud/?path=/docs/form-elements-slider--docs)

DEV handover done with: @JoCa96 

**Testing**
- [x] covered by visual screenshot tests
- [x] covered by functional tests (Playwright or unit test)

**Documentation**
- [x] "New" badge in Storybook is added, see [storybook-addon-tag-badges](https://github.com/Sidnioulz/storybook-addon-tag-badges/?tab=readme-ov-file#-usage)
- [x] auto-generated Storybook code snippets are checked
- [x] component status has been updated in the [docs overview](https://onyx.schwarz/#components)
- [x] updated version is released
- [x] updated documentation is deployed

### UX handover checklist

UX handover done with: @jannick-ux 

- [x] changes of this ticket approved by UX

PR #4579, opened 2025-12-22
feat(OnyxSlider): stabilize component API

larsrickert:
Relates to #4457

`OnyxUnstableSlider` is renamed to `OnyxSlider` and is now considered to have a stable API.
Therefore, no breaking changes will be introduced from now on within major versions. The `OnyxUnstableSliderControl` component has been removed and integrated into the OnyxSlider directly.

Changes to the last unstable version:

- when `control="icon"` is set, the icon buttons will no longer be focusable via keyboard because its redundant (slider can already be changed with arrow keys). When clicking the icon buttons, the tooltip will now be shown automatically
- when labelled marks are used, the tooltip will now be positioned top
- remove `disableTooltip` property in favor of new `tooltip` property that also allows customizing the tooltip value
- removed `errorMessages` and `successMessages` property which were inconsistent with other form components. Use `error` and `success` instead
- fix: error message not shown when slider is touched

## Checklist

### DIFF
--- a/packages/headless/src/composables/slider/createSlider.ts
+++ b/packages/headless/src/composables/slider/createSlider.ts
@@ -105,8 +105,8 @@ export const createSlider = createBuilder(
     });
 
     const focusThumb = (index: number) => {
-      if (index === 0) firstThumbRef.value.focus();
-      else if (index === 1) secondThumbRef.value.focus();
+      if (index === 0) firstThumbRef.value?.focus();
+      else if (index === 1) secondThumbRef.value?.focus();
     };
 
     const shiftStep = computed(() => {


packages/sit-onyx/src/components/OnyxSlider/OnyxSlider.vue
```diff
@@ -82,6 +74,13 @@ const {
   shiftStep: toRef(props, "shiftStep"),
   onChange: (newValue) => (modelValue.value = newValue),
 });
+
+const hasMarkLabels = computed(() => marks.value.some((mark) => !!mark.label));
+
+const handleControlUpdate = (value: number, index: number, focus?: boolean) => {
+  updateValue(value, index);
+  if (focus) focusThumb(index);
```
JoCa96: Unfortunately moving the focus is a bad practice, because it is disruptive to the user experience. E.g. when using the keyboard and focus is moved after triggering the "+" button: I have to reorient myself and suddenly need to use a different key to change the value.

larsrickert: This is not trivial since the tooltip relies on focusin and focusout events to correctly show/hide itself. Even when manually forcing the tooltip open state after pressing the icon control via keyboard to open the tooltip, the tooltip is not hidden when e.g. clicking outside because it is not focused and does not receive the focusout event.

My suggestion here would be (to keep things simple) to make the icon buttons not focusable via keyboard. The slider itself already has keyboard support to adjust the value with arrow left/right so additionally focusing the icon buttons via keyboard does not make sense to me. They are mostly useful for mouse users imo. What do you think?

larsrickert: As discussed, I added `tabindex="-1"` to the icon buttons so they are not focusable via keyboard. This is redudant since the slider itself can already be controlled using the keyboard

packages/headless/src/composables/slider/createSlider.ts
JoCa96, line 1: There are currently 3 different methods that can handle the modelValue (step) increase for the OnyxSlider:
- [OnyxSliderControl.vue => handleIconClick](https://github.com/SchwarzIT/onyx/blob/fb5640bbe934ee326de5afdaaff8328f0a16942f/packages/sit-onyx/src/components/OnyxSliderControl/OnyxSliderControl.vue#L30)
- [createSlider.ts => shiftStep](https://github.com/SchwarzIT/onyx/blob/fb5640bbe934ee326de5afdaaff8328f0a16942f/packages/headless/src/composables/slider/createSlider.ts#L111)
- [OnyxSliderControl => OnyxStepper](https://github.com/SchwarzIT/onyx/blob/1b810bb1d9fbaf1043c6dec6138ea804bcf56503/packages/sit-onyx/src/components/OnyxSliderControl/OnyxSliderControl.vue#L82)

These should be unified to reduce edge-cases. Imho it is probably fine to leave the `OnyxStepper` logic out - but `handleIconClick` and `shiftStep` should not both individually handle the step logic.

packages/sit-onyx/src/components/OnyxSlider/OnyxSlider.vue
JoCa96, line 175: The trigger is not bound to a valid element:
```vue
 <OnyxVisuallyHidden is='input'>
    :id="index === 0 ? inputId : undefined"
    v-custom-validity
    :class="[
      'onyx-slider__native',
      { 'onyx-slider__native--touched': wasTouched },
    ]"
    :tabindex="props.control === 'input' ? -1 : undefined"
    v-bind="mergeVueProps(thumbInput({ value, index }), trigger)"
    :disabled="disabled"
```

packages/headless/src/composables/slider/createSlider.ts
```diff
@@ -102,10 +99,14 @@ export const _unstableCreateSlider = createBuilder(
...
+      focusThumb(newThumbIndex);
+    });
+
+    const focusThumb = (index: number) => {
       Array.from(sliderRef.value.querySelectorAll<HTMLElement>('[role="slider"]'))
-        .at(newThumbIndex)
+        .at(index)
         ?.focus();
-    });
+    };
```
JoCa96, line 110: using a ref here (applied by the element beneath) would be preferable over a `querySelectorAll`

packages/sit-onyx/src/components/OnyxSlider/OnyxSlider.vue
```diff
@@ -153,9 +165,9 @@ const {
               class="onyx-slider__thumb"
```
JoCa96, line 185: (Optional): Imho adding a simple transition would make dragging the slider feel a lot smoother :)

packages/sit-onyx/src/components/OnyxSlider/OnyxSlider.vue
JoCa96: There is still an accessibility issue: While `tabindex="-1"` moves it out of the the tab-order, it still allows an element to be focused. So in this case the issue of the moving focus still persists. 

My suggestion would be to handle the tooltips `open` (=> the opening and closing) prop inside of the `OnyxSlider` component. E.g. adding an open `ref` and managing it here.

`tabindex="-1"` is still a good idea as it removes redundant tabs for the user.

larsrickert: I created #4619 as a follow up so this PR is not blocked by this. As discussed, we will use `tabindex="-1"` for now and check the accessibility in #4619

packages/sit-onyx/src/components/OnyxSlider/OnyxSlider.vue
larsrickert: I like the idea :) But implementing the transition is not so trivial since it needs to work for both single and range slider so just adding a transition to the width of the track and the left position of the thumb does not work.

I created a follow up ticket: #4620

Issue #2778, opened 2025-02-18
Implement keyboard support for OnyxDataGrid


JoCa96:
### Why?

As a user I want to be able to use the table with the keyboard.

### Design

[Figma](https://www.figma.com/design/YfEUBOHk4J4nYrk04geswG/Onyx-Component-Library?m=auto&node-id=8936-76014)

![Image](https://github.com/user-attachments/assets/ccfc2dae-e70a-4013-962f-8c56974518eb)

### Implementation details

<-- Add technical details / implementation proposals if necessary, e.g.:
  - API proposal
  - relevant web APIs
  - learnings from other implementations
  - code snippets
  -->

### Definition of Done

- The following component requirements are implemented:
  - [ ] skeleton
  - [ ] density
  - [ ] example usage added to `apps/demo-app/src/views/HomeView.vue`

- Should be covered by tests:
  - [ ] functional tests (Playwright or unit test)
  - [ ] visual tests (Playwright screenshots)

- Make sure, that
  - [ ] follow-up tickets were created if necessary
  - [ ] updated version + documentation is deployed
  - [ ] Storybook can show the feature
  - [ ] Storybook code snippet of new/changed examples are checked that they are generated correctly
  - [ ] Namings are aligned with Figma

#### Approval

[Storybook]()
[Documentation]()

- Checked and approved by
  - [ ] designer 
  - [ ] dev

PR #4584, opened 2025-12-22
feat(createDataGrid): basic implementation of the headless data grid composable

JoCa96:
Relates to #2778

### DIFF
--- a/packages/headless/src/composables/dataGrid/createDataGrid.testing.ts
+++ b/packages/headless/src/composables/dataGrid/createDataGrid.testing.ts
@@ -31,7 +31,7 @@ const expectLabel = async (page: Page, grid: Locator) => {
 
 const expectMaybeDescription = async (page: Page, grid: Locator) => {
   const describedBy = await grid.getAttribute("aria-describedby");
-  const caption = await grid.getByRole("caption");
+  const caption = grid.getByRole("caption");
 
   if (describedBy) {
     expect(
--- a/packages/headless/src/composables/dataGrid/createDataGrid.ts
+++ b/packages/headless/src/composables/dataGrid/createDataGrid.ts
@@ -226,11 +226,10 @@ export const createDataGrid = createBuilder(
     };
 
     const onKeydown = (event: KeyboardEvent) => {
-      const cellElement: HTMLTableCellElement | null = (event.target as HTMLElement).closest(
-        "td, th",
-      );
-      const rowElement = (event.target as HTMLElement).closest("tr");
-      const tableElement = (event.target as HTMLElement).closest("table");
+      const target = event.target as HTMLElement;
+      const cellElement: HTMLTableCellElement | null = target.closest("td, th");
+      const rowElement = target.closest("tr");
+      const tableElement = target.closest("table");
 
       if (!cellElement || !rowElement || !tableElement) {
         return;


packages/headless/tsconfig.json
```diff
@@ -4,7 +4,7 @@
   "compilerOptions": {
     "rootDir": "./src",
     "baseUrl": ".",
-    "lib": ["DOM", "DOM.Iterable", "ES2022"],
+    "lib": ["DOM", "DOM.Iterable", "ES2024"],
```
larsrickert, line 7: Why is this needed? Doesn't we force all using projects to add this to their tsconfig as well?

packages/headless/src/composables/dataGrid/createDataGrid.ts
```diff
@@ -0,0 +1,329 @@
...
+        /**
+         * total number of rows that are viewable, this includes rows that are hidden because they are out of view
+         */
+        totalRows: number | "unknown";
+        /**
+         * total number of columns that are viewable, this includes columns that are hidden because they are out of view
+         */
+        totalCols: number | "unknown";
+
+        requestLazyLoad: RequestLazyLoad;
```
larsrickert, line 108: Is it fine to put a function inside of a ref?

packages/headless/src/composables/dataGrid/createDataGrid.ts
```diff
@@ -0,0 +1,329 @@
...
+    };
+
+    let mutationObserver: MutationObserver;
+
+    onMounted(() => {
+      ensureTabTarget();
+
+      mutationObserver = new MutationObserver(ensureTabTarget);
+
+      watch(
```
larsrickert, line 201: Why is this needed?

packages/headless/src/composables/dataGrid/createDataGrid.ts
```diff
@@ -0,0 +1,329 @@
...
+      setSelected(event.target as HTMLElement);
+      focusQueue.cancel();
+    };
+
+    const onKeydown = (event: KeyboardEvent) => {
+      const cellElement: HTMLTableCellElement | null = (event.target as HTMLElement).closest(
+        "td, th",
+      );
+      const rowElement = (event.target as HTMLElement).closest("tr");
+      const tableElement = (event.target as HTMLElement).closest("table");
```
larsrickert:
```suggestion
      const target = event.target as HTMLElement;
      const cellElement: HTMLTableCellElement | null = target.closest("td, th");
      const rowElement = target.closest("tr");
      const tableElement = target.closest("table");
```

packages/headless/src/composables/dataGrid/createDataGrid.testing.ts
```diff
@@ -0,0 +1,170 @@
...
+
+  if (labelledBy) {
+    expect(
+      label,
+      `"aria-labelledby" and "aria-label" must not both be defined on the grid element!`,
+    ).toBeFalsy();
+    expect(
+      page.locator(`#${labelledBy}`),
+      `"aria-labelledby" must point to an existing element!`,
+    ).toBeAttached();
```
larsrickert, line 19:
```suggestion
    ).toBeVisible();
```

Shouldn't it also be visible to the user?

packages/headless/src/composables/dataGrid/createDataGrid.testing.ts
```diff
@@ -0,0 +1,170 @@
...
+const expectNthCellToBeFocused = async (
+  { page, nth, row }: ExpectNthCellToBeFocusedOptions,
+  message?: string,
+) =>
+  expect(async () => {
+    const focused = page.locator("*:focus");
+    const cells = row.getByRole("cell");
+    const targetHandle = await cells.nth(nth).elementHandle();
+    expect(await focused.evaluate((a, b) => a === b, targetHandle), message).toEqual(true);
+  }).toPass();
```
larsrickert:
```suggestion
const expectNthCellToBeFocused = (
  { nth, row }: ExpectNthCellToBeFocusedOptions,
  message?: string,
) => {
  const cells = row.getByRole("cell");
  return expect(cells.nth(nth), message).toBeFocused();
};
```

Isn't this the same but simpler?

larsrickert (changes requested):

packages/headless/src/composables/dataGrid/createDataGrid.testing.ts
JoCa96: Not necessarily. From an ARIA perspective it just needs to exist, but doesn't have to visible.

packages/headless/src/composables/dataGrid/createDataGrid.testing.ts
JoCa96: looks good :)

packages/headless/src/composables/dataGrid/createDataGrid.ts
JoCa96: to setup the mutation observer and update if the tableElement is re-rendered

packages/headless/src/composables/dataGrid/createDataGrid.ts
JoCa96: I don't see a reason why not. Do you have a specific issue in mind?

packages/headless/tsconfig.json
JoCa96: I want to use `Promise.withResolvers`, which is Baseline since March 2024. So in accordance with our `.browserlistrc`, this is fine.

packages/headless/src/composables/dataGrid/createDataGrid.ts
larsrickert: Nope, was just curious

packages/headless/tsconfig.json
larsrickert: Yes thats fine with out browserlist but doesn't it force all projects to also change this anyways?

PR #4595, opened 2025-12-29
feat(useMatrixScreenshotTest): support passing a custom `test` function

larsrickert:
Feature request from trinity. To easily support custom test fixtures (e.g. to globally mock images), we now offer an option to pass in a custom `test` function.

## Checklist

### DIFF
--- a/packages/playwright-utils/src/screenshots/index.tsx
+++ b/packages/playwright-utils/src/screenshots/index.tsx
@@ -14,7 +14,7 @@ import { escapeGridAreaName } from "./utils.js";
  * Useful for capturing a single screenshot/image that contains multiple variants of a component.
  */
 export const useMatrixScreenshotTest = <TContext extends HookContext = HookContext>(
-  globalOptions: UseMatrixScreenshotTestOptions<TContext>,
+  globalOptions: UseMatrixScreenshotTestOptions<TContext> = {},
 ) => {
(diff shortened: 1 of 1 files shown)


packages/playwright-utils/src/screenshots/index.tsx
```diff
@@ -13,9 +13,11 @@ import { escapeGridAreaName } from "./utils.js";
  * Creates a screenshot utility that can be used to capture matrix screenshots.
  * Useful for capturing a single screenshot/image that contains multiple variants of a component.
  */
-export const useMatrixScreenshotTest = <TContext extends HookContext = HookContext>({
-  defaults,
-}: UseMatrixScreenshotTestOptions<TContext>) => {
+export const useMatrixScreenshotTest = <TContext extends HookContext = HookContext>(
+  globalOptions: UseMatrixScreenshotTestOptions<TContext>,
```
BoppLi: Is globalOptions required now? That would be a breaking change if no parameter was needed before and you now have to call `useMatrixScreenshotTest({})`, right?

BoppLi (changes requested):

packages/playwright-utils/src/screenshots/index.tsx
BoppLi: nevermind, was required before :)

Issue #4444, opened 2025-11-12
Implement truncation / max. results for global search


larsrickert:
### Why?

Follow up for #4275. If there are a lot of options / search results for the global search, it should be possible to show a "Show all results" button that either displays all options or e.g. opens a search result page.

### Design

[Figma](https://www.figma.com/design/j7Hy7Zk8MiCmJBJf0ArQRm/Onyx-Documentation?node-id=16438-24376&m=dev)

### Acceptance criteria

- [x] custom content can be displayed at the bottom of the search. The content is NOT sticky and scrolls with the content
- [x] the custom content integrates correctly into the aria pattern of the component (the Playwright a11y checks succeed)
- [x] an example is created that showcases a "Show all results" button if there are more than 10 options

### DEV handover checklist

[Storybook](https://onyx-storybook-dev.apps.01.cf.eu01.stackit.cloud/?path=/story/search-filter-globalsearch--end-of-list-example)

DEV handover done with: @larsrickert 

**Testing**
- [x] covered by visual screenshot tests

**Documentation**
- [x] updated version is released
- [x] updated documentation is deployed

### UX handover checklist

UX handover done with: @Guergchen 

**Implementation is checked**
- [x] all features of Figma are implemented or a follow-up ticket is refined
- [x] responsiveness / wrapping behavior

PR #4603, opened 2026-01-05
feat(GlobalSearch): implemented end of List slot + example

ChristianBusshoff:
- added end of List slot + show all Results example 

Relates to #4444

### DIFF
--- a/packages/sit-onyx/src/components/OnyxGlobalSearch/OnyxGlobalSearch.vue
+++ b/packages/sit-onyx/src/components/OnyxGlobalSearch/OnyxGlobalSearch.vue
@@ -131,7 +131,6 @@ const getOptionProps = (value: string) => {
     role: "option",
     id: headless.internals.getOptionId(value),
     "aria-selected": isActive,
-    class: isActive ? "active" : undefined,
   };
 };
 
@@ -170,7 +169,7 @@ provide(GLOBAL_SEARCH_INJECTION_KEY, { headless, activeValue });
     >
       <slot></slot>
 
-      <div v-if="!!slots.endOfList" role="none" class="onyx-global-search__end-of-list">
+      <div v-if="!!slots.endOfList" class="onyx-global-search__end-of-list">
         <slot
           name="endOfList"
           :get-option-props="getOptionProps"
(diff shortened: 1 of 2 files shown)


packages/sit-onyx/src/components/OnyxGlobalSearch/OnyxGlobalSearch.vue
```diff
@@ -140,8 +169,15 @@ provide(GLOBAL_SEARCH_INJECTION_KEY, { headless, activeValue });
       v-bind="headless.elements.listbox.value"
     >
       <slot></slot>
-    </div>
 
+      <div v-if="!!slots.endOfList" role="none" class="onyx-global-search__end-of-list">
```
JoCa96: `role="none"` should not be necessary here
```suggestion
      <div v-if="!!slots.endOfList" class="onyx-global-search__end-of-list">
```

packages/sit-onyx/src/components/OnyxGlobalSearch/OnyxGlobalSearch.vue
```diff
@@ -106,6 +118,23 @@ const headless = createComboBox({
...
+ * @returns An object containing the required ARIA roles, IDs, and the current active state.
+ */
+const getOptionProps = (value: string) => {
+  const isActive = activeValue.value === value;
+
+  return {
+    role: "option",
+    id: headless.internals.getOptionId(value),
+    "aria-selected": isActive,
+    class: isActive ? "active" : undefined,
```
JoCa96: We don't need the `active` class ourselves, so we should not provide it.
Instead the application developer should set it.

```suggestion
```

packages/sit-onyx/src/components/OnyxGlobalSearch/examples/EndOfListExample.vue
```diff
@@ -0,0 +1,211 @@
...
+.show-all-button {
+  width: 100%;
+
+  /**
+   * The .active class is applied by getOptionProps when the user 
+   * navigates to this button using the arrow keys.
+   * Since the browser focus stays in the input field, we must 
+   * visually simulate the focus state here.
+   */
+  &.active {
```
JoCa96:
```suggestion
  /**
   * The `aria-selected` attribute is applied by getOptionProps when the user 
   * navigates to this button using the arrow keys.
   * Since the browser focus stays in the input field, we must 
   * visually simulate the focus state here.
   */
  &[aria-selected="true"] {
```

PR #4604, opened 2026-01-05
feat: support clickable property for OnyxBadge and implement new OnxyFilterBadge component

ChristianBusshoff:
Relates to #4512

support clickable property for OnyxBadge and implement new OnxyFilterBadge component

### DIFF
--- a/.changeset/fair-taxes-push.md
+++ b/.changeset/fair-taxes-push.md
@@ -2,7 +2,7 @@
 "sit-onyx": minor
 ---
 
-feat(OnyxBadge): add `clickable` and `active` property to support interactive badges
+feat(OnyxBadge): add `clickable` property to support interactive badges
 feat: implemented new OnyxUnstableFilterBadge component
 
 For now, the OnyxUnstableFilterBadge is marked as experimental/unstable which means that it's still under active development and the API might change in patch or minor releases. Keep an eye on the [changelog](https://onyx.schwarz/development/packages/changelogs/sit-onyx.html) when using them.


packages/sit-onyx/src/components/OnyxFilterTag/types.ts
```diff
@@ -6,4 +6,10 @@ export type OnyxFilterTagProps = Omit<OnyxTagProps, "color" | "clickable"> & {
    * If `true` the filter is selected, shows an 'x' icon and can be removed on click.
    */
   active?: Nullable<boolean>;
+  /**
+   * An icon displayed on the right side of the label.
+   * Should be used to indicate an action (e.g., dismissing).
+   * @default iconXSmall
+   */
+  actionIcon?: Nullable<string>;
```
JoCa96: Imho this is a opinionated variant of the OnyxTag (it is designed for a specific usage/context) we should not allow this customization here. This way we can also ensure a consistent design language over multiple applications.

JoCa96 (changes requested):

.changeset/fair-taxes-push.md
```diff
@@ -0,0 +1,9 @@
+#---
+"sit-onyx": minor
+
+---
```
larsrickert:
```suggestion
---
"sit-onyx": minor
---
```

Something looks broke here :)

packages/sit-onyx/src/components/OnyxBadge/OnyxBadge.ct.tsx
```diff
@@ -75,3 +75,33 @@ test("should truncate text", async ({ mount }) => {
   // ASSERT
   await expect(component).toHaveScreenshot("truncation-ellipsis.png");
 });
+
+test("should render interactive Badge with clickable prop", async ({ mount }) => {
+  const component = await mount(<OnyxBadge clickable="clickable">Badge</OnyxBadge>);
+  const interactiveTag = component.getByRole("button", { name: "Badge" });
+  await expect(interactiveTag).toContainClass("onyx-badge--interactive");
```
larsrickert: Would prefer to test if its actually a (role) button / clickable

packages/sit-onyx/src/components/OnyxBadge/OnyxBadge.ct.tsx
```diff
@@ -75,3 +75,33 @@ test("should truncate text", async ({ mount }) => {
...
+  mount,
+}) => {
+  const component = await mount(
+    <OnyxBadge clickable="clickable" selected>
+      Badge
+    </OnyxBadge>,
+  );
+  const badge = component.getByRole("button", { name: "Badge" });
+
+  await expect(badge).toContainClass("onyx-badge--selected");
```
larsrickert: Screenshot test would be great :) Also for states like hover, focus etc.

packages/sit-onyx/src/components/OnyxBadge/OnyxBadge.stories.ts
```diff
@@ -65,3 +66,36 @@ export const WithTruncation = {
     style: "max-width: 16rem",
   },
 } satisfies Story;
+
+/**
+ * This example shows the interactive tag.
+ */
+export const Interactive = {
```
larsrickert: As discussed, we would prefer to merge all these 3 new examples into one "Clickable" example. The same for the OnyxTag

packages/sit-onyx/src/components/OnyxBadge/OnyxBadge.vue
```diff
@@ -78,24 +118,63 @@ defineSlots<{
       }
     }
 
+    &--interactive {
```
larsrickert: Would prefer `:is(button)` so we can remove the custom class and bind the styles more to what the element actually is :)

packages/sit-onyx/src/components/OnyxBadge/OnyxBadge.vue
```diff
@@ -19,16 +45,24 @@ defineSlots<{
...
-  >
+  <OnyxTooltip v-if="props.clickable" :text="tooltipText">
+    <template #default="{ trigger }">
+      <button v-bind="trigger" :class="badgeClasses" type="button" :aria-pressed="props.selected">
+        <template v-if="!props.dot">
+          <OnyxIcon v-if="props.icon" class="onyx-badge__icon" :icon="props.icon" />
+          <slot v-else></slot>
+
+          <OnyxIcon
+            v-if="typeof props.clickable === 'object' && props.clickable.actionIcon"
```
larsrickert: If we replace the tooltipText computed with this clickable computed, we can use it in all v-ifs here and don't need to check if its an object here because we already normalize it into and object:

```ts
const clickable = computed(() => {
  if (!props.clickable) return;
  if (typeof props.clickable === "object") return props.clickable;
  return { label: props.clickable };
});
```

packages/sit-onyx/src/components/OnyxBadge/types.ts
```diff
@@ -15,4 +15,29 @@ export type OnyxBadgeProps = DensityProp & {
...
+        label: string;
+        /**
+         * An icon displayed on the right side of the label.
+         * Should be used to indicate an action (e.g., dismissing).
+         */
+        actionIcon?: string;
+      };
+  /**
+   * Whether the badge should visually marked as selected
+   * If the badge is also clickable, it will automatically handle aria-pressed.
```
larsrickert:
```suggestion
```

I am not sure if this information is relevant for the user

packages/sit-onyx/src/components/OnyxBadge/types.ts
```diff
@@ -15,4 +15,29 @@ export type OnyxBadgeProps = DensityProp & {
...
+   */
+  selected?: boolean;
```
larsrickert: Is it also possible to have a selected badge without being clickable? If not, we can move the selected prop into the clickable prop to ensure that it must be clickable if wanting to set it to selected

packages/sit-onyx/src/components/OnyxFilterBadge/OnyxFilterBadge.ct.tsx
```diff
@@ -0,0 +1,31 @@
+import { DENSITIES } from "../../composables/density.js";
+import { test } from "../../playwright/a11y.js";
+import { executeMatrixScreenshotTest } from "../../playwright/screenshots.jsx";
+import OnyxFilterBadge from "./OnyxFilterBadge.vue";
+
+test.describe("Screenshot tests", () => {
+  const state = ["default", "hover", "focus", "active"];
+  executeMatrixScreenshotTest({
+    name: "Badge",
```
larsrickert:
```suggestion
    name: "Filter badge",
```

packages/sit-onyx/src/components/OnyxFilterBadge/OnyxFilterBadge.ct.tsx
```diff
@@ -0,0 +1,31 @@
...
+        active={row === "active"}
+        style={{ margin: "0 3rem 2rem 0" }}
+      />
+    ),
+    hooks: {
+      beforeEach: async (component, _page, _column, row) => {
+        const badge = component.getByRole("button", { name: "Badge" });
+        if (row === "hover") {
+          await badge.hover();
+          await new Promise((resolve) => setTimeout(resolve, 200));
```
larsrickert: Why is the timeout needed? Timeouts can be very flaky so we should avoid them

packages/sit-onyx/src/components/OnyxFilterBadge/OnyxFilterBadge.vue
```diff
@@ -0,0 +1,53 @@
...
+const { t } = injectI18n();
+const badgeProps = useForwardProps(props, OnyxBadge);
+
+const active = useVModel({
+  props,
+  emit,
+  key: "active",
+  default: false,
+});
+const tooltipLabel = computed(() =>
```
larsrickert, line 36: Imo the labels for both the OnyxFilterTag and OnyxFilterBadge are little weird because "Add Tag filter" and "Add Badge filter" seem confusing to me 🤔  Would prefer to make them more generic like "Add filter" and "Remove filter" since it should be irrelevant to the user whether its a tag or badge

larsrickert (approved):

Issue #4480, opened 2025-11-21
Implement year and month selection for OnyxCalendar


larsrickert:
### Why?

The OnyxCalendar should support changing the year and month.

### Design

[Figma](https://www.figma.com/design/YfEUBOHk4J4nYrk04geswG/Onyx-Components?node-id=19389-83696&m=dev)
[Documentation](https://www.figma.com/design/j7Hy7Zk8MiCmJBJf0ArQRm/Onyx-Documentation?node-id=19754-30544&m=dev)

### Acceptance criteria

- when clicking on the current month/year in the calendar header a flyout is opened (for the large calendar) or the view is changed in-place (for the small calendar) that allows to:
    - [ ] first select a year
    - [ ] after a year is selected, a month can be selected
    - [ ] after the month is selected, the view is switched back to the regular calendar view and the selected month and year is selected as viewMonth
- [ ] the month names are translated using JavaScript's Intl API
- [ ] the current year / month is highlighted visually
- [ ] when clicking the header area again or clicking outside, the view reverts back to the regular view
- [ ] the actual selected date(s) are NOT changed, this feature just changes the viewed month/year

### DEV handover checklist

[Storybook](ADD_LINK_HERE)

DEV handover done with: @ADD_DEV_NAME_HERE

**Testing**
- [ ] covered by visual screenshot tests
- [ ] covered by functional tests (Playwright or unit test)

**Documentation**
- [ ] "New" badge in Storybook is added, see [storybook-addon-tag-badges](https://github.com/Sidnioulz/storybook-addon-tag-badges/?tab=readme-ov-file#-usage)
- [ ] auto-generated Storybook code snippets are checked
- [ ] component status has been updated in the [docs overview](https://onyx.schwarz/#components)
- [ ] updated version is released
- [ ] updated documentation is deployed

**Other**
- [ ] follow-up tickets were created if necessary (add links below)

### UX handover checklist

UX handover done with: @ADD_UX_NAME_HERE

**Naming is aligned with Figma**
- [ ] component name
- [ ] which support components are used and how they are named
- [ ] all property names and types

**Implementation is checked**
- [ ] all features of Figma are implemented or a follow-up ticket is refined
- [ ] skeleton
- [ ] densities
- [ ] light and dark mode
- [ ] truncation
- [ ] responsiveness / wrapping behavior
- [ ] min / max width

**Figma**
- [ ] Figma component is published

**Documentation**
- [ ] Storybook category is aligned with Figma
[...]

PR #4628, opened 2026-01-12
feat(OnyxCalendar): implement year and month selection

ChristianBusshoff:
Relates to #4480

- added year & month Picker
- changed next/prev Month from IconButtons to SystemButtons

## Checklist


packages/sit-onyx/playwright/snapshots/components/OnyxCalendar/Calendar-yearMonthPicker--edge-linux.png
larsrickert, line 1: In Figma, we are using the OnyxSystemButton for the years/month which is smaller and has a little different font styles. We currently use the regular OnyxButton, is that intended? Do we need to sync with UX?

larsrickert, line 1: I think the Today button should only be displayed in the small calendar but still be enabled in the big one. At least it looks like this in Figma

larsrickert, line 1: The month buttons are way bigger (taller) than the year buttons, is that intended?

packages/sit-onyx/src/components/OnyxCalendar/OnyxCalendar.vue
```diff
@@ -178,6 +214,16 @@ const calendarWeeksDisplay = computed(
   () =>
     `${t.value("calendar.calendarWeek")} ${weeksToRender.value[0]?.weekNumber} - ${weeksToRender.value[weeksToRender.value.length - 1]?.weekNumber} `,
 );
+
+const picker = useTemplateRef("pickerRef");
+const pickerSmall = useTemplateRef("pickerSmallRef");
+const pickerSmallButton = useTemplateRef("pickerSmallButtonRef");
+useOutsideClick({
```
larsrickert, line 221: Is this `useOutsideClick` only needed to close the small calendar month/year select when clicking outside? In this case, wouldn't it be enough to use one ref on the calendar / wrapper itself?

For the big calendar, the flyout/popover already handles the outside click, right?

packages/sit-onyx/src/components/OnyxCalendar/OnyxCalendar.vue
```diff
@@ -196,34 +242,71 @@ const calendarWeeksDisplay = computed(
         <OnyxSystemButton
           :label="t('calendar.todayButton.label')"
           class="control-container__today-btn"
-          :disabled="disabled === true"
+          :disabled="disabled === true || isPickerOpen"
           @click="goToToday"
         />
+        <div v-if="calendarSize === 'small'" ref="pickerSmallButtonRef">
```
larsrickert: We can omit the nested divs here by placing the v-if and v-else on the OnyxButton and OnyxBasicPopover directly, right?

packages/sit-onyx/src/components/OnyxCalendar/OnyxCalendar.vue
```diff
@@ -196,34 +242,71 @@ const calendarWeeksDisplay = computed(
...
+                isPickerOpen = !isPickerOpen;
+                handlePickerUpdate(isPickerOpen);
+              }
+            "
+          />
+        </div>
+        <div v-else ref="pickerRef">
+          <OnyxBasicPopover
+            v-model:open="isPickerOpen"
+            role="dialog"
```
larsrickert:
```suggestion
```

dialog is already the default role for the popover :)

packages/sit-onyx/src/components/OnyxCalendar/OnyxCalendar.vue
```diff
@@ -196,34 +242,71 @@ const calendarWeeksDisplay = computed(
...
+              <OnyxButton
+                color="neutral"
+                :label="d(viewMonth, { month: 'long', year: 'numeric' })"
+                mode="plain"
+                v-bind="trigger"
+              />
+            </template>
+
+            <template #content>
+              <div>
```
larsrickert: Is the div needed here?

packages/sit-onyx/src/components/OnyxCalendar/OnyxCalendar.vue
```diff
@@ -232,7 +315,17 @@ const calendarWeeksDisplay = computed(
     </div>
 
     <div class="onyx-calendar__body">
-      <table v-bind="tableProps">
+      <div v-if="calendarSize === 'small' && isPickerOpen" ref="pickerSmallRef">
```
larsrickert: We can omit the div here by moving the v-.if and ref to the OnyxMonthYearPickerGrid :)

packages/sit-onyx/src/components/OnyxCalendar/types.ts
```diff
@@ -72,6 +72,22 @@ export type OnyxCalendarProps<TSelection extends OnyxCalendarSelectionMode> = De
    * Whether to show week numbers in the calendar.
    */
   showCalendarWeeks?: boolean;
+
+  /**
+   * Whether to show a month / year picker in the calendar.
+   */
+  monthYearPicker?: {
```
larsrickert: As discussed, to keep things simple we want to remove this prop for now and use a pre-defined 100 year range in the past and future from the current year. If the users sets a min and/or max date on the calendar itself, we consider them as borders

packages/sit-onyx/src/components/OnyxCalendar/OnyxMonthYearPickerGrid.vue
```diff
@@ -0,0 +1,114 @@
...
+  return Array.from({ length: 12 }, (_, i) => ({
+    label: d.value(new Date(2000, i, 1), { month: "short" }),
+    value: i,
+  }));
+});
+</script>
+
+<template>
+  <div class="onyx-component onyx-calendar__picker-grid">
+    <template v-if="props.mode === 'year'">
```
larsrickert, line 69: We can place the v-if / v-else also on the OnyxButton directly so we don't need the template tag

packages/sit-onyx/src/components/OnyxCalendar/OnyxMonthYearPickerGrid.vue
```diff
@@ -0,0 +1,114 @@
...
+
+    &--inline {
+      width: 100%;
+      aspect-ratio: 1;
+      background-color: var(--onyx-color-base-background-blank);
+      border: var(--onyx-1px-in-rem) solid var(--onyx-color-component-border-neutral);
+      border-radius: var(--onyx-radius-md);
+    }
+
+    * {
```
larsrickert: Would prefer to not use `*` since it really affects ALL deeply nested elements and can have performance or unintended issues. Instead we should add a class to the elements where we need these styles and use the class instead

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxCalendar/OnyxCalendar.vue
```diff
@@ -196,34 +242,71 @@ const calendarWeeksDisplay = computed(
...
-        />
-        <OnyxIconButton
-          class="control-container__next-month-button"
-          :label="t('calendar.nextMonthButton')"
-          color="neutral"
-          :icon="iconChevronRightSmall"
-          :disabled="disabled === true"
-          @click="goToMonthByOffset(1)"
-        />
+        <div class="control-container__month-buttons">
```
larsrickert, line 246: The buttons are jumping around (moving left/right) when clicking them multiple times because the  length of the current month changes. We should prevent this :) Maybe we need to sync with UX here?

packages/sit-onyx/src/components/OnyxCalendar/OnyxCalendar.stories.ts
```diff
@@ -30,6 +30,7 @@ export const Default = {} satisfies Story;
 export const SingleSelect = {
   args: {
     selectionMode: "single",
+    monthYearPicker: { max: new Date(2050, 0, 1) },
```
larsrickert: This prop does no longer exist, right?

packages/sit-onyx/src/components/OnyxCalendar/OnyxCalendar.vue
```diff
@@ -178,6 +214,16 @@ const calendarWeeksDisplay = computed(
...
 );
+
+const picker = useTemplateRef("pickerRef");
+const pickerSmall = useTemplateRef("pickerSmallRef");
+const pickerSmallButton = useTemplateRef("pickerSmallButtonRef");
+useOutsideClick({
+  inside: () => [picker.value, pickerSmall.value, pickerSmallButton.value],
+  onOutsideClick: () => {
+    if (isPickerOpen.value) isPickerOpen.value = false;
+  },
```
larsrickert:
```suggestion
  onOutsideClick: () => (isPickerOpen.value = false),
```

(optional) Vue does not track changes if the actual value does not change so we can just set it to false here. If the value is already false, Vue will not trigger any reactive effects

packages/sit-onyx/src/components/OnyxCalendar/OnyxCalendar.vue
```diff
@@ -196,34 +242,71 @@ const calendarWeeksDisplay = computed(
...
+            <OnyxButton
+              color="neutral"
+              :label="d(viewMonth, { month: 'long', year: 'numeric' })"
+              mode="plain"
+              v-bind="trigger"
+            />
+          </template>
+
+          <template #content>
+            <div>
```
larsrickert: Is the div needed here?

packages/sit-onyx/src/components/OnyxCalendar/OnyxMonthYearPickerGrid.vue
```diff
@@ -0,0 +1,141 @@
...
+    const containerHeight = container.offsetHeight;
+
+    container.scrollTop = elementTop - containerHeight / 2 + elementHeight;
+  }
+};
+
+watch(
+  () => props.open,
+  (open) => {
+    if (open === true) {
```
larsrickert:
```suggestion
    if (open) {
```

Since open is a boolean, we don't need the `===` check

packages/sit-onyx/src/components/OnyxCalendar/OnyxMonthYearPickerGrid.vue
```diff
@@ -0,0 +1,141 @@
...
+    const container = gridRef.value;
+    const element = buttonRefs.value[activeIndex].$el || buttonRefs.value[activeIndex];
+
+    const elementTop = element.offsetTop;
+    const elementHeight = element.offsetHeight;
+    const containerHeight = container.offsetHeight;
+
+    container.scrollTop = elementTop - containerHeight / 2 + elementHeight;
+  }
+};
```
larsrickert, line 55:
```suggestion
const scrollToActiveYear = async () => {
  if (props.mode !== "year") return;
  await nextTick();
  const activeIndex = years.value.indexOf(currentViewYear.value);
  const selectedYearButton: HTMLElement | undefined = buttonRefs.value?.[activeIndex]?.$el;
  if (!selectedYearButton) return;
  selectedYearButton.scrollIntoView({ block: "center" });
};
```

We can make use of the native `scollIntoView()` method so we don't need to calculate here :) This way, we can also remove the gridRef

packages/sit-onyx/src/components/OnyxCalendar/OnyxMonthYearPickerGrid.vue
```diff
@@ -0,0 +1,141 @@
...
+      scrollToActiveYear();
+    }
+  },
+);
+</script>
+
+<template>
+  <div ref="grid" class="onyx-component onyx-calendar__picker-grid">
+    <template v-if="props.mode === 'year'">
+      <!-- TODO: Change to SystemButton after styles are implemented -->
```
larsrickert, line 70: Please create a follow up ticket for this

packages/headless/src/composables/helpers/useOutsideClick.ts
```diff
@@ -37,7 +38,10 @@ export const useOutsideClick = <TCheckOnTab extends boolean | undefined>({
     if (!target) return true;
     const raw = toValue(inside);
     const elements = Array.isArray(raw) ? raw : [raw];
-    return !elements.some((element) => element?.contains(target as Node | null));
+    return !elements.some((element) => {
```
larsrickert, line 41: We can create a changeset for this with something like:
feat(useOutsideClick): support passing component refs to `inside` option

packages/sit-onyx/src/components/OnyxCalendar/OnyxMonthYearPickerGrid.vue
```diff
@@ -0,0 +1,141 @@
...
+
+const { d } = injectI18n();
+
+const currentViewYear = computed(() => props.viewMonth.getFullYear());
+const currentViewMonthIndex = computed(() => props.viewMonth.getMonth());
+
+const years = computed(() => {
+  const currentYear = new Date().getFullYear();
+  const start = currentYear - 102;
+  const end = currentYear + 53;
```
larsrickert:
```suggestion
  const count = 100;
  const currentYear = new Date().getFullYear();
  const start = currentYear - count;
  const end = currentYear + count;
```

Is there a reason for these "strange" numbers? :D

packages/sit-onyx/src/components/OnyxCalendar/OnyxMonthYearPickerGrid.vue
ChristianBusshoff: Created it: #4643

PR #4665, opened 2026-01-14
chore: add `.nuxtrc` to gitignore

larsrickert:
Currently, Nuxt generates ´.nuxtrc` files frequenly in the monorepo ([see Nuxt docs](https://nuxt.com/docs/4.x/directory-structure/nuxtrc#usage)) which are annoying in the git changes. This PR adds them go the .gitignore since they are not relevant

### DIFF
--- a/packages/nuxt-docs/.nuxtrc
+++ b/packages/nuxt-docs/.nuxtrc
@@ -1,1 +1,0 @@
-typescript.includeWorkspace = true


.gitignore
```diff
@@ -38,6 +38,7 @@ blob-report/
 eslint-results.sarif
 .nuxt
 nuxt.d.ts
+.nuxtrc
```
JoCa96, line 41: We already have a [.nuxtrc](https://github.com/SchwarzIT/onyx/blob/main/packages/nuxt-docs/.nuxtrc) checked in, do we remove that too?

JoCa96 (commented):
Based on [the nuxt docs](https://nuxt.com/docs/4.x/directory-structure/nuxtrc#global-nuxtrc-file:~:text=Nuxt%20automatically%20adds%20a%20setups%20section%20to%20track%20module%20installation%20and%20upgrade%20state%2E%20This%20is%20used%20internally%20for%20module%20lifecycle%20hooks%20and%20should%20not%20be%20modified%20manually) I would think we should commit the file instead

.gitignore
larsrickert: Sure 👍

JoCa96 (approved):

Issue #4465, opened 2025-11-19
Implement OnyxSearch component


larsrickert:
Might depend on #2072 

### Why?

We want to offer a unified component for search (and filter) functionality for various data (e.g. for a data grid or table)

### Open questions / TODOs

- should we unify the "Personal filters" modal or is it completely up to the project to implement it?

### Design

[Figma](https://www.figma.com/design/YfEUBOHk4J4nYrk04geswG/Onyx-Components?node-id=15763-175102&m=dev)
[Documentation](https://www.figma.com/design/j7Hy7Zk8MiCmJBJf0ArQRm/Onyx-Documentation?node-id=16321-9078&m=dev)

### Acceptance criteria

- [x] an input is displayed where a search value can be entered
- [x] supports the colors blank (default) and tinted
- [x] supports border radius strong (default) and soft
- [ ] optionally a keyboard shortcut can be set which triggers / focuses the search and is displayed in the component
- [ ] optionally a filter icon button can be displayed. There are the following layout options (can be changed as a property)
    - [ ] filters beside search (inline), default
    - [ ] filters below search
    - [ ] filters inside temporary sidebar
    - [ ] filters inside a modal
    - [ ] manual (component itself does nothing, just emits an event)
- [ ] optionally a "Personal filters" icon button can be displayed. It just emits an event, it is up to the developer to e.g. show a modal with pre-defined user filters

### Implementation details

- use the OnyxInput internally and use the leading and trailing slot to add the custom icons / actions

### Applicable ARIA Pattern

[Disclosure pattern](https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/)

### DEV handover checklist

[Storybook](ADD_LINK_HERE)

DEV handover done with: @ADD_DEV_NAME_HERE

**Testing**
- [ ] covered by visual screenshot tests
- [ ] covered by functional tests (Playwright or unit test)

**Documentation**
- [ ] "New" badge in Storybook is added, see [storybook-addon-tag-badges](https://github.com/Sidnioulz/storybook-addon-tag-badges/?tab=readme-ov-file#-usage)
- [ ] auto-generated Storybook code snippets are checked
- [ ] component status has been updated in the [docs overview](https://onyx.schwarz/#components)
- [ ] updated version is released
- [ ] updated documentation is deployed

**Other**
- [ ] follow-up tickets were created if necessary (add links below)

### UX handover checklist

UX handover done with: @ADD_UX_NAME_HERE

[...]

Issue #4650, opened 2026-01-14
Implement OnyxInputLayout Support component


JoCa96:
### Why?

We want to ensure, that all `input` based components behave identically and have the same layout.

### Depends on

- #4506 

### Design

[Figma](https://www.figma.com/design/j7Hy7Zk8MiCmJBJf0ArQRm/Onyx-Documentation?node-id=20855-13108&t=G6Mv0VntwtpxTVoE-4)

### Acceptance criteria

- [ ] A new support component `InputLayout` is implemented
- [ ] When the input icon is clicked, it should also focus the input itself
- [ ] The following components have been updated to use the new `InputLayout` component:
  - [ ] `OnyxInput`
  - [ ] `OnyxSelect`
  - [ ] `OnyxStepper`
  - [ ] `OnyxDatepicker`
  - [ ] `OnyxTimepicker`

### Implementation details

- Make use of the already existing `OnyxInputAction` component

### DEV Handover Checklist

DEV handover done with: @ADD_DEV_NAME_HERE

**Testing**
- [ ] covered by visual screenshot tests
- [ ] covered by functional tests (Playwright or unit test)

**Documentation**
- [ ] "New" badge in Storybook is added, see [storybook-addon-tag-badges](https://github.com/Sidnioulz/storybook-addon-tag-badges/?tab=readme-ov-file#-usage)
- [ ] auto-generated Storybook code snippets are checked
- [ ] component status has been updated in the [docs overview](https://onyx.schwarz/#components)
- [ ] updated version is released
- [ ] updated documentation is deployed

**Other**
- [ ] follow-up tickets were created if necessary (add links below)

### UX handover checklist

UX handover done with: @ADD_UX_NAME_HERE

**Naming is aligned with Figma**
- [ ] component name
- [ ] which support components are used and how they are named
- [ ] all property names and types

**Implementation is checked**
- [ ] all features of Figma are implemented or a follow-up ticket is refined
- [ ] skeleton
- [ ] densities
- [ ] light and dark mode
- [ ] truncation
- [ ] responsiveness / wrapping behavior
- [ ] min / max width

**Figma**
- [ ] Figma component is published

**Documentation**
- [ ] Storybook category is aligned with Figma
- [ ] Storybook examples are checked

**Other**
- [ ] follow-up tickets were created if necessary (add links below)

PR #4668, opened 2026-01-15
feat(OnyxSearch): implement basic OnyxSearch input

JoCa96:
Relates to #4465 

- feat(createToggleButton): implement createToggleButton composable and testing helper
  - used by the `OnyxInputAction`, which will be used by the `OnyxSearch` after ticket #4651

- `OnyxSearch` is for now hidden from Storybook, as it doesn't offer any value to our users yet and I don't want to have too many `Unstable` components.

### DIFF
--- a/packages/sit-onyx/src/components/OnyxSearch/types.ts
+++ b/packages/sit-onyx/src/components/OnyxSearch/types.ts
@@ -2,6 +2,10 @@ import type { OnyxInputProps } from "../OnyxInput/types.js";
 
 export type OnyxSearchProps = Pick<
   OnyxInputProps,
+  | "message"
+  | "maxlength"
+  | "error"
+  | "readonly"
   | "disabled"
   | "density"
   | "id"


packages/sit-onyx/src/components/OnyxLoadingIndicator/OnyxCircleSpinner.vue
larsrickert, line 1: The 1px padding originally was added in this fix PR: #4522
Is this intended now to remove it again?

packages/headless/src/composables/toggleButton/createToggleButton.ts
```diff
@@ -0,0 +1,27 @@
...
+      /**
+       * A html button element that is supposed to act as a toggle button.
+       */
+      button: computed(
+        () =>
+          ({
+            "aria-pressed": toValue(options.isPressed),
+            onClick: options.onToggle,
+          }) as const satisfies VBindAttributes<ButtonHTMLAttributes>,
+      ),
```
larsrickert, line 30:
```suggestion
      button: computed(() => ({
        "aria-pressed": toValue(options.isPressed),
        onClick: options.onToggle,
      })),
```

Why is the type cast needed here? Seems to work fine without :)

packages/headless/src/composables/toggleButton/createToggleButton.ts
```diff
@@ -0,0 +1,27 @@
+import { computed, toValue, type ButtonHTMLAttributes, type MaybeRefOrGetter } from "vue";
+import { createBuilder, type VBindAttributes } from "../../utils/builder.js";
+
+type CreateToggleButtonOptions = {
```
larsrickert, line 4: Would be nice to add some JSDoc to the options

packages/sit-onyx/src/components/OnyxInput/modules/OnyxInputAction.vue
```diff
@@ -0,0 +1,122 @@
...
+      >
+        <OnyxIcon
+          class="onyx-input-action__icon"
+          size="24px"
+          :icon="props.icon"
+          color="currentColor"
+        />
+      </button>
+    </template>
+    <template #tooltip>{{ props.label }}</template>
```
larsrickert: If its just the label, we can also use `:text="props.label"` on the OnyxTooltip. The slot is mainly intended for custom content afaik

packages/sit-onyx/src/components/OnyxInput/modules/OnyxInputAction.vue
```diff
@@ -0,0 +1,122 @@
...
+<style lang="scss">
+@use "../../../styles/mixins/layers.scss";
+
+@include layers.component() {
+  .onyx-input-action {
+    all: initial;
+    color: var(--onyx-color-text-icons-neutral-medium);
+    padding: var(--onyx-density-xs) var(--onyx-density-sm);
+    margin-right: -12px;
+    margin-left: -12px;
```
larsrickert: Would prefer rem here. Can we compute the value somehow with CSS variables instead of the magic numbers?

packages/sit-onyx/src/components/OnyxSearch/types.ts
```diff
@@ -0,0 +1,34 @@
+import type { OnyxInputProps } from "../OnyxInput/types.js";
+
+export type OnyxSearchProps = Pick<
```
larsrickert, line 3: What about the other input props like required etc.?

packages/sit-onyx/src/components/OnyxSkeleton/OnyxSkeleton.vue
```diff
@@ -16,6 +16,7 @@ onMounted(() => requestAnimationSync("onyx-skeleton"));
   @include layers.component() {
     border-radius: var(--onyx-radius-sm);
     pointer-events: none;
+    cursor: wait;
```
larsrickert: The cursor has no effect for me 🤔  Does it work for you when hovering the skeleton?

packages/sit-onyx/src/styles/mixins/input.scss
```diff
@@ -170,6 +171,7 @@
     line-height: inherit;
     padding: 0;
     height: 100%;
+    caret-color: var(--onyx-color-base-primary-500);
```
larsrickert: afaik we intentionally did not use the primary color for the caret for better visibility

packages/sit-onyx/src/components/OnyxSearch/OnyxSearch.vue
```diff
@@ -0,0 +1,105 @@
...
+
+defineExpose({ input });
+</script>
+
+<template>
+  <OnyxInput
+    ref="inputRef"
+    v-bind="inputProps"
+    v-model="modelValue"
+    :loading="props.loading"
```
larsrickert: Aren't the loading and label prop already passed with the `v-bind`?

packages/sit-onyx/src/components/OnyxSearch/OnyxSearch.vue
```diff
@@ -0,0 +1,105 @@
...
+    :class="['onyx-search', `onyx-search--${props.color}`, `onyx-search--${props.cornerRadius}`]"
+    :label="props.label"
+    :placeholder="props.label"
+    type="search"
+    hide-label
+  >
+    <template #leading>
+      <OnyxIcon v-if="!props.loading" class="onyx-search__icon" :icon="iconSearch" />
```
larsrickert: Just clicking the icon focus the input?

packages/sit-onyx/src/components/OnyxInput/modules/OnyxInputAction.vue
larsrickert, line 1: Whats this component intended for? I don't see it used anywhere? :)

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxInput/modules/OnyxInputAction.vue
JoCa96: Good point! Actually we are not going to need them anymore, as they are only needed to work with the current `OnyxInput` slots.

packages/sit-onyx/src/components/OnyxInput/modules/OnyxInputAction.vue
JoCa96: It is intended to be used with/for the OnyxInputLayout from [#4650](https://github.com/SchwarzIT/onyx/issues/4650). I will add a comment to that ticket.

packages/sit-onyx/src/components/OnyxSearch/OnyxSearch.vue
JoCa96: I don't understand the question.

packages/sit-onyx/src/components/OnyxSkeleton/OnyxSkeleton.vue
JoCa96: No, because of `pointer-events: none;`

packages/sit-onyx/src/styles/mixins/input.scss
JoCa96: Ok, noticed it in Figma, so will ask UX about it :)

packages/sit-onyx/src/components/OnyxSearch/OnyxSearch.vue
larsrickert: Ups sorry, I typed to fast :D So the question was: When clicking on the icon, should we focus the input element? So similar to the regular OnyxInput where clicking the label will also move focus to the native input element so you can start typing

packages/sit-onyx/src/styles/mixins/input.scss
JoCa96: updated it after discussion with UX

packages/sit-onyx/src/components/OnyxInput/modules/OnyxInputAction.vue
```diff
@@ -0,0 +1,119 @@
...
+}>();
+
+const active = useVModel({
+  props,
+  emit,
+  key: "active",
+  default: false,
+});
+
+const toggleAttrs = computed(() =>
```
larsrickert, line 49: Currently this is no problem but composables should not be called inside computed since there is no active component instance. You can test this by using "onMounted" inside "createToggleButton" and then toggle the "type" property in Storybook between "toggle" and "button". You will see a Vue warning in the console.

I think we have the same issue currently for the OnyxTooltip when dynamically switching between tooltip and toggletip 👀 

In this case we can solve it by always calling createToggleButton at the top and only bind it when `props.type === "toggle"`

packages/sit-onyx/src/components/OnyxSearch/OnyxSearch.stories.ts
```diff
@@ -0,0 +1,75 @@
...
+import type { Meta, StoryObj } from "@storybook/vue3-vite";
+import OnyxSearch from "./OnyxSearch.vue";
+
+/**
+ * This type of searchbar is ideal for simpler, less complex, and smaller datasets. It provides a straightforward text search functionality, allowing users to type in a keyword or phrase and receive results that directly match their input. This option is best suited for cases where data retrieval is straightforward and filtering isn’t necessary.
+ */
+const meta: Meta<typeof OnyxSearch> = {
+  title: "Search & Filter/Search",
+  component: OnyxSearch,
+  // TODO: unhide from storybook after https://github.com/SchwarzIT/onyx/issues/4651
```
larsrickert, line 11: Its currently not hidden, right?

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxInput/modules/OnyxInputAction.vue
larsrickert: image

packages/sit-onyx/src/components/OnyxSearch/OnyxSearch.vue
JoCa96: Yes, but we should do that generically. I added it to the AC of issue #4650

packages/sit-onyx/src/components/OnyxInput/modules/OnyxInputAction.vue
JoCa96: Interesting... I would prefer not to always call the composable, because it will create unnecessary global listener.
I would not allow the users to change these props at runtime.

JoCa96: Created a follow-up in #4719

packages/sit-onyx/src/components/OnyxSearch/types.ts
JoCa96: The OnyxSearch acts as a complete form instead of a single element, so not all properties make sense in that context.
I adapted the code to show what I mean.

Issue #4229, opened 2025-10-07
Define Security Hardenings for onyx


JoCa96:
### Why?

With npmjs recently being targeted repeatedly by multiple supply-chain attacks, we need to research and define which steps we can take to minimize our attack surfaces. Especially as enterprise software product.

See:
- 16.09.2025 [Self-Replicating Worm Hits 180+ Software Packages](https://krebsonsecurity.com/2025/09/self-replicating-worm-hits-180-software-packages/)
- 08.09.2025 [18 Popular Code Packages Hacked, Rigged to Steal Crypto](https://krebsonsecurity.com/2025/09/18-popular-code-packages-hacked-rigged-to-steal-crypto/)
- 27.08.2025 [Popular nx packages compromised on npm](https://www.aikido.dev/blog/popular-nx-packages-compromised-on-npm)

### Acceptance criteria

- [ ] Created a docs page with Securtiy Best-Practices for us to follow
  - [ ] Sources have been included (as links)
- [ ] These have been implemented
- [ ] Or follow-up tickets have been created

### Implementation details

- We have Snyk licenses in the company, we should consider using snyk
- See https://pnpm.io/supply-chain-security
- See https://nodejs.org/en/learn/getting-started/security-best-practices#openssf-tools
- See e.g. https://docs.stepsecurity.io/orchestrate-security

[This StepSecurity blog entry](https://www.stepsecurity.io/blog/ctrl-tinycolor-and-40-npm-packages-compromised#implement-security-controls) also has some interesting suggestions:

> ## Implement Security Controls
> ### GitHub Security Hardening
>
> Review and remove unnecessary GitHub Apps and OAuth applications
> Audit all repository webhooks for unauthorized additions
> Check deploy keys and repository secrets for all projects
> Enable branch protection rules to prevent force-pushes
> Turn on GitHub Secret Scanning alerts
> Enable Dependabot security updates
>
> ### Ongoing Monitoring
>
> Set up alerts for any new npm publishes from your organization
> Monitor CloudTrail/GCP audit logs for secret access patterns
> Implement regular credential rotation policies
> Use separate, limited-scope tokens for CI/CD pipelines
>
> [ ... ] 
> 
> ### Use NPM Package Cooldown Check
> 
[...]

PR #4694, opened 2026-01-22
chore: harden pipelines and implement pnpm supply chain security recommendations

JoCa96:
Relates to #4229

- Add [harden-runner](https://github.com/step-security/harden-runner) to all jobs
  - It uses [eBPF](https://en.wikipedia.org/wiki/EBPF) to audit network and file system calls.
  - It will currently run in `audit` mode to create a baseline. The baseline is used to create a whitelist for network calls and detect anomalous behavior.
  - E.g. you can check the audit results for this run at https://app.stepsecurity.io/github/SchwarzIT/onyx/actions/runs/21247198404

- Update pipeline permissions to follow the rule of "least privileged" and fix the [Token-Permissions](https://github.com/SchwarzIT/onyx/security/code-scanning) code scanning issues.
- Implement pnpm's [Mitigating supply chain attacks](https://pnpm.io/supply-chain-security) recommendations:
  - [Block risky postinstall scripts](https://pnpm.io/supply-chain-security#block-risky-postinstall-scripts)
  - [Prevent exotic transitive dependencies](https://pnpm.io/supply-chain-security#prevent-exotic-transitive-dependencies)
  - [Delay dependency updates](https://pnpm.io/supply-chain-security#delay-dependency-updates)
  - [Enforce trust with trustPolicy](https://pnpm.io/supply-chain-security#enforce-trust-with-trustpolicy)

- Update dependabot configuration to also use two day cooldown, so it's in sync with the pnpm `minimumReleaseAge` configuration

### DIFF
--- a/.github/workflows/scorecards.yml
+++ b/.github/workflows/scorecards.yml
@@ -47,11 +47,11 @@ jobs:
         with:
           results_file: results.sarif
           results_format: sarif
-          # TODO: Enable after we checked first results:
+          # Publish results:
           #   - Publish results to OpenSSF REST API for easy access by consumers
           #   - Allows the repository to include the Scorecard badge.
           #   - See https://github.com/ossf/scorecard-action#publishing-results.
-          publish_results: false
+          publish_results: true
 
       # Upload the results to GitHub's code scanning dashboard.
       - name: "Upload to code-scanning"


.github/workflows/check.yml
```diff
@@ -29,7 +29,13 @@ jobs:
     permissions:
       contents: read
     steps:
-      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+      - name: 🛃 Harden the runner (Audit all outbound calls)
```
larsrickert: Would prefer to unify these harden runner steps by creating a reusable workflow and use it here instead so we don't need to maintain the version, parameters etc. in so many places :)

larsrickert: We could also move the checkout step there

JoCa96: I don't know about that, not every workflow uses the checkout with the same parameters. And having just a template for a single step with a single parameter seems unnecessary tbh.

larsrickert: Even if its just one step, it still allows us to control the version and egress-policy etc. in one single place. Otherwise we'd need to change it in 12 files + if a new version is released, dependabot will update 12 files instead of just a single one :)

I just fear that our pipelines will get very messy with all the new actions + duplication added recently

JoCa96: as discussed

Issue #3063, opened 2025-04-01
Implement hover effect when resizing columns in the OnyxDataGrid


MajaZarkova:
=> WIP branch: 3063/Fix-hover-styles

### Why?

This is a FoUp -> #2972 

### Acceptance criteria

- [x] While clicking and holding to adjust the width the hover of the table should be locked to the column. Right now, when the cursor moves down during the resizing process, the hover state changes to the row. But during the resizing, it is intended to be the column hover exclusively, even if the cursor is in the level of the row.
- [x] ~~While resizing (moving the cursor to the left) the gap between the cursor and the border is increasing -> Investigate if we can freeze the cursor. Answer: No, the cursor is controlled by the operating system~~
- [x] visual tests (Playwright screenshots)

### Definition of Done

- Should be covered by tests:
  - [x] visual tests (Playwright screenshots)

#### Approval

[Storybook](https://onyx-storybook-dev.apps.01.cf.eu01.stackit.cloud/?path=/story/data-datagrid-features--resizing)

- Checked and approved by
  - [x] designer @jannick-ux 
  - [x] dev @JoCa96

PR #4715, opened 2026-01-23
fix(OnyxDataGrid): lock column hover effect while resizing

larsrickert:
Relates to #3063

While actively resizing a column in the data grid, ensure hover styles on active column and prevent hover styles on other columns and rows

## Checklist


packages/sit-onyx/src/components/OnyxDataGrid/features/resizing/resizing.ts
```diff
@@ -11,7 +11,7 @@ export const RESIZING_FEATURE = Symbol("Resizing");
 export const FILLER_COLUMN = Symbol("FILLER_COLUMN");
 export const useResizing = <TEntry extends DataGridEntry>(options?: ResizingOptions<TEntry>) =>
   createFeature((ctx) => {
-    const resizingCol = ref<Readonly<InternalColumnConfig<TEntry>>>();
+    const resizingCol = ref<TEntry["id"]>();
```
JoCa96: `id` would identify a row, not a column
```suggestion
    const resizingCol = ref<keyof TEntry>();
```

packages/sit-onyx/src/components/OnyxDataGrid/features/resizing/resizing.scss
```diff
@@ -15,4 +15,15 @@
       }
     }
   }
+
+  // styles to apply to the whole data grid while actively resizing a column
+  .onyx-data-grid {
+    &:has(.onyx-data-grid-resize-cell--active) {
```
JoCa96, line 21: Can we use `onyx-resize-handle--active` class?
Then we might not need the "prevent infinite loop" workaround.

larsrickert: Sure we can use the class but we still need the infinite loop fix since the "hovered" class is added.

I'd consider the infinite loop change a fix, not a workaround

JoCa96 (approved):

Issue #4616, opened 2026-01-09
[Bug]: Demo app sidebar not working in mobile


larsrickert:
### Version

1.5.0

### Description

The mobile sidebar toggle buttons do not work in the demo app

### Reproduction steps

https://demo.onyx.schwarz/forms

### Confirmations

- [x] I have checked that there is not already a bug report for my issue.
- [x] I confirm that are no sensitive information included in this bug report. e.g. credentials, internal project details (for employees of the Schwarz Group) etc.

PR #4716, opened 2026-01-23
fix(demo-app): fixed sidebar opening in mobile demo app

ChristianBusshoff:
Relates to #4616

## Checklist

### DIFF
--- a/apps/demo-app/layers/blueprint/app/layouts/sidebar.vue
+++ b/apps/demo-app/layers/blueprint/app/layouts/sidebar.vue
@@ -25,6 +25,8 @@ const filteredItems = computed(() => {
   return props.sidebarItems.filter((item) => normalizedIncludes(item.label, searchTerm));
 });
 
+const { width } = useWindowSize();
+const isMobile = computed(() => width.value < ONYX_BREAKPOINTS.sm);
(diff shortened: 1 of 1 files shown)


apps/demo-app/layers/blueprint/app/layouts/sidebar.vue
```diff
@@ -46,13 +46,21 @@ globalFAB.add(
...
 </script>
 
 <template>
   <OnyxPageLayout v-bind="props">
     <template #sidebar>
       <!-- using v-show here instead of v-if so the search value is kept when toggling the open state -->
-      <OnyxSidebar v-show="isOpen" class="sidebar" :label="$t('blueprint.sidebar')">
+      <OnyxSidebar
+        v-show="isOpen"
+        :temporary="width < ONYX_BREAKPOINTS.sm ? { open: isOpen } : undefined"
```
larsrickert: Currently the sidebar is initially opened when switching to mobile (or reloading the page). It would be nice if it is only opened initially on desktop but closed by default in mobile (we could use two open refs here)

Also, should we add a small comment here that we disable the collapse-sidebar and handle it on our own because we manage the floating action button on our own?

Issue #4663, opened 2026-01-14
Internal demo app nav bar not using theme font familiy


larsrickert:
### Version

1.5.0

### Description

The nav bar of our [internal demo app](https://demo-internal.onyx.schwarz) does not use the correct font family for the nav bar. It seems like it always uses the font family of the last theme, even when switching between different themes

### Reproduction steps

1. open [internal demo app](https://demo-internal.onyx.schwarz)
2. check nav bar font family
3. switch theme
4. check that nav bar font family didn't change

### Confirmations

- [x] I have checked that there is not already a bug report for my issue.
- [x] I confirm that are no sensitive information included in this bug report. e.g. credentials, internal project details (for employees of the Schwarz Group) etc.

PR #4755, opened 2026-01-30
fix(demo-app): correctly apply density to whole app

larsrickert:
Relates to #4663

Correctly apply demo app density to `<html>` element so it also affects global styles

## Checklist

### DIFF
--- a/apps/demo-app/app/app.vue
+++ b/apps/demo-app/app/app.vue
@@ -2,8 +2,16 @@
 import App from "#layers/blueprint/app/app.vue";
 
 const settingsStore = useSettingsStore();
+
+// we need to apply the CSS class to the `<html>` element so the density is correctly defined for all components
+// otherwise (if we would e.g. just bind it to the app below), global styles (e.g. the vertical grid margin etc. would not consider the corresponding density)
+useHead({
+  htmlAttrs: {
+    class: computed(() => `onyx-density-${settingsStore.settings.density}`),
+  },
+});
 </script>
 
 <template>
-  <App :class="`onyx-density-${settingsStore.settings.density}`" />
+  <App />
 </template>


JoCa96:
@larsrickert Is this fix only regarding `--onyx-grid-margin-vertical`, or are there more cases?

larsrickert:
> @larsrickert Is this fix only regarding `--onyx-grid-margin-vertical`, or are there more cases?

@JoCa96 Not 100% sure but I think currently it seems like the only variable yes. But we might are additional global variables in the future that depend on the densityx

apps/demo-app/app/app.vue
```diff
@@ -2,8 +2,16 @@
...
+useHead({
+  htmlAttrs: {
+    class: computed(() => `onyx-density-${settingsStore.settings.density}`),
+  },
+});
 </script>
 
 <template>
-  <App :class="`onyx-density-${settingsStore.settings.density}`" />
+  <App />
```
JoCa96, line 16: Densities should work correctly even when applied only to nested element. So in this case of `--onyx-grid-margin-vertical` we need to assign the custom-property wherever the densities are changed:

```css
  :where(:root, .onyx-density-cozy, .onyx-density-compact, .onyx-density-default) {
    --onyx-grid-margin-vertical: var(--onyx-density-3xl);
  }
```

larsrickert: Hmm I am not sure about this since this doesn't work in such cases:
- we currently define a custom value for `--onyx-grid-margin-vertical` for the OnyxSidebar. If the user uses another density for the sidebar or sidebar content, this custom value is ignored an the unintentional global 3xl value is used
- if the user adds custom CSS to set `--onyx-grid-margin-vertical` on any element, defining the density on any child element will break this (same reason as above)

PR #4808, opened 2026-02-10
fix(OnyxKey): use key name for screen readers instead of symbol

larsrickert:
Relates to #4655

Add visually hidden key name to the OnyxKey so screen readers are announcing the key name instead of the unicode symbol

## Checklist

### DIFF
--- a/packages/sit-onyx/src/components/OnyxKey/OnyxKey.stories.ts
+++ b/packages/sit-onyx/src/components/OnyxKey/OnyxKey.stories.ts
@@ -5,9 +5,11 @@ import OnyxKey from "./OnyxKey.vue";
 /**
  * The key component displays a single keyboard key with support for specific operating system symbols (macOS, Windows).
  * Useful for documentation, tutorials and displaying keyboard shortcuts.
+ *
+ * For representing keyboard shortcuts, use the [OnyxShortcut](/docs/basic-shortcut--docs) component.
  */
 const meta: Meta<typeof OnyxKey> = {
-  title: "Basic/Key",
+  title: "Support/Key",
   component: OnyxKey,
   tags: ["unstable"],
 };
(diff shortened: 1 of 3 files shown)


packages/sit-onyx/src/components/OnyxKey/OnyxKey.vue
```diff
@@ -92,9 +93,11 @@ defineExpose({
       { 'onyx-key--highlighted': isHighlighted },
     ]"
   >
-    <span class="onyx-truncation-ellipsis">
+    <span aria-hidden="true" class="onyx-truncation-ellipsis">
       {{ visualLabel }}
     </span>
+
+    <OnyxVisuallyHidden>{{ props.name }}</OnyxVisuallyHidden>
   </kbd>
```
JoCa96, line 105: We could consider using a tooltip on the kbd element, then it could also help more people

packages/sit-onyx/src/components/OnyxKey/OnyxKey.vue
```diff
@@ -92,9 +93,11 @@ defineExpose({
       { 'onyx-key--highlighted': isHighlighted },
     ]"
   >
-    <span class="onyx-truncation-ellipsis">
+    <span aria-hidden="true" class="onyx-truncation-ellipsis">
       {{ visualLabel }}
```
JoCa96, line 101: Would be good to embed it into a translation, e.g. `t('key.label', props.name)` which would result in `"Escape" Key`

JoCa96 (changes requested):

packages/sit-onyx/src/components/OnyxKey/OnyxKey.vue
larsrickert: I like the idea! Imo a tooltip might be too prominent. I added the "title" attribute instead so it shows the native browser tooltip when hovering a while, what do you think?

JoCa96: The plan is to replace all `title` usages in the near-term, as they are not accessible.
So a tooltip would be the better choice here.
But would be interesting to know what you mean with "too prominent"?

larsrickert: I checked with UX and the tooltip should not be shown by default and instead be up to the user. Is that also fine for you for now? I removed the title attribute for now

JoCa96: not a fan of that decision; but this PR is fine :)

Issue #4751, opened 2026-01-28
Optimize showcase content FL


mj-hof:
### Why?

- Check our existing content and rewrite and optimize it in our introduction section.
- Consider the comments to each section in our Google sheet

- [x] Foundation > Breakpoints & Grid
- [x] Foundation > Layout

Issue #4749, opened 2026-01-28
Optimize showcase content JK


mj-hof:
### Why?

- Check our existing content and rewrite and optimize it in our introduction section.
- Consider the comments to each section in our Google sheet

- [ ] Foundation > Colors & Themes
- [ ] Foundation > Variables

jannick-ux:
Colors topic finished

Issue #4748, opened 2026-01-28
Optimize showcase content LR


mj-hof:
### Why?

- Check our existing content and rewrite and optimize it in our introduction section.
- Consider the comments to each section in our Google sheet

- [ ] Getting started > Installation
- [ ] Getting started > First steps

PR #4826, opened 2026-02-11
docs(showcase): update content

larsrickert:
Relates to #4751, #4749, #4748

- include updated showcase content
- update `@sit-onyx/nuxt-docs` to provide some more useful markdown components and updated spacings

## Checklist


packages/nuxt-docs/app/composables/useIcon.ts
```diff
@@ -0,0 +1,31 @@
...
+ * Composable for resolving the icon content based on the icon name.
+ */
+export const useIcon = (iconName: MaybeRef<string | undefined>) => {
+  const icon = ref<string>();
+
+  watch(
+    toRef(iconName),
+    (newIconName) => {
+      icon.value = newIconName
+        ? ALL_ICONS[getIconImportName(newIconName) as keyof typeof ALL_ICONS]
```
JoCa96, line 14: Might want to consider using [`dynamic imports`](https://vite.dev/guide/features#dynamic-import) to support lazy loading at runtime

larsrickert: Good idea :) For now I'd keep it as it is because the app is server side rendered anyway so the client will only receive the already resolved icons and not all imported ones :) Because when making this async, the icon will be missing on initial hydration and then be resolved afterwards which might lead to a visual "flicker"

PR #4850, opened 2026-02-16
chore: automate translations

larsrickert:
Relates to #4860 

Add a new GitHub action that can be triggered to add translations for all missing keys for all languages.

- prepared [AITxt template](https://ai-txt-gen.digits.schwarz/ai-txt/?userMode=advanced&template=262)
- Example pipeline run: https://github.com/SchwarzIT/onyx/actions/runs/22569439199/job/65373237894
- Example generated PR: https://github.com/SchwarzIT/onyx/pull/4953

## Checklist

### DIFF
--- a/.github/scripts/translate.ts
+++ b/.github/scripts/translate.ts
@@ -126,8 +126,8 @@ function mergeTranslations(base: TranslationJson, update: TranslationJson) {
     // check if both values are objects; if so, recurse
     if (isObject(baseValue) && isObject(updateValue)) {
       output[key] = mergeTranslations(baseValue, updateValue);
-    } else {
-      // otherwise, overwrite the value (handles string updates or type changes)
+    } else if (!output[key]) {
+      // otherwise, overwrite the value (handles string updates or type changes) if no value exists yet
       output[key] = updateValue;
     }
   });


.github/workflows/translate.yml
```diff
@@ -0,0 +1,39 @@
...
+      - name: 🛃 Harden the runner (Audit all outbound calls)
+        uses: step-security/harden-runner@a90bcbc6539c36a85cdfeb73f7e2f433735f215b # v2.15.0
+        with:
+          egress-policy: audit
+
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - uses: ./.github/templates/node-setup
+
+      - run: pnpx tsx .github/scripts/translate.ts
```
JoCa96: We should pin the version of the `tsx` or put into a lockfile, to not accidentally use a vulnerable version or have breaking changes.
Alternatively this script should be able to work with `node --experimental-strip-types` or `node --experimental-transform-types another-example.ts`, so we only need to rely on node.js. See: https://nodejs.org/en/learn/typescript/run-natively

.github/scripts/translate.ts
```diff
@@ -0,0 +1,158 @@
...
+    const baseValue = output[key];
+    const updateValue = update[key];
+
+    // check if both values are objects; if so, recurse
+    if (isObject(baseValue) && isObject(updateValue)) {
+      output[key] = mergeTranslations(baseValue, updateValue);
+    } else {
+      // otherwise, overwrite the value (handles string updates or type changes)
+      output[key] = updateValue;
+    }
```
JoCa96: We should only add new/empty values, otherwise we might overwrite manually adapted/entered messages. Also, LLMs don't produce deterministic output, so it might overwrite existing keys with slightly different messages in every PR. 

```suggestion
    } else if (!output[key]) {
      // overwrite the value (handles string updates or type changes), if no value exists yet
      output[key] = updateValue;
    }
```

larsrickert: The AI prompt is configured to only include keys in the response that are new. Should we still implement a double check here?

.github/workflows/translate.yml
larsrickert: Nice 👍  Using node works just fine, I removed the tsx usage

.github/scripts/translate.ts
JoCa96: yes, please

larsrickert: Done :)

Issue #4397, opened 2025-11-06
Define and implemention actions slot / API for OnyxDataGrid


larsrickert:
### Why?

Currently the OnyxDataGrid does not support passing custom actions to the "actions" slot of the underlying OnyxTable.

We want to introduce a unified API for data grid features to define custom actions.

[...]

PR #4857, opened 2026-02-17
feat(OnyxDataGrid): implement actions slot api

ChristianBusshoff:
Relates to #4397

- implement api

### DIFF
--- a/packages/sit-onyx/src/components/OnyxDataGrid/features/index.ts
+++ b/packages/sit-onyx/src/components/OnyxDataGrid/features/index.ts
@@ -666,23 +666,9 @@ export const useDataGridFeatures = <
   const createSlots = () => {
     const slots: InternalDataGridSlots = {};
 
-    const hasActions = features.some((f) => f.actions != null);
-
-    if (hasActions) {
-      slots.actions = () => {
-        const allActions = features.flatMap((f) => {
-          if (!f.actions) return [];
-          return f.actions();
-        });
-
-        if (allActions.length === 0) return [];
-
-        return [
-          h(DataGridActions, {
-            actions: allActions,
-          }),
-        ];
-      };
+    const actions = features.flatMap((f) => f.actions?.()).filter((action) => action != undefined);
+    if (actions.length) {
+      slots.actions = () => [h(DataGridActions, { actions })];
     }
 
     features.forEach((feature) => {
--- a/packages/sit-onyx/src/components/OnyxDataGrid/features/dataGridActions/DataGridActionButton.vue
+++ b/packages/sit-onyx/src/components/OnyxDataGrid/features/dataGridActions/DataGridActionButton.vue
@@ -31,14 +31,12 @@ const iconButtonProps = useForwardProps(props, OnyxIconButton);
 <template>
   <template v-if="isVisible">
     <OnyxSeparator v-if="props.showSeparator" orientation="vertical" />
-
     <OnyxButton
       v-if="props.displayAs === 'button'"
       ref="componentRef"
       v-bind="buttonProps"
       @click="props.onClick"
     />
-
     <OnyxTooltip v-else :text="props.label">
       <template #default="{ trigger }">
         <OnyxIconButton


packages/sit-onyx/src/components/OnyxDataGrid/features/actionButton/ActionButton.vue
```diff
@@ -0,0 +1,53 @@
...
+      showSeparator: boolean;
+    }
+  >(),
+  {
+    displayAs: "iconButton",
+  },
+);
+
+// eslint-disable-next-line vue/no-setup-props-reactivity-loss -- not relevant for this component
+const restAttr = useForwardProps(props, props.displayAs === "button" ? OnyxButton : OnyxIconButton);
```
larsrickert: Ideally we use this in a reactive way :) Or define them like this:
```ts
const buttonProps = useForwardProps(props, OnyxButton);
const iconButtonProps = useForwardProps(props, OnyxIconButton);
```

packages/sit-onyx/src/components/OnyxDataGrid/features/actionButton/ActionButton.vue
```diff
@@ -0,0 +1,53 @@
...
+</script>
+
+<template>
+  <template v-if="props.displayAs === 'button' && isVisible">
+    <OnyxSeparator v-if="props.showSeparator" orientation="vertical" />
```
larsrickert, line 33: Should we move this to the parent / data grid to add the separators?

packages/sit-onyx/src/components/OnyxDataGrid/features/actionButton/ActionButton.vue
```diff
@@ -0,0 +1,53 @@
...
+    />
+  </template>
+
+  <template v-else-if="isVisible">
+    <OnyxSeparator v-if="props.showSeparator" orientation="vertical" />
+    <OnyxTooltip :text="props.label">
+      <template #default="{ trigger }">
+        <OnyxIconButton
+          ref="componentRef"
+          v-bind="{ ...restAttr, ...trigger }"
```
larsrickert: We can use `mergeVueProps` here

packages/sit-onyx/src/components/OnyxDataGrid/features/actionButton/types.ts
```diff
@@ -0,0 +1,62 @@
...
+   * order inside the group.
+   * @default 0
+   */
+  order?: number;
+  /**
+   * callback function to execute when the action is clicked.
+   * @param config
+   */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any -- TODO: add proper Types
+  onClick: (config: ColumnConfig<any, any, any>[]) => void;
```
larsrickert:
```suggestion
  onClick?: (config: ColumnConfig<any, any, any>[]) => void;
```

Should be optional if e.g. a link is used. Why do we need the column config here?

packages/sit-onyx/src/components/OnyxDataGrid/features/index.ts
```diff
@@ -221,6 +230,10 @@ export type DataGridFeatureDescription<
      */
     order?: number;
   };
+  /**
+   * Allows modifying the action Slot above of the dateGrid
```
larsrickert:
```suggestion
   * Allows modifying the action slot above of the data grid
```

packages/sit-onyx/src/components/OnyxDataGrid/features/index.ts
```diff
@@ -658,9 +671,120 @@ export const useDataGridFeatures = <
...
+        const shouldShowSeparator = (list: ActionProps[], currentIndex: number): boolean => {
+          if (currentIndex <= 0) return false;
+          const getGroupName = (g: ActionGroup | undefined) =>
+            typeof g === "object" && g !== null ? g.name : (g ?? "default");
+          return (
+            getGroupName(list[currentIndex - 1]?.group) !== getGroupName(list[currentIndex]?.group)
+          );
+        };
+
+        return [
```
larsrickert: It might be easier for us if we just create a separate component to include all of this logic where we just pass a list of actions instead of needing to define all these h functions etc.

packages/sit-onyx/src/components/OnyxDataGrid/OnyxDataGridRenderer/OnyxDataGridRenderer.vue
```diff
@@ -76,7 +76,12 @@ const columnStyle = computed(() => {
   .onyx-data-grid {
     width: 100%;
     max-width: 100%;
-
+    .onyx-table-wrapper__actions {
```
larsrickert: Should we move this to the OnyxTable itself? Seems to be general styles

larsrickert (changes requested):
Question: Currently the actions are defined as array so they don't allow for dynamic / different actions depending on state, right? Because they would always be the same for a feature. 

Maybe we should change it to be a function returning the actions instead so they are re-considered on each data grid re-render and can return dynamic / different actions depending on the state

packages/sit-onyx/src/components/OnyxDataGrid/features/index.ts
```diff
@@ -658,9 +666,27 @@ export const useDataGridFeatures = <
...
+
+        if (allActions.length === 0) return [];
+
+        return [
+          h(DataGridActions, {
+            actions: allActions,
+          }),
+        ];
+      };
+    }
```
larsrickert:
```suggestion
    const actions = features.flatMap((f) => f.actions?.()).filter((action) => action != undefined);
    if (actions.length) {
      slots.actions = () => [h(DataGridActions, { actions })];
    }
```

We can already call the action functions at the beginning so we don't have to check for the actions twice :)

PR #4893, opened 2026-02-23
feat: implement `OnyxFormElementV2`

larsrickert:
Relates to #4650

Implement new unified form element component with support for additional slots and label alignments.

## Checklist

### DIFF
--- a/.changeset/green-shrimps-bow.md
+++ b/.changeset/green-shrimps-bow.md
@@ -3,3 +3,7 @@
 ---
 
 feat: implement new `OnyxUnstableFormElementV2` component
+
+This component will replace the `OnyxFormElement` in the future and be the unified foundation for other form elements.
+
+For now, the `OnyxUnstableFormElementV2` component is marked as experimental/unstable which means that it is still under active development and the API might change in patch or minor releases. Keep an eye on the [changelog](https://onyx.schwarz/development/packages/changelogs/sit-onyx.html) when using it.


packages/sit-onyx/src/components/OnyxFormElementV2/OnyxFormElementV2.vue
```diff
@@ -0,0 +1,372 @@
...
+  /**
+   * Optional popover content. If set, a popover will be wrapped around the main input area.
+   * Note: The input should typically be readonly or disabled when using a popover.
+   */
+  popover?(): unknown;
+}>();
+
+const { densityClass } = useDensity(props);
+const formContext = useFormContext(props);
+const errorClass = useErrorClass(formContext.showError);
```
JoCa96, line 29: The validity state itself needs to be managed by the using component?

packages/sit-onyx/src/components/OnyxFormElementV2/OnyxFormElementV2Bottom.vue
```diff
@@ -0,0 +1,118 @@
...
+  };
+};
+
+const message = computed(() => {
+  return (
+    normalizeMessage("danger", props.error) ??
+    normalizeMessage("success", props.success) ??
+    normalizeMessage("neutral", props.message)
+  );
+});
```
JoCa96: All messages should be rendered to the DOM, even if they might not be visible.
This way the messages can still be associated to the input using aria properties.
Otherwise the neutral messages might vanish, even though it is still relevant.

packages/sit-onyx/src/components/OnyxFormElementV2/OnyxFormElementV2Bottom.vue
```diff
@@ -0,0 +1,118 @@
...
+    ..._message,
+    color,
+    // if tooltipText is unset, use the message label so the tooltip is always shown (so the full text is readable even if label is truncated)
+    tooltipText: _message.tooltipText ?? _message.label,
+  };
+};
+
+const message = computed(() => {
+  return (
+    normalizeMessage("danger", props.error) ??
```
JoCa96: We want to use the CSS native `:user-invalid` state to toggle the display state of the error message, so this will not work then: E.g. has not been interacted with, but is required and empty => so actually invalid, but should show default message.

packages/sit-onyx/src/components/OnyxFormElementV2/OnyxFormElementV2.vue
larsrickert: Yes thats the idea for now so the layout is separated from the logic :) But I'm open on this so we can also check in the future how much effort it is to get the validity state integrated and might move this to the layout as well

Issue #4766, opened 2026-02-02
Implement changed icon size of OnyxSystemButton


larsrickert:
### Why?

Follow up for #4670.

### Design

[Figma Component](https://www.figma.com/design/YfEUBOHk4J4nYrk04geswG/Onyx-Components?node-id=7860-25394&m=dev)

### Acceptance criteria

- [x] the icon size is changed to 16px
- [x] the overall size remains at 24px
- [x] other components that use the OnyxSystemButton internally are checked if they still look correctly

### DEV handover checklist

[Storybook](https://storybook.onyx.schwarz/?path=/story/buttons-systembutton--default&globals=outline:!true)

DEV handover done with: @ChristianBusshoff 

**Testing**
- [x] covered by visual screenshot tests
- [x] covered by functional tests (Playwright or unit test)

### UX handover checklist

UX handover done with: @ADD_UX_NAME_HERE

PR #4904, opened 2026-02-24
fix(OnyxSystemButton): Fix incorrect icon size from 20px to 16px

JoCa96:
Relates to #4766


packages/sit-onyx/src/components/OnyxSystemButton/OnyxSystemButton.vue
```diff
@@ -111,7 +111,7 @@ const skeleton = useSkeletonContext(props);
     }
 
     .onyx-icon {
-      --icon-size: 1.125rem;
+      --icon-size: 1rem;
```
larsrickert, line 114: Isn't this obsolete now that we use one of the default OnyxIcon sizes?

JoCa96: This way it also works with slot-content e.g. OnyxTextEditor

PR #4925, opened 2026-02-25
chore: fix flaky tests and update playwright pipeline config

JoCa96:
Relates to #4945

- fix vitest failure, where `GMT+0` is used instead of `GMT`: Apparently caused by a mismatch of the underlying system libraries. As a workaround we align this in the test itself.
- skip navbar headless test, which fails in firefox
- fix flaky `OnyxFilterTag` screenshot test by always hiding the popover
  - added `screenshotOptions` to `playwright-utils` to allow these options from the test setup to the playwright screenshot function
- reduced shards from 16 to 10 to not run into the job limit as often and reduce total runtime


packages/sit-onyx/src/components/OnyxFilterTag/OnyxFilterTag.ct.tsx
```diff
@@ -10,6 +10,9 @@ test.describe("Screenshot tests", () => {
     name: "Filter tag",
     columns: DENSITIES,
     rows: state,
+    screenshotOptions: {
```
larsrickert: Can't we use the existing beforeEach hook / Playwrights test.beforeEach for this?
```ts
        await page.addStyleTag({
          content: `.onyx-tooltip { display: none !important; }`,
        });
```

JoCa96: That would have worked, too :)

Issue #4139, opened 2025-09-15
Fix hydration mismatch errors


JoCa96:
### Version

1.0.0

### Description

The new SSR test in `packages/sit-onyx/src/components.spec.ts` revealed some hydration mismatches and error.
These need to be addressed.

- [ ] Fix all hydration errors
- [ ] Also make sure they are not run in the browser or any virtual DOM environment (e.g., `jsdom`)
- [ ] Check current uses of `onbeforemount`
- [ ] Remove SSR warning/error filtering in SSR tests: 
  - [ ] https://github.com/SchwarzIT/onyx/blob/main/packages/sit-onyx/src/components.spec.ts#L354-L360

### Implementation Details

- Use [`onbeforemount`](https://vuejs.org/api/composition-api-lifecycle.html#onbeforemount) so that it is excluded from SSR (if it makes sense)
- Check current uses of `onbeforemount`

### Reproduction steps

Remove the `filter` functions from the test in `packages/sit-onyx/src/components.spec.ts`:
[errorSpy.mock.calls.filter((c) => c.at(0) !== "Hydration completed but contains mismatches."),](https://github.com/SchwarzIT/onyx/blob/main/packages/sit-onyx/src/components.spec.ts#L283)
[warningSpy.mock.calls.filter((c) => !c.at(0).startsWith("[Vue warn]: Hydration")),](https://github.com/SchwarzIT/onyx/blob/main/packages/sit-onyx/src/components.spec.ts#L286)

### Test result
```
❯  sit-onyx  src/components.spec.ts (93 tests | 12 failed) 473ms
   ✓ components > should mount 'OnyxAccordion' without errors and console logs 15ms
   ✓ components > should mount 'OnyxAccordionItem' without errors and console logs 5ms
   ✓ components > should mount 'OnyxAlertModal' without errors and console logs 2ms
   ✓ components > should mount 'OnyxAppLayout' without errors and console logs 3ms
   ✓ components > should mount 'OnyxAvatar' without errors and console logs 2ms
   ✓ components > should mount 'OnyxAvatarStack' without errors and console logs 1ms
   ✓ components > should mount 'OnyxBadge' without errors and console logs 1ms
   ✓ components > should mount 'OnyxBottomBar' without errors and console logs 1ms
   ✓ components > should mount 'OnyxBreadcrumb' without errors and console logs 9ms
   ✓ components > should mount 'OnyxBreadcrumbItem' without errors and console logs 2ms
[...]

PR #4973, opened 2026-03-03
fix: prevent hydration mismatches

larsrickert:
Relates to #4139

Enable unit tests for hydration mismatches and fix all related errors/warnings.

## Checklist

### DIFF
--- a/packages/sit-onyx/src/components/OnyxVisuallyHidden/OnyxVisuallyHidden.vue
+++ b/packages/sit-onyx/src/components/OnyxVisuallyHidden/OnyxVisuallyHidden.vue
@@ -12,7 +12,7 @@ const props = withDefaults(defineProps<OnyxVisuallyHiddenProps>(), { is: "span" 
 </script>
 
 <template>
-  <!-- this v-if check is needed to prevent hydration errors in SSR -->
+  <!-- This v-if check is needed to prevent hydration errors in SSR: It ensures that self-closing elements (like `<input />`) are rendered correctly -->
   <component :is="props.is" v-if="slots.default" class="onyx-component onyx-visually-hidden">
     <slot></slot>
   </component>


packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxColorSchemeMenuItem/OnyxColorSchemeMenuItem.vue
```diff
@@ -32,7 +36,7 @@ const isOpen = ref(false);
 
     <!-- the menu button renders a <li> and <button> so we need to teleport the dialog
       to not nest it inside the button -->
-    <Teleport to="body">
+    <Teleport v-if="isMounted" to="body">
```
JoCa96, line 39: Would `defer` work here? https://vuejs.org/guide/built-ins/teleport.html#deferred-teleport

```suggestion
    <Teleport defer to="body">
```

packages/sit-onyx/src/composables/useAnchorPositionPolyfill.ts
```diff
@@ -134,7 +135,7 @@ export const useAnchorPositionPolyfill = ({
    * SSR safe composable for checking whether the browser supports the Anchor API.
    * TODO: can be removed after anchor is implemented in all common browsers.
    */
-  onBeforeMount(() => {
+  onMounted(() => {
```
JoCa96, line 138: Why does it make a difference, afaik `onBeforeMount` is also not run on the server?

packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxColorSchemeMenuItem/OnyxColorSchemeMenuItem.vue
larsrickert: I tried this already but it didn't work in this case

packages/sit-onyx/src/composables/useAnchorPositionPolyfill.ts
larsrickert: tbh, I don't know 😅  I thought the same but using onBeforeMount still causes hydration errors

JoCa96: I assume the first render result (so after `onBeforeMounted`) is compared, so having a second render cycle works around the hydration mismatch

packages/sit-onyx/src/components/OnyxVisuallyHidden/OnyxVisuallyHidden.vue
```diff
@@ -1,18 +1,23 @@
...
-  default: () => unknown;
+  default?: () => unknown;
 }>();
 
 const props = withDefaults(defineProps<OnyxVisuallyHiddenProps>(), { is: "span" });
 </script>
 
 <template>
-  <component :is="props.is" class="onyx-component onyx-visually-hidden"><slot></slot></component>
+  <!-- this v-if check is needed to prevent hydration errors in SSR -->
```
JoCa96:
```suggestion
  <!-- This v-if check is needed to prevent hydration errors in SSR: It ensures that self-closing elements (like `<input />`) are rendered correctly -->
```

Issue #4737, opened 2026-01-27
Implement range mode for TimePicker


ChristianBusshoff:
### Why?

we want to offer a range mode

### Design

[Figma Documentation](https://www.figma.com/design/j7Hy7Zk8MiCmJBJf0ArQRm/Onyx-Documentation?node-id=17388-13890&m=dev)
[Figma Component](https://www.figma.com/design/YfEUBOHk4J4nYrk04geswG/Onyx-Components?node-id=511-199&p=f&m=dev)

### Acceptance criteria

- [ ] add range type
- [ ] the flyout opens when the timepicker gets focused
- [ ] the input is done via the flyout
- [ ] the timepicker input just shows the value (can't be edited)

### DEV handover checklist

[Storybook](ADD_LINK_HERE)

DEV handover done with: @ADD_DEV_NAME_HERE

**Testing**
- [ ] covered by visual screenshot tests
- [ ] covered by functional tests (Playwright or unit test)

**Documentation**
- [ ] "New" badge in Storybook is added, see [storybook-addon-tag-badges](https://github.com/Sidnioulz/storybook-addon-tag-badges/?tab=readme-ov-file#-usage)
- [ ] Storybook code snippets are checked
- [ ] component status has been updated in the [docs overview](https://onyx.schwarz/#components)

**Other**
- [ ] follow-up tickets were created if necessary (add links below)

### UX handover checklist

UX handover done with: @ADD_UX_NAME_HERE

**Naming is aligned with Figma**
- [ ] component name
- [ ] which support components are used and how they are named
- [ ] all property names and types

**Implementation is checked**
- [ ] all features of Figma are implemented or a follow-up ticket is refined
- [ ] skeleton
- [ ] densities
- [ ] light and dark mode
- [ ] truncation
- [ ] responsiveness / wrapping behavior
- [ ] min / max width

**Figma**
- [ ] Figma component is published

**Documentation**
- [ ] Storybook category is aligned with Figma
- [ ] Storybook examples are checked

**Other**
- [ ] follow-up tickets were created if necessary (add links below)

PR #4989, opened 2026-03-05
feat(TimePicker): implement range mode

ChristianBusshoff:
Relates to #4737

- added range mode, 
- enabled Keyboard test for timepicker flyout

Questions:
 - for the range mode, the error never show if showErrror = touched.


packages/sit-onyx/playwright/snapshots/components/OnyxTimePicker/Time-picker-range-default--chrome-linux.png
larsrickert, line 1: Shouldn't the flyout be open on focus?

packages/sit-onyx/src/components/OnyxTimePicker/OnyxTimePickerInput.vue
```diff
@@ -49,6 +54,10 @@ const emit = defineEmits<{
    * Emitted when the input changes focus.
    */
   "update:isFocused": [focused: boolean];
+  /**
+   * Emitted when the input is clicked or focused and enter is pressed.
+   */
+  toggleOpen: [event: MouseEvent | KeyboardEvent];
```
larsrickert: Usually we have a `open` property and corresponding `update:open` event. Should we align this here?

packages/sit-onyx/src/components/OnyxTimePicker/OnyxTimePickerInput.vue
```diff
@@ -57,15 +66,25 @@ const emit = defineEmits<{
 
 defineOptions({ inheritAttrs: false });
 const { rootAttrs, restAttrs } = useRootAttrs();
-const { vCustomValidity, errorMessages } = useFormElementError({ props, emit });
+const error = computed(() => props.error);
+const mappedProps = reactive({
+  ...props,
+
+  type: computed(() => (props.type === "range" ? "text" : "date")) as unknown as "text",
```
larsrickert:
```suggestion
  type: computed(() => (props.type === "range" ? "text" : "date")),
```

The type cast does not seem to be needed, why is it needed?

packages/sit-onyx/src/components/OnyxTimePicker/OnyxTimePickerInput.vue
```diff
@@ -163,6 +182,22 @@ useAutofocus(useTemplateRef("input"), props);
             @keydown.space.prevent
             @focus="emit('update:isFocused', true)"
             @blur="emit('update:isFocused', false)"
+            @click="emit('toggleOpen', $event)"
+            @paste="(e) => props.type === 'range' && e.preventDefault()"
+            @keydown="
```
larsrickert: Would prefer to move this to the `<script>` part + we can use the same logic as in the OnyxSelectInput:

```ts
const blockTyping = (event: KeyboardEvent) => {
  if (navigationalKeys.includes(event.key)) return;
  event.preventDefault();
};
```

packages/sit-onyx/src/components/OnyxTimePicker/OnyxTimeInput.vue
```diff
@@ -27,12 +25,16 @@ const emit = defineEmits<{
...
-const hourInput = useTemplateRef<InputRef>("hourInput");
-const minuteInput = useTemplateRef<InputRef>("minuteInput");
-const secondInput = useTemplateRef<InputRef>("secondInput");
-const root = useTemplateRef("root");
+const timePickerGroup =
+  useTemplateRef<InstanceType<typeof OnyxTimePickerGroup>>("timePickerGroupRef");
+const starttimePickerGroup =
+  useTemplateRef<InstanceType<typeof OnyxTimePickerGroup>>("startTimePickerGroupRef");
+const endtimePickerGroup =
+  useTemplateRef<InstanceType<typeof OnyxTimePickerGroup>>("endTimePickerGroupRef");
```
larsrickert: Why are the generics needed here? Removing them seems to work fine

packages/sit-onyx/src/components/OnyxTimePicker/OnyxTimeInput.vue
```diff
@@ -59,138 +61,96 @@ const parseTimeSeconds = (timeString?: string): number | null => {
 const minTimeSeconds = computed(() => parseTimeSeconds(props.min));
 const maxTimeSeconds = computed(() => parseTimeSeconds(props.max));
 
-const clampTime = (newParts: string[]): string[] => {
-  const currentTotalSeconds = partsToTotalSeconds(newParts);
+const clampTime = (time: string | undefined): string | undefined => {
```
larsrickert:
```suggestion
const clampTime = (time?: string): string | undefined => {
```

packages/sit-onyx/src/components/OnyxTimePicker/OnyxTimeInput.vue
```diff
@@ -59,138 +61,96 @@ const parseTimeSeconds = (timeString?: string): number | null => {
...
 
-  if (e.key === "Tab") {
-    if (currentSegment !== availableSegments.value[availableSegments.value.length - 1]) {
-      e.preventDefault();
-      jumpSegment(currentSegment, 1);
-    }
-    return;
-  }
+  await nextTick();
+  await nextTick();
```
larsrickert: Why do need to wait twice?

packages/sit-onyx/src/components/OnyxTimePicker/OnyxTimeInput.vue
```diff
@@ -249,58 +267,46 @@ const showClearButton = computed(() => {
...
-                hide-label
-                hide-clear-icon
-                hide-success-icon
-                hide-buttons
-                @click="handleSegmentFocus(secondInput)"
-                @keydown="(e: KeyboardEvent) => handleInputChange('second', e)"
-              />
-            </template>
-          </div>
+          <template v-else>
```
larsrickert: Can we move the v-if to the OnyxTimePickerGroup?

packages/sit-onyx/src/components/OnyxTimePicker/OnyxTimePickerGroup.vue
```diff
@@ -0,0 +1,184 @@
...
+
+  // Jump immediately if the first digit is greater than the max possible first digit
+  // e.g. for hours, if the user types "3", it can only be "03", so we can jump.
+  const maxFirstDigit = currentSegment === "hour" ? 2 : 5;
+  if (Number.parseInt(digit, 10) > maxFirstDigit) {
+    setTimeout(() => emit("jump-segment", currentSegment, 1), 0);
+    return;
+  }
+
+  setTimeout(() => {
```
larsrickert, line 110: Why do we need the setTimeout? Also in the code above

larsrickert (changes requested):
Currently the `modelValue` for range mode is still a string. Imo it would be a better DX if we would e.g. use an object or and array with 2 values

packages/sit-onyx/src/components/OnyxTimePicker/OnyxTimePickerGroup.vue
ChristianBusshoff: To jump directly after the second number got inputted

packages/sit-onyx/src/components/OnyxTimePicker/OnyxTimeInput.vue
ChristianBusshoff: to wait for the popover to be open.

Issue #4736, opened 2026-01-27
Implement AM/PM for OnyxTimepicker


ChristianBusshoff:
### Why?

We want to provide a Am|Pm variant for the timePicker

### Open questions

- [x] @Guergchen How to handle the AM/PM selection flyout? => We don't have a Flyout currently
- [x] @Guergchen Are there any other use-cases for the toggle component that is used by the AM/PM thing

=> will be adjusted by ux in #4866 

<img width="52" height="57" alt="Image" src="https://github.com/user-attachments/assets/551ed9db-fed9-43d8-8342-131c0758d560" />

### Depends on

#4734

### Design

[Figma Documentation](https://www.figma.com/design/j7Hy7Zk8MiCmJBJf0ArQRm/Onyx-Documentation?node-id=17388-13890&m=dev)
[Figma Component](https://www.figma.com/design/YfEUBOHk4J4nYrk04geswG/Onyx-Components?node-id=511-199&p=f&m=dev)

### Acceptance criteria

- [ ] add `locale` prop, which accepts string or Intl.Locale
  - [ ] injected by default
  - [ ] use `hourCycle` to determine if AM/PM should be shown
  - [ ] add comment: currently only affects AM/PM
- [ ] AM/PM toggle should always toggle, independently if Am or Pm is clicked
- [ ] implement am|pm functionality

### Implementation details

Align with Intl API: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl

### DEV handover checklist

[Storybook](https://onyx-storybook-dev.apps.01.cf.eu01.stackit.cloud/?path=/story/form-elements-timepicker--range&args=open:!true;showAmPm:!true)

DEV handover done with: @larsrickert 

**Testing**
- [ ] covered by visual screenshot tests
- [ ] covered by functional tests (Playwright or unit test)

**Documentation**
- [ ] "New" badge in Storybook is added, see [storybook-addon-tag-badges](https://github.com/Sidnioulz/storybook-addon-tag-badges/?tab=readme-ov-file#-usage)
- [ ] Storybook code snippets are checked
- [ ] component status has been updated in the [docs overview](https://onyx.schwarz/#components)

**Other**
- [ ] follow-up tickets were created if necessary (add links below)

### UX handover checklist

UX handover done with: @ADD_UX_NAME_HERE

**Naming is aligned with Figma**
- [ ] component name
- [ ] which support components are used and how they are named
- [ ] all property names and types

[...]

PR #5087, opened 2026-03-18
feat(TimePicker): implement am/pm

ChristianBusshoff:
Relates to #4736

- added am/pm

Question: FormElementV2 provides the aria-label for the input, but I got an Error without : vuejs-accessibility/form-control-has-label


packages/sit-onyx/src/components/OnyxTimePicker/OnyxTimePicker.stories.ts
```diff
@@ -31,6 +31,12 @@ export const WithSeconds = {
   },
 } satisfies Story;
 
+export const WithAmPm = {
+  args: {
```
larsrickert, line 36:
```suggestion
  tags: ["new:feature"],
  args: {
```

packages/sit-onyx/src/components/OnyxTimePicker/OnyxTimePickerInput.vue
```diff
@@ -106,15 +103,121 @@ const modelValue = useVModel({
...
 });
 
+const timeSuffix = ref("am");
+
+watch(
+  modelValue,
+  (newVal) => {
+    if (props.type === "range") return;
+
+    if (!props.showAmPm || !newVal) return;
```
larsrickert:
```suggestion
    if (props.type === "range" || !newVal) return;
```

We should still execute the code even when "showAmPm" is false because it might be enabled afterwards. Currently when amPm is disabled and the user enters e.g. "14:30", when enabling apPm, the value is incorrectly set to am although it should be pm :) Can you also create Playwright tests für the AM/PM functionality in general?

packages/sit-onyx/src/components/OnyxTimePicker/TimeSuffixToggle.vue
```diff
@@ -0,0 +1,107 @@
...
+
+    &__slider {
+      position: absolute;
+      top: var(--track-padding);
+      left: var(--track-padding);
+      right: var(--track-padding);
+      height: calc(50% - var(--track-padding));
+      background-color: var(--onyx-color-base-background-blank);
+      border-radius: var(--onyx-radius-sm);
+      transition: transform 0.3s cubic-bezier(0.4, 0, 0.2, 1);
```
larsrickert:
```suggestion
      transition: transform var(--onyx-duration-sm) cubic-bezier(0.4, 0, 0.2, 1);
```

PR #5131, opened 2026-03-24
chore: increase playwright performance by creating traces on retry instead of retain

JoCa96:
- Traces for playwright are not captured by default anymore, instead only on retries. A retry is performed once on failure. Flaky tests count as failed tests.
- Updated `gh:show-report` script to accept the run-id as argument. This makes it easy to check specifc executions like for merge group requests.
- Fix(useMatrixScreenshotTest): reduce flakiness by waiting for all images to have been loaded before performing the screenshot comparison
- Fix(expectEmit): reduce flakiness by allowing retries when checking for the emit


larsrickert (approved):

packages/playwright-utils/src/emits/index.ts
larsrickert, line 1: Shouldn't we also make `expectEmit` itself async and await `test.step()` so the async check is awaited correctly?

larsrickert (approved):

PR #5133, opened 2026-03-24
feat(OnyxTimePicker): implement small changes

ChristianBusshoff:
Relates to #4736

- implemented focus outline for am/pm Toggle
- The toggle now automatically jumps from AM to PM if a value greater than 12 is entered for range mode.


packages/sit-onyx/src/components/OnyxTimePicker/types.ts
```diff
@@ -78,7 +78,7 @@ export type OnyxTimePickerProps<TType extends TimePickerType = "default"> = Pick
     /**
      * Whether to use am/pm
```
larsrickert:
```suggestion
     * Whether to use am/pm.
     * If set to "auto", the current locale will be used to determine if it is am/pm.
```

packages/sit-onyx/src/components/OnyxTimePicker/OnyxTimePickerInput.vue
larsrickert, line 1: Should we check with UX what should be the default value? false or "auto"?

ChristianBusshoff: I checked it. Should be set to auto

larsrickert: So now all our examples show the am/pm since we use en-US as default^^

ChristianBusshoff: yes, should I force false for the storybook examples

Issue #5138, opened 2026-03-24
Implement updated border radius variables


larsrickert:
### Why?

In #5036, UX concepted new border radius tokens for different themes.
We now want to implement them in the corresponding components

### Design

<img width="443" height="639" alt="Image" src="https://github.com/user-attachments/assets/1874b01f-58c0-4114-8d17-179c74431a94" />

### Acceptance criteria

- [x] the [Import Figma variables](https://github.com/SchwarzIT/onyx/actions/workflows/import-figma.yml) action is triggered to get the new variables
- [x] all new border radius variables are implemented in their corresponding components
- [x] all internal onyx themes are updated to include the new variables

### DEV handover checklist

[Storybook](https://onyx-storybook-dev.apps.01.cf.eu01.stackit.cloud/?path=/docs/buttons-button--docs)

DEV handover done with: @ADD_DEV_NAME_HERE

**Testing**
- [ ] covered by visual screenshot tests
- [ ] covered by functional tests (Playwright or unit test)

**Documentation**
- [ ] "New" badge in Storybook is added, see [storybook-addon-tag-badges](https://github.com/Sidnioulz/storybook-addon-tag-badges/?tab=readme-ov-file#-usage)
- [ ] Storybook code snippets are checked
- [ ] component status has been updated in the [docs overview](https://onyx.schwarz/#components)

**Other**
- [ ] follow-up tickets were created if necessary (add links below)

### UX handover checklist

UX handover done with: @ADD_UX_NAME_HERE

**Naming is aligned with Figma**
- [ ] component name
- [ ] which support components are used and how they are named
- [ ] all property names and types

**Implementation is checked**
- [ ] all features of Figma are implemented or a follow-up ticket is refined
- [ ] skeleton
- [ ] densities
- [ ] light and dark mode
- [ ] truncation
- [ ] responsiveness / wrapping behavior
- [ ] min / max width

**Figma**
- [ ] Figma component is published

**Documentation**
- [ ] Storybook category is aligned with Figma
- [ ] Storybook examples are checked

**Other**
- [ ] follow-up tickets were created if necessary (add links below)

PR #5175, opened 2026-03-27
feat(theme): implement updated border radius variables

ChristianBusshoff:
Relates to #5138

- implemented new radius variables


packages/sit-onyx/src/components/OnyxFormElementV2/OnyxFormElementV2.vue
```diff
@@ -162,7 +162,7 @@ const popoverLayoutProps = useForwardProps(props, MaybePopoverLayout);
 .onyx-form-element-v2 {
   @include layers.component() {
     --onyx-form-element-v2-gap: var(--onyx-density-3xs);
-    --onyx-form-element-v2-border-radius: var(--onyx-radius-sm);
+    --onyx-form-element-v2-border-radius: var(--onyx-radius-component-input);
```
larsrickert, line 165: Do we also need to change this in the OnyxFormElement so the "old" components that are not yet using the OnyxFormElementV2 are updated?

packages/sit-onyx/src/styles/variables/themes/onyx.css
```diff
@@ -428,9 +427,24 @@
   --onyx-font-size-xl: var(--onyx-font-values-500);
   --onyx-font-weight-regular: var(--onyx-font-weight-400);
   --onyx-font-weight-semibold: var(--onyx-font-weight-600);
+  --onyx-radius-component-badge: var(--onyx-number-radius-600);
+  --onyx-radius-component-button: var(--onyx-number-radius-200);
+  --onyx-radius-component-checkbox: var(--onyx-number-radius-200);
+  --onyx-radius-component-container-sharp: var(--onyx-number-radius-200);
+  --onyx-radius-component-container-soft: var(--onyx-number-radius-300);
```
larsrickert, line 434: Can you check with UX if those container variables are needed? I remember a discussion with Flo that they are actually unused and we might remove them

packages/sit-onyx/src/components/OnyxBasicDialog/OnyxBasicDialog.vue
```diff
@@ -121,7 +121,7 @@ defineExpose({
 .onyx-basic-dialog {
   @include layers.component() {
     --onyx-basic-dialog-screen-gap: var(--onyx-grid-margin);
-    --onyx-basic-dialog-border-radius: var(--onyx-radius-md);
+    --onyx-basic-dialog-border-radius: var(--onyx-radius-component-flyout);
```
larsrickert, line 124: The OnyxBasicDialog is used for the modal etc, is it intended by UX to apply the same radius than for flyouts?

packages/sit-onyx/src/components/OnyxBasicPopover/OnyxBasicPopover.vue
```diff
@@ -240,7 +240,7 @@ const popoverStyles = computed(() => {
 
     &__dialog {
       position: fixed;
-      border-radius: var(--onyx-radius-md);
+      border-radius: var(--onyx-radius-component-flyout);
```
larsrickert, line 243: Should this also apply to the Onyxtooltip?

packages/sit-onyx/src/styles/variables/themes/onyx.css
ChristianBusshoff: I checked it with Flo. The vars ar not needed at the moment, as well as --onyx-radius-component-search vars. But in the future there will be probably used.

packages/sit-onyx/src/components/OnyxBasicDialog/OnyxBasicDialog.vue
ChristianBusshoff: Yes

packages/sit-onyx/src/components/OnyxBasicPopover/OnyxBasicPopover.vue
ChristianBusshoff: no, the tooltip should stay the same

Issue #5127, opened 2026-03-24
Implement basic Onyx MCP Server


mj-hof:
### Why?

…

### Acceptance criteria

Basic MCP functionality has been implemented:

- [ ] ~getTokens~
- [x] getComponents
- [x] getComponent: API
- [x] getIcons

### Implementation details

- https://modelcontextprotocol.io/docs/getting-started/intro

### Reference implementations

- https://vuetifyjs.com/en/getting-started/upgrade-guide/#quick-start-with-vuetify-mcp
- https://ui.nuxt.com/docs/getting-started/ai/mcp

### DEV handover checklist

[Storybook](ADD_LINK_HERE)

DEV handover done with: @ADD_DEV_NAME_HERE

**Testing**
- [x] Covered by functional tests (Playwright or unit test)

**Documentation**
- [x] Has been created

**Other**
- [ ] Follow-up tickets were created if necessary (add links below)

PR #5213, opened 2026-04-01
feat: Generate and publish 'component-meta.json' which includes the onyx API as json

JoCa96:
Relates to #5127 

- Implement `extractComponentMeta` plugin to generate `component-meta.json` which can be used to e.g. generate MCP content
- Removed `tsx` work around in `build/index.js` and migrated script to TypeScript.
- Use `vite.config.prod.ts` for storybook build and remove workaround in `vite.config.ts`
- Use `pnpm exec` instead of `npx` in PR template

### DIFF
--- a/packages/sit-onyx/build/extract-component-meta.ts
+++ b/packages/sit-onyx/build/extract-component-meta.ts
@@ -173,7 +173,7 @@ export async function extractComponentMeta(
       this.emitFile({
         type: "asset",
         fileName,
-        source: JSON.stringify(json, null, 2),
+        source: JSON.stringify(json),
       });
     },
   };


packages/sit-onyx/build/extract-component-meta.ts
```diff
@@ -0,0 +1,358 @@
...
+    generateBundle() {
+      const json = collect
+        .flat()
+        .filter(filterMeta)
+        .map(({ sourceFiles: _, ...meta }) => meta);
+
+      this.emitFile({
+        type: "asset",
+        fileName,
+        source: JSON.stringify(json, null, 2),
```
larsrickert:
```suggestion
        source: JSON.stringify(json),
```

We could shrink the JSON file size by ~40% if we omit whitespaces. The file will most likely be used by code anyway so it will not be "looked at" by users, right?

JoCa96:
> Should we add a changeset?

Yep, forgot to push

Issue #2997, opened 2025-03-20
Implement `drilldown` mode `external` for OnyxFlyoutMenu


JoCa96:
### Why?

As a designer I want to be able to have flyouts with drill-down lists, so that nested options in navigation menus can be defined.

### Depends on

- #2996 
- #2982 

### Design

[Figma](https://www.figma.com/design/EBi98T51T9fGfjrYYfsulg/onyx---Archive?node-id=3427-53313&m=dev)

![Image](https://github.com/user-attachments/assets/3b6b8857-3396-4677-ba7b-1a39eba35548)

### Acceptance criteria

- [x] The property `drilldownMode` is extended with the `external` option
- [x] Nested items are passed via the `children` slot of the `OnyxListItem` component
- [x] The HTML/DOM structure reflects the hierachy of the items
- [x] Initially only the parent element is shown in the list
- [x] Hovering a parent opens the popover with the children 
- [x] Nesting can be done recursively without a limit
- [x] When a parent is closed, all nested parents are also closed

In OnyxNavBar, the MoreList dropdown supports this, too:
- [ ] if hidden elements contain children, they should still be shown in the more indicator

**out-of-scope:**
- animation
- dynamic/smart placment of the child flyouts

### Implementation details

We want to use nested popovers: 
https://play.vuejs.org/#eNp9kj1PwzAQhv/KyRNIVTPAFNJKgDqUARB09FKSa+ri2JY/QqUo/x3bqdMW0W539z5nv/ZdRx6VmrYOSU4KU2qmLBi0Ts2pYI2S2sKzbBRstGyAkmkWssBTQkWRDR2e9YnFRvG1RZ8BFIGL0XmcsiIelICz5K8+hkV2cgWZEGtKKTasnu6MFN5/F2hKSk8zjvpNWSaFoSSHqARtzbn8eYk1qx1OUr3cYvn9T31n9qFGybtGg7r1zx41u9Y12kFefL7i3sej2MjKcU9fET/QSO6CxwF7cqLytk+46HYZp8BEvTKLvUVh0qOC0UD2kafEDyV81KWnH+3eTe9jHxW9/8U00MsL0IEzuKygT1sQx/9Ahf9+Y4FVMBuIm1tfvLYUFWvTlL+ctVJArqSSLerB3YwSVlEyX8m65gh2i3DQoevCRX1fZENnOscfCTmrDp0JP26b4dJCNi5acpCi853qfwFcsABD

### Reference implementations

[...]

PR #5244, opened 2026-04-08
feat(FlyoutMenu): implement external drilldown

ChristianBusshoff:
Relates to #2997

- implement external drilldown


packages/sit-onyx/playwright/snapshots/components/OnyxNavBar/modules/OnyxFlyoutMenu/nested-external-open-level-2-chrome-linux.png
larsrickert, line 1: UX question: Is it intended that the nested flyouts do not have the top/bottom padding outside of the primary color?

packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxFlyoutMenu/OnyxFlyoutMenu.ct.tsx
```diff
@@ -293,3 +294,104 @@ test("should keep nested children open state when toggling", async ({ page, moun
...
+
+  // ACT
+  await page.keyboard.press("ArrowDown");
+
+  // ASSERT
+  await expect(firstItem).toBeFocused();
+  await expect(component).toHaveScreenshot("nested-external-open-level-1.png");
+
+  // ACT
+  await page.keyboard.press("ArrowRight");
```
larsrickert, line 369: If the drilldown opens to the left (e.g. because there is no space available on the right), does it correctly navigate with ArrowLeft then or does it still expect arrow right?

packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxFlyoutMenu/OnyxFlyoutMenu.stories.ts
```diff
@@ -133,3 +133,55 @@ export const Nested = {
...
+
+    options: () => [
+      h(
+        OnyxMenuItem,
+        { label: "Item 1", nested: "external" },
+        {
+          children: () => [
+            h(
+              OnyxMenuItem,
+              { label: "Item 1.1", nested: "external" },
```
larsrickert: It would be nice if the nested items would inherit the mode automatically so the developer does not have to set nested="external" to every menu item :)

packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxMenuItem/types.ts
```diff
@@ -28,4 +28,10 @@ export type OnyxMenuItemProps = WithLinkProp & {
    * Main color of the item content.
    */
   color?: Extract<OnyxColor, "primary" | "danger">;
+  nested?: "internal" | "external";
```
larsrickert: Would be great to have some JSDoc here that explains the two modes

packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxMenuItem/OnyxMenuItem.vue
```diff
@@ -89,18 +115,63 @@ const childrenClickHandler = computed(() =>
...
+  handleTriggerMouseEnter();
+  parentMenu?.onHoverEnter();
+};
+
+const handlePopoverMouseLeave = () => {
+  handleTriggerMouseLeave();
+  parentMenu?.onHoverLeave();
+};
+
+provide<NestedMenuContext>("onyx-nested-menu", {
```
larsrickert: We should use an injection key instead of a string for better type safety and to prevent naming colissions:

```ts
export const MENU_ITEM_INJECTION_KEY = Symbol() as InjectionKey<{
  onHoverEnter: () => void;
  onHoverLeave: () => void;
}>;

// when injecting
const parentMenu = inject(MENU_ITEM_INJECTION_KEY, undefined);

// when providing
provide(MENU_ITEM_INJECTION_KEY, {
  onHoverEnter: handlePopoverMouseEnter,
  onHoverLeave: handlePopoverMouseLeave,
});
```

packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxMenuItem/OnyxMenuItem.vue
```diff
@@ -52,17 +62,35 @@ const open = useVModel({
...
 const {
   elements: { listItem, menuItem },
 } = createMenuItems({
   onOpen: async () => {
     if (!hasChildren.value) return;
     open.value = true;
+
     await nextTick();
-    backButton.value?.$el.querySelector("button").focus();
+    await nextTick();
```
larsrickert: Why is a second nextTick needed here?

packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxMenuItem/OnyxMenuItem.vue
```diff
@@ -52,17 +62,35 @@ const open = useVModel({
...
+
+    if (!isExternal.value) {
+      backButton.value?.$el.querySelector("button")?.focus();
+    } else {
+      // Focus the first actionable item in the external popover slot
+      const firstFocusable = externalChildrenRef.value?.querySelector("button, a") as
+        | HTMLElement
+        | undefined;
```
larsrickert:
```suggestion
      const firstFocusable = externalChildrenRef.value?.querySelector<HTMLElement>("button, a");
```

packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxFlyoutMenu/OnyxFlyoutMenu.stories.ts
larsrickert, line 1: When using keyboard navigation and navigating from the second flyout back to the first one, a flyout glitches at the top

https://github.com/user-attachments/assets/f7db6952-9106-4fcd-8759-7fc9c22926b3

packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxMenuItem/OnyxMenuItem.vue
```diff
@@ -89,18 +115,63 @@ const childrenClickHandler = computed(() =>
...
+  if (event.key === "ArrowLeft") {
+    event.preventDefault();
+    event.stopPropagation();
+    await closeAndFocusTrigger();
+  }
+};
+
+let hoverTimeout: ReturnType<typeof setTimeout>;
+
+onBeforeUnmount(() => {
```
larsrickert: Is it possible to get rid of the timeouts by using the "debounce" helper from @sit-onyx/headless?

packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxMenuItem/OnyxMenuItem.vue
```diff
@@ -110,44 +181,102 @@ const handleBackButtonKeydown = async (event: KeyboardEvent) => {
...
+          @click.stop="open = false"
+        >
+          <OnyxIcon :icon="iconArrowSmallLeft" />
+          {{ t("back") }}
+        </OnyxMenuItem>
+        <slot name="children"></slot>
+      </ul>
+    </template>
+
+    <template v-else>
```
larsrickert: Since we only have one element, we can move the `v-else` to the OnyxBasicPopover directly :)

packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxMenuItem/OnyxMenuItem.vue
```diff
@@ -110,44 +181,102 @@ const handleBackButtonKeydown = async (event: KeyboardEvent) => {
...
+        position="auto-inline"
+        alignment="auto"
+        class="onyx-menu-item__popover"
       >
-        <OnyxIcon :icon="iconArrowSmallLeft" />
-        {{ t("back") }}
-      </OnyxMenuItem>
-      <slot name="children"></slot>
-    </ul>
+        <template #default>
```
larsrickert: WE can remove this wrapper since just placing the content inside the OnyxBasicPopover will already use the "default" slot

packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxMenuItem/OnyxMenuItem.vue
```diff
@@ -110,44 +181,102 @@ const handleBackButtonKeydown = async (event: KeyboardEvent) => {
...
+                <span class="onyx-truncation-ellipsis">
+                  {{ props.label }}
+                </span>
+                <OnyxExternalLinkIcon v-bind="props.link ? extractLinkProps(props.link) : {}" />
+              </span>
+            </slot>
+
+            <div v-if="hasChildren" class="onyx-menu-item__chevron">
+              <OnyxIcon :icon="iconChevronRightSmall" size="24px" />
+            </div>
```
larsrickert: This code is somewhat duplicated to the internal drilldown. Maybe we can extract it to a support component or something so we can maintain it in a single place?

.changeset/every-drinks-warn.md
```diff
@@ -0,0 +1,5 @@
+---
+"sit-onyx": minor
+---
+
+feat(OnyxFlyoutMenu): implement external drilldown mode
```
larsrickert:
```suggestion
feat(OnyxFlyoutMenu): implement external drilldown mode using the new `nested="external"` property
```

larsrickert (dismissed):

packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxMenuItem/OnyxMenuItem.vue
ChristianBusshoff: Yes, otherwise it don't work

packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxMenuItem/types.ts
```diff
@@ -28,4 +36,17 @@ export type OnyxMenuItemProps = WithLinkProp & {
    * Main color of the item content.
    */
   color?: Extract<OnyxColor, "primary" | "danger">;
+  /**
+   * Defines how nested menu items are rendered and behave.
+   * * - `internal`: Nested items expand within current menu.
```
JoCa96:
```suggestion
   * - `internal`: Nested items expand within current menu.
```

packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxMenuItem/OnyxMenuItem.vue
```diff
@@ -52,18 +65,60 @@ const open = useVModel({
...
+
+const handleOpen = async () => {
+  if (!hasChildren.value) return;
+  open.value = true;
+
+  await nextTick();
+  await nextTick();
+
+  if (!isExternal.value) {
+    backButton.value?.$el.querySelector("button")?.focus();
```
JoCa96: If we need access to internals of components like specific elements or functions, we should use `defineExpose` or a `useTemplateRef`.
Using `querySelector` is usually a big indicator for an Anti-Pattern.

```suggestion
    backButton.value?.$el.querySelector("button")?.focus();
```

packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxMenuItem/OnyxMenuItem.vue
```diff
@@ -110,44 +232,81 @@ const handleBackButtonKeydown = async (event: KeyboardEvent) => {
...
+          @click.stop="closeAndFocusTrigger"
+        >
+          <OnyxIcon :icon="iconArrowSmallLeft" />
+          {{ t("back") }}
+        </OnyxMenuItem>
+        <slot name="children"></slot>
+      </ul>
+    </template>
+
+    <OnyxBasicPopover
```
JoCa96, line 278: `OnyxBasicPopover` or rather `role="dialog"`/`role="popover"` are the wrong semantic elements here. If we need the component for it's logic, we should extract it into a composable and use that instead.

packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxMenuItem/OnyxMenuItem.vue
```diff
@@ -110,44 +232,81 @@ const handleBackButtonKeydown = async (event: KeyboardEvent) => {
...
+      v-else
+      :label="props.label ?? 'external drill down'"
```
JoCa96: no inline translations/texts

packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxFlyoutMenu/OnyxFlyoutMenu.stories.ts
```diff
@@ -133,3 +133,55 @@ export const Nested = {
...
+            ),
+            h(OnyxMenuItem, { label: "Item 1.3" }),
+          ],
+        },
+      ),
+      h(OnyxMenuItem, () => "Item 2 "),
+      h(OnyxMenuItem, () => "Item 3 "),
+    ],
+  },
+} satisfies Story;
```
JoCa96, line 188: In the example please don't mix the different opening modes

JoCa96 (changes requested):

JoCa96 (approved):

PR #5270, opened 2026-04-09
feat: Add "resourcesAsTools" flag to support gemini CLI

JoCa96:
Relates to #5127

- feat: Add "resourcesAsTools" flag to support Gemini CLI
  - Gemini CLI seems to not be able to make use of [MCP Resources ](https://modelcontextprotocol.io/specification/2025-11-25/server/resources), only [Tools](https://modelcontextprotocol.io/specification/2025-11-25/server/tools)
  - As a workaround this special option was added, so that the resources can also be used as tools
- Added README.md
- Added default config for [`@modelcontextprotocol/inspector`](https://github.com/modelcontextprotocol/inspector)


packages/modelcontextprotocol/README.md
```diff
@@ -0,0 +1,65 @@
+# onyx demo app
```
larsrickert:
```suggestion
# onyx MCP
```

larsrickert (approved):

PR #5328, opened 2026-04-20
chore: refactor playwright pipeline strategy

JoCa96:
- Updated playwright pipeline:
  - Now uses one shard per package for playwright tests (Except `sit-onyx` package, which uses 10 shards). This avoids overhead of building and running tests for every on every shard.
  - On the PR check: Only affected packages are tested. And only `sit-onyx` uses the `--only-changed` flag, as it doesn't make sense for downstream packages .
- Updated `gh:show-report` script to accept the run-id as argument. This makes it easy to check specific executions like for merge group requests.
- Fix(useMatrixScreenshotTest): reduce flakiness by waiting for all images to have been loaded before performing the screenshot comparison
- Fix(expectEmit): reduce flakiness by allowing retries when checking for the emit

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -21,6 +21,11 @@ jobs:
     outputs:
       packages: ${{ steps.queryPackages.outputs.packages }}
     steps:
+      - name: "🛒 Checkout Repository"
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          fetch-depth: 0
+
       - name: "🛃 Harden the runner (Audit all outbound calls)"
         uses: step-security/harden-runner@f808768d1510423e83855289c910610ca9b43176 # v2.17.0
         with:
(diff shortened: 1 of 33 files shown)


.github/workflows/playwright.yml
```diff
@@ -15,18 +15,79 @@ permissions:
...
-        # note:
-        # - GitHub's shard limit is 256 parallel (see https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#using-a-matrix-strategy)
-        # - Github's job limit is 20 parallel (see https://docs.github.com/en/actions/reference/limits#job-concurrency-limits-for-github-hosted-runners)
-        shard: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
+        package: ${{ fromJSON(needs.prepare.outputs.packages) }}
+        shard: [1]
+        include:
+          - package: "sit-onyx"
+            shard: 1
+            totalShards: 10
```
larsrickert, line 63: Can we use a variable or something so we don't need to repeat the "10" in each shard?

larsrickert (approved):

.github/workflows/playwright.yml
JoCa96: Doesn't seem possible to use envs there: `Available expression contexts: github, inputs, vars, needs`

Issue #4072, opened 2025-09-04
Fix selection column not sticky when used with sticky columns


JoCa96:
### Why?

![Image](https://github.com/user-attachments/assets/08de6f18-b7bc-4a60-a1aa-daac96ddd939)

### Design

[Figma](ADD_FIGMA_LINK_HERE)

### Acceptance criteria

- [ ] When sticky columns are on the left side:
  - [ ] Selection column should be sticky
  - [ ] Selection column should be the first column

### Implementation details

<-- Add technical details / implementation proposals if necessary, e.g.:
  - API proposal
  - relevant web APIs
  - learnings from other implementations
  - code snippets
  -->

### DEV handover checklist

[Storybook](ADD_LINK_HERE)

DEV handover done with: @ADD_DEV_NAME_HERE

**Testing**
- [ ] covered by visual screenshot tests
- [ ] covered by functional tests (Playwright or unit test)

**Documentation**
- [ ] "New" badge in Storybook is added, see [storybook-addon-tag-badges](https://github.com/Sidnioulz/storybook-addon-tag-badges/?tab=readme-ov-file#-usage)
- [ ] auto-generated Storybook code snippets are checked
- [ ] component status has been updated in the [docs overview](https://onyx.schwarz/#components)
- [ ] updated version is released
- [ ] updated documentation is deployed

**Other**
- [ ] follow-up tickets were created if necessary (add links below)

### UX handover checklist

UX handover done with: @ADD_UX_NAME_HERE

**Naming is aligned with Figma**
- [ ] component name
- [ ] which support components are used and how they are named
- [ ] all property names and types

**Implementation is checked**
- [ ] all features of Figma are implemented or a follow-up ticket is refined
- [ ] skeleton
- [ ] densities
- [ ] light and dark mode
- [ ] truncation
- [ ] responsiveness / wrapping behavior
- [ ] min / max width

**Figma**
- [ ] Figma component is published

**Documentation**
- [ ] Storybook category is aligned with Figma
- [ ] Storybook examples are checked

**Other**
- [ ] follow-up tickets were created if necessary (add links below)

Sinan-SchwarzIT:
Cant wait to get that fix :)

PR #5340, opened 2026-04-21
fix(StickyColumn): fixed selection column to be sticky when used with sticky columns

ChristianBusshoff:
Relates to #4072

- added the possibility to set the position on each column
- implemented selection column to be sticky


packages/sit-onyx/src/components/OnyxDataGrid/features/selection/selection.ts
larsrickert, line 1: Would prefer to use a second symbol for the column since the `SELECTION_FEATURE` is used for the feature itself, not the column :)

packages/sit-onyx/src/components/OnyxDataGrid/features/stickyColumns/stickyColumns.ct.tsx
```diff
@@ -163,3 +163,79 @@ positions.forEach((position) => {
...
+  const stickyColJ = component.getByRole("columnheader", { name: "j" });
+  const stickyColK = component.getByRole("columnheader", { name: "k" });
+
+  await expect(stickyColJ).toContainClass("onyx-data-grid-sticky-columns--sticky");
+  await expect(stickyColJ).toContainClass("right");
+  await expect(stickyColJ).toHaveCSS("right", /[0-9]+px/);
+
+  await expect(stickyColK).toContainClass("onyx-data-grid-sticky-columns--sticky");
+  await expect(stickyColK).toContainClass("right");
+  await expect(stickyColK).toHaveCSS("right", /[0-9]+px/);
```
larsrickert, line 302: Would be nice to also capture a screenshot here

packages/sit-onyx/src/components/OnyxDataGrid/features/index.ts
```diff
@@ -526,7 +531,9 @@ export const useDataGridFeatures = <
     skeleton,
   }: UseDataGridFeaturesOptions<TEntry, TColumnGroup, TTypes>,
 ) => {
-  const features = featureDefinitions.map((f) => f({ async, i18n, skeleton }));
+  const activeFeatures = reactive(new Set<symbol>());
+  const features = featureDefinitions.map((f) => f({ async, i18n, skeleton, activeFeatures }));
+  features.forEach((f) => activeFeatures.add(f.name));
```
larsrickert: Isn't this a chicken in the egg problem? Because we are initializing the features to know the names of active features but pass in the activeFeatures at the same time.

Do we even need this information for this stick columns fix? Can't we just always add the selection column key? If the columns does not exist, it should not be an issue, right?

packages/sit-onyx/src/components/OnyxDataGrid/features/stickyColumns/stickyColumns.ts
```diff
@@ -13,114 +13,170 @@ import "./stickyColumns.scss";
...
+      const rawColumns = toValue(options?.columns) ?? [];
+      const normalized = rawColumns.map((col) => {
+        if (typeof col === "object" && col !== null && "key" in col) {
+          return {
+            key: col.key as PropertyKey,
+            position: col.position ?? globalPosition.value,
+          };
+        }
+        return { key: col as PropertyKey, position: globalPosition.value };
+      });
```
larsrickert, line 32:
```suggestion
      const normalized = rawColumns.map((col) => {
        if (typeof col === "object") {
          return { key: col.key, position: col.position ?? globalPosition.value };
        }
        return { key: col, position: globalPosition.value };
      });
```

packages/sit-onyx/src/components/OnyxDataGrid/features/stickyColumns/stickyColumns.ts
```diff
@@ -13,114 +13,170 @@ import "./stickyColumns.scss";
...
+
+      const selectionSymbol = Array.from(ctx.activeFeatures).find(
+        (name) => typeof name === "symbol" && name.description === "Selection",
+      );
+      if (selectionSymbol) {
+        normalized.unshift({ key: selectionSymbol, position: "left" });
+      }
+
```
larsrickert:
```suggestion
normalized.unshift({ key: SELECTION_FEATURE, position: "left" });
```

We should import the SELECTION_FEATURE symbol and should not rely on comparing the symbol description :) Also can't we just always add the selection feature? If it does not exists, it is just ignored so we don't need the activeFeatures

packages/sit-onyx/src/components/OnyxDataGrid/features/stickyColumns/stickyColumns.ts
```diff
@@ -13,114 +13,170 @@ import "./stickyColumns.scss";
...
+      if (index === -1) return;
+
+      const width = relevantKeys
+        .map((colKey) => elementWidths.value[colKey] || 0)
+        .reduce((acc, currentWidth, i) => {
+          if (i < index) {
+            return acc + currentWidth;
+          }
+          return acc;
+        }, 0);
```
larsrickert:
```suggestion
      const width = relevantKeys
        .slice(0, index)
        .reduce<number>((acc, colKey) => acc + (elementWidths.value[colKey] || 0), 0);
```

We can use .slice() to get the part of the array that is actual relevant so we don't need to iterate through the whole array :)

packages/sit-onyx/src/components/OnyxDataGrid/features/stickyColumns/stickyColumns.ts
```diff
@@ -13,114 +13,170 @@ import "./stickyColumns.scss";
...
       if (!("ResizeObserver" in window)) return;
 
       const resizeObserver = new ResizeObserver(() => {
-        Object.entries(elementsToStyle.value).forEach(
-          ([column, el]: [PropertyKey, HTMLElement]) =>
-            (elementWidths.value[column] = el.getBoundingClientRect().width),
-        );
+        Reflect.ownKeys(elementsToStyle.value).forEach((column) => {
+          const el = elementsToStyle.value[column as PropertyKey];
+          elementWidths.value[column as PropertyKey] = el?.getBoundingClientRect().width || 0;
```
larsrickert:
```suggestion
          const el = elementsToStyle.value[column];
          elementWidths.value[column] = el?.getBoundingClientRect().width || 0;
```

Type casts seem to be redundant

packages/sit-onyx/src/components/OnyxDataGrid/features/stickyColumns/stickyColumns.ts
```diff
@@ -13,114 +13,170 @@ import "./stickyColumns.scss";
...
       });
 
       watch(
         elementsToStyle,
         () => {
           resizeObserver.disconnect();
-          Object.values(elementsToStyle.value).forEach((el) => resizeObserver.observe(el));
+
+          Reflect.ownKeys(elementsToStyle.value).forEach((column) => {
+            const el = elementsToStyle.value[column as PropertyKey];
```
larsrickert:
```suggestion
            const el = elementsToStyle.value[column];
```

Type casts seem to be redundant

larsrickert (changes requested):

packages/sit-onyx/playwright/snapshots/components/OnyxDataGrid/features/stickyColumns/data-grid-sticky-left-right-chrome-linux.png
larsrickert, line 1: The right column is not display sticky for chrome and firefox but for webkit 👀  Maybe the columns ar enot scrolled enough?

larsrickert (approved):

Issue #5159, opened 2026-03-25
Fix deprecated eslint config


JoCa96:
### Why?

- We use [`tsEslint.config`](https://github.com/SchwarzIT/onyx/blob/main/eslint.config.js#L231), which was deprecated.
- Also the `typescript-eslint` rules don't seem to be (correctly?) applied:
  - E.g. the rule [no-floating-promises](https://typescript-eslint.io/rules/no-floating-promises) is not checked
  - But when the deprecated function is replaced, I suddenly get `typescript-eslint` warnings in the IDE:

We should resolve this soon, as there are quite a few useful `typescript-eslint` linting rules, that we should enforce.

### Acceptance criteria

- [x] Migrate deprecated`tsEslint.config`
- [x] Check `typescript-eslint` linting rules and fix errors

### Implementation details

- https://typescript-eslint.io/packages/typescript-eslint/#migrating-to-defineconfig

### DEV handover checklist

DEV handover done with: @JoCa96

PR #5379, opened 2026-04-24
chore: fix deprecated eslint config

larsrickert:
Relates to #5159

Remove deprecated `tsEslint.config()` and fix resulting open linting errors.


packages/headless/src/composables/listbox/createListbox.ts
```diff
@@ -34,27 +34,27 @@ export type CreateListboxOptions<TValue extends ListboxValue, TMultiple extends
   /**
    * Hook when an option is selected.
    */
-  onSelect?: (value: TValue) => void;
```
JoCa96, line 37: It is preferable to keep the `void` type, so that it is clear that the return value is not relevant and not used.
To resolve the eslint issue, check the rule examples: https://typescript-eslint.io/rules/no-misused-promises

JoCa96 (approved):

JoCa96 (changes requested):
see comment

packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxMenuItem/OnyxMenuItem.vue
```diff
@@ -123,7 +123,7 @@ const getOpeningArrowDirection = () => {
 const {
   elements: { listItem, menuItem },
 } = createMenuItems({
-  onOpen: handleOpen,
+  onOpen: () => handleOpen,
```
JoCa96: function should be called

```suggestion
  onOpen: () => handleOpen(),
```

JoCa96 (approved):

PR #5406, opened 2026-04-29
feat(useMatrixScreenshotTest): add `fastNoIsolation` flag for useMatrixScreenshotTest

JoCa96:


packages/playwright-utils/src/screenshots/types.ts
```diff
@@ -40,6 +41,11 @@ export type MatrixScreenshotTestOptions<
    * Function that returns the component for the given column and row.
    */
   component: (column: TColumn, row: TRow) => JSX.Element;
+  /**
+   * Renders the component matrix without isolation, which drastically improves the test duration.
+   * On the downside the hooks will not be executed.
+   */
```
larsrickert, line 50:
```suggestion
  /**
   * By default a combined matrix screenshot is created that includes the screenshots for every column-row combination.
   * Every combination is mounted individually, which allows to perform pointer or keyboard interactions before taking a screenshot.
   *
   * If the isolation is not necessary, consider enabling this option which is way faster, as it mounts all combinations together.
   * On the downside the hooks will not be executed.
   */
```

Personally, I find the docs from the changelog very helpful to understand the difference

packages/playwright-utils/src/screenshots/types.ts
```diff
@@ -59,7 +65,14 @@ export type MatrixScreenshotTestOptions<
...
   screenshotOptions?: LocatorScreenshotOptions;
-};
+} & (TIsolation extends true
+  ? {
+      fastNoIsolation: true;
+      hooks?: never;
+    }
+  : {
+      fastNoIsolation?: false;
+    });
```
larsrickert, line 78: Question: Can't we just always run the screenshots with "fastNoIsolation" when no hooks are passed so we have a performant default?

Or would you consider this a breaking change since the screenshots somehow differ slightly in text rendering?

larsrickert (approved):

packages/playwright-utils/src/screenshots/types.ts
JoCa96: Yes, I would consider that a Breaking Change. Because of changed screenshots and in cases like modals/flyouts which might not behave predictable.

Issue #3162, opened 2025-04-16
Implement pattern error message for OnyxInput


larsrickert:
### Why?

We want to support a more easily custom error message when using patterns/regex.

Currently, the user has to listen to the validityChange event, check the error type and then set and pass their error message to the input which is not the best experience. [Playground example](https://playground.onyx.schwarz/?onyxVersion=1.0.0-beta.194#eNp9U8Fu2zAM/RVCp3ao48N28pKuXZFDB3Qt1mGXeRhUmXHU2bIg0l6yIP8+Wo6TtAh6CGDxPYKP7zEbde39pGtRZWpKJljPUGlXznLFlCsg5NZf5s7WvgkMG7h369Wt8y3DFhahqSFXZDlppJyrj0fEgIsDRQZENHemccTQ6apFmPWks1zl6lywAfGaGYObh9CEOyTS5WveyGzcw8D9oStbWF7fLEV4zz4j1owZjMBj/zyH2SVscgenRkxGQbFzsmPcWao1m2XfBPBJ9riuquYvyehqDfKFwWhCMEsdtJEOQQKQ10a2HZoyaYqbb+U3TQeHxU95MNa+kmnyApgWtgNTaSJxvjczKYMtEtmUtXUYchVpQtwHMLwBuqRuCqykLy4xTgbJ8SmWdzadFHxgZ7ulpSH9qZN/8OtdeoSalripE+xNE8oJDw/kKxESjU9kkEQSVzoZ1tiTDiak4oJ8TdMjb9SFGm4qqbWfPFPj5FRjjvkOkCWyIdm+drjG3vsls6csTU3hpFdssl2YOOTU+TodqVdPyFpmE6fWFbgS5iBsK7nJfCbJYWHLV9NNU3tbYbj3bOUiX6jQ/aF8iTUOLV6MdbNE8+dE/Zl2gh8CEoZOnNljrEOJPMDzx6+4ku89KNm3lbDfAL8hNVXbaxxon1tXiOwjXlR7G820rvxO8xWjo3GpXmh0I/Ljn/nmjdUPct9PPhy5+LuTexO2GNhHS6y2/wHpvYR7)

### Acceptance criteria

- [ ] when passing a custom pattern, optionally also a error message can be passed that is only displayed if the input has a pattern error

### Implementation details

API proposal:

<OnyxInput :pattern="{ regex: /[a-z ]*/, errorMessage: 'Allows only lowercase characters or space' }" />

### Definition of Done

[...]

PR #5480, opened 2026-05-12
Feat(Input): implement pattern error message

ChristianBusshoff:
Relates to #3162

- added pattern error message


packages/sit-onyx/src/components/OnyxInput/types.ts
```diff
@@ -14,7 +15,7 @@ export type OnyxInputProps = Omit<SharedFormElementProps, "label"> &
     /**
      * Pattern the value must match to be valid.
      */
-    pattern?: string | RegExp;
+    pattern?: string | RegExp | { value: string | RegExp; errorMessage: CustomMessageType };
```
larsrickert:
```suggestion
    pattern?: InputPattern | { value: InputPattern; error: CustomMessageType }
    
    type InputPattern = string | RegExp;
```

Would prefer to add an extra type so we don't have to maintain the input pattern type twice :) Also we could rename errorMessage to just error so its aligned with the regular "error" property of the input itself

packages/sit-onyx/src/components/OnyxInput/OnyxInput.vue
```diff
@@ -82,6 +72,46 @@ defineExpose({
...
+      source: pattern.value instanceof RegExp ? pattern.value.source : pattern.value,
+      errorMessage: pattern.errorMessage,
+    };
+  }
+
+  return {
+    source: pattern instanceof RegExp ? pattern.source : pattern,
+    errorMessage: undefined,
+  };
+});
```
larsrickert, line 87:
```suggestion
const normalizedPattern = computed(() => {
  const { pattern } = props;
  if (!pattern) return;

  const isConfigObj = typeof pattern === "object" && !(pattern instanceof RegExp);
  const value = isConfigObj ? pattern.value : pattern;
  const errorMessage = isConfigObj ? pattern.errorMessage : undefined;

  return {
    source: value instanceof RegExp ? value.source : value,
    errorMessage,
  };
});
```

We can simplify this a bit so we only have to check the regexp once

packages/sit-onyx/src/components/OnyxInput/OnyxInput.vue
```diff
@@ -82,6 +72,46 @@ defineExpose({
...
+
+const patternSource = computed(() => normalizedPattern.value?.source);
```
larsrickert: I guess this computed is obsolete now since its just used once in the `<input />` so we can just use `:pattern="normalizedPattern?.source"` there

packages/sit-onyx/src/components/OnyxInput/OnyxInput.vue
```diff
@@ -82,6 +72,46 @@ defineExpose({
...
+const error = computed(() => {
+  if (props.error) return props.error;
+  if (maxLengthError.value) return maxLengthError.value;
+
+  const pattern = normalizedPattern.value;
+  if (
+    pattern?.errorMessage &&
+    modelValue.value !== undefined &&
+    input.value?.validity.patternMismatch
+  ) {
```
larsrickert:
```suggestion
  if (pattern?.errorMessage && input.value?.validity.patternMismatch) {
```

I am not sure if we should really check the modelValue here. Since when a pattern is provided and there is a patternMismatch error, we should always return the error here. Showing the error (e.g. only when touched etc.) is responsibility of the form element

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxInput/PatternTestCase.vue
larsrickert, line 1: Should we rename this file to PatternTestCase.ct.vue so its clear that this test case is only used for the pattern? So we might add other test cases in the future

packages/sit-onyx/src/components/OnyxInput/OnyxInput.vue
```diff
@@ -82,6 +72,41 @@ defineExpose({
...
+  const { pattern } = props;
+  if (!pattern) return;
+
+  const isConfigObj = typeof pattern === "object" && !(pattern instanceof RegExp);
+  const value = isConfigObj ? pattern.value : pattern;
+  const errorMessage = isConfigObj ? pattern.error : undefined;
+
+  return {
+    source: value instanceof RegExp ? value.source : value,
+    errorMessage,
```
larsrickert:
```suggestion
  const error = isConfigObj ? pattern.error : undefined;

  return {
    source: value instanceof RegExp ? value.source : value,
    error,
```

So its aligned with the property

packages/sit-onyx/src/composables/useFormElementError.ts
```diff
@@ -46,12 +62,19 @@ export type FormValidationProps = {
...
   UseFormValidityOptions<FormValidationProps>,
   "error" | "props"
 > & {
-  props: MaybeRef<Omit<FormValidationProps, "error"> & { error?: AnyFormError }>;
+  props: MaybeRef<
+    Omit<FormValidationProps, "error" | "customErrorMessages"> & {
+      error?: AnyFormError;
+      customErrorMessages?: CustomValidationMessages;
+    }
+  >;
```
larsrickert:
```suggestion
  props: MaybeRef<Omit<FormValidationProps, "error"> & { error?: AnyFormError }>;
```

custom errors are not passed as property, right?

packages/sit-onyx/src/composables/useFormElementError.ts
```diff
@@ -27,15 +27,31 @@ export type FormMessages = {
...
+  | "tooLong"
+  | "tooShort"
+  | "typeMismatch"
+  | "valueMissing";
+
+export type CustomValidationMessages = Partial<Record<ValidationKey, AnyFormError>>;
+
 export type FormValidationProps = {
   error?: CustomMessageType;
+  customErrorMessages?: CustomValidationMessages;
```
larsrickert:
```suggestion
```

There is no such property, right?

packages/sit-onyx/src/composables/useFormElementError.ts
```diff
@@ -27,15 +27,31 @@ export type FormMessages = {
...
+  | "patternMismatch"
+  | "rangeOverflow"
+  | "rangeUnderflow"
+  | "stepMismatch"
+  | "tooLong"
+  | "tooShort"
+  | "typeMismatch"
+  | "valueMissing";
+
+export type CustomValidationMessages = Partial<Record<ValidationKey, AnyFormError>>;
```
larsrickert:
```suggestion
export type CustomValidationMessages = Partial<Record<keyof ValidityState, AnyFormError>>;
```

.changeset/yellow-vans-obey.md
```diff
@@ -0,0 +1,5 @@
+---
+"sit-onyx": minor
+---
+
+feat(OnyxInput): Added support for custom error messages in OnyxInput pattern validation. The pattern prop now accepts an object structure { value: string | RegExp; errorMessage: string }, allowing developers to display a specific error message when the input value does not match the provided pattern.
```
larsrickert:
```suggestion
feat(OnyxInput): Added support for custom error messages in OnyxInput pattern validation. The pattern prop now accepts an object structure { value: string | RegExp; error: string }, allowing developers to display a specific error message when the input value does not match the provided pattern.
```

larsrickert (changes requested):

larsrickert (approved):

Issue #2977, opened 2025-03-19
Define sticky columns hover effect


ChristianBusshoff:
### Why?

The Column hover is not working for the sticky Columns. the hole column should have an hover-effect.

Using `::column` could be an option to remove the pseudo-element workaround.

<img width="436" alt="Image" src="https://github.com/user-attachments/assets/d411eea7-ae25-4624-9c37-a5f0d6904443" />

PR #5482, opened 2026-05-13
feat(DataGrid) fixed hover effect for sticky Columns

ChristianBusshoff:
Relates to #2977

- fixed stickyColumn hoverEffect


packages/sit-onyx/src/components/OnyxTable/OnyxTable.vue
```diff
@@ -221,6 +221,19 @@ const headlineId = computed(() => (slots.headline ? _headlineId : undefined));
 
     @include define-borders();
 
+    &:not(:has(.onyx-table__empty)) {
+      @for $i from 1 through 99 {
```
larsrickert: This would create a lot of CSS output + does not work when using more than 99 columns (which should not be the case but is still a limitation). Would prefer to find another solution for this 🤔

larsrickert: Or if its too complex to implement the hover styles for sticky columns, we could also talk with UX to remove them for sticky columns completely

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxTable/OnyxTable.vue
ChristianBusshoff: I couldn't find a cleaner way to solve this with pure CSS. So I talked with Jonathan and he suggestion using the @for loop. I googled and the generated CSS size is easily handled by browser compression and we already have somewhere else a 99col limitation inside the DataGrid.
But we can also talk with UX to remove them.

larsrickert (approved):
Would be great to also have a screenshot test for this :)

larsrickert (commented):
And a changeset :D

PR #5528, opened 2026-05-19
fix(MoreList): export moreList composables

ChristianBusshoff:
- exporting moreList composables to make the moreList compoent useable outside of onyx


packages/sit-onyx/src/index.ts
```diff
@@ -320,6 +320,7 @@ export { createOnyx } from "./utils/plugin.js";
 
 export * from "./components/illustrations/index.js";
 export * from "./composables/themeTransition.js";
+export { useMoreList, useMoreListChild } from "./composables/useMoreList.js";
```
larsrickert:
```suggestion
export * from "./composables/useMoreList.js";
```

Is there a reason why we don't export everything (including types)?

larsrickert (dismissed):

JoCa96 (approved):

Issue #3394, opened 2025-05-21
Implement feature + column config helper types for OnyxDatagrid


JoCa96:
### Why?

We need to provide an easy way to type the `ColumnConfig` for the `OnyxDataGrid` with consideration to the features. 

E.g. something that works like this

```ts
const withCustomType = createFeature(() => ({
  name: Symbol("example feature name"),
  typeRenderer: {
   customType: createTypeRenderer<{ specialOption: string }>(...)
  } satisfies TypeRenderMap<Entry>,
}));

const features = [withCustomType, withAnotherFeature];

const columns: ColumnConfig<Entry, typeof features> = [{
  key: "a",
  type: { name: "customType", options: { specialOption: "hello world!" } }
}]
```

### Acceptance criteria

- [ ] There is a helper type that enriches the ColumnConfig based on the provided Features
- [ ] Examples and docs are updated

### Definition of Done

- The following component requirements are implemented:
  - [ ] skeleton
  - [ ] density
  - [ ] example usage added to `apps/demo-app/src/views/HomeView.vue`

- Should be covered by tests:
  - [ ] functional tests (Playwright or unit test)
  - [ ] visual tests (Playwright screenshots)

- Make sure, that
  - [ ] follow-up tickets were created if necessary
  - [ ] updated version + documentation is deployed
  - [ ] Storybook can show the feature
  - [ ] Storybook code snippet of new/changed examples are checked that they are generated correctly
  - [ ] Namings are aligned with Figma
  - [ ] Storybook sidebar badge for "New" is added on component or story level if needed. See: https://github.com/Sidnioulz/storybook-addon-tag-badges/?tab=readme-ov-file

#### Approval

[Storybook]()
[Documentation]()

- Checked and approved by
  - [ ] dev

PR #5601, opened 2026-06-02
feat(DataGrid): Enhanced ColumnConfig type

ChristianBusshoff:
Relates to #3394

- Enhanced ColumnConfig type to dynamically extract supported column types directly from a feature function or a array of features.

### DIFF
--- a/.changeset/some-donuts-go.md
+++ b/.changeset/some-donuts-go.md
@@ -2,4 +2,10 @@
 "sit-onyx": minor
 ---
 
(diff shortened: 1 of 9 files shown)


JoCa96:
@larsrickert @ChristianBusshoff What about this PR?

larsrickert:
> @larsrickert @ChristianBusshoff What about this PR?

Looks good to me in general but maybe we can get rid of the ExtractTypesFromFeature and IsFeatureArrayOrFeature types and add the `extends` conditions for the generics of ColumnConfig 🤔

PR #5639, opened 2026-06-12
fix(useClearButton): show clear button if value is 0 or false

larsrickert:
Fix the clear button logic to consider the number "0" and "false" as "being defined" to also show a clear button in this case.

## Checklist

### DIFF
--- a/packages/sit-onyx/src/composables/useClearButton.ts
+++ b/packages/sit-onyx/src/composables/useClearButton.ts
@@ -27,7 +27,7 @@ export const useClearButton = (options: UseClearButtonOptions) => {
 };
 
 /**
- * Checks if the given value is considered defines in order to show the clear button.
+ * Checks if the given value is considered "defined" in order to show the clear button.
  */
 function isValueDefined(value: unknown): boolean {
   // any number expect NaN (including 0 should be considered defined)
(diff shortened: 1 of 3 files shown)


packages/sit-onyx/src/composables/useClearButton.ts
```diff
@@ -20,8 +20,19 @@ export const useClearButton = (options: UseClearButtonOptions) => {
...
     const value = toValue(options.modelValue);
-    return Array.isArray(value) ? value.length > 0 : !!value;
+    return isValueDefined(value);
   });
 
   return { showClearButton };
 };
+
+/**
+ * Checks if the given value is considered defines in order to show the clear button.
```
JoCa96:
```suggestion
 * Checks if the given value is considered "defined" in order to show the clear button.
```

packages/sit-onyx/src/composables/useClearButton.ts
```diff
@@ -20,8 +20,18 @@ export const useClearButton = (options: UseClearButtonOptions) => {
...
   });
 
   return { showClearButton };
 };
+
+/**
+ * Checks if the given value is considered "defined" in order to show the clear button.
+ */
+function isValueDefined(value: unknown): boolean {
+  // any number expect NaN (including 0 should be considered defined)
```
JoCa96: comment is outdated

JoCa96 (approved):

Issue #4214, opened 2025-09-30
Fix `onyx-grid-center` not working when used with OnyxSidebar


JoCa96:
### Version

1.0.0

### Description

- onyx-grid-center doesn't work with OnyxSidebar, even in temporary mode

PR #5644, opened 2026-06-15
Feat(DataGrid): implement link column type for onyx data grid

ChristianBusshoff:
Relates to #4214

- added link column type


packages/sit-onyx/src/components/OnyxDataGrid/features/sorting/defaults.ts
```diff
@@ -34,6 +34,36 @@ export const TIME_COMPARE = (a: unknown, b: unknown, collator: Intl.Collator) =>
...
+        const linkObj = value as { link: string; label?: string };
+        return linkObj.label ?? linkObj.link ?? "";
+      }
+      return JSON.stringify(value);
+    }
+
+    return "";
+  };
+
+  return collator.compare(getSortString(a), getSortString(b));
```
larsrickert:
```suggestion
  const getLabel = (value: unknown): unknown => {
    if (!value || typeof value !== "object") return value;
    if ("label" in value && typeof value.label === "string") return value.label;
    if ("link" in value && typeof value.link === "string") return value.link;
    return value;
  };

  return STRING_COMPARE(getLabel(a), getLabel(b), collator);
```

Would prefer to re-use the default string compare here

packages/sit-onyx/src/components/OnyxDataGrid/features/renderer.ts
```diff
@@ -165,6 +167,50 @@ export const STRING_RENDERER = createTypeRenderer<StringCellOptions>({
   },
 });
 
+export type LinkCellOptions = Partial<OnyxLinkProps> & {
+  /**
+   * Fallback value to display when the value is undefined.
+   * Defaults to "-" if not provided.
+   */
+  fallback?: string;
```
larsrickert: For other column types we currently do not support such fallback property so I would prefer to remove it for now :) But I like the idea, maybe we can add this option generically to all column types

packages/sit-onyx/src/components/OnyxDataGrid/features/renderer.ts
```diff
@@ -165,6 +167,50 @@ export const STRING_RENDERER = createTypeRenderer<StringCellOptions>({
...
+export const LINK_RENDERER = createTypeRenderer<LinkCellOptions>({
+  header: { component: HeaderCell },
+  cell: {
+    component: ({ column: columnKey, row, metadata, modelValue, ...rest }) => {
+      if (metadata?.editable) {
+        return h(DataGridFormElementWrapper, {
+          ...rest,
+          label: getEditingLabel(row.id, columnKey),
+          is: OnyxInput,
+          modelValue,
```
larsrickert:
```suggestion
          modelValue,
          type: "url",
```

Should we only allow URLs? + please create a Playwright test for this when editing, e.g. in DataGridFormElementWrapper

packages/sit-onyx/src/components/OnyxDataGrid/examples/DefaultExample.vue
larsrickert, line 1: Imo we should keep the default example simple and not add too many different columns types here :)

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxDataGrid/features/editing/editing.ct.tsx
```diff
@@ -0,0 +1,43 @@
...
+          { link: { link: "https://example.com", label: "link label" } },
+          { link: "https://example.com/2" },
+        ]}
+      />,
+    );
+
+    const websiteCell = component.getByRole("button", { name: "link label" });
+    const popover = component
+      .getByRole("dialog")
+      .getByText("Display labelClear inputLinkClear input");
```
larsrickert:
```suggestion
    const popover = component.getByRole("dialog", { name: "Edit link" });
```

We can get the dialog by its name directly + should we change the label to be more clear such as "Edit link"?

packages/mdc/playwright/snapshots/components/OnyxMarkdownRenderer/link-chrome-linux.png
larsrickert, line 1: I guess the margin here changed unintentionally?

packages/sit-onyx/playwright/snapshots/components/OnyxDataGrid/features/editing/link-editor-popover-chrome-linux.png
larsrickert, line 1: Should we align the editing popover here with the one that we already have for the text editor? https://onyx-showcase-dev.apps.01.cf.eu01.stackit.cloud/components/form-elements/text-editor#basic

Ideally we can re-use it

packages/sit-onyx/src/components/OnyxDataGrid/features/editing/LinkEditor.vue
```diff
@@ -0,0 +1,98 @@
...
+const { t } = injectI18n();
+const popover = useTemplateRef("popoverRef");
+
+useOutsideClick({
+  inside: popover,
+  checkOnTab: true,
+  disabled: computed(() => !open.value),
+  onOutsideClick: () => (open.value = false),
+});
+const isObjectLink = computed(() => {
```
larsrickert: We need this mapping logic 3 times currently: In the renderer, the sorter and here. Would be nice to just have 1 single utility for this :)

packages/sit-onyx/playwright/snapshots/components/OnyxDataGrid/features/editing/link-editor-popover-chrome-linux.png
larsrickert: Maybe this should be discussed with UX

larsrickert (changes requested):

larsrickert (approved):

Issue #5614, opened 2026-06-09
Fix empty OnyxDataGrid placed inside OnyxModal causes resizing behaviour


boedigheimer:
### Version

1.14.0

### Description

We are facing an issue when using the OnyxDataGrid with empty data, especially when placed inside an OnyxModal. The Modal is slowly stretching horizontally when opened (and on every re-open). This can be currently be mitigated by unsetting the width on the onyx empty or by setting a fixed modal width. 

### Reproduction steps

[...]

PR #5670, opened 2026-06-18
fix(OnyxTable, OnyxDataGrid): prevent empty state from causing infinite growth

larsrickert:
Relates to #5614

fix(OnyxTable, OnyxDataGrid): prevent empty state from causing infinite growth inside `fit-content` parents (e.g. `OnyxModal` without an explicit width)

## Checklist

### DIFF
--- a/packages/sit-onyx/src/components/OnyxDataGrid/OnyxDataGrid.ct.tsx
+++ b/packages/sit-onyx/src/components/OnyxDataGrid/OnyxDataGrid.ct.tsx
@@ -88,8 +88,8 @@ test("should not resize in a loop when empty and used in a modal without explici
   // ACT
   const box1 = (await modal.boundingBox())!;
 
-  // eslint-disable-next-line playwright/no-wait-for-timeout -- relevant for this test
-  await page.waitForTimeout(2_000);
+  const handle = await modal.elementHandle();
+  await handle!.waitForElementState("stable");
 
   const box2 = (await modal.boundingBox())!;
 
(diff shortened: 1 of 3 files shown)


packages/sit-onyx/src/components/OnyxDataGrid/OnyxDataGrid.ct.tsx
```diff
@@ -59,3 +61,38 @@ test.describe("Screenshot tests", () => {
...
+  // ACT
+  const box1 = (await modal.boundingBox())!;
+
+  // eslint-disable-next-line playwright/no-wait-for-timeout -- relevant for this test
+  await page.waitForTimeout(2_000);
+
+  const box2 = (await modal.boundingBox())!;
+
+  // ASSERT
+  expect(box1.width).toBe(box2.width);
```
JoCa96, line 97: Playwright has a builtin stability check:
```suggestion
  const modal = page.getByRole("dialog", { name: "Example label" });

  // ASSERT
  await expect(modal).toBeVisible();
  const handle = await modal.elementHandle();
  await handle!.waitForElementState("stable", { timeout: 2000 });
```

JoCa96 (approved):

packages/sit-onyx/src/components/OnyxDataGrid/OnyxDataGrid.ct.tsx
larsrickert: I like the approach :) The only issue this can have is that the test would still succeed if the width grows fast enough :D So if it grows from e.g. 500px to 800px in 1 second and is then stable at 800px, the test would succeed although it should actually fail

PR #5675, opened 2026-06-19
fix(docs): determine browserlist support offline

larsrickert:
Currently all our CI builds fail because the [browsersl.ist](https://browsersl.ist/) website is down. This PR determines the supported browser versions offline using the browserslist npm package instead of relying on the online API.

image

### DIFF
--- a/apps/docs/src/.vitepress/browser-loader.data.ts
+++ b/apps/docs/src/.vitepress/browser-loader.data.ts
@@ -1,15 +1,24 @@
-import { readFile } from "fs/promises";
-import { fileURLToPath } from "url";
+import browserslist from "browserslist";
 import { defineLoader } from "vitepress";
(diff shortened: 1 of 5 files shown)


apps/docs/src/.vitepress/browser-loader.data.ts
```diff
@@ -1,54 +1,95 @@
...
+    }, [] as Browser[]);
+
+    return { browsers };
   },
 });
+
+/**
+ * Sorts the given versions ascending.
+ */
+function sortVersions(versions: string[]) {
```
JoCa96, line 74: We also have the same functionality in `packages/modelcontextprotocol/src/util/version-compare.ts`, maybe we can create a FoUp to put this into the shared package.

JoCa96 (approved):

Issue #4445, opened 2025-11-12
Implement quick filters for global search


larsrickert:
### Why?

Follow up for #4275. The OnyxGlobalSearch should support showing project-defined quick filters (w.g. with a badge or tag) that the user can select.

### Open questions / TODOs

- the design uses badges but they are not clickable. Should we use tags instead or some other component? Or should badges support being clickable in general?

### Design

[Figma](https://www.figma.com/design/YfEUBOHk4J4nYrk04geswG/Onyx-Components?node-id=15981-20709&m=dev)
[Documentation](https://www.figma.com/design/j7Hy7Zk8MiCmJBJf0ArQRm/Onyx-Documentation?node-id=16435-19123&m=dev)

### Acceptance criteria

- [ ] a OnyxGlobalSearchGroup can be displayed with badges
- [ ] the aria pattern linked be low is correctly integrated with the badges
- [ ] the badges become scrollable if not all fit into the available width

### Applicable ARIA Pattern

[Combobox](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/)

### DEV handover checklist

[Storybook](ADD_LINK_HERE)

DEV handover done with: @ADD_DEV_NAME_HERE

**Testing**
- [ ] covered by visual screenshot tests
- [ ] covered by functional tests (Playwright or unit test)

**Documentation**
- [ ] "New" badge in Storybook is added, see [storybook-addon-tag-badges](https://github.com/Sidnioulz/storybook-addon-tag-badges/?tab=readme-ov-file#-usage)
- [ ] auto-generated Storybook code snippets are checked
- [ ] component status has been updated in the [docs overview](https://onyx.schwarz/#components)
- [ ] updated version is released
- [ ] updated documentation is deployed

**Other**
- [ ] follow-up tickets were created if necessary (add links below)

### UX handover checklist

UX handover done with: @ADD_UX_NAME_HERE

**Naming is aligned with Figma**
- [ ] component name
- [ ] which support components are used and how they are named
- [ ] all property names and types

**Implementation is checked**
- [ ] all features of Figma are implemented or a follow-up ticket is refined
- [ ] skeleton
- [ ] densities
- [ ] light and dark mode
- [ ] truncation
- [ ] responsiveness / wrapping behavior
- [ ] min / max width

**Figma**
- [ ] Figma component is published

**Documentation**
- [ ] Storybook category is aligned with Figma
- [ ] Storybook examples are checked

**Other**
[...]

PR #5688, opened 2026-06-23
feat(OnyxGlobalSearch): implement filters for global search

ChristianBusshoff:
Relates to #4445

- implement filters for global search


apps/showcase/.cloud-foundry/manifest-local.yml
larsrickert, line 1: Should this be added?

packages/sit-onyx/playwright/snapshots/components/OnyxGlobalSearch/multiple-groups-chrome-linux.png
larsrickert, line 1: Should we increase the screenshot height here so the full content is visible?

packages/sit-onyx/src/components/OnyxFilterBadge/OnyxFilterBadge.vue
larsrickert, line 1: Would be nice to add the skeleton tot he OnyxFilterBadge screenshot test

packages/sit-onyx/src/components/OnyxFilterBadge/types.ts
```diff
@@ -10,4 +11,8 @@ export type OnyxFilterBadgeProps = Omit<OnyxBadgeProps, "dot" | "icon" | "clicka
    * If `true` the filter is active/selected, shows an 'x' icon and can be removed on click.
    */
   active?: Nullable<boolean>;
+  /**
+   * Whether to show a skeleton option.
```
larsrickert:
```suggestion
   * Whether to show a skeleton badge.
```

packages/sit-onyx/src/components/OnyxGlobalSearchFilterGroup/OnyxGlobalSearchFilterGroup.vue
larsrickert, line 1: Do we need an extra component for this? It seems to be basically the same as OnyxGlobalSearchGroup but with different skeleton styles (we could maybe add a slot for this instead)

packages/sit-onyx/src/components/OnyxGlobalSearch/OnyxGlobalSearch.vue
```diff
@@ -42,6 +42,10 @@ const slots = defineSlots<{
     /** The value of the currently highlighted option. */
     activeValue?: string;
   }): unknown;
+  /**
+   * Slot to pass filters. Only `OnyxGlobalSearchFilterGroup` components should be used.
+   */
+  filters?(): unknown;
```
larsrickert: Can't the user simply add the group to the existing default slot? Or is there a reason why the filters should be placed outside the listbox?

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxGlobalSearch/OnyxGlobalSearch.vue
ChristianBusshoff: I thought the listItem is the wrong aria pattern for it. That's why I did it this way.

packages/sit-onyx/playwright/snapshots/components/OnyxImage/Image-chrome-linux.png
larsrickert, line 1: The default error screenshot does not seem correct 🤔  Maybe thats a flaky issue

packages/sit-onyx/src/components/OnyxGlobalSearch/OnyxGlobalSearch.vue
```diff
@@ -157,21 +161,21 @@ provide(GLOBAL_SEARCH_INJECTION_KEY, { headless, activeValue });
...
-    >
-      <slot></slot>
-
-      <div v-if="!!slots.endOfList" class="onyx-global-search__end-of-list">
-        <slot
-          name="endOfList"
-          :get-option-props="getOptionProps"
-          :active-value="activeValue"
-        ></slot>
+    <div v-show="!!slots.default" class="onyx-global-search__body">
```
larsrickert: Shouldn't the v-show be placed on the listbox itself?

packages/sit-onyx/src/components/OnyxGlobalSearchGroup/OnyxGlobalSearchGroup.vue
```diff
@@ -26,31 +30,54 @@ const skeletonCount = computed(() => {
...
+    <component
+      :is="props.is"
+      v-if="direction === 'row'"
+      class="onyx-global-search-group__content--row"
+    >
+      <template v-if="skeletonCount <= 0">
+        <slot></slot>
+      </template>
+      <template v-else>
+        <OnyxFilterBadge
```
larsrickert, line 57: Hmm I am not sure if I like placing the filter badges here 🤔  Maybe we should leave this up to the project?

packages/sit-onyx/src/components/OnyxGlobalSearchGroup/types.ts
```diff
@@ -9,4 +9,13 @@ export type OnyxGlobalSearchGroupProps = DensityProp & {
...
   skeleton?: boolean | number;
+  /** Alignment of the group items.
+   * @default "column"
+   */
+
+  direction?: "column" | "row";
+  /** Whether to render as an HTML list (ul/li) or div.
+   * @default "li"
+   */
+  is?: "li" | "div";
```
larsrickert: Idea: We could combine those two props in one "orientation" property that can be set to vertical (default) or horizontal. For horizontal we automatically use div instead of li. Because the props are kind of related two each other imo so that it does not make sense to have horizontal layout if items

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxGlobalSearch/OnyxGlobalSearch.vue
ChristianBusshoff: no? It should be on the body, like before to not render empty content, or?

packages/sit-onyx/src/components/OnyxGlobalSearchGroup/OnyxGlobalSearchGroup.vue
ChristianBusshoff: This is for the skeleton, should I use the OnyxGlobalSearchOption Skelton (like for the vertical mode) instead or should I remove the skeleton for the horizontal mode completely?

packages/sit-onyx/playwright/snapshots/components/OnyxGlobalSearch/default-chrome-linux.png
larsrickert, line 1: We should still perform screenshots for the default state (without any content) to make sure that the body+footer is hidden correctly when no slot content exist :)

We can add a new screenshot for the filters then

.changeset/twelve-mammals-exist.md
larsrickert, line 1: The changeset content is no longer correct, right?

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxGlobalSearch/OnyxGlobalSearch.vue
```diff
@@ -42,6 +42,10 @@ const slots = defineSlots<{
     /** The value of the currently highlighted option. */
     activeValue?: string;
   }): unknown;
+  /**
+   * Slot to pass leading Content. It's displayed before the List.
```
larsrickert:
```suggestion
   * Slot to pass leading content that is displayed before the `default` slot.
```

packages/sit-onyx/src/components/OnyxGlobalSearch/OnyxGlobalSearch.vue
```diff
@@ -56,6 +60,7 @@ const searchTerm = useVModel({
 
 const dialog = useTemplateRef("dialog");
 const dialogElement = computed(() => dialog.value?.dialog);
+const hasContent = () => !!slots.default || !!slots.endOfList || !!slots.leading;
```
larsrickert, line 63:
```suggestion
const hasContent = computed(() => !!slots.default || !!slots.endOfList || !!slots.leading);
```

Is there a reason why this is not a computed? Since the slots are reactive :)

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxGlobalSearch/OnyxGlobalSearch.vue
ChristianBusshoff: computed didn't worked inside the tests.

larsrickert (approved):

PR #5702, opened 2026-06-24
perf: replace prettier with oxfmt

larsrickert:
Replace prettier with the new [oxfmt](https://oxc.rs/docs/guide/usage/formatter.html) which is much faster and covers the same formatting. If you haven'T done it yet, install the [oxc vscode extension](https://marketplace.visualstudio.com/items?itemName=oxc.oxc-vscode)

In GitHub CI, prettier formatting checks take around [30 seconds](https://github.com/SchwarzIT/onyx/actions/runs/28016405642/job/82921963227) while oxfmt is done in [8 seconds](https://github.com/SchwarzIT/onyx/actions/runs/28077209871/job/83124694148).

FYI: oxfmt also has native support for import sorting and JSdoc formatting which might be nice to have. I haven't enabled them yet in this PR so there are not too many changes.

### DIFF
--- a/.github/workflows/import-figma-assets.yml
+++ b/.github/workflows/import-figma-assets.yml
@@ -49,9 +49,8 @@ jobs:
         working-directory: packages/icons
 
       # needed to prevent creating empty PR when nothing but the formatting changed
-      - name: Format with prettier
-        run: pnpm prettier --write src/metadata.json
-        working-directory: packages/icons
+      - name: Format
+        run: pnpm format:all packages/icons/src/metadata.json
 
       - name: Create pull request
         uses: peter-evans/create-pull-request@5f6978faf089d4d20b00c7766989d076bb2fc7f1 # v8.1.1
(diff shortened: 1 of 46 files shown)


JoCa96 (changes requested):
Can we remove prettier completely? Or is there a reason to keep it around?

larsrickert:
> Can we remove prettier completely? Or is there a reason to keep it around?

Good point, I removed prettier now where possible. We still need to keep it for the Storybook utils to format the code snippets since oxfmt does not run in the browser, only in Node. I moved the dependency to the storybook-utils package

JoCa96:
> > Can we remove prettier completely? Or is there a reason to keep it around?
> 
> Good point, I removed prettier now where possible. We still need to keep it for the Storybook utils to format the code snippets since oxfmt does not run in the browser, only in Node. I moved the dependency to the storybook-utils package

Okay, then please create a follow-up ticket, so that we don't forget to remove the dependency.

JoCa96 (approved):

PR #5719, opened 2026-06-24
chore: deactivate automatic snapshot release

JoCa96:
- Deactivated release on push to main
- Instead, the pipeline can be triggered manual or by pushing release tags
- We do this to stop spamming the npm registry with development versions

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,8 +3,9 @@
 name: Release
 on:
   push:
-    branches:
-      - "main"
+    tags:
+      - "sit-onyx@*"
+      - "@sit-onyx@*"
   workflow_dispatch:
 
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -3,8 +3,9 @@
 name: Release
 on:
   push:
-    branches:
-      - "main"
+    tags:
```
larsrickert, line 6: Wouldn't this trigger the release pipeline for every published package since each packages has a tag?

Can't we use the stage (prod/dev) to just not trigger the release steps?

larsrickert: Also deployment to dev should still be done, right?

larsrickert (changes requested):

Issue #3576, opened 2025-06-18
Implement row click feature for OnyxDataGrid


larsrickert:
### Why?

Feature request from #3512, #5458

The data grid should support row clicks so that custom actions can be triggered when clicking a row.

### Open questions / TODOs

- should it be possible to disable the "clickability" for specific rows or should it be enabled for all rows?

### Design

[Figma](https://www.figma.com/design/YfEUBOHk4J4nYrk04geswG/Onyx-Component-Library?node-id=4435-45283&t=uOfOiDP9Dlx5LtPk-4)

### Acceptance criteria

- [x] a new `useRowClick` data grid feature is implemented
- [x] when a row is clickable, the cursor becomes a "pointer"
- [x] a callback function is called when the row is clicked that includes:
    - [x] row data
    - [x] click event
- [x] nice to have: when text is selected, the click is not triggered so cell content can still be copied

### Definition of Done

- Should be covered by tests:
  - [x] functional tests (Playwright or unit test)

- Make sure, that
  - [x] Storybook can show the feature
  - [x] Storybook code snippet of new/changed examples are checked that they are generated correctly
  - [x] Storybook sidebar badge for "New" is added on component or story level if needed. See: https://github.com/Sidnioulz/storybook-addon-tag-badges/?tab=readme-ov-file

#### Follow ups

- #5769

#### Approval

[Storybook](https://onyx-storybook-dev.apps.01.cf.eu01.stackit.cloud/?path=/story/data-datagrid-features--row-click)
[Showcase](https://onyx-showcase-dev.apps.01.cf.eu01.stackit.cloud/components/data/data-grid#row-click)

- Checked and approved by
  - [x] designer @Guergchen 
  - [x] dev @JoCa96

PR #5725, opened 2026-06-25
feat(OnyxDataGrid): implement `useRowClick` feature

larsrickert:
Relates to #3576

Implement a new row click feature for the data grid:
- supports clicking a row and trigger a callback function then
- ignores clicks then (text) selection exist in the row so values can be copied without trigger the click
- feature can be enabled/disabled for specific rows, columns and cells
- fix: downgrade Vite to run Nuxt server again, see https://github.com/nuxt/nuxt/issues/35450


packages/sit-onyx/src/components/OnyxDataGrid/features/rowClick/types.ts
```diff
@@ -0,0 +1,31 @@
...
+  enabled?:
+    | MaybeRef<boolean | undefined>
+    | ((entry: TEntry, column: keyof TEntry) => boolean | undefined);
+  /**
+   * Whether the click is also triggered when the user has a [selection](https://developer.mozilla.org/en-US/docs/Web/API/Selection)
+   * inside the row (e.g. marked text to copy).
+   *
+   * @default false
+   */
+  ignoreSelection?: boolean;
```
JoCa96: IMHO we should always block the click when the user is selecting. Is there a use-case to make it configurable?

packages/sit-onyx/src/components/OnyxDataGrid/features/rowClick/types.ts
```diff
@@ -0,0 +1,31 @@
+import type { MaybeRef, MaybeRefOrGetter } from "vue";
+import type { DataGridEntry } from "../../types.js";
+
+export type RowClickOptions<TEntry extends DataGridEntry> = {
+  /**
+   * Label that describes the click action. Required for accessibility.
+   *
+   * @example "Show details"
+   */
+  label: MaybeRefOrGetter<string>;
```
JoCa96, line 10: We can also add the link prop for the follow-up ticket:
```suggestion
  label: MaybeRefOrGetter<string>;
  /**
   * If set, the button will be rendered as link.
   * Note that not all button properties are supported when used as link, e.g. `disabled`, `type` and `autofocus` are not supported.
   */
  link?: WithLinkProp["link"]
```

JoCa96 (changes requested):
see comments

packages/sit-onyx/src/components/OnyxDataGrid/features/rowClick/types.ts
larsrickert: I don't think so. I removed it 👍

packages/sit-onyx/src/components/OnyxDataGrid/features/rowClick/types.ts
larsrickert: Would prefer to add this later when we actually implement the button :) Because instead of only support the link property, I'd prefer to allow setting every OnyxIconButton props (icon, link etc.)

JoCa96 (approved):

packages/sit-onyx/src/components/OnyxDataGrid/features/rowClick/types.ts
JoCa96: Okay, because the label was already added, but not used

larsrickert: Yes because the label is a required property which would be a breaking change if we'd add it later on :)

Issue #5713, opened 2026-06-24
FoUp: implement option to pass all link props individual to the link column type


ChristianBusshoff:
### Relates to:

#4124  

### Why?

- we want to offer the possibility to pass all link props individual

### Acceptance criteria
- add an option to pass all link props
- add a gab between link and icon inside editing

### DEV handover checklist

DEV handover done with: @JoCa96

PR #5726, opened 2026-06-25
feat(DataGird): implemented all link options for link column type

ChristianBusshoff:
Relates to #5713

- add support for target and withExternalIcon options for column link type

### DIFF
--- a/packages/sit-onyx/src/components/OnyxEditLinkDialog/utils.ts
+++ b/packages/sit-onyx/src/components/OnyxEditLinkDialog/utils.ts
@@ -1,4 +1,4 @@
-import { LINK_TARGETS } from "../OnyxRouterLink/types.js";
+import { LINK_TARGETS, type LinkTarget } from "../OnyxRouterLink/types.js";
 import type { EditLinkValue } from "./types.js";
 
 /**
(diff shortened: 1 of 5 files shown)


packages/sit-onyx/src/components/OnyxEditLinkDialog/utils.ts
```diff
@@ -10,6 +11,26 @@ export const parseLinkValue = (value: unknown): EditLinkValue | undefined => {
...
   if (!href) return;
-  return { href, label };
+
+  const result: EditLinkValue = { href };
+  if (label) result.label = label;
+
+  if (
+    "target" in value &&
+    typeof value.target === "string" &&
+    (LINK_TARGETS as readonly string[]).includes(value.target)
```
JoCa96:
```suggestion
    (LINK_TARGETS).includes(value.target as LinkTarget)
```

packages/sit-onyx/src/components/OnyxEditLinkDialog/utils.ts
```diff
@@ -10,6 +11,26 @@ export const parseLinkValue = (value: unknown): EditLinkValue | undefined => {
...
+  }
+
+  if ("withExternalIcon" in value) {
+    const iconValue = value.withExternalIcon;
+    if (typeof iconValue === "boolean" || iconValue === "auto") {
+      result.withExternalIcon = iconValue;
+    }
+  }
+
+  return result;
```
JoCa96: We can format/order the code here a bit more:
```ts
  const href = "href" in value && typeof value.href === "string" ? value.href : undefined;
  if (!href) return;

  const result: EditLinkValue = {href};
  if ("label" in value && typeof value.label === "string") {
    result.label = value.label;
  }
  if (
    "target" in value &&
    typeof value.target === "string" &&
    (LINK_TARGETS as readonly string[]).includes(value.target)
  ) {
    result.target = value.target as EditLinkValue["target"];
  }

  if ("withExternalIcon" in value) {
    const iconValue = value.withExternalIcon;
    if (typeof iconValue === "boolean" || iconValue === "auto") {
      result.withExternalIcon = iconValue;
    }
  }

  return result;
```

JoCa96 (approved):

Issue #5631, opened 2026-06-10
Implement external link behavior for OnyxLink


larsrickert:
### Why?

Currently links are always opened in the current tab by default. We want to update this behavior to be more intuitive and open external links in a new tab by default if not defined explicitly.

### Acceptance criteria

- [x] the behavior for the OnyxLink, OnyxRouterLink and useLink are updated to open external links in a new tab by default if not explicitly defined
- [x] the current "workarounds" are removed from:
    - apps/demo-app/app/components/LinkCard.vue
    - apps/showcase/app/components/content/LinkCard.vue
    - apps/showcase/app/components/content/TypographyCard.vue
    - packages/mdc/src/components/prose/ProseA.vue

### DEV handover checklist

[Storybook](https://onyx-storybook-dev.apps.01.cf.eu01.stackit.cloud/?path=/docs/navigation-link--docs)

DEV handover done with: @larsrickert 

**Testing**
- [x] covered by functional tests (Playwright or unit test)

PR #5730, opened 2026-06-26
 feat: Links that are determined to be external will now open in a new tab per default

JoCa96:
Relates to #5631

### DIFF
--- a/apps/showcase/app/components/content/TypographyCard.vue
+++ b/apps/showcase/app/components/content/TypographyCard.vue
@@ -1,5 +1,5 @@
 <script lang="ts" setup>
-import { isInternalLink, type SharedLinkProps } from "sit-onyx";
+import { type SharedLinkProps } from "sit-onyx";
 import type { DetailsItem } from "../DetailsList.vue";
 
 const props = defineProps<{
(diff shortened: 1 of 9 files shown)


apps/showcase/app/components/content/TypographyCard.vue
larsrickert, line 1: We can also remove the same logic in:
- apps/showcase/app/components/content/LinkCard.vue

packages/sit-onyx/src/components/OnyxRouterLink/OnyxRouterLink.vue
```diff
@@ -14,6 +16,16 @@ defineSlots<{
...
+
+const computedTarget = computed<LinkTarget>(() => {
+  if (props.target !== "auto") {
+    return props.target;
+  }
+  if (isInternalLink(props.href)) {
+    return "_self";
+  }
+  return "_blank";
+});
```
larsrickert, line 23:
```suggestion
const computedTarget = computed<LinkTarget>(() => {
  if (props.target !== "auto") return props.target;
  return isInternalLink(props.href) ? "_self" : "_blank";
});
```

packages/sit-onyx/src/components/OnyxRouterLink/types.ts
```diff
@@ -12,10 +12,28 @@ export type SharedLinkProps = {
    * Where to display the linked URL (same tab, new tab etc.).
    * For `_blank`, the `rel="noreferrer"` will be set automatically.
```
larsrickert, line 13:
```suggestion
   * For `_blank`, the `rel="noreferrer"` will be set automatically.
   * If set to "auto", external links will be opened with "_blank" and internal ones with "_self".
```

larsrickert (changes requested):
Would be great to also have a test for this :)

apps/showcase/app/components/content/TypographyCard.vue
JoCa96: Unfortunately we need the computed target there, so we cannot remove it

larsrickert (approved):
Playwright tests for this would be great 👍

JoCa96:
> Playwright tests for this would be great 👍

done

larsrickert (approved):

Issue #5711, opened 2026-06-24
Implement height fix for flyout with internal drilldown


larsrickert:
### Why?

Based on #5519. Fix the flyout menu with internal drilldown so it is not closed when the next layer has less items than the current one (because the cursor is no longer hovering over the flyout menu then)

### Acceptance criteria

- [x] in internal drilldown mode, the maximum height of a layout is kept when navigating to the next one. If the next layer has less items, empty space is shown at the bottom
- [x] the "Back" button is updated to show the parent menu item instead

### DEV handover checklist

[Storybook](https://onyx-storybook-dev.apps.01.cf.eu01.stackit.cloud/?path=/story/basic-flyoutmenu--nested&args=trigger:hover;open:!false)

DEV handover done with: @JoCa96 

**Testing**
- [x] covered by visual screenshot tests

### UX handover checklist

UX handover done with: @Guergchen

PR #5733, opened 2026-06-29
Feat(FlyoutMenut): implement height fix for flyout with internal drilldown

ChristianBusshoff:
Relates to #5711

- implement height fix for flyout with internal drilldown
- updated the back Button to display the parentName


packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxFlyoutMenu/OnyxFlyoutMenu.vue
```diff
@@ -81,6 +91,40 @@ const {
...
+
+onBeforeUnmount(() => {
+  observer?.disconnect();
+});
+
+watch(isExpanded, (newVal) => {
+  if (!newVal) {
+    minHeight.value = undefined;
+  }
+});
```
larsrickert:
```suggestion
const menuRef = useTemplateRef<HTMLElement>("menuRef");
const minHeight = ref<number>();

const { height } = useResizeObserver(menuRef);
watch(
  height,
  (newHeight) => {
    if (!isExpanded.value || props.drilldownMode !== "internal") {
      minHeight.value = undefined;
      return;
    }

    if (!minHeight.value || minHeight.value < newHeight) {
      minHeight.value = newHeight;
    }
  },
  { immediate: true },
);

watch(isExpanded, (newValue) => {
  if (!newValue) {
    minHeight.value = undefined;
  }
});
```

We can/should re-use our existing useResizeObserver composable :)

We could also extend it to support a "disabled" option where we can pass in the `!isExpanded.value || props.drilldownMode !== "internal"` condition which unobserves the menu and sets the height to 0 automatically.

packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxFlyoutMenu/OnyxFlyoutMenu.vue
```diff
@@ -112,7 +156,11 @@ provide<NestedMenuDrilldownModeContext>(MENU_ITEM_DRILLDOWN_INJECTION_KEY, {
       <ul
         v-if="slots.options"
         v-bind="menu"
+        ref="menuRef"
         class="onyx-flyout-menu__wrapper onyx-flyout-menu__group"
+        :style="{
+          minHeight: props.drilldownMode === 'internal' && minHeight ? `${minHeight}px` : undefined,
+        }"
```
larsrickert, line 143:
```suggestion
        :style="{ minHeight: minHeight ? `${minHeight}px` : undefined }"
```

Would prefer to not duplicate the condition here and instead move it to the resize observer (see suggestion above). Also I would prefer defining a CSS variable here and use in in the CSS to prevent inline CSS

packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxFlyoutMenu/OnyxFlyoutMenu.vue
larsrickert: Also we only need this logic when the trigger prop is set to "hover", for click there is no issue with this, right?

packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxMenuItem/OnyxMenuItem.vue
larsrickert, line 1: If possible it would be nice to reset the stored height when clicking the back button since in the current "Nested" Storybook example, it feels weird if the height is 3 items initially but when drilling in and then going back, there is an empty space at the bottom of the second layer contained more items

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxFlyoutMenu/OnyxFlyoutMenu.vue
```diff
@@ -77,10 +86,35 @@ const {
...
+  minHeight.value = undefined;
+};
+
 // Provide the context so that all OnyxMenuItems within this flyout adapt properly to the nested mode.
 provide<NestedMenuDrilldownModeContext>(MENU_ITEM_DRILLDOWN_INJECTION_KEY, {
   drilldownMode: computed(() => props.drilldownMode),
+  resetMinHeight,
 });
+
+const menuRef = useTemplateRef<HTMLElement>("menuRef");
```
larsrickert:
```suggestion
const menuRef = useTemplateRef("menuRef");
```

Is the explicit type needed?

packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxFlyoutMenu/OnyxFlyoutMenu.vue
```diff
@@ -77,10 +86,35 @@ const {
...
+const minHeight = ref<number>();
+
+const { height } = useResizeObserver(menuRef);
+
+watch(
+  [height, isExpanded, () => props.drilldownMode],
+  ([newHeight, expanded, mode]) => {
+    if (!expanded || mode !== "internal") {
```
larsrickert:
```suggestion
watch(
  [height, isExpanded, () => props.drilldownMode, () => props.trigger],
  ([newHeight, expanded, mode, trigger]) => {
    if (!expanded || mode !== "internal" || trigger !== "hover") {
```

We only need to use the minHeight if the trigger is hovered, right?
Also we could extend the "useResizeObserver" to support a disabled option with this condition here since currently the resize observer is always active, even if we don't need it

packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxMenuItem/OnyxMenuItem.vue
```diff
@@ -201,7 +202,7 @@ provide<NestedMenuContext>(MENU_ITEM_INJECTION_KEY, {
       >
         <OnyxMenuItem ref="backButton" class="onyx-menu-item__back" v-bind="backButtonProps">
           <OnyxIcon :icon="iconArrowSmallLeft" />
-          {{ t("back") }}
+          {{ props.label || t("back") }}
```
larsrickert: This should usually be fine but the label can also be passed as slot, should we also consider this?

packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxFlyoutMenu/OnyxFlyoutMenu.vue
larsrickert, line 1: A test for this would be great :)

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxFlyoutMenu/OnyxFlyoutMenu.ct.tsx
```diff
@@ -165,7 +165,7 @@ test("should behave correctly with nested items (via mouse)", async ({ page, mou
   const trigger = page.getByRole("button", { name: "Trigger" });
   const firstItem = page.getByRole("menuitem", { name: "Item 1", exact: true });
   const nestedChild = page.getByRole("menuitem", { name: "Item 1.1" });
-  const backButton = page.getByRole("menuitem", { name: "Back" });
+  const backButton = page.locator(".onyx-menu-item__back").getByRole("menuitem");
```
larsrickert: We should target it via a label instead :)

packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxFlyoutMenu/OnyxFlyoutMenu.vue
```diff
@@ -63,6 +64,12 @@ const slots = defineSlots<{
 }>();
 
 const popover = ref<ComponentInstance<typeof OnyxBasicPopover>>();
+const menuRef = ref<ComponentInstance<typeof HTMLUListElement>>();
+
+const setMenuRef = (el: unknown) => {
+  menuRef.value = el as ComponentInstance<typeof HTMLUListElement>;
+};
```
larsrickert:
```suggestion
const menuRef = ref<HTMLElement>();
const setMenuRef: VNodeRef = (el) => (menuRef.value = el as typeof menuRef.value);
```

packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxFlyoutMenu/OnyxFlyoutMenu.vue
```diff
@@ -111,8 +146,11 @@ provide<NestedMenuDrilldownModeContext>(MENU_ITEM_DRILLDOWN_INJECTION_KEY, {
 
       <ul
         v-if="slots.options"
-        v-bind="menu"
+        v-bind="mergeVueProps(menu, { ref: setMenuRef as VNodeRef | undefined })"
```
larsrickert:
```suggestion
        v-bind="mergeVueProps(menu, { ref: setMenuRef })"
```

packages/sit-onyx/src/composables/useResizeObserver.ts
```diff
@@ -7,6 +7,11 @@ export type UseResizeObserverOptions = {
    * @default 'content-box'
    */
   box?: ResizeObserverBoxOptions;
+  /**
+   * Whether the observer is disabled.
+   * * @default false
+   */
+  disabled?: MaybeRef<boolean>;
```
larsrickert:
```suggestion
  disabled?: Ref<boolean>;
```

Imo it does not make sense to allow non-ref values since it would always be enabled or disabled then

packages/sit-onyx/src/composables/useResizeObserver.ts
```diff
@@ -45,21 +52,29 @@ export const useResizeObserver = (
...
-        if (oldTarget) observer?.unobserve(oldTarget);
-        if (newTarget) observer?.observe(newTarget, { box });
-        else {
-          // target was removed (e.g. with v-if so we need to reset the size manually)
+        if (newTarget) {
+          observer.observe(newTarget, { box });
+        } else {
           width.value = 0;
           height.value = 0;
         }
```
larsrickert:
```suggestion
    const targetElement = target ?? document.documentElement;

    watch(
      [() => toValue(targetElement), isDisabled],
      ([newTargetRef, disabled], [oldTargetRef]) => {
        const newTarget = getTemplateRefElement(newTargetRef);
        const oldTarget = getTemplateRefElement(oldTargetRef);

        if (oldTarget) observer.unobserve(oldTarget);

        // target was removed (e.g. with v-if) or observer is disabled so we need to reset the size manually
        if (!newTarget || disabled) {
          width.value = 0;
          height.value = 0;
          return;
        }

        observer.observe(newTarget, { box });
```

packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxFlyoutMenu/OnyxFlyoutMenu.vue
```diff
@@ -77,10 +84,38 @@ const {
...
+      minHeight.value = undefined;
+      return;
+    }
+
+    if (newHeight && (!minHeight.value || minHeight.value < newHeight)) {
+      minHeight.value = newHeight;
+    }
+  },
+  { immediate: true },
+);
```
larsrickert, line 108:
```suggestion
watch(
  [height, isObserverDisabled],
  ([newHeight, isDisabled]) => {
    minHeight.value = isDisabled ? undefined : Math.max(minHeight.value ?? 0, newHeight);
  },
  { immediate: true },
);
```

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxNavBar/modules/OnyxFlyoutMenu/OnyxFlyoutMenu.ct.tsx
ChristianBusshoff: The label is changing because it's now the label of the parent item. That's why I change it. 
Is there a better way to select the backButton?

packages/sit-onyx/src/composables/useResizeObserver.ts
```diff
@@ -7,6 +7,11 @@ export type UseResizeObserverOptions = {
    * @default 'content-box'
    */
   box?: ResizeObserverBoxOptions;
+  /**
+   * Whether the observer is disabled.å
+   * * @default false
```
larsrickert:
```suggestion
   * Whether the observer is disabled.
   * @default false
```

larsrickert (approved):

Issue #4810, opened 2026-02-10
Implement eslint update to version 10


JoCa96:
### Why?

eslint had a major version update, which broke some dependencies

- https://github.com/amilajack/eslint-plugin-compat/issues/675

### Acceptance criteria

- [x] `eslint` is updated to v10
- [x] If new eslint rules pop up, create a FoUp to "fix" them
- [x] Update `dependabot.yaml` and remove ESLint `ignore`

### Implementation details

Relevant issues:
- https://github.com/amilajack/eslint-plugin-compat/issues/675

### DEV handover checklist

DEV handover done with: @ChristianBusshoff 

**Other**
- [x] follow-up tickets were created if necessary (add links below)

PR #5745, opened 2026-07-01
chore: update eslint to v10

JoCa96:
Relates to #4810 

- Updated ESLint and related packages
- Fixed new lint issues
- Fixed breaking changes in ESLint-plugin
- Fixed broken storybook page because of too long code example in description
- Removed dependabot ESLint update rule


pnpm-workspace.yaml
```diff
@@ -51,3 +51,4 @@ overrides:
   vite: "catalog:"
   vue: "catalog:"
   sass-embedded: "catalog:"
+  "@microsoft/eslint-formatter-sarif@3.1.0>eslint": "10"
```
larsrickert: Why is this override needed?

larsrickert (approved):

pnpm-workspace.yaml
JoCa96: "@microsoft/eslint-formatter-sarif@3.1.0" has an explicit dependency on eslint 8; I added an comment.

PR #5747, opened 2026-07-02
feat(theme): improve color contrasts for progress item, segmented control and switch

ChristianBusshoff:
Relates to #5710

- Updated formElement border color (normal, hover)
- updated segmented control, tableOfContent und progress item hover effect


packages/sit-onyx/playwright/snapshots/components/OnyxSegmentedControl/Segmented-Control-chrome-linux.png
larsrickert, line 1: The hover state looks very dark, is that intended?

packages/sit-onyx/playwright/snapshots/components/OnyxSwitch/Switch-invalid--chrome-linux.png
larsrickert, line 1: Shouldn't the hover border color be red here?

larsrickert (changes requested):
We should also add a changeset for this + rename the PR title to something like:
feat(theme): improve color contrasts for progress item, segmented control and switch


Since this change is relevant for all themes, not just digits

packages/sit-onyx/playwright/snapshots/components/OnyxSegmentedControl/Segmented-Control-chrome-linux.png
ChristianBusshoff: I think so, in figma it's the same

packages/sit-onyx/src/components/OnyxSelectDialog/OnyxSelectDialog.vue
```diff
@@ -155,10 +155,10 @@ const handleApply = () => {
         &:hover {
           border-color: var(--onyx-color-component-border-primary);
         }
-
+        .onyx-text--small,
         .onyx-select-dialog__label,
         .onyx-select-dialog__icon {
-          color: var(--onyx-color-text-icons-primary-intense);
+          color: var(--onyx-color-text-icons-neutral-intense);
```
larsrickert, line 163: Isn't neutral intense the default color so we can remove this block of CSS? If we still need it, please add a dedicated class instead of. "onyx-text--small"

packages/sit-onyx/src/components/OnyxFormElementAction/OnyxFormElementAction.vue
```diff
@@ -115,6 +115,7 @@ const toggleAttrs = computed(() =>
 
     &__button {
       all: initial;
+      color-scheme: inherit;
```
larsrickert: Could you extend why this is needed?

packages/sit-onyx/src/components/OnyxPagination/OnyxSelectPagination.vue
```diff
@@ -162,6 +162,9 @@ const hasReachedMax = computed(() => props.modelValue >= props.pages);
     max-width: 100%;
 
     &__select {
+      .onyx-form-element-v2__input-container {
```
larsrickert, line 165: The vertical border color overlaps the outer border

.changeset/six-towns-stare.md
```diff
@@ -0,0 +1,5 @@
+---
+"sit-onyx": minor
+---
+
+feat(theme): improved color contrasts for default and hover states.
```
larsrickert:
```suggestion
feat(theme): improve color contrasts for several components
```

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxFormElementAction/OnyxFormElementAction.vue
ChristianBusshoff: 'all: initial' resets color-scheme, so re-inheriting is necessary.

larsrickert: I see. Maybe we should not use all instead? And manually reset background color, fonts, border etc?

packages/sit-onyx/playwright/snapshots/components/OnyxNavBar/burger-children-chrome-linux.png
larsrickert, line 1: Shouldn't the active text color also be neutral here just like the "regular" nav item?

packages/sit-onyx/src/components/OnyxPagination/OnyxSelectPagination.vue
larsrickert, line 1: Personally, I am not sure if we should adjust the pagination components this much only to have some minor change for the border color. What do you think?

larsrickert (approved):

larsrickert (approved):

PR #5764, opened 2026-07-03
fix(OnyxDataGrid): Fix column groups breaking lazy loading

### LINKED ISSUE #5612: Fix Lazy Loading not working in combination with ColumnGrouping in DataGrid
### Version

1.14.0

### Description

When using the OnyxDataGrid with a Column Group and Lazy Loading to achieve infinite scrolling for pagination, the infinite scrolling part doesnt work correctly anymore. If i start scrolling the first ne "page" does get loaded, but then the scroll bar is stuck and it doesnt load new data. If i then scroll a bit upwards and then downwards again the next batch gets loaded, but after that its stuck again. When removing the group from the column definition it works fine again.

I would expect that both features work together without breaking the the other one.

### Reproduction steps

I created a playground with the example code from the onyx documentation regarding the lazy loading feature and added the column group logic. If you test the lazy loading / infinite scrolling you see it isnt working correctly like described above. If you remove the column group from the column definition it starts working again.

[Playground Example](https://playground.onyx.schwarz/?onyxVersion=1.14.0#eNp9VVlz2zYQ/itbTh/olqJsN53psLZnVEf1pM3hqf3QNso4MLkSEYMgC4A6otF/7y5ASlSuJxKLPb799sA2mjRNumwxyqILmxvZOLDo2gaU0IvLWeTsLLqaaVk1tXGwBYPzBFbC5SXsYG7qCmYRmc+iXwdKz4UTN0YWv6NwrUGbwBu9WffSBNymQbiuVVvp61rP5eLgy0o3qknZO5xpr3k/1c5s4BK2Mw0giwx0Wz2iIQ0ALSrMwDoj9cILxILOe4Wdd5PX2jqY/j15dfty+vB8cj/JOq9v35HfiTFikzKCeAsK9cKVGZyd/wK7BOKHBKQucH0Cl1cBQfAmC7L0N/AjnPnQhpgzmgjgJAOw96/pA99vZbF7nwRsZMiodicDaLeTm+nD3Yt/p+Tz/OfBRUGkkYxov+gBX8XDRFKrZI7xaXLwccKeO5D2ZS0Koib4iOdCWRwGbsRCauFkre+ccNiF+rSA6e2x2lUciGiNQe2Iq4SP5AttBq+EK9McpTrGGXiF8QDn3upOfiRi9hck79jxnRazWuz5/wRuuhSqxbTD4d0Ju9E5xJ2IcOOhcNQ8PR/BktJ1pkUfiq/HY7CyahUz8SjyJ9QFFKjEBko0GHTESkgHGldwSw0jLcYxMVSrZQhE03MvK6xb14sTODs9PfVFYXuu6D76lys5AA8/DPja4/w8DV9YH2FHNAwrnPs5o7oMB65rpivf/m/ZbAtPuMloArlvZ1FCC+ARFQu4g1kQHN2Yum3+DKoL/j+bRT7mwAXhPvIwCeevODjvHLwbgF5JVx56jjB+1pGtxYNCaEdeFuxViY8bitc117BfvIyK3AhreSikBlciqG5GrB+BVYm6a6ODOVEOFLLgb9hzQtOqrIGWZq0U4JIKZkEYhFLoQpFmXlMVc6c2HLQLQdPfl+7Q5CHleZcZF+Q4fSbmYhy2M+1iOjisGm5SOgFcfDcawQp97NanJXr8uBakSJCoe2Fem5At0dPjSeGfuoWcDRQlQxkeXfs6UCt37lalpL2/kpzvf61QsJTG8TeQwAYUQKcwGgVkw7UPuSLW6UnhARgtSEJ1L1EUZIcknu6xBhHdZl3vQubXYLanKMAZU5CL8YALOlq3Ufyb7qOE0a/EelSiXJS0rc6fGax4A3tavUGU0EtHdaDRSD/YWtNz6M1mUV5XjVRo3jRcCnoNs36XUJ8rVa/+8DJeI767vE2J+dMX5B8svWvcobeUBJolT0V/54RZoAvX07vXuKb//WVVF60i7W9c/sXLpmWMQe23lrvQDPQ82he+d6lS93a6dqhtnxQD7dcHa9Ojfv2N1A9wf0qfeTvik1h8WKJhn0QgF8W6aPc/faHZHQ==)

### Confirmations

JoCa96:
Closes #5612 

*Cause:* The `colgroup` element is not explicitly into the grid and was therefor placed into the last cell. The scroll position was anchored to that element. This caused the scroll position always to be at the end, even when new rows were added. 

*Fix:* In a regular table the [`colgroup`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/colgroup) is a special element that can be used to group columns. To emulate this behavior in our `display: grid;` we use the fact, that the colgroup can also specify the groupings via [`col`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/col) elements. This way we can define the `colgroup` as a subgrid spanning the whole grid and the `col` elements spanning the relevant columns. 

Also:
- Enabled `css` minification for prod build
- Added missing `onyx.utility` layer for z-index specifications

### DIFF
--- a/packages/sit-onyx/src/components/OnyxDataGrid/OnyxDataGridRenderer/OnyxDataGridRenderer.vue
+++ b/packages/sit-onyx/src/components/OnyxDataGrid/OnyxDataGridRenderer/OnyxDataGridRenderer.vue
@@ -89,6 +89,12 @@ const columnStyle = computed(() => {
         grid-column: 1 / -1;
         grid-row: 1 / -1;
 
+        > colgroup {
+          // Prevents the "colgroup" element from participating in the CSS grid.
+          // This fixes the issue https://github.com/SchwarzIT/onyx/issues/5612
+          position: fixed;
+        }
+
         > thead {
           display: grid;
           grid-template-columns: subgrid;
--- a/packages/sit-onyx/src/components/OnyxDataGrid/OnyxDataGridRenderer/OnyxDataGridRenderer.vue
+++ b/packages/sit-onyx/src/components/OnyxDataGrid/OnyxDataGridRenderer/OnyxDataGridRenderer.vue
@@ -92,7 +92,21 @@ const columnStyle = computed(() => {
         > colgroup {
           // Prevents the "colgroup" element from participating in the CSS grid.
           // This fixes the issue https://github.com/SchwarzIT/onyx/issues/5612
-          position: fixed;
+          display: grid;
+          grid-template-columns: subgrid;
+          grid-template-rows: subgrid;
+          grid-column: 1 / -1;
+          grid-row: 1 / -1;
+
+          > col {
+            grid-row: 1 / -1;
+          }
+
+          @for $i from 1 through 99 {
+            > col[span="#{$i}"] {
+              grid-column: span $i;
+            }
+          }
         }
 
         > thead {
--- a/packages/sit-onyx/vite.config.prod.ts
+++ b/packages/sit-onyx/vite.config.prod.ts
@@ -13,6 +13,7 @@ export default defineConfig({
     bundlerConfig,
     {
       build: {
+        cssMinify: true,
         sourcemap: false,
         lib: {
           fileName: "index",
--- a/packages/sit-onyx/vite.config.prod.ts
+++ b/packages/sit-onyx/vite.config.prod.ts
@@ -13,7 +13,6 @@ export default defineConfig({
     bundlerConfig,
     {
       build: {
-        cssMinify: true,
         sourcemap: false,
         lib: {
           fileName: "index",
--- a/.changeset/dull-oranges-enjoy.md
+++ b/.changeset/dull-oranges-enjoy.md
@@ -1,0 +1,5 @@
+---
+"sit-onyx": patch
+---
+
+fix(DataGridFeatures.usePagination): Fix Lazy Loading not working in combination with ColumnGrouping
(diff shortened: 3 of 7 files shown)


packages/sit-onyx/src/components/OnyxDataGrid/OnyxDataGridRenderer/OnyxDataGridRenderer.vue
```diff
@@ -89,6 +89,26 @@ const columnStyle = computed(() => {
         grid-column: 1 / -1;
         grid-row: 1 / -1;
 
+        > colgroup {
```
larsrickert, line 92: Can you add a changeset for this?

packages/sit-onyx/vite.config.prod.ts
```diff
@@ -13,6 +13,7 @@ export default defineConfig({
     bundlerConfig,
     {
       build: {
+        cssMinify: true,
```
larsrickert: We have/had issues with lightnincss breaking light-dark() CSS variables, are we sure we want to enable this? Could break styles

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxDataGrid/OnyxDataGridRenderer/OnyxDataGridRenderer.vue
JoCa96: done

packages/sit-onyx/vite.config.prod.ts
JoCa96: done

larsrickert (approved):

larsrickert (approved):

Issue #4713, opened 2026-01-23
Fix OnyxDataGrid having an empty column / gap when using the resizing feature


larsrickert:
### Version

1.6.0

### Description

When using the [resize feature](https://storybook.onyx.schwarz/?path=/story/data-datagrid-features--resizing) of the data grid, it shows an empty column / blank space as the last column

<img width="676" height="689" alt="Image" src="https://github.com/user-attachments/assets/c3df17cd-a9a8-4fda-b2bf-8d9421a7a0d4" />

### Reproduction steps

1. Open [resize example](https://storybook.onyx.schwarz/?path=/story/data-datagrid-features--resizing) in Storybook
2. Scroll the data grid to the very end
3. See the empty column

### Confirmations

- [x] I have checked that there is not already a bug report for my issue.
- [x] I confirm that are no sensitive information included in this bug report. e.g. credentials, internal project details (for employees of the Schwarz Group) etc.

PR #5783, opened 2026-07-07
fix(DataGridFeatures.useResizing): Fix onyx data grid having an empty column gap when using the resizing feature

ChristianBusshoff:
Relates to #4713

- removed empty column for resizing feature


.changeset/hungry-impalas-scream.md
```diff
@@ -0,0 +1,5 @@
+---
+"sit-onyx": minor
+---
+
+fix(OnyxDataGrid:Resizing): removed empty column gap for the resizing feature.
```
JoCa96:
```suggestion
fix(DataGridFeatures.useResizing): Removed empty column gap for the resizing feature.
```

packages/sit-onyx/src/components/OnyxDataGrid/features/resizing/resizing.ts
```diff
@@ -118,28 +174,11 @@ export const useResizing = <TEntry extends DataGridEntry>(options?: ResizingOpti
...
       header: {
         wrapper:
-          (cols, i, { length }) =>
+          (cols) =>
           (_, { slots }) => {
-            const isLastColumn = i === length - 1;
-            const isFillerColumn = cols.type.name === FILLER_COLUMN;
-            return renderWrapper(slots, cols, isLastColumn || isFillerColumn);
+            return renderWrapper(slots, cols, false);
           },
```
JoCa96: If we don't need the third parameter anymore, we should remove it from the function.
```suggestion
          (_, { slots }) => renderWrapper(slots, cols);
```

packages/sit-onyx/src/components/OnyxDataGrid/features/resizing/resizing.ts
```diff
@@ -32,12 +34,22 @@ export const useResizing = <TEntry extends DataGridEntry>(options?: ResizingOpti
...
+            scrollContainer.value.scrollWidth - scrollContainer.value.clientWidth;
+        }
       },
       { flush: "post", deep: true },
     );
 
     const modifyColumns = (cols: Readonly<InternalColumnConfig<TEntry>[]>) => {
-      const columns = cols.map((column) => {
+      if (cols.length > 0) {
+        lastColumnKey.value = cols[cols.length - 1]!.key;
```
JoCa96: Prefer `.at`
```suggestion
        lastColumnKey.value = cols.at(-1)!.key;
```

packages/sit-onyx/src/components/OnyxDataGrid/features/resizing/resizing.scss
```diff
@@ -18,6 +14,27 @@
...
+        .onyx-resize-handle {
+          right: 0;
+          &::after,
+          &::before {
+            left: auto;
+            right: 0;
+          }
+        }
+      }
+    }
```
JoCa96: What is this needed for?

packages/sit-onyx/src/components/OnyxDataGrid/features/resizing/resizing.ts
```diff
@@ -82,26 +90,74 @@ export const useResizing = <TEntry extends DataGridEntry>(options?: ResizingOpti
...
+              // Prevents shrinking the table below the container size when dragging
+              const finalWidth = Math.max(lastColumnActiveMinWidth.value, width);
+              resizeState.value.set(column.key, `${finalWidth}px`);
+            } else {
+              resizeState.value.set(column.key, `${width}px`);
+            }
+          },
+          onAutoSize: () => {
+            if (column.key === lastColumnKey.value) {
+              resizeState.value.set(column.key, "minmax(max-content, 1fr)");
```
JoCa96, line 200: autosizing for the last column doesn't work

JoCa96 (changes requested):

packages/sit-onyx/src/components/OnyxDataGrid/features/resizing/resizing.ts
JoCa96, line 1: https://github.com/user-attachments/assets/cb1c8636-43e1-49e7-8d46-77d85c789129

The resizing doesn't work correctly when

1. The last column is resized to a bigger size
2. The second last column is resized
3. => then the size of the last column is reset

JoCa96, line 1: The last column can't be resized to smaller size, when the seconds last column was shrunk before.

https://github.com/user-attachments/assets/d9932271-379d-46e8-881d-bcfd27cc6c30

packages/sit-onyx/src/components/OnyxDataGrid/features/resizing/resizing.scss
ChristianBusshoff: to position the resize-handle for the last column completely inside the last-column so the user can grab it better.

packages/sit-onyx/src/components/OnyxDataGrid/features/resizing/resizing.ts
ChristianBusshoff: only works if the last column gets bigger or is allowed to shrink

packages/sit-onyx/src/components/OnyxDataGrid/features/resizing/resizing.ts
JoCa96: Sorry, should have described it better. There is no use-case for the `noResizeHandle` parameter of the `renderWrapper` anymore, so the parameter should be removed completely from the `renderWrapper` function:

```ts
    const renderWrapper = (
      slots: Slots,
      column: Readonly<InternalColumnConfig<TEntry>>
    ) => {
```

packages/sit-onyx/src/components/OnyxDataGrid/features/resizing/resizing.ts
```diff
@@ -17,27 +17,55 @@ export const useResizing = <TEntry extends DataGridEntry>(options?: ResizingOpti
     const { isEnabled } = useFeatureContext(ctx, options);
     const resizeState: Ref<ResizeState<TEntry>> = toRef(options?.resizeState ?? new Map());
     const scrollContainer = ref<HTMLElement>();
+    const lastColumnKey = ref<keyof TEntry>();
+    const lastColumnActiveMinWidth = ref(MIN_COLUMN_WIDTH);
+    const lastColumnStartWidth = ref(MIN_COLUMN_WIDTH);
+
+    const getContainer = (el?: HTMLElement) => {
+      return el?.closest<HTMLElement>(".onyx-table-wrapper__container") || scrollContainer.value;
+    };
```
JoCa96: scrollContainer is the same element as the one with this class `onyx-table-wrapper__container`, so the scrollContainer can always be used.
```suggestion
```

packages/sit-onyx/src/components/OnyxDataGrid/features/resizing/resizing.scss
```diff
@@ -18,6 +14,27 @@
 
   // styles to apply to the whole data grid while actively resizing a column
   .onyx-data-grid {
+    th:last-child,
+    td:last-child {
+      width: 100%;
+    }
+    &:has(.onyx-data-grid-resize-cell) {
```
JoCa96: This effects every cell, is it necessary?

ChristianBusshoff: image

Yes, otherwise this can happen

JoCa96 (approved):

PR #5809, opened 2026-07-08
feat: improve onyx-mcp capabilities

JoCa96:
Relates to #5718

- implemented semver utility to correctly compare and parse semver strings
- fix(list-components): ignore pre-release versions
- fix(get-component-api): Add hint when no `events`, `slots`, `props` or `exposed` are not defined
- feat: Add `onyx-install-mcp` skill


.changeset/green-heads-drive.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sit-onyx/modelcontextprotocol": patch
+---
+
+fix(get-component-api): Add hint when no `events`, `slots`, `props` or `exposed` are not defined
```
larsrickert:
```suggestion
fix(get-component-api): Add hint when no `events`, `slots`, `props` or `exposed` are defined
```

packages/modelcontextprotocol/src/index.ts
larsrickert, line 1: Aren't CLI flags usually named in kebab-case?

larsrickert (approved):

packages/modelcontextprotocol/src/index.ts
JoCa96: Yes, often

larsrickert: So shouldn't we do the same? Currently they are written in pascalCase (also this change here)

JoCa96: For now I would like to avoid a breaking change and still keep it consistent

PR #5811, opened 2026-07-09
fix(OnyxTab): Fix incorrect font-family usage

JoCa96:
The font-family for the OnyxTabs was incorrect.
We can just remove the definition, as It's already defined via the headline style.


packages/sit-onyx/src/components/OnyxTab/OnyxTab.vue
larsrickert, line 1: If we remove the font-family, the tab button does not define any font family, right? Is this intended?

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxTab/OnyxTab.vue
JoCa96: Its already defined via the headline style, also this is the wrong font-family type

larsrickert (approved):

PR #5838, opened 2026-07-15
chore: remove old "new" tags in storybook

JoCa96:


larsrickert (approved):
Would be great if you can also change this for the showcase :)

PR #5854, opened 2026-07-16
docs: update installation page

larsrickert:
Update the "Installation" page of our new showcase

## Checklist


apps/showcase/nuxt.config.ts
```diff
@@ -44,6 +47,14 @@ export default defineNuxtConfig({
...
       },
     },
+    server: {
+      fs: {
+        allow: [monorepoRoot],
+      },
+    },
+  },
+  alias: {
+    "@repo-root": monorepoRoot,
```
JoCa96: Would prefer to use the convention with the `#` prefix, so it can't be confused with a package:

```suggestion
    "#root": monorepoRoot,
```

larsrickert: Done 👍

JoCa96 (approved):

PR #5869, opened 2026-07-17
Fix favicon crash in showcase and downgrade to fixed changeset version

JoCa96:


packages/flags/package.json
```diff
@@ -31,7 +31,7 @@
     "generate:changeset": "tsx ./scripts/changeset.ts"
   },
   "devDependencies": {
-    "@changesets/write": "1.0.0-next.6",
+    "@changesets/write": "^0.4.0",
```
larsrickert, line 34:
```suggestion
    "@changesets/write": "~0.4.0",
```

should we use ~here since its a 0.x version? Same for packages/icons/package.json

larsrickert (approved):

packages/flags/package.json
JoCa96: The `^` already behaves this way, a it refers to leftmost non-zero digit

PR #5878, opened 2026-07-20
feat(Switch): implement FormElementV2

ChristianBusshoff:
Relates to #5802

-implement FormElementV2 inside Switch


packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.vue
```diff
@@ -43,52 +38,48 @@ const emit = defineEmits<{
...
   default: false,
 });
 
+const { formElementV2Props: legacyProps } = useLegacyFormElementProps({ props, errorMessages });
+
+const formElementV2Props = computed(() => {
+  return {
+    ...legacyProps.value,
+    loading: false,
+    label: props.descriptionLabel ?? {
```
larsrickert: What do you think if we remove the descriptionLabel property and use the position: "right" as default here? Since the form element already supports right aligned labels, it would be nice to use this here for consistency

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxSwitch/types.ts
```diff
@@ -1,11 +1,19 @@
...
-  "value"
-> & {
-  /**
-   * Whether the switch should be checked or not.
-   */
-  modelValue?: Nullable<boolean>;
-};
+  "value" | "label"
+> &
+  Pick<OnyxFormElementV2Props, "label"> & {
```
larsrickert, line 10: Would be nice to also support new properties like message and success

packages/sit-onyx/src/components/OnyxSwitch/types.ts
```diff
@@ -1,11 +1,19 @@
...
+    /**
+     * Whether the switch should be checked or not.
+     */
+    modelValue?: Nullable<boolean>;
+    /**
+     * The text label displayed directly next to the switch.
+     * Can be a single string for a static label, or an object with `truthy` and `falsy` properties
+     * to display different texts based on the checked state of the switch.
+     */
+    valueLabel?: string | { truthy: string; falsy: string };
```
larsrickert:
```suggestion
    valueLabel?:{ truthy: string; falsy: string };
```

I am not sure if a static label makes sense here since the regular label should be used for this and the value label is true/false specific

packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.vue
```diff
@@ -43,62 +37,84 @@ const emit = defineEmits<{
...
+    return { label: props.label, position: "right" as const, hidden };
+  }
+  if (props.label) {
+    return {
+      ...props.label,
+      hidden: hidden || props.label.hidden,
+    };
+  }
+  return props.label;
+});
```
larsrickert:
```suggestion
const normalizedLabel = computed<FormElementV2LabelOptions>(() => {
  const labelObject =
    typeof props.label === "string" ? { label: props.label } : props.label;
  return { position: "right", hidden: props.hideLabel, ...labelObject };
});
```

Isn't this the same?

packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.vue
```diff
@@ -43,62 +37,84 @@ const emit = defineEmits<{
...
+
+const legacyProps = new Proxy(props, {
```
larsrickert: Why is this needed? Can't we just pass the props to useLegacyFormElementProps and override the label property in the template?

packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.vue
```diff
@@ -43,62 +37,84 @@ const emit = defineEmits<{
...
+
+const { formElementV2Props: legacyFormElementProps } = useLegacyFormElementProps({
+  props: legacyProps,
+  errorMessages,
+});
+
+const formElementV2Props = computed(() => {
+  return {
+    ...legacyFormElementProps.value,
+    loading: false,
```
larsrickert: Could you add a short comment why we disable this here?

packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.vue
```diff
@@ -114,33 +130,33 @@ useAutofocus(input, props);
...
         <span
-          v-if="!props.hideLabel"
+          v-if="displayValueLabel"
           class="onyx-switch__label"
-          :class="[
-            `onyx-truncation-${props.truncation}`,
-            // shows the required marker inline for multiline labels
-            props.truncation === 'multiline' ? requiredMarkerClass : undefined,
-          ]"
+          :class="[`onyx-truncation-${props.truncation}`]"
```
larsrickert:
```suggestion
          v-if="displayValueLabel"
          aria-hidden="true"
          :class="['onyx-switch__label', `onyx-truncation-${props.truncation}`]"
```

Aria pattern recommends to set aria-hidden to avoid redundant announcement of the checked state

packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.vue
```diff
@@ -114,33 +130,33 @@ useAutofocus(input, props);
...
+      height: var(--onyx-switch-frame-height);
+      width: calc(2 * var(--onyx-switch-icon-size) - 2 * var(--onyx-switch-container-padding));
+      border-radius: var(--onyx-radius-full);
+      margin: var(--onyx-switch-label-padding-vertical) 0;
+      position: relative;
+    }
+  }
+}
+
+.onyx-switch {
```
larsrickert, line 147: Would be great to keep the spacings. Currently the gap is bigger than before

Currently:
Image

Before:

Image

packages/sit-onyx/src/components/OnyxSwitch/types.ts
larsrickert, line 1: Could you add screenshots tests for the top, left and right label as well as the new valueLabel?

packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.stories.ts
```diff
@@ -23,6 +23,13 @@ export const Default = {
   },
 } satisfies Story;
 
+export const WithValueLabel = {
+  args: {
```
larsrickert, line 28:
```suggestion
  tags: ["new:feature"],
  args: {
```

Please also add this example to the new showcase as well as the different label positions + messages

.changeset/warm-experts-sort.md
```diff
@@ -0,0 +1,5 @@
+---
+"sit-onyx": minor
+---
+
+feat(OnyxSwitch): implement OnyxFormElementV2 inside the Switch component
```
larsrickert:
```suggestion
feat(OnyxSwitch): use OnyxFormElementV2 internally. This now supports: top, left and right aligned label,`message`, `success` and `valueLabel` property
```

larsrickert (changes requested):

apps/showcase/content/en/components/02.form-elements/switch/index.md
```diff
@@ -25,3 +25,16 @@ The switch can be disabled to indicate that its action is currently not availabl
...
+
+The switch label can be positioned in several ways to support a wide variety of layouts.
+
+:component-example{layout="grow" name="LabelPositions" orientation="vertical"}
+
+### Value label
+
+The switch value label can be used to display different labels depending on its state.
+
+:component-example{layout="grow" name="ValueLabel"}
```
larsrickert:
```suggestion
:component-example{name="LabelPositions" orientation="vertical"}

### Value label

The switch value label can be used to display different labels depending on its state.

:component-example{name="ValueLabel"}
```

Would prefer to remove the grow so the examples are aligned with the other examples :)

apps/showcase/content/en/components/02.form-elements/switch/index.md
larsrickert, line 1: Would be great to also have examples for message, error and success :) Can also be done in a follow up PR

packages/sit-onyx/playwright/snapshots/components/OnyxFormElementV2/Form-element-v2-label-positions--chrome-linux.png
larsrickert, line 1: The left and right label are no longer vertically aligned :/ I guess this is related to the `margin-block: auto` change

packages/sit-onyx/playwright/snapshots/components/OnyxSwitch/Switch-invalid--chrome-linux.png
larsrickert, line 1: The label is not vertically centered with the switch here

packages/sit-onyx/playwright/snapshots/components/OnyxSwitch/Switch-label-position--chrome-linux.png
larsrickert, line 1: The left label has a bigger gap than the right label, is that intended?

packages/sit-onyx/playwright/snapshots/components/OnyxSwitch/Switch-truncation-required-optional--chrome-linux.png
larsrickert, line 1: multiline truncation seems to be broken here

packages/sit-onyx/src/components/OnyxFormElementV2/OnyxFormElementV2.vue
```diff
@@ -277,9 +277,7 @@ const popoverLayoutProps = useForwardProps(props, MaybePopoverLayout);
 
       > .onyx-form-element-v2__label {
         width: max-content;
-        padding-block: calc(
-          var(--onyx-form-element-v2-padding-block) + var(--onyx-form-element-v2-border-size)
-        );
+        margin-block: auto;
```
larsrickert: See comments on the screenshots. This change propaply misalignes the label vertically with the input. What was the reason why you removed the padding here?

larsrickert (changes requested):

apps/showcase/content/en/components/02.form-elements/switch/index.md
```diff
@@ -19,9 +19,29 @@ The loading state is used after a user interaction to indicate that the triggere
 
 :component-example{name="Loading"}
 
+### Message
+
+An optional message, error or success message can be displayed. Each message supports showing an info tooltip with further information.
+When multiple message types are defined at once, only the most relevant will be displayed (e.g. error message takes precedent over the regular message).
+
+:component-example{name="Message" layout="grow" orientation="vertical"}
```
larsrickert, line 27:
```suggestion
:component-example{name="Message" orientation="vertical"}
```

Would remove grow so the example is aligned with the other examples

apps/showcase/content/en/components/02.form-elements/switch/index.md
```diff
@@ -19,9 +19,29 @@ The loading state is used after a user interaction to indicate that the triggere
...
+
+The switch label can be positioned in several ways to support a wide variety of layouts.
+
+:component-example{ name="LabelPositions" orientation="vertical"}
+
+### Value label
+
+The switch value label can be used to display different labels depending on its state.
+
+:component-example{ name="ValueLabel"}
```
larsrickert, line 46: Can we move this example below Basic?

apps/showcase/content/en/components/02.form-elements/switch/examples/Message.example.vue
```diff
@@ -0,0 +1,23 @@
...
+  <!--
+    "show-error" is only set for this example so the error is displayed immediately.
+    Remove it if not needed so the error is only displayed after user interaction.
+  -->
+  <OnyxSwitch
+    :label="{ label: 'Example label', position: 'top' }"
+    :error="{ shortMessage: 'Custom error', longMessage: 'Error tooltip content' }"
+    show-error
+  />
+  <OnyxSwitch
```
larsrickert, line 19: Can you please check the hover styles for the success state? They look off:

Image

Image

packages/sit-onyx/playwright/snapshots/components/OnyxSwitch/Switch-truncation-required-optional--chrome-linux.png
larsrickert, line 1: For multiline, the required asterisk should still be placed at the end of the text, right?
Also the (optional) text in the top right screenshot seems to be overlapping

packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.vue
```diff
@@ -360,23 +403,22 @@ $input-width: calc(2 * var(--onyx-switch-icon-size) - 2 * var(--onyx-switch-cont
...
-      height: var(--onyx-switch-frame-height);
-      border-radius: var(--onyx-radius-full);
-      width: $input-width;
+    &.onyx-form-element-v2--label-right:not(:has(.onyx-switch__display-label)),
+    &.onyx-form-element-v2--label-left:not(:has(.onyx-switch__display-label)) {
+      cursor: pointer;
+      gap: var(--onyx-switch-label-padding-vertical);
+      .onyx-truncation-ellipsis {
+        cursor: inherit;
+      }
```
larsrickert: Why is this needed?

packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.vue
```diff
@@ -360,23 +403,22 @@ $input-width: calc(2 * var(--onyx-switch-icon-size) - 2 * var(--onyx-switch-cont
...
     }
-
-    &__label {
-      height: var(--onyx-spacing-md);
-      width: var(--onyx-spacing-3xl);
+    &:has(.onyx-truncation-multiline) {
```
larsrickert: Can we use dedicated CSS classes instead of using `onyx-truncation-multiline`?

packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.vue
```diff
@@ -360,23 +403,22 @@ $input-width: calc(2 * var(--onyx-switch-icon-size) - 2 * var(--onyx-switch-cont
...
+      .onyx-form-element-v2__content {
+        height: auto;
+      }
```
larsrickert, line 414: Should we apply this inside the OnyxFormElementV2?

packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.vue
```diff
@@ -360,23 +403,22 @@ $input-width: calc(2 * var(--onyx-switch-icon-size) - 2 * var(--onyx-switch-cont
...
+    }
+    // Truncates messages if the label is on the right to prevent layout shifts
+    &.onyx-form-element-v2--label-right .onyx-form-element-v2__bottom {
```
larsrickert, line 417: Should the same be applied to left label?

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.vue
ChristianBusshoff: Don't need to, or? The error starts with the formElement, so there is no layout shift for left.

packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.vue
ChristianBusshoff: To have a pointer cursor while hovering over the value-label.

larsrickert: For other form elements, the label also has a regular cursor instead of a selection-specific one so I'd prefer to keep this aligned. What do you think?

ChristianBusshoff: It label had the pointer curser before, but I'm also fine to align it.

larsrickert: Then lets align it for now an keep the regular cursor :)

packages/sit-onyx/src/components/OnyxSwitch/OnyxSwitch.vue
ChristianBusshoff: this is for the valueLabel so only needed inside the switch.

packages/sit-onyx/playwright/snapshots/components/OnyxFormElementV2/Form-element-v2-label-positions--chrome-linux.png
larsrickert, line 1: The label shouldn't be truncated here, right? 🤔

packages/sit-onyx/playwright/snapshots/components/OnyxSwitch/Switch-truncation-required-optional--chrome-linux.png
larsrickert, line 1: Shouldn't the label be position right but aligned left? As in the previous screenshot?

larsrickert (changes requested):

larsrickert (approved):

PR #5884, opened 2026-07-21
feat(FileUpload): implement new form element layout for file upload

ChristianBusshoff:
Relates to #5802

- implemented  form element v2 inside file upload
- made the small variant matches the other formElements hights.


packages/sit-onyx/src/components/OnyxFileUpload/OnyxFileUpload.vue
```diff
@@ -431,6 +418,39 @@ const shouldShowFileList = computed(() => {
...
+      width: calc(20rem + 2 * var(--onyx-density-xl));
+    }
+
+    &--medium .onyx-form-element-v2__content-skeleton {
+      height: calc(3.125rem + 2 * var(--onyx-density-xs));
+      width: calc(16rem + 2 * var(--onyx-density-xs));
+    }
+    &--small .onyx-form-element-v2__content-skeleton {
+      width: calc(6rem + 2 * var(--onyx-density-sm));
+    }
```
larsrickert, line 428:
```suggestion
    &--large .onyx-form-element-v2__content-skeleton {
      --onyx-form-element-v2-content-height: 7lh;
    }

    &--medium .onyx-form-element-v2__content-skeleton {
      --onyx-form-element-v2-content-height: 2lh;
    }
    &--small .onyx-form-element-v2__content-skeleton {
      --onyx-form-element-v2-content-height: 1lh;
    }
```

Would prefer to use the existing CSS variable to define the height. Also we should not limit the width imo

packages/sit-onyx/src/components/OnyxFileUpload/OnyxFileUpload.vue
```diff
@@ -431,6 +418,39 @@ const shouldShowFileList = computed(() => {
...
+    &.onyx-form-element-v2--label-left,
+    &.onyx-form-element-v2--label-right {
+      .onyx-form-element-v2__label {
+        margin-block: auto;
+      }
+    }
+  }
+
+  .onyx-file-upload-container {
+    --onyx-file-upload-max-files: 3;
```
larsrickert: Would prefer to keep the CSS variable on the parent element (wrapper) since if projects defined this variable with a custom value, moving it to a child would override the value set by the project

packages/sit-onyx/src/components/OnyxFileUpload/types.ts
```diff
@@ -4,9 +4,17 @@ import type { Nullable } from "../../types/utils.js";
...
+import type { FormElementV2LabelOptions } from "../OnyxFormElementV2/types.js";
 
 export type OnyxFileUploadProps<TMultiple extends boolean> = DensityProp &
-  Pick<SharedFormElementProps, "name" | "required" | "showError"> & {
+  Pick<
+    SharedFormElementProps,
+    "name" | "required" | "showError" | "requiredMarker" | "reserveMessageSpace"
+  > & {
+    /**
+     * Label to show for the file upload.
```
larsrickert:
```suggestion
     * Label to show for the file upload.
     * Will be required in onyx version 2.
```

+ please add this to a v2 follow up :)

packages/sit-onyx/src/components/OnyxFileUpload/types.ts
```diff
@@ -4,9 +4,17 @@ import type { Nullable } from "../../types/utils.js";
 import type { BinaryPrefixedSize } from "../../utils/numbers.js";
 import type { FormInjected } from "../OnyxForm/OnyxForm.core.js";
 import type { SharedFormElementProps } from "../OnyxFormElement/types.js";
+import type { FormElementV2LabelOptions } from "../OnyxFormElementV2/types.js";
 
 export type OnyxFileUploadProps<TMultiple extends boolean> = DensityProp &
-  Pick<SharedFormElementProps, "name" | "required" | "showError"> & {
+  Pick<
```
larsrickert: What about other form element properties such as error, message and success?

packages/sit-onyx/src/components/OnyxFileUpload/OnyxFileUpload.vue
```diff
@@ -111,6 +103,20 @@ const { vCustomValidity, errorMessages } = useFormElementError({
   error: requiredError,
 });
 
+const legacyProps = new Proxy(props, {
```
larsrickert: To prevent needing a proxy, we could either support computed values for the "props" of useLegacyFormElementProps or pass in the regular props and then override the when v-binding the props

packages/sit-onyx/src/components/OnyxFileUpload/OnyxFileUpload.stories.ts
larsrickert, line 1: Can you also create example for the new showcase for the different label positions (top, left, right) and message, success and error?

.changeset/swift-trees-rescue.md
```diff
@@ -0,0 +1,5 @@
+---
+"sit-onyx": minor
+---
+
+feat(OnyxFileUpload): implement OnyxFormElementV2 inside the File upload component
```
larsrickert, line 5: We should prefer documenting what actually changed for the user so e.g. which new properties are available now

larsrickert (changes requested):

apps/showcase/content/en/components/02.form-elements/file-upload/examples/Message.example.vue
```diff
@@ -0,0 +1,12 @@
+<script lang="ts" setup>
+import { OnyxFileUpload } from "sit-onyx";
+</script>
+
+<template>
+  <OnyxFileUpload :message="{ shortMessage: 'Message', longMessage: 'Message tooltip content' }" />
+  <OnyxFileUpload
+    :error="{ shortMessage: 'Custom error', longMessage: 'Error tooltip content' }"
+    show-error
+  />
```
larsrickert, line 14:
```suggestion
  <!--
    "show-error" is only set for this example so the error is displayed immediately.
    Remove it if not needed so the error is only displayed after user interaction.
  -->
  <OnyxFileUpload
```

packages/sit-onyx/src/components/OnyxFileUpload/OnyxFileUpload.ct.tsx
```diff
@@ -303,11 +302,12 @@ test("should show required error message", async ({ mount, page }) => {
   );
 
   const button = component.getByRole("button", { name: "Click to select" });
-  const errorMessage = component.locator(".onyx-file-upload__required_error");
+  const errorMessage = component.getByText(
+    "You need to upload at least one fileYou need to upload at least one fileYou",
+  );
```
larsrickert:
```suggestion
  const errorMessage = component.getByText("You need to upload at least one file").first();
```

packages/sit-onyx/src/components/OnyxFileUpload/OnyxFileUpload.stories.ts
```diff
@@ -17,6 +17,13 @@ export const Default = {
   },
 } satisfies Story;
 
+export const WithLabel = {
+  args: {
```
larsrickert, line 22:
```suggestion
  tags: ["new:feature"],
  args: {
```

packages/sit-onyx/src/components/OnyxFileUpload/OnyxFileUpload.vue
```diff
@@ -111,6 +103,17 @@ const { vCustomValidity, errorMessages } = useFormElementError({
...
+const legacyFormElementProps = computed(() => {
+  return {
+    ...props,
+    label: props.label ?? { label: t.value("fileUpload.clickToUpload"), hidden: true },
+  };
+});
+const { formElementV2Props } = useLegacyFormElementProps({
+  props: legacyFormElementProps,
+  errorMessages,
+});
```
larsrickert:
```suggestion
const { formElementV2Props } = useLegacyFormElementProps({
  props: computed(() => ({
    ...props,
    label: props.label ?? { label: t.value("fileUpload.clickToUpload"), hidden: true },
  })),
  errorMessages,
});
```

(optional) If we don't need to access legacyFormElementProps, we can also inline it

packages/sit-onyx/src/components/OnyxFileUpload/types.ts
```diff
@@ -1,12 +1,28 @@
...
+    | "message"
+    | "error"
+    | "skeleton"
+    | "disabled"
+  > & {
+    /**
+     * Label to show for the file upload.
+     * Will be required in onyx version 2.
+     */
+    label?: string | FormElementV2LabelOptions;
```
larsrickert:
```suggestion
  > &
  Partial<Pick<OnyxFormElementV2Props, "label">> & {
```

(optional) we can also pick this from the OnyxFormElementV2 so we don't need to maintain the type manually

packages/sit-onyx/src/components/OnyxFileUpload/types.ts
```diff
@@ -1,12 +1,28 @@
...
+    | "required"
+    | "showError"
+    | "requiredMarker"
+    | "reserveMessageSpace"
+    | "success"
+    | "message"
+    | "error"
+    | "skeleton"
+    | "disabled"
+  > & {
```
larsrickert:
```suggestion
  Pick<SharedFormElementProps, "name" | "disabled"> &
  Pick<
    OnyxFormElementV2Props,
    | "required"
    | "showError"
    | "requiredMarker"
    | "reserveMessageSpace"
    | "success"
    | "message"
    | "error"
    | "skeleton"
  > & {
```

Would prefer to pick as much as possible from OnyxFormElementV2Props since "SharedFormElementProps" are the old props where e.g. the message types are legacy

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxFileUpload/types.ts
ChristianBusshoff: For the other components we did the same, I thought it's this way to support the old prop structure till the V2?

larsrickert: Exactly but for the file upload, there were no label, message, success and error properties before so we can directly use the new v2 props and do not need to support the old props

packages/sit-onyx/playwright/snapshots/components/OnyxFileUpload/File-upload-medium-firefox-linux.png
larsrickert, line 1: The medium upload got much bigger. Is this intended?

packages/sit-onyx/playwright/snapshots/components/OnyxFileUpload/File-upload-medium-max-file-sizes--chrome-linux.png
larsrickert, line 1: Is it intended that the size grew that much?

larsrickert (changes requested):

packages/sit-onyx/playwright/snapshots/components/OnyxFileUpload/File-upload-medium-firefox-linux.png
ChristianBusshoff: I adjusted the hight, the width got intended bigger for the tests.

packages/sit-onyx/playwright/snapshots/components/OnyxFileUpload/File-upload-medium-max-file-sizes--chrome-linux.png
ChristianBusshoff: I adjusted the hight, the width got intended bigger for the tests.

larsrickert (approved):

PR #5920, opened 2026-07-28
feat(nuxt): generate locales at build time

### LINKED ISSUE #1417: Implement build-time locales for the Nuxt module
So we could generate those files at build time / pre-publish of the nuxt module so we don't need to maintain them manually?

_Originally posted by @larsrickert in https://github.com/SchwarzIT/onyx/pull/1402#discussion_r1654436944_

larsrickert:
closes #1417

Generate the locale files for the Nuxt module dynamically at build time so we don't have to maintain them manually. Available locales are loaded from sit-onyx.

## Checklist

### DIFF
--- a/packages/nuxt/build.config.ts
+++ b/packages/nuxt/build.config.ts
@@ -1,38 +1,46 @@
 import fs from "node:fs/promises";
 import { createRequire } from "node:module";
 import path from "node:path";
-import { defineBuildConfig } from "unbuild";
+import { defineBuildConfig, type RollupOptions } from "unbuild";
 
 export default defineBuildConfig({
   hooks: {
-    "build:before": async (ctx) => {
+    "rollup:options"(ctx, options) {
+      options.plugins.push(createLocalesPlugin());
+    },
+  },
+});
+
+function createLocalesPlugin(): RollupOptions["plugins"][number] {
+  return {
+    name: "onyx:build-locales",
+    async buildStart() {
       const locales = await getOnyxLocales();
 
-      const targetPath = path.resolve(ctx.options.rootDir, "src/runtime/locales");
+      locales.map((locale) => {
+        this.emitFile({
+          type: "asset",
+          fileName: `runtime/locales/${locale}.js`,
+          source: createLocaleModule(locale),
+        });
+      });
+    },
+  };
+}
 
-      // clear folder so no longer existing locales are removed
-      await fs.rm(targetPath, { recursive: true, force: true });
-      await fs.mkdir(targetPath, { recursive: true });
+/**
+ * Builds the ESM wrapper module that exposes an onyx locale to `@nuxtjs/i18n`.
+ */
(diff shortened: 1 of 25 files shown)


pnpm-lock.yaml
JoCa96, line 1: Please run `pnpm dedupe`

packages/nuxt/build.config.ts
```diff
@@ -0,0 +1,50 @@
...
+
+          const fileContent = `// @ts-check
+// generated at build time, do not edit
+import ${importName} from "sit-onyx/locales/${locale}.json";
+
+export default { onyx: ${importName} };
+`;
+
+          const localePath = path.join(targetPath, `${locale}.js`);
+          await fs.writeFile(localePath, fileContent, "utf-8");
```
JoCa96: Files, which are checked into git, must not be changed at build-time.
Instead we should use something like [`emitFile`](https://rolldown.rs/reference/Interface.PluginContext#in-depth-type-asset) for generating files at buildtime.

JoCa96 (changes requested):

packages/nuxt/build.config.ts
```diff
@@ -0,0 +1,57 @@
...
+  },
+});
+
+function createLocalesPlugin(): RollupOptions["plugins"][number] {
+  return {
+    name: "onyx:build-locales",
+    async buildStart() {
+      const locales = await getOnyxLocales();
+
+      locales.map((locale) => {
```
JoCa96: The result of `.map` is not used, prefer `forEach`:
```suggestion
      locales.forEach((locale) => {
```

packages/nuxt/build.config.ts
```diff
@@ -0,0 +1,57 @@
...
+  };
+}
+
+/**
+ * Builds the ESM wrapper module that exposes an onyx locale to `@nuxtjs/i18n`.
+ */
+function createLocaleModule(locale: string) {
+  const importName = locale.replace("-", "");
+
+  return `// @ts-check
```
JoCa96: The `@ts-check` doesn't have an effect here, so it can be removed.

```suggestion
  return `
```

packages/nuxt/build.config.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export default defineBuildConfig({
+  hooks: {
+    "rollup:options"(ctx, options) {
+      options.plugins.push(createLocalesPlugin());
+    },
+  },
+});
+
+function createLocalesPlugin(): RollupOptions["plugins"][number] {
```
JoCa96: There is no harm here adding `rollup` directly as a devDependency, so we can use a cleaner type here:

```suggestion
function createLocalesPlugin(): Plugin {
```

JoCa96 (approved):

JoCa96 (approved):

Issue #5900, opened 2026-07-22
Fix overflowing content for resizing datagrid


ChristianBusshoff:
### Why?

<img width="1071" height="557" alt="Image" src="https://github.com/user-attachments/assets/7b0312a8-7734-4a8f-afaf-07e8b1009740" />

### Acceptance criteria

- [ ] fix the bug

**Other**
- [ ] follow-up tickets were created if necessary (add links below)

PR #5931, opened 2026-07-30
fix(DataGrid): fixed overflowing cell content 

ChristianBusshoff:
Relates to #5900


packages/sit-onyx/src/components/OnyxCalendar/OnyxCalendar.vue
```diff
@@ -404,6 +403,7 @@ useOutsideClick({
         background-color: var(--onyx-color-base-background-blank);
 
         th {
+          overflow-wrap: normal;
```
larsrickert: Nice that fixes our current issue in the showcase where the weekdays are cut off :)

Image

But I am not sure if this is the correct fix since the issue still seems to be that the weekday cells do not have enough width for their content, thats why they overflow at all. Would be great to adjust the width instead

Image

packages/sit-onyx/src/components/OnyxTable/OnyxTable.vue
```diff
@@ -242,9 +242,9 @@ const headlineId = computed(() => (slots.headline ? _headlineId : undefined));
         &:hover:not(.onyx-table__empty, .onyx-table__standalone-row) > td::before {
           background-color: var(--onyx-color-base-neutral-200);
         }
-
         > th,
         > td {
+          overflow-wrap: break-word;
```
larsrickert, line 247: Could you please add a screenshot test for the cell and header overflow? So that we can ensure they behave correctly

.changeset/stale-hands-attend.md
```diff
@@ -0,0 +1,5 @@
+---
+"sit-onyx": major
+---
+
+fix(OnyxTable/OnyxDataGrid): Prevent long words from overflowing table cells.
```
larsrickert:
```suggestion
- fix(OnyxTable/OnyxDataGrid): Prevent long words from overflowing table cells.
- fix(OnyxCalendar): ensure weekday labels do not wrap

```

.changeset/stale-hands-attend.md
```diff
@@ -0,0 +1,5 @@
+---
+"sit-onyx": major
```
larsrickert:
```suggestion
"sit-onyx": patch
```

We do not want to publish onyx 2.0 with this 😅

larsrickert (changes requested):

packages/sit-onyx/playwright/snapshots/components/OnyxDataGrid/features/base/headline-filtered-chrome-linux.png
larsrickert, line 1: Is this change intended? Shouldn't the header actions be placed directly next to the label?

packages/sit-onyx/src/components/OnyxCalendar/OnyxCalendar.vue
```diff
@@ -371,6 +371,8 @@ useOutsideClick({
     $calendar-week-column-width: 2.5rem;
     $calendar-day-number-display-width: 2rem;
 
+    min-width: 18rem;
```
larsrickert, line 374: The min width might not always work, e.g. when switching the other languages where the weekday names are different. We can keep this for now tho but ideally the min-width is based on the content

larsrickert (changes requested):

packages/sit-onyx/playwright/snapshots/components/OnyxDataGrid/features/base/headline-filtered-chrome-linux.png
ChristianBusshoff: If the label doesn't get a width of 100% the overflowing is not working. That's why i changed it.

larsrickert: But then we should make sure that the content is still placed next to the label

ChristianBusshoff: Okay, by setting max-width instead it's also working. I change it.

larsrickert (approved):

PR #5936, opened 2026-07-30
fix(useAutofocus): autofocus is not performed when the target element is not visible or any other input element still has focus

JoCa96:


ChristianBusshoff (approved):

packages/sit-onyx/src/composables/useAutoFocus.ts
```diff
@@ -9,8 +29,32 @@ export const useAutofocus = (
...
   const performAutoFocus = () => {
-    const input = Array.isArray(ref) ? ref[0] : ref;
-    input.value?.focus();
+    const elem: Ref<Nullable<HTMLElement>> = Array.isArray(ref) ? ref[0] : ref;
+    if (elem.value === document.activeElement) {
+      return;
+    }
+    const isVisible = elem.value?.checkVisibility({
+      checkOpacity: true,
+      checkVisibilityCSS: true,
```
larsrickert:
```suggestion
      opacityProperty: true,
      visibilityProperty: true,
```

Those are historical aliases so we should not use them: https://developer.mozilla.org/en-US/docs/Web/API/Element/checkVisibility#parameters

packages/sit-onyx/src/composables/useAutoFocus.ts
```diff
@@ -9,8 +29,32 @@ export const useAutofocus = (
...
+    }
+    if (anyInputFocused()) {
+      return userConsole?.warn(
+        "Did not perform autofocus on Element ",
+        elem.value,
+        ". Because the ",
+        document.activeElement,
+        " is already focused!",
+      );
+    }
```
larsrickert, line 45: I am not sure if this is the desired behavior. What happens if another element is focused, a modal is then opened programatically that should autofocus another element inside the modal?

larsrickert (changes requested):

larsrickert (approved):

PR #5971, opened 2026-08-05
fix(demo-app): preserve theme query on navigation

### LINKED ISSUE #5573: Fix URL parameters for theme selection
### Why?

When navigating inside the [Demo App](https://demo-internal.onyx.schwarz/?theme=onyx), the ?theme=onyx parameter disappears from the URL on subpages. The router must keep this parameter on every page change so the selected theme stays active across all pages.

### DEV handover checklist

[Demo app](https://demo-internal.onyx.schwarz/?theme=lidl)

DEV handover done with: @ChristianBusshoff
UX handover done with: @mj-hof 

### UX handover checklist

UX handover done with: @ADD_UX_NAME_HERE

alectimison-maker:
## Description

Adds a global Demo App route middleware that carries the current `theme` query value to the next route. It preserves the target route's existing query and hash, and it does not overwrite a theme explicitly set by the target navigation.

Closes #5573

## Checklist

- [x] Targeted formatting and ESLint checks pass
- [x] Verified theme inheritance, explicit target themes, and navigation without a source theme using a mocked Nuxt middleware runtime
- [x] No changeset needed because this only changes the private Demo App

## Validation

- `pnpm exec oxfmt --check apps/demo-app/app/middleware/preserve-theme.global.ts`
- `pnpm exec eslint apps/demo-app/app/middleware/preserve-theme.global.ts`
- `git diff --check`
- 3 focused middleware behavior cases passed

The full Demo App dependency build is blocked locally on Windows by the existing `@sit-onyx/vite-plugin-svg` build script invoking the POSIX-only `cp` command.

### DIFF
--- a/apps/demo-app/app/middleware/preserve-theme.global.ts
+++ b/apps/demo-app/app/middleware/preserve-theme.global.ts
@@ -1,0 +1,14 @@
+export default defineNuxtRouteMiddleware((to, from) => {
+  const theme = from.query.theme;
+
+  if (theme == null || "theme" in to.query) return;
+
+  return navigateTo(
+    {
+      path: to.path,
+      query: { ...to.query, theme },
+      hash: to.hash,
+    },
+    { replace: true },
+  );
+});


larsrickert:
Thanks for the contribution! Since #5573 is related only to our Schwarz group internal demo application, I will close this PR and move the implementation to our internal repository

Issue #6045, opened 2026-08-19
Implement new variable for OnyxSlider track


Marko-Kordic:
### Why?

The OnyxSlide track color should be changed from **onyx/color/base/neutral/200** to **onyx/color/base/neutral/300** so that it no longer disappears on hover.

### Open questions / TODOs

The OnyxSlide track color should be changed from **onyx/color/base/neutral/200** to **onyx/color/base/neutral/300** so that it no longer disappears on hover.

### Design

"[Figma Documentation](ADD_FIGMA_LINK_HERE)"
"[Figma Component](ADD_FIGMA_LINK_HERE)"

### Acceptance criteria

The OnyxSlide track color should be changed from **onyx/color/base/neutral/200** to **onyx/color/base/neutral/300** so that it no longer disappears on hover.

### DEV handover checklist

[Storybook](https://onyx-storybook-dev.apps.01.cf.eu01.stackit.cloud/?path=/story/form-elements-slider--disabled)

DEV handover done with: @larsrickert 

**Testing**
- [x] covered by visual screenshot tests

### UX handover checklist

UX handover done with: @Marko-Kordic 

**Implementation is checked**
- [ ] disabled state

**Figma**
- [ ] Figma component is published

**Documentation**
- [ ] Storybook category is aligned with Figma
- [ ] Storybook examples are checked

**Other**
- [ ] follow-up tickets were created if necessary (add links below)

PR #6052, opened 2026-08-19
feat(Slider): implement new track variable

ChristianBusshoff:
Relates to #6045


packages/sit-onyx/playwright/snapshots/components/OnyxSlider/Slider-disabled--chrome-linux.png
larsrickert, line 1: The new rail color does not work for the disabled state

larsrickert (changes requested):

larsrickert (approved):

PR #6085, opened 2026-08-25
fix(OnyxDatePicker): preserve keyboard input while typing the year segment

figub:
## Summary

`OnyxDatePicker` (type `date` / `datetime-local`) destroys keyboard input while the user types the **year** segment. Typing `15.06.2026` digit by digit makes the year "dance":

```
year digit "2" → 0002-06-15
year digit "0" → 0020-06-15
year digit "2" → 0202-06-15
year digit "6" → 2026-06-15
```

For end users this looks like the field resetting on every keystroke (e.g. `00.00.0001 → 00.00.0000 → 00.00.0002 …`). It is most visible in locales the browser renders as `dd/mm/yyyy` (e.g. **es-ES**), where the segment order and reformat collide most noticeably.

## Root cause

A native `<input type="date">` fires `input` on **every** segment change as soon as the entered segments form a *valid* date. While the year is typed digit by digit, the field briefly holds valid dates with a 1–3 digit year. The `value` computed's setter fed each of those through `dateToISOString`, which zero-pads the year (`padStart(4, "0")`) and wrote the result **straight back into the same input** via `v-model` — overwriting the segment the user was still typing.

`v-model.lazy` does **not** fix this: the native input also fires `change` on every valid intermediate state.

## Fix

Guard the setter: ignore intermediate values whose **numeric** year is still incomplete (`< 1000`). Because `modelValue` is left unchanged, the getter keeps returning the previous string and Vue performs no DOM writeback, so the native input keeps the user's in-progress input.

Note: the native input already **zero-pads** the serialised year while typing (`"0002-06-15"`), so the year part is always 4 chars — the guard therefore compares the numeric year, not the string length.

Complete dates, calendar-popup selections and clearing the field all fall through unchanged.

## Verification

Verified in a browser against the real component (segment-by-segment typing of `15.06.2026`): the field ends on `2026-06-15` and emits **only** the final value — no intermediate padded years.

Added component tests:
- `does not emit zero-padded intermediate years while typing the year segment` (uses `pressSequentially`, which the existing `.fill()`-based test could not catch)
- `still commits a complete date and supports clearing (guard does not over-block)`

## Notes

- Scope: only the deprecated v1 `OnyxDatePicker`. `OnyxDatePickerV2` is calendar-based and unaffected.
- A changeset (`sit-onyx: patch`) is included.


JoCa96:
Hi @figub! We are unable to reproduce/comprehend the error, which you are describing here.
The `OnyxDatepicker` uses the browser native date input, so it might be a browser specific issue?
Although, we were unable to reproduce the issue in Chrome and Firefox.

To get rid of any misunderstanding on our side, please provide a [Playground example](https://playground.onyx.schwarz) or a screen recording of the issue.

JoCa96 (changes requested):
Please provide a reproducible error example.

figub:
Root cause: the native <input type="date"> is bound with v-model="value", which Vue compiles to :value + @input. A native date input fires input on every segment change as soon as the entered segments form a valid date, not just on blur. While the user types the year digit by digit, the field briefly holds valid dates with a 1 - 3 digit year (2, 20, 202). The computed setter runs on each of these:

set: (value) => {
  const newDate = new Date(value ?? "");
  modelValue.value = dateToISOString(newDate, "date");
}

dateToISOString zero-pads the year via padStart(4, "0") (2 -> 0002), assigns it to modelValue, the getter returns it, and Vue writes it back into the same native input through :value. That write-back overwrites the segment the user is currently typing, so the year "dances" 0002 -> 0020 -> 0202….

It surfaces most visibly in dd/mm/aaaa locales (e.g. es-ES, Spanish colleagues brought it up to us), but the destructive get/set/write-back roundtrip happens in every locale (visible via the raw input value / emit log in the playground repro regardless of format). However the date picker getting unusable only happened for spanish colleagues. 

The fix guards the setter so intermediate (incomplete-year) values are not committed and written back, leaving the user's in-progress input untouched. complete dates, calendar-popup selections and clearing are unaffected.

https://playground.onyx.schwarz/?onyxVersion=1.20.0#eNqFVu1u2zYUfRVC/VEbsGTFadzWdd11XYp1aNei6bofVTDQ0rXNRKI0kvJHkwB7iD3DHmxPskPKkiW3yBIgEe+9PLyfh7zxXhZFsC7Jm3hTHStRGKbJlAVLuVw+jzyjI28WSZEVuTLshuXyXV5KQ8mAKVqwO7ZQecYiDxCR96xl+F7utj9xQx9EfE3qYKiF8XPonHUk41xqw9Y8LYk9t5BTbZSQS3bLSpnQQkhKZr0+bCtLyoR5my87tl8uZ70vlwcbxTevBaXJ5wNqL/Iiz1pEclHK2IhcIpTfigQe9tYT9u2hfXYTSVZ5FtT+rYHAah8qaVCUetW7cPt76749484e0ySq1+uz57MKDal9A5nCzraYscpxIYvS4Jgkj8uMpAn+LEntLiil2ORq+vOnd2/fWJPzlKx61nvodnwxu4JQKxtM5F0+tD5Y0E4emhjclhf75YsXKIkrBWN3A3YShi4A+2c6rPoB1cfCUFakwMeKsWki1tUR2uxSe3S1YiwRGma7CVuktN27wdzCT4RCGEj8BNGmZSYb9ZIXE3aiKGskGd/6G5GY1YSdhl2FWgogdKwLniTIPqTB6KytmPP4eqlQhmTCHizcT6ODD7mC+GQxOh09OWzJVUKQnxRbpvNUJOxBEiaPEzqy8BVPRKkn7EnRijOXxl/wTKTIgOZS+5qUqM+ssuQSiBSuRk3y6qBCjNrR2Pz719/smnbznKtk3x7zcjkdrkauLA6p+C5Q7dNr9JJmZkVsYXuBcZkw2y+MM9svbDqfaVradmLzHRrUfU6H89kE+t2AZYhpNbAA0qHsiKsaG3sTsRRup/uw+5zPFCyD6phpnCc0OzmbDt3HYC8Ix43AQjeITjba62rb7vJIWwMFNcbv3MQrZBvtv4/Z9XrkVbEjBjvALAWLXJVZAR5gtMacudCqOPZY02FxSPNRZVI+pxQ5tyJATzL4kPruJEjrE38oHcUcaWvmgcHw/+r4EVFIbsSa9vV3IBOb+5ubI6q7vUXYPUyq2fWBfXdn61EFcc8Jr0qlbPmdjw6oRm9xRMOLR7j3Ip8jz2DACkezzUqkZLvCDmvbLUV17Y755GiGT8j+NvN2mOJHFD+dhy1Fwwlh8LjDCd8M8Lg1wAeCaWO5sdbiK/ISBk+7YCly4q9ILFfG8s9ZS4VwDfm64DE2IkZ/o3jR6OsYbZ7rC+UqF9LeVEir199XU+ZmZa+mHZm6psC6N/PPmrScPR6HY/6sG8ETm48WQ5xvCxAzIb9VaZoxd35pzEe62xOI5Gk9hOFo7IdjvxnsYDpXrpsr0JexKWHcA1v1AVwj2ktKxoRHhaMJKz8aUzeTIKy8Hv4wHHUOGnS5Auqwq68VcLGjwIn/MK4ZcdDDYdaZwGlgKrxo5hTzUqNFV4eGJHSwYhuFWlY0+pVU7tv2QmMXXBmBMB2a7VTMKDy3Zm5YA/Yut88bocUcrS8kWCfmaYXUXCgq3+CewDMFM6a0dbGKIEmGWTbk+Kkj6DleJe2fX/Q7HGX/41LG13TYuqq9AR5weFssBJpL5xKvPPfgiLw4zwpMo3pf2CsZ9ZjUT5HI42mab35xMqNK2icce1YUX39HfqXxmpvg44MiBLIGrzU6g6YkU6nPL36lLb4bJRinTGF9j/Ij4RYurY+V2Y8gArjdsnPevnHPTjTvJ32+NSR1HZR1tHrcVHHjnfrqntAP7p4Gj9w+vIaQxT9wQVhMJNCmVhsrM/pzIx0HYXDq3f0H9tCjbA==

JoCa96:
@figub Thanks for the quick response! But we don't see this as a bug/issue, this behaves as expected.
You have the same behavior in the native text input with an e.g. `minlength` requirement:
While the user types the intermediate and invalid values are also emitted.
If you have to avoid this you can use a debounce or wait for a change event.

Can you clarify what becoming unusable means? I am unable to recreate a case in which I was unable to continue typing or it wasn't the expected input.

figub:
@JoCa96 
https://github.com/user-attachments/assets/f2630a64-df2f-4cdb-a656-2b2d6a32d8e3

i have attached a video of the bug the colleagues in Spain encounter, the colleague entered "01072026" see for yourself. i was not able to recreate it either, but was able to fix it this way...

larsrickert:
@figub Thats definitely a weird issue 🤔 However, since the "OnyxDatePicker" is already deprecated, I'd recommend to migrate to the [OnyxDatePickerV2](https://storybook.onyx.schwarz/?path=/docs/form-elements-datepickerv2--docs) which will replace the "OnyxDatePicker" in onyx version 2

figub:
It's not very desirable, our power users want to enter their dates via keyboard, and V2 doesn't support that or did I miss anything?

larsrickert:
> It's not very desirable, our power users want to enter their dates via keyboard, and V2 doesn't support that or did I miss anything?

It has keyboard support so it can be navigated using the keyboard but it does not yet support to directly type in the date. We should make a feature request to the OnyxDatePickerV2 then since the current "OnyxDatePicker" will be removed in onyx version 2 so you should not rely on it :)

JoCa96:
> @JoCa96 https://github.com/user-attachments/assets/f2630a64-df2f-4cdb-a656-2b2d6a32d8e3
> 
> i have attached a video of the bug the colleagues in Spain encounter, the colleague entered "01072026" see for yourself. i was not able to recreate it either, but was able to fix it this way...

Thanks, that already helps a lot to understand whats going on! Can you find which locale and browser is used?

figub:
As i said, locale is es-ES, browser is chrome. i can get you in touch with someone who can produce the problem.

Issue #6000, opened 2026-08-11
Implement vertical nav bar integration for OnyxNavButton


larsrickert:
### Why?

The OnyxNavButton component that is used inside the nav bar for custom actions such as language selection etc. (see our [demo app](https://demo.onyx.schwarz)) should adapt automatically when used inside a vertical nav bar. This means, the nav button should always show a label when the vertical nav bar is expanded. Currently it looks "broken" if the developer does not handle this case manually

<img width="1023" height="960" alt="Image" src="https://github.com/user-attachments/assets/56cc8d46-9e9b-43b7-8eae-d08a56371192" />

### Acceptance criteria

- [ ] the vertical nav bar injects the information whether it is expanded to the OnyxNavButton
- [ ] the OnyxNavButton ensures the label is always shown if it is used inside a vertical nav bar that is expanded (even if hideLabel is set)

### Implementation details

- [ ] support new "auto" mode (default) for the "hideLabel" property of the OnyxNavButton

### DEV handover checklist

[Storybook](ADD_LINK_HERE)

DEV handover done with: @ADD_DEV_NAME_HERE

**Testing**
- [ ] covered by visual screenshot tests
- [ ] covered by functional tests (Playwright or unit test)

**Documentation**
- [ ] "New" badge in Storybook is added, see [storybook-addon-tag-badges](https://github.com/Sidnioulz/storybook-addon-tag-badges/?tab=readme-ov-file#-usage)
- [ ] Storybook code snippets are checked
- [ ] component status has been updated in the [docs overview](https://onyx.schwarz/#components)

**Other**
- [ ] follow-up tickets were created if necessary (add links below)

### UX handover checklist

UX handover done with: @ADD_UX_NAME_HERE

**Naming is aligned with Figma**
- [ ] component name
- [ ] which support components are used and how they are named
- [ ] all property names and types

**Implementation is checked**
- [ ] all features of Figma are implemented or a follow-up ticket is refined
- [ ] skeleton
- [ ] densities
- [ ] light and dark mode
- [ ] truncation
- [ ] responsiveness / wrapping behavior
- [ ] min / max width

**Figma**
- [ ] Figma component is published

**Documentation**
- [ ] Storybook category is aligned with Figma
- [ ] Storybook examples are checked

**Other**
- [ ] follow-up tickets were created if necessary (add links below)

PR #6057, opened 2026-08-20
docs(showcase): implement examples for some "Navigation" components

larsrickert:
Relates to #5500

Implement showcase component examples for:
- bottom bar
- breadcrumb
- headline
- link
- nav bar

The remaining components will follow in a separate PR

## Checklist

- [x] A changeset is added with `pnpm exec changeset add` if your changes should be released as npm package (because they affect the library usage)
- [x] I have performed a self review of my code ("Files changed" tab in the pull request)
- [x] All relevant changes are documented in the documentation app (folder `apps/docs`) if needed

PR #6101, opened 2026-08-26
feat(NavButton): implement vertical nav bar integration

ChristianBusshoff:
Relates to #6000


packages/sit-onyx/src/components/OnyxNavButton/OnyxNavButton.vue
```diff
@@ -24,6 +26,14 @@ defineSlots<{
   default?(): unknown;
 }>();
 
+const isExpanded = inject(NAV_BAR_IS_EXPANDED_INJECTION_KEY, undefined);
+
+const isLabelHidden = computed(() => {
```
larsrickert, line 31: I accidentaly added the "auto" property in #6057 :/
I'd prefer to remove the auto mode and always force to show the label when used inside an expanded vertical nav bar since I don't see any valid reason why there should be only an icon when expanded :)

What do you think?

apps/demo-app/app/components/NavBar.vue
```diff
@@ -69,10 +69,19 @@ const orientation = computed(() => (isVertical.value ? "vertical" : "horizontal"
       <DensitySwitch />
       <LocaleSwitch />
       <GlobalSearch />
-      <OnyxSeparator v-if="!isVertical" orientation="vertical" />
-      <NotificationCenter />
+      <OnyxSeparator
```
larsrickert: Should we change this inside the nav bar / separator itself? Since this is propably an issue that every project is facing when using the vertical nav bar, right?

.changeset/social-forks-deny.md
larsrickert, line 1: Isn't this changeset obsolete since this feature is already added and not new to this PR?

larsrickert (changes requested):

apps/demo-app/app/components/NavBar.vue
ChristianBusshoff: Good Point, I will implement it.

apps/demo-app/app/components/NotificationCenter.vue
```diff
@@ -1,14 +1,14 @@
...
-import { useGlobalFAB, useNotification } from "sit-onyx";
+import { OnyxIcon, OnyxUnstableNavButton, useGlobalFAB, useNotification } from "sit-onyx";
 import type { MyNotification } from "~/stores/notification-store";
 
 const store = useNotificationStore();
 const { t } = useI18n();
 
 const isSidebarOpen = ref(false);
-
+const props = defineProps<{ isVertical?: boolean; expanded: boolean }>();
```
larsrickert:
```suggestion
const props = defineProps<{ isVertical?: boolean; expanded?: boolean }>();
```

packages/sit-onyx/src/components/OnyxNavButton/types.ts
```diff
@@ -3,10 +3,9 @@ import type { OnyxButtonProps } from "../OnyxButton/types.js";
 
 export type OnyxNavButtonProps = Pick<OnyxButtonProps, "disabled" | "link" | "label" | "icon"> & {
   /**
-   * Whether the label should (visually) be hidden. If set to "auto", the label will be hidden
-   * automatically when used inside a vertical nav bar which is collapsed.
+   * Whether the label should (visually) be hidden.
```
larsrickert, line 6:
```suggestion
   * Whether the label should (visually) be hidden.
   * If used inside a vertical nav bar, the label is shown depending on the expanded state.
```

packages/sit-onyx/src/components/OnyxNavButton/OnyxNavButton.vue
```diff
@@ -24,6 +26,14 @@ defineSlots<{
...
+  if (isExpanded?.value) {
```
larsrickert:
```suggestion
  if (isExpanded) {
```

If the vertical nav bar is collapsed, we should still force the label to be hidden

packages/sit-onyx/src/components/OnyxNavBar/types.ts
```diff
@@ -75,10 +75,15 @@ export const NAV_BAR_MORE_LIST_TARGET_INJECTION_KEY = Symbol() as InjectionKey<
...
- * nav bar to communicate child components whether they should render horizontal or vertical.
+ * nav bar to communicate child components whether the vertical Navbar is expanded.
  */
 export const NAV_BAR_IS_EXPANDED_INJECTION_KEY = Symbol() as InjectionKey<Ref<boolean>>;
 
+/**
+ * [Vue injection key](https://vuejs.org/guide/components/provide-inject) that is provided by the
+ * nav bar to communicate child components whether they should render horizontal or vertical.
+ */
+export const NAV_BAR_ORIENTATION_INJECTION_KEY = Symbol() as InjectionKey<Ref<NavBarOrientation>>;
```
larsrickert: We already have a lot of nav bar injection keys that just represent a single value 🤔  Should we simplify them to just have a "NAV_BAR_INJECTION_KEY" that is an object with all the context details?

packages/sit-onyx/src/components/OnyxSeparator/OnyxSeparator.vue
larsrickert, line 1: Why are the changes in this file needed? Isn't there always only one separator orientation in the nav bar? So the user could just set the orientation accordingly?

I would try to prevent as much "magic" logic inside components as possible

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxNavButton/OnyxNavButton.vue
larsrickert, line 1: Please add a test for this behavior :)

packages/sit-onyx/src/components/OnyxNavBar/OnyxNavBar.ct.tsx
```diff
@@ -612,3 +612,29 @@ test("should switch to mobile correctly", async ({ mount, page }) => {
...
+  // ASSERT
+  await expect(labelElement).toBeHidden();
+  await expect(navButton).toHaveAttribute("title", "Settings");
+
+  // ACT
+  await component.getByRole("menuitem", { name: "Expand" }).click();
+
+  // ASSERT
+  await expect(labelElement).toBeVisible();
+  await expect(labelElement).toHaveText("Settings");
```
larsrickert:
```suggestion
  const navButton = component.getByRole("button", { name: "Settings" });
  const labelElement = navButton.getByText("Settings");

  // ASSERT
  await expect(labelElement).toBeHidden();
  await expect(navButton).toHaveAttribute("title", "Settings");

  // ACT
  await component.getByRole("menuitem", { name: "Expand" }).click();

  // ASSERT
  await expect(labelElement).toBeVisible();
```

We can use getByText() instead of the CSS class locator so if the locator is renamed or changed in the future, the test doesn't break

packages/sit-onyx/src/components/OnyxNavButton/OnyxNavButton.vue
larsrickert, line 31: Can you add a changeset for this?

larsrickert (approved):

Issue #6106, opened 2026-08-27
Implement SSR support for OnyxTabs


larsrickert:
### Why?

The tabs component is not fully SSR (server side rendering) compatible since we are using a Vue Teleport to move the tab content to the correct DOM position. This has the downside that when using with SSR (e.g. Nuxt), the content can not be server side rendered and also scroll behavior is not working (so e..g opening https://onyx-showcase-dev.apps.01.cf.eu01.stackit.cloud/components/data/data-grid#features does not scroll to the correct position)

### Acceptance criteria

- [ ] the OnyxTabs is compatible with SSR and renders the content also server side

### DEV handover checklist

[Storybook](ADD_LINK_HERE)

DEV handover done with: @ADD_DEV_NAME_HERE

**Testing**
- [ ] covered by visual screenshot tests
- [ ] covered by functional tests (Playwright or unit test)

**Other**
- [ ] follow-up tickets were created if necessary (add links below)

PR #6107, opened 2026-08-27
fix(nuxt-docs): implement custom scroll behavior

larsrickert:
Our showcase / the Nuxt documentation template does not handle scroll behavior correctly. Since onyx uses a custom scroll container inside OnyxAppLayout, the default Vue Router scroll behavior does not work. Therefore, opening a page with a hash (e.g. https://onyx-showcase-dev.apps.01.cf.eu01.stackit.cloud/introduction/foundation/colors#usage) does not scroll to the correct position and when navigating to another page, the page scroll position is kept and not scrolled correctly to the top of the new page.

This PR fixes this by implementing a custom scroll behavior tailored to the OnyxAppLayout.

Note: The scrolling does currently NOT work for our component example pages since the OnyxTags component is not SSR compatible (the content does only render after mounted). A follow up is created for this: #6106
 
## Checklist

### DIFF
--- a/.changeset/fresh-needles-follow.md
+++ b/.changeset/fresh-needles-follow.md
@@ -1,0 +1,5 @@
+---
+"@sit-onyx/nuxt": minor
+---
+
+feat: implement `useOnyxScrollBehavior` that can be used to correctly scroll to anchors on initial page load and scroll to the top of the page when navigating to a new page
--- a/.changeset/tricky-books-enter.md
+++ b/.changeset/tricky-books-enter.md
@@ -1,0 +1,5 @@
+---
+"@sit-onyx/nuxt-docs": minor
+---
+
+feat: use scroll behavior to correctly scroll to anchors on initial page load and scroll to the top of the page when navigating to a new page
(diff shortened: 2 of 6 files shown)


JoCa96 (changes requested):
Unfortunately this is only fixing the symptom, but not the actual underlying issue.
You can check the dev version with different pages and browsers and see, e.g. that the auto-scrolling works *sometimes* and on page reload in Chrome, and more reliably in Firefox.

We should instead focus on resolving missing hydrations and layout-shifts, which will allow the browser to handle the hash links correctly.

larsrickert:
> We should instead focus on resolving missing hydrations and layout-shifts, which will allow the browser to handle the hash links correctly.

I am not sure if this will work since the browser and the Vue Router [scroll behavior](https://router.vuejs.org/guide/advanced/scroll-behavior) use the window as scroll container. But in our case we have a custom scroll container inside the OnyxAppLayout so I don't think that this will work. I tried by just creating a simple Nuxt page that statically renders content. Reloading the page does not jump to the correct section

Issue #6110, opened 2026-08-27
Fix mobile height not correct on some devices (iphone, ipad)


larsrickert:
### Version

1.20.0

### Description

Some heights (e.g. app layout height) is not correctly displayed on some devices such as iphones and ipads since they have a dynamic URL bar. The reason is propably that onyx is using 100vh internally which should be replaced with [100dvh](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/length#dynamic_viewport_units)

<img width="585" height="1266" alt="Image" src="https://github.com/user-attachments/assets/bd2e215a-deb8-437e-8d8d-77a56c48fa67" />

### Reproduction steps

1. Open the demo app on an iPhone: https://demo.onyx.schwarz/forms
2. See that the bottom bar is not fully visible

### Confirmations

- [x] I have checked that there is not already a bug report for my issue.
- [x] I confirm that are no sensitive information included in this bug report. e.g. credentials, internal project details (for employees of the Schwarz Group) etc.

PR #6143, opened 2026-09-03
fix(core): changed vh to dvh to fix mobile height

ChristianBusshoff:
Relates to #6110

- changed vh to dvh


packages/sit-onyx/src/components/OnyxAppLayout/OnyxAppLayout.vue
larsrickert, line 1: We can add a custom rule to our stylelint config to forbid the "vh" unit: https://stylelint.io/user-guide/rules/unit-disallowed-list/

larsrickert (changes requested):
There are a few more places where we use "vh", could you also take a look at them? :)

Image

larsrickert (approved):

Issue #5947, opened 2026-08-02
Implement migration from `@sit-onyx/mdc` to `@sit-onyx/comark`


JoCa96:
### Why?

> Currently we are using `@nuxtjs/mdc` for markdown rendering. It has recently been deprecated in favor of [Comark](https://comark.dev/kb/migration-from-mdc)

- `@sit-onyx/comark` was implemented in #5521

### Acceptance criteria

- [ ] the `@sit-onyx/mdc` package is updated
  - [ ] Added deprecation notice to `README.md`
  - [ ] Deprecated `@sit-onyx/mdc` package
- [ ] the `@sit-onyx/nuxt-docs` is updated
  - [ ]  All dependencies on `@sit-onyx/mdc` are replaced with `@sit-onyx/comark`
- [ ] documentation is updated

### Implementation details

- https://comark.dev/kb/migration-from-mdc

For deprecation notice, you can use the "[!IMPORTANT]" marker:

```md
> [!IMPORTANT]
> **`@sit-onyx/mdc` is being deprecated in favor of [@sit-onyx/comark](https://www.npmjs.com/package/@sit-onyx/comark).**
>
> [Comark](https://comark.dev) is the successor to MDC, which is being deprecated itself.
>
> **[Migration Guide: from MDC to Comark](https://comark.dev/kb/migration-from-mdc)**
```

### DEV handover checklist

[Storybook](ADD_LINK_HERE)

DEV handover done with: @ADD_DEV_NAME_HERE

**Testing**
- [ ] covered by visual screenshot tests
- [ ] covered by functional tests (Playwright or unit test)

**Documentation**
- [ ] "New" badge in Storybook is added, see [storybook-addon-tag-badges](https://github.com/Sidnioulz/storybook-addon-tag-badges/?tab=readme-ov-file#-usage)
- [ ] Storybook code snippets are checked
- [ ] component status has been updated in the [docs overview](https://onyx.schwarz/#components)

**Other**
- [ ] follow-up tickets were created if necessary (add links below)

PR #6157, opened 2026-09-09
feat!: Replace deprecated `@sit-onyx/mdc` dependency with `@sit-onyx/comark`

JoCa96:
Relates to #5947

- Added deprecation notice for `@nuxtjs/mdc`
- Replace deprecated `@sit-onyx/mdc` components with `@sit-onyx/comark` components in `@sit-onyx/nuxt-docs`

- We cannot remove `@nuxtjs/mdc` yet, because of the error https://github.com/nuxt-content/mdc/issues/174 . Instead of the suggested workaround of installing all sub-dependencies individually, we keep a direct dependency on `@nuxtjs/mdc`.
  - We will need to wait for https://github.com/nuxt/content/discussions/3776#discussioncomment-18049799 until it can be completely removed.
- The `mdc-unwrap` directives are also kept, as `@nuxt/content` has a direct dependency on `@nuxtjs/mdc`, so *mdc* ist still used internally.


packages/nuxt-docs/package.json
```diff
@@ -55,9 +56,10 @@
     "@nuxt/content": ">= 3",
     "@nuxtjs/color-mode": ">= 4",
     "@nuxtjs/i18n": ">= 10",
-    "@nuxtjs/mdc": ">= 0.20.2",
+    "@shikijs/themes": ">= 4",
```
larsrickert: Shouldn't this be a regular / direct dependency?

packages/nuxt-docs/package.json
```diff
@@ -55,9 +56,10 @@
...
     "@sit-onyx/icons": "workspace:^",
     "@sit-onyx/nuxt": "workspace:^",
+    "comark": "^0.6.2",
```
larsrickert:
```suggestion
    "comark": ">= 0.6.0",
```

Should we be more permissive?

packages/nuxt-docs/app/components/HighlightedCode.vue
```diff
@@ -12,7 +18,11 @@ ${props.code}
 `;
 });
 
-const { data: ast } = await useAsyncData(snippet, () => parseMarkdown(snippet.value));
+const { data: value } = await useAsyncData(snippet, () =>
+  parseMarkdown(snippet.value, {
+    plugins: [shiki({ themes: { light, dark } })],
```
larsrickert, line 23: Should we move this to `@sit-onyx/comark` itself so that we define the highlighting there and the user does not have to care about it manually?

.changeset/tender-cycles-serve.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sit-onyx/nuxt-docs": minor
+---
+
+feat!: Replace deprecated `@sit-onyx/mdc` dependency with `@sit-onyx/comark`
```
larsrickert, line 5: Can you also update our Storybook docs on packages/sit-onyx/.storybook/pages/markdown-renderer.mdx?

packages/nuxt-docs/package.json
```diff
@@ -43,6 +43,7 @@
     "@sit-onyx/icons": "workspace:^",
     "@sit-onyx/nuxt": "workspace:^",
     "@sit-onyx/shared": "workspace:^",
+    "comark": "^0.6.2",
```
larsrickert, line 47: I noticed that we expose the parser result for the deprecated MDC `OnyxMarkdownRenderer` component so the user can easily access the TOC etc. without needing to re-parse it. Should we do the same for the new comark based `OnyxMarkdown` component?

Can ofc be done in a separate PR

larsrickert (changes requested):

packages/nuxt-docs/app/components/HighlightedCode.vue
```diff
@@ -1,4 +1,10 @@
 <script lang="ts" setup>
+import light from "@shikijs/themes/github-dark";
+import dark from "@shikijs/themes/github-light";
```
larsrickert: Seems like this is not the correct theme? WE use the HighlightedCode component for the pnpm tabs but they look weird / different than the other code tabs

Image

JoCa96: fixed

packages/nuxt-docs/package.json
JoCa96: Good idea, but with comark this is not necessary anymore: The `OnyxMarkdown` and `OnyxMarkdownDocument` components accept parser results themselves, so they don't need to expose it again.

.changeset/tender-cycles-serve.md
JoCa96: done

packages/comark/README.md
```diff
@@ -79,18 +79,20 @@ export async function getContentDocument(slug: string) {
 
 ### 2. Render the parsed tree
 
-```vue
+```html
```
larsrickert, line 82:
```suggestion
```vue
```

Shouldn't this be vue?

packages/nuxt-docs/app/components/HighlightedCode.vue
```diff
@@ -1,4 +1,10 @@
 <script lang="ts" setup>
+import dark from "@shikijs/themes/github-dark";
```
larsrickert, line 2: Do you know why the theme does not look correct in dark mode?

Image

larsrickert (changes requested):

PR #6187, opened 2026-09-21
feat(DataGrid:RowClick): implement data grid row click integration with other features

ChristianBusshoff:
Relates to #5954

- Prevent rowClick feature from being triggered when interacting with expandableRows, selection and rearrange feature columns


packages/sit-onyx/src/components/OnyxDataGrid/features/rowClick/rowClick.ts
```diff
@@ -43,7 +54,10 @@ export const useRowClick = <TEntry extends DataGridEntry>(options: RowClickOptio
       if (hasSelection(rowElement)) return;
 
       // check if cell is configured to not be clickable
-      const cellElement = target.closest("td");
+      const path = event.composedPath();
+      const cellElement =
```
larsrickert, line 58: Why is this change needed? 🤔

larsrickert (approved):

packages/sit-onyx/src/components/OnyxDataGrid/features/rowClick/rowClick.ts
ChristianBusshoff: The expandable row creates a new <tr>. Without this fix, path is undefined and the click event triggers unintentionally.

PR #6188, opened 2026-09-21
fix(useSkeletonState): skeleton prop injection through onyx page layout and onyx form

ChristianBusshoff:
Relates to #1955

- fix OnyxForm inherit skeleton context from OnyxPageLayout


packages/sit-onyx/src/composables/useSkeletonState.ts
```diff
@@ -49,27 +49,50 @@ export type SKELETON_INJECTED = symbol; // we can't use `typeof SKELETON_INJECTE
...
+        return false;
+      }
+
+      if (parentElementProps?.skeleton !== undefined) {
+        return parentElementProps.skeleton === true ? 3 : false;
+      }
+
+      if (parentSkeleton) {
+        return parentSkeleton.value;
       }
```
larsrickert, line 81: Don't we need to do the same for props.skeleton?

packages/sit-onyx/src/components/OnyxPageLayout/OnyxPageLayout.vue
```diff
@@ -6,6 +6,7 @@ import type { OnyxPageLayoutProps } from "./types.js";
 
 const props = withDefaults(defineProps<OnyxPageLayoutProps>(), {
   footerAlignment: "full",
+  skeleton: undefined,
```
larsrickert, line 9: Is the `SKELETON_INJECTION_KEY` no longer needed?

larsrickert (changes requested):

packages/sit-onyx/src/components/OnyxPageLayout/OnyxPageLayout.vue
ChristianBusshoff: It's still needed inside useSkeletonContext and provideSkeletonContext, like before :)

